PR #45, opened 2016-02-08
image.py cog implementation

initiumSrc:
This cog handles image related commands.
Currently implemented commands:

imgur
gif

Imgur has various subcommands.
Gif is an implementation of the redv1 gif (less code because of new text (tuple) handling).

aiohttp does not collide with any of the imgurpython->ImgurClient dependencies.


initiumSrc:
Found some bugs, re

PR #59, opened 2016-02-14
Fixed is_owner_check

sedders123:
Was originally checking against a (as far as I can tell) undefined variable, and wasn't working for me. Using the owner id from the settings dictionary fixed my problem


Twentysix26:
Thanks

PR #60, opened 2016-02-16
Added !list queue command

sedders123:
Lists songs in the queue, with item 0 being the song currently playing. Works by checking to see whether the link in the queue is a SoundCloud or YouTube link and then using relevant API to get songs title.

SoundCloud API requires a Client ID, which can be obtained from this link http://soundcloud.com/you/apps/new, and set using !audioset soundcloud <client_id>. SoundCloud API key is optional and if not given the link will be listed as 'Could not get Title'.

Also gives more information about links added to the queue. Instead of 'Link added to queue' it will now try to respond with '\<song_name\> has been put into the queue by \<user\>'.


Twentysix26:
This is very well done. Much appreciated

PR #62, opened 2016-02-17
Update !gifr to use Giphy's random API call

scummboy:
This works far better than randomly selecting from 25 results returned by the search API call.


Twentysix26:
:+1:

PR #70, opened 2016-02-22
fixed hug command lol

NotSoSuper:


Twentysix26:
Granted, it is indeed bugged because when I made it I lazily copy pasted another command (probably !info) and then forgot to edit the parameters but your fix is not correct.
!hug is meant to be used with other users, yours only work with the author. Thanks for making me notice

NotSoSuper:
Ah I see, I could change it for that however it seems you've got it.
Thanks.

Twentysix26:
https://github.com/Twentysix26/Red-DiscordBot/commit/3c338eeab5574c13de399e900f1172470333ae0f
I think I may have went a bit overboard

ghost:
10 intensity levels of hugs? Oh my...

PR #77, opened 2016-02-27
Multiserver roles et al

tekulvw:
- Rewrote bot settings (data/red/settings.json) interface (cogs.utils.settings), all modifications to the settings should be done through `from __main__ import settings` object
- Adds server specific roles
  - Updates old setting format
- Adds command aliases (all future commands should be alias-proof)
- Customcommands is now case insensitive
- Better help messages for missing *.exe's *.dll's for audio
- fileIO pretty saves json
- ping in general
- chat formatting commands at `from cogs.utils.chat_formatting import *`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,5 @@ __pycache__
 data
 !data/trivia/*
 !data/audio/playlists/*
+*.exe
+*.dll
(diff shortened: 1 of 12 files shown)


jaw3l:
can u add support for aliases, multiple commands to execute? 
for example '' `!alias add test command1,command2`'' and then,  when we type test two commands work with one alias^

Issue #71, opened 2016-02-22
Suggestion: Better Dice Rolling


ghost:
I would like to have actual dice rolling in the bot, not just "roll something less than 100." :)

For more detail, I'd like to quote [a similar request made for Gravebot](https://github.com/Gravebot/Gravebot/issues/135).

> I'd like to request an expansion to Gravebot's current dice function. These features are partially inspired by the popular [Roll20](https://wiki.roll20.net/Dice_Reference) site.
> 
> Change syntax to **XdY+MOD**. **X** being the number of dice, **Y** being the die size, and **MOD** being the modifier that gets added to a given roll. This allows the bot to be used quickly by any tabletop gamer as the aforementioned syntax has almost become standard.
> 
> Automated re-rolls aka "_Exploding dice_". This means that when - for example - a number of six-sided dice are rolled, any die that comes up a 6 gets re-rolled. Let us say I roll 4d6 (4 six-sided dice).If one of these comes up as a 6 I roll it again, which ends up putting out five die results instead of just four.
> 
> Success-based rolls. Some systems such as **Vampire: The Masquarade** or **Shadowrun** simply roll a certain number of dice and evaluate only how many of them come up above a certain number. For example, Vampire considers any roll above 8 a success. If I roll 8 10-sided dice and 5 come up 8, 9 or 10 this means **5 successes** have been achieved.

Even if we could simply get `!roll 2d6` or something, that would be perfect. :+1:

Shigbeard:
+1, I can see this being really easy to implement with the equivalent of a for loop.

tekulvw:
No plans to add this functionality to the default bot. Good idea for a custom cog though.

orels1:
@xnaas I will try and make a custom cog for that. will get back with results if any

orels1:
Basically implemented the thing. Uses the following logic:
By default it rolls 4d20 (so 4 dices with 20 sides) and with mod = 0, so nothing is added.
You can roll with success check, by default rolls 4d20 with 10 as a check value. If half or more of the rolls are exceeding the success value - the roll will count as a success.

The extra-rolling thing is always present (need to clarify if it's necessary), so if any of the rolls are matching the maximum for the dice (so 6, for 6-sided, 12 for 12-sided, etc), it will roll extra and add them at the end of the roll.

Will clear it up and add as a PR to Red-Cogs, so everyone could use it

PR #83, opened 2016-03-07
added d20 option for dice rolls

thunderbrew:
Added d20 option referenced here 
#71


tekulvw:
Negative numbers kill the regex and can make their way down to the range()'s throwing errors

tekulvw:
Closing because author hasn't fixed cited issue

PR #91, opened 2016-03-15
leaderboard paging

irdumbs:
Why switch to paging:
- messages over 2000 characters get stopped by discord, so !leaderboard [large_number] currently breaks for large banks
- small pages look nicer and avoid spam
- easier to find what you're looking for

TODO:
   ~~!leaderboard [page or user] - user: display page where user is at.
      must use mentions in case of usernames of only numbers~~

### DIFF
--- a/cogs/economy.py
+++ b/cogs/economy.py
@@ -121,18 +121,34 @@ class Economy:
         else:
             await self.bot.say("{} You need an account to receive credits.".format(author.mention))
 
-    @commands.command()
-    async def leaderboard(self, page : int=1):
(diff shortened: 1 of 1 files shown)


Twentysix26:
I prefer the current system. `!leaderboard` is meant to display the top users and it should be limited to that.

irdumbs:
just know that the current way will error out if the leaderboard ends up being >2k characters
ex. !leaderboard 1000 will error out if there are that many entries in bank.json

Twentysix26:
Not anymore

PR #93, opened 2016-03-17
display game being played (only twitch)

irdumbs:
also `twitch_online()` changed to `get_twitch_data()`


Twentysix26:
I'll probably do it some other way if I'm gonna implement it

PR #104, opened 2016-03-25
accepts "y" for installation

ghost:
Irdumbs PR added only for update, now works for "cog install" too.


Twentysix26:
`elif answer.content.lower().strip() == ["yes", "y"]:`

I doubt a string will ever be equal to a list ;p
Adding it myself.

PR #106, opened 2016-03-27
Allow listing whitelisted users

Borks:


lucasvienna:
Wrote a less elegant solution, but one that work with all users, not just member of the server in which the command was run. https://github.com/Twentysix26/Red-DiscordBot/pull/112

Borks:
@lmvienna Isnt this a bit overcomplicating things? I definitely need to add a "No members in whitelist" message, but why would you even want to see all whitelisted members, not server based whitelisted users? @Twentysix26 Input on this?

lucasvienna:
@Borks well, since whitelist is not per server (just look at whitelist.json, it's simply an array) I don't see a reason to make it per server. I tried using get_member() and it gives a "None" for users not in the current server. Hence, get_all_members(). It may be a little overly, but I preffer seeing all whitelisted users. The best way, I think, would be if I did it separating users per server, as in:
Server 1:
- User1

Server 2:
- User2
  .... etc

Twentysix26:
@Borks The whitelist is global and as such it makes sense to show all users, instead of only the ones in the server. It would make things confusing otherwise.
The other PR also takes into account IDs that can not be retrieved. I had already noticed that both of those things were missing and I was gonna add them myself if I were going to accept your PR

PR #112, opened 2016-04-01
Added []whitelist show and []whitelist clear methods to mod.py

lucasvienna:
Someone suggested a method to see who is in the whitelist, so here it is. Also, there is an included method to clear the list, independent of server and who is in it.


tekulvw:
Why not do the same for blacklist?

tekulvw:
Closing as #213 does have both

ddday:
#213 Was closed because I was PRing to the wrong spot. The new one is #222 .

PR #121, opened 2016-04-05
added sound effects

irdumbs:
`!sfx` - plays the specified sound file from the `data/audio/sfx` directory
`!audioset sfx` - toggles sfx for server
`!list sfx` - lists sound effects

Gives responsibility of `!stop`-ping a stuck player to the user, like the other audio commands do
- [x] add >2k character support to `!list sfx`
- [x] list only the sfx toggle for the current server
- [ ] add other extensions to the supported list?
- [ ] add some default sfx?
- [x] don't require multiword sfx to be put in "qoutes"
- [ ] ~~sfx playing while song is playing~~ (probably will never support)
- [ ] disable/enable sfx usage per person :P 
- [ ] case-insensitive (requested by Kowlin)

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -49,6 +49,7 @@ class Audio:
     def __init__(self, bot):
         self.bot = bot
         self.music_player = EmptyPlayer()
+        self.sfx_player = EmptyPlayer()
         self.settings = fileIO("data/audio/settings.json", "load")
         self.queue_mode = False
         self.queue = []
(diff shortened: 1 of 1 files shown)


irdumbs:
sfx are never paused. They are played until finished or stopped.

define sfx functionality of player commands 
- [x] `!play` owner: stops sfx, plays song
- [x] `!playlist` | `!mix` | `!sing` | `!queue` doesn't start until sfx is finished
- [x] `!prev` nothing
- [x] `!pause` nothing
- [x] `!resume` | `!skip` stops sfx, resumes song (no majority vote needed)
- [x] `!stop` owner: stops all audio~~, non-owner: stops sfx, resumes song~~

~~these behaviors are not implemented yet~~

irdumbs:
there may have been a bug with overlapping audio when
!queue something
!sfx something
!playlist something

That may just be my non-vanilla audio though. will test on vanilla audio later
fixed: 162916d

irdumbs:
#132 <- any audio command, includes sfx

PR #130, opened 2016-04-09
Stop the ping spam

tekulvw:
It's a debug command anyways, there's no real good reason why anyone but the owner should be able to use it.


Twentysix26:
I respectfully disagree :^)

PR #138, opened 2016-04-11
Added support for Bot tokens

Kowlin:
Feel free to give some comments about the code and instructions before pulling in

I also recommend waiting  for the multible audio support ;)
- [x] Make logintype work
- [x] Keep backwards compatibility
- [x] Make the bot state clearly the difference's and give user links
- [x] Test it at least 10 times, in any situation possible!
- [ ] Explain Oauth in the bot after installing

### DIFF
--- a/red.py
+++ b/red.py
@@ -162,7 +162,7 @@ async def debug(ctx, *, code : str):
 
     result = python.format(result)
     if not ctx.message.channel.is_private:
-        censor = (settings.login, settings.password)
+        censor = (settings.email, settings.password)
         r = "[EXPUNGED]"
         for w in censor:
             result = result.replace(w, r)
@@ -348,16 +348,16 @@ def check_configs():
         print("Red - First run configuration")
         print("If you don't have one, create a NEW ACCOUNT for Red. Do *not* use yours. (https://discordapp.com)")
         print("Alternatively, you can use a bot token. Register one in the Discord API docs. (https://discordapp.com/developers/applications/me)")
-        settings.login = input("\nEmail or Token> ")
+        settings.email = input("\nEmail or Token> ")
 
-        if len(settings.login) is 59 and "@" not in settings.login:
+        if len(settings.email) is 59 and "@" not in settings.email:
             print("Token found, We're going to use token for authentication")
             settings.logintype = "token"
         else:
-            settings.logintype = "email"
             print("Email found. We're going to use email for authentication")
+            settings.logintype = "email"
             settings.password = input("\nPassword> ")
-            if not settings.login or not settings.password:
+            if not settings.email or not settings.password:
                 input("Email/Token and password cannot be empty. Restart Red and repeat the configuration process.")
                 exit(1)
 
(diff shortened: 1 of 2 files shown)


tekulvw:
You have to add a logintype property to the settings module, it doesn't just work on it's own

Kowlin:
Surprisingly no. It adds some weird kwerks I could not find out. The login type is a way of avoiding that

tekulvw:
I wrote the settings module? It does require you to add a property for logintype in order for the settings module to be able to read/write from it. If it's working for you then I have absolutely no idea how

Kowlin:
I don't even know lol, but it seems I need to fix another merge conflict

red.py
```diff
@@ -488,8 +491,11 @@ def main():
     if settings.owner == "id_here":
         print("Owner has not been set yet. Do '{}set owner' in chat to set yourself as owner.".format(bot.command_prefix[0]))
     else:
-        owner.hidden = True # Hides the set owner command from help
-    yield from bot.login(settings.email, settings.password)
+        owner.hidden = True  # Hides the set owner command from help
+    if settings.logintype == "email":
```
tekulvw: I promise, this ain't gonna work unless you make the necessary changes [here](https://github.com/Twentysix26/Red-DiscordBot/blob/develop/cogs/utils/settings.py)

Kowlin:
Tested this like... 14 times... it should work without a problem now.

franc-ist:
Can confirm it works

irdumbs:
that auth length thing is scurry
Why not just ask them straight out
input('<description of the differences between bot/not> Do you want a BOT tag or do you want to keep your bot a user account? (bot/user)')

irdumbs:
official request to keep the info for both auth and user/pass
and add a command to restart the bot as one or the other

Kowlin:
Ehh... Why? This still is base code to work on. But I dont see the need for 2logins

red.py
```diff
@@ -347,16 +347,22 @@ def check_configs():
...
         print("If you don't have one, create a NEW ACCOUNT for Red. Do *not* use yours. (https://discordapp.com)")
-        settings.email = input("\nEmail> ")
-        settings.password = input("\nPassword> ")
+        print("Alternatively, you can use a bot token. Register one in the Discord API docs. (https://discordapp.com/developers/applications/me)")
+        settings.email = input("\nEmail or Token> ")
 
-        if not settings.email or not settings.password:
-            input("Email and password cannot be empty. Restart Red and repeat the configuration process.")
-            exit(1)
+        if len(settings.email) is 59 and "@" not in settings.email:
```
tekulvw: We've got no clue if it's always gonna be 59 chars and you shouldn't use `is` here

tekulvw:
Honestly I'd feel much better about making the user specify whether they wanna use token or email login. Mostly because we don't have a deterministic way to determine if a string is a valid token.

edit: on that note, awaiting a response from Jake to see if they have an internal method for checking token validity

edit2: you know...we could do it by logging the bot in right there, if it throws an error we know its an invalid token, unless of course they don't have internet during setup

Kowlin:
I think someone answered that question in the api channel. But feel free to make a PR on my branch

irdumbs:
I agree with @tekulvw let the user tell us which they're using
that way when it breaks, it's just them not following instructions.

Kowlin:
Well alright. But @tekulvw will need to make that part then since I do not understand the damn settings at all. This is the best I can do

Kowlin:
currently making the logintype setting work. I'm close... very close
Edited my main post with a todo list.

PR #141, opened 2016-04-12
Added lists for StreamAlerts that are active.

HeatherEliza:


cogs/streams.py
```diff
@@ -367,6 +367,48 @@ def __init__(self, bot):
...
+        await self.bot.say("Twitch Streams set up for Alerts:")
+        twitempty = 1
+        for item in self.twitch_streams:
+            await self.bot.say(item["NAME"])
+            twitempty = 0
+
+        if twitempty == True:
+            await self.bot.say("No Twitch streams are set to Alert in this channel.")
+
+        fileIO("data/streams/twitch.json", "r", self.twitch_streams)
```
irdumbs, line 383: ooo inline comments.

what's this fileIO doing here?

also, propose to change !hitboxlist to !hitlist  :3

tekulvw:
Yeah I'm agreeing with @irdumbs, the fileIO reads really should _not_ be here. Also you really don't want to send a separate message for each feed. Just put em into a list and send the whole thing at once.

Twentysix26:
This PR needs serious fixing... I'm also pretty sure that it doesn't work correctly since it pulls all the streams, not the ones for the current channel. Let me know if you want to fix it up

PR #158, opened 2016-04-15
Actual *user* counter

tekulvw:
Members are unique to a server, not to each other. Let's get rid of the duplicates, shall we?

### DIFF
--- a/red.py
+++ b/red.py
@@ -40,7 +40,7 @@ lock = False
 
 @bot.event
 async def on_ready():
-    users = str(len([m for m in bot.get_all_members()]))
+    users = str(len(set(bot.get_all_members())))
     servers = str(len(bot.servers))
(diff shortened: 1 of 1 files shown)


irdumbs:
can you explain?

tekulvw:
Basically each server has a set of members that belong to that server. Any one given discord user can be a member on multiple servers, and we end up counting them multiple times if we just check the size of `bot.get_all_members()`. However, using set() will make sure each user is only counted once.

irdumbs:
why would get_all_members return a list of all members including duplicates if there's no way to distinguish which server they came from with that call?

tekulvw:
There is a way, because a discord.Member has a `server` attribute. Using `set()` will only keep the members with unique user id's

irdumbs:
i see. ty

PR #161, opened 2016-04-16
More info on !join w/ BOT tag

irdumbs:
more convenient for the user and for #support \o/


Twentysix26:
Missing an await. It was missing it even before this PR

Kowlin:
HA hahaha.... oops >.<

PR #166, opened 2016-04-17
custom command args

irdumbs:
prints test msg after add/edit

hidden 4th arg messages
also all args can be referenced by number instead
in case people are used to welcome
author - 0
server - 1
channel - 2
message - 3

edge-cases:
all custom commands that already had '{something else}' in them will break
^ commands won't send any feedback to the user. They must look in console to find error

TODO:
- [x] make a doc explaining further how to use. cause some people won't understand [Red-Docs#7](https://github.com/Twentysix26/Red-Docs/pull/7)

### DIFF
--- a/cogs/customcom.py
+++ b/cogs/customcom.py
@@ -153,15 +153,37 @@ def check_folders():
         print("Creating data/customcom folder...")
         os.makedirs("data/customcom")
 
-def check_files():
+def check_files(bot):
     f = "data/customcom/commands.json"
     if not fileIO(f, "check"):
         print("Creating empty commands.json...")
-        fileIO(f, "save", {})
+        fileIO(f, "save", {"FORMAT":True})
+        return
+    print('hi')
+    current = fileIO(f, "load")
+    if "FORMAT" not in current: # consistency check
+        print("Custom Commands have been updated with format string support.")
+        print("Updating old custom commands. Any brackets {} found will be escaped {{}}")
+        for sid, commands in current.items():
+            msg = ""
+            for cmd, response in commands.items():
+                if "{" in response or "}" in response:
+                    commands[cmd] = response.replace("{","{{").replace("}","}}")
+                    #use [p] if loaded before settings/prefixes were
+                    pfx = "[p]" if bot.command_prefix[0] == "_" else bot.command_prefix[0]
+                    msg += "{}{}\n".format(pfx,cmd)
+            if msg:
+                # can't get server cause not logged in..
+                svr = bot.get_server(sid)
+                if svr is None:
+                    svr = sid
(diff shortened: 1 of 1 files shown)


Twentysix26:
> all custom commands that already had '{something else}' in them will break

That doesn't sound good at all :/

franc-ist:
https://twentysix26.github.io/Red-Docs/red_guide_command_args/

irdumbs:
kk will go through all old commands to find any {} pairs.
~~customcom json file will have a useless header though as a result~~ just change the name of a key
bug me about this in discord, will ya?

irdumbs:
should I silently escape all the {} in the old customcommands?
some people have already started trying to make some customcommands with format strings.
should I escape those too?

should I print to console, "This is the first time running customcom since format string update. editing old commands to make sure their outputs don't change" or something like that?

irdumbs:
will spit out to console that commands are being changed.
~~will spit out which server is being changed,~~ cog is most likely loaded before bot logs in, so no server name. can print id, but don't think it's needed
and which commands in those servers

irdumbs:
can't print the prefix..
image

irdumbs:
ready for review

output from 1st time loading the updated customcom.py will be something like this if brackets are found.

```
Loading cogs...

Custom Commands have been updated with format string support.
Updating old custom commands. Any brackets {} found will be escaped {{}}
Server: 133049272517001216 - custom commands altered:
--------
[p]stash
[p]blurp
--------

Server: 143153940500971520 - custom commands altered:
--------
[p]stash
--------
```

irdumbs:
@tekulvw and @Kowlin have brought up a few concerns
- format strings may leave room for security flaws
- erroring out on addcom/editcom is very unwanted even if the user is told what they entered isn't valid
- restricts usage of {}
- too complex for users

@tekulvw's proposed solution:
- template strings would be simpler, safer, wouldn't error out and would avoid edge cases

My concerns with the solution:
- limits functionality
- different syntax than users may be used to / is documented
- ${different}syntax for no spaces

My proposed compromise:
- Continue to use consistent and familiar {arg.attr} but, 
- limit functionality for security by only allowing an exact list of argument/attributes, defined by [Red-Docs#7](https://github.com/Twentysix26/Red-Docs/pull/7). 
- Substrings not in the approved list are ignored and just printed as literal strings so as to not restrict {} usage and to prevent erroring out.
- The backend of how this works is independent of the functionality. Could use format strings, template strings, or neither.

tekulvw:
Can we close this for now?

irdumbs:
yeh. I'll get back to it eventually. I think 26 was working on a similar solution anyway

PR #168, opened 2016-04-17
more specific owner check

irdumbs:
in case someone can hijack a message and replace their author.id with "id_here", then use !debug to do whatever they want
although..
if they know the owner, they could just replace it with the owner's id ¯\_(ツ)_/¯


Twentysix26:
That... just cannot happen, assuming discord's servers don't get compromised. And in that case it would probably be possible to impersonate the owner directly.

PR #184, opened 2016-04-24
Added Youtube Live Broadcast alerts to streams.py

Redjumpman:
This requires an API key in order to work, but instructions to obtain one, and where to put it have been added to the code. All other functions of stream.py have no conflicts with the addition of youtube broadcast live.


Kowlin:
It's better to redo this PR and only stage the files that are deliberate to be changed. There is a lot of trash in the changes.

tekulvw:
This should be rebased, not PR'd. I will update the branch this afternoon

PR #192, opened 2016-04-25
Google TTS

Fr6jDJF:
Google TTS

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -297,39 +297,37 @@ class Audio:
         tts.save(ttsFile)
 
         msg = ctx.message
-        localsfx = True
-        if localsfx:
-            if os.path.exists(ttsFile):
-                if await self.check_voice(msg.author, ctx.message, True):
-                    try:
-                        if self.music_player.is_playing():
-                            self.music_player.paused = True
-                            self.music_player.pause()
+        if os.path.exists(ttsFile):
+            if await self.check_voice(msg.author, ctx.message, True):
+                try:
+                    if self.music_player.is_playing():
+                        self.music_player.paused = True
+                        self.music_player.pause()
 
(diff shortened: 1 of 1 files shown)


cogs/audio.py
```diff
@@ -263,6 +271,63 @@ def __init__(self, bot):
         else:
             await self.bot.say("There are no valid sound effects in the `data/audio/sfx` folder.")
 
+    @commands.command(pass_context=True, no_pm=True)
+    async def tts(self, ctx, lang: str, *, text: str="Please try again."):
+        """Convert your text to speech (Google TTS).
+        """
+        
+        text = self.chop(ctx, text)
```
irdumbs, line 279: `!tts en my string`
text is "my string"
~~from the looks of it, chop will return: "ring"~~
ah nvmd.
but is text ever None?
#chop `if message == None:`

irdumbs:
doesn't look like it accounts for resuming audio while a tts is playing

cogs/audio.py
```diff
@@ -923,6 +985,20 @@ def is_playlist_name_valid(self, name):
             result = {"title": "A song "}
         return result
 
+    def chop(self, ctx, stuff:str):
+        message = stuff
+        if message == None:
+            message = ctx.message.author.name
+        else:
+            message = ctx.message.clean_content
+            message = message[(len(ctx.prefix)+len(ctx.command.name))+1:]
```
irdumbs, line 994: I would be careful of using this.
I believe it bugs out if a prefix is a mention.
I recommend just going through `stuff` and converting the mentions in that.

irdumbs:
please PR this to cog_changes

tekulvw:
I'm seconding that. Closing in anticipation of a PR to `cog-changes`.

Issue #187, opened 2016-04-24
General list of things to do


tekulvw:
Feel free to comment with suggestions
- [ ] ~~Import plugins as modules~~
- [ ] Nested commands
- [ ] ~~Better command parsing~~
- [x] ~~Separate Owner commands into Owner plugin #188~~
- [ ] Scheduled file flushers
- [ ] Base plugin class
  - infilter
  - outfilter
- [ ] `--debug` flag
  - And start flags in general
  - `--daemon`?
- [x] Global permission changes (see tekulvw/Squid-Plugins)
- [ ] Global config
- [ ] commands costing currency
- [x] ~~`aliaslist` to `alias list` #195~~
- [x] Server specific bank \* #201
- [x] ~~Make aliases case insensitive #195~~
- [x] ~~Auto generate oauth URL for bot accounts #196~~
- [x] Add `bot` to globals in debug
- [ ] Override `on_error` to provide some better permissions error messages
- [ ] Choose which cogs to update
- [ ] Audio improvements? https://github.com/Twentysix26/Red-DiscordBot/issues/271

\* _denotes currently in development_

irdumbs:
- commands optionally costing economy currency
- customizable command permissions #41
  - per user
  - per role
  - per channel
  - per server
  - per time??

franc-ist:
restrict `!bank set` to owner only. i'm fed up of people fucking up my leaderboard by cheating

irdumbs:
perhaps choose your admins more carefully 😉

franc-ist:
@irdumbs its not that. somehow my bot has made its way onto 5 servers i dont own, and people are cheating

irdumbs:
lolol
That's actually an issue with economy.py
It's bank should be server-specific

franc-ist:
Remove user details from cogs (e.g. economy) when the bot leaves a server

franc-ist:
#219

franc-ist:
`NameError: name 'bot' is not defined` when calling a debug alias

irdumbs:
fix debug not having bot in lambdas
`[p]debug (lambda: bot.say('hi'))()`
should work to be backward compatible

kanahanazawa:
hiya. cant add people to blacklist that have a space in their name. guessing that would also count for whitelist? tired of certain users playing troll songs. also, i tried using the new nickname feature in discord to remove the space but that doesn't work guessing cause its not their actual name.

[...]

PR #201, opened 2016-04-27
Turn economy to server specific

tekulvw:
Needs testers!

Fix for #187 server specific bank

### DIFF
--- a/cogs/economy.py
+++ b/cogs/economy.py
@@ -9,6 +9,8 @@ import os
 import time
 import logging
 
+log = logging.getLogger('red.economy')
+
 slot_payouts = """Slot machine payouts:
     :two: :two: :six: Bet * 5000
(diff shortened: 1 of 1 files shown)


irdumbs:
problems I have with this:
- requires all users to re-register
- moves multi-server users' balances to one server, removing their balance from all other servers
- the 1st server they re-register to will be the one with their balance. does not warn the user of this at all
- should be done without users noticing. Too much user involvement because of the above issues

tekulvw:
@irdumbs latest commit 663c2d2 gives the original account value to a user once they use an economy command on a new server

irdumbs:
ty <3

ddday:
tested it, users did have to re-register and commands had to be used twice to go through

tekulvw:
Closing because of impending doom.

(economy rewrite)

Kowlin:
https://youtu.be/aXU8w336oGs

PR #204, opened 2016-04-27
Updated filters on Audio.py

Kowlin:
A actual fix for filters?! What?!


tekulvw:
[source](https://libav.org/libavfilter.html#Examples-1)

Issue #219, opened 2016-04-30
Make aliaslist actually list aliases instead of spamming the chat.


agtr:
![fulllength](https://cloud.githubusercontent.com/assets/12876103/14938390/dc6a887c-0f21-11e6-9757-f08921b3cf75.PNG)

franc-ist:
#187

tekulvw:
I'm confused as to what you're actually asking.

If you're asking to show what each alias is, that can be done with `!alias
show <name>`. If you're asking to show what each alias is NEXT to the name
in the list, that's likely something I'm not going to implement. Like you
said, we're trying to avoid spamming the chat and having it show what the
alias is in `!alias list` is gonna wreak havoc especially if you have long
aliases.
On Apr 30, 2016 4:27 PM, "Francis Taylor" notifications@github.com wrote:

#187 https://github.com/Twentysix26/Red-DiscordBot/issues/187

irdumbs:
!customcommands sends them in a PM,
maybe !aliaslist could do that as well?

ddday:
`  @commands.command(pass_context=True, no_pm=True)
    async def aliaslist(self, ctx):
        server = ctx.message.server
        if server.id in self.aliases:
            message = "```Alias list:\n"
            for alias in sorted(self.aliases[server.id]):
                if len(message) + len(alias) + 3 > 2000:
                    await self.bot.say(message)
                    message = "```\n"
                message += "\t{}\n".format(alias)
            if len(message) > 4:
                message += "```"
                await self.bot.say(message)`


To 


`    @commands.command(pass_context=True, no_pm=True)
    async def aliaslist(self, ctx):
        server = ctx.message.server
        if server.id in self.aliases:
            message = "```Alias list:\n"
            for alias in sorted(self.aliases[server.id]):
                if len(message) + len(alias) + 3 > 2000:
                    await self.bot.say(message)
                    message = "```\n"
                message += "\t{}\n".format(alias)
            if len(message) > 4:
                message += "```"
                await self.bot.whisper(message)`

tekulvw:
#195 7e4c41f

PR #221, opened 2016-05-01
Aliaslist to pms

ddday:
A solution to #219 , instead of !aliaslist showing up in the chat, it will show up in your pms.

### DIFF
--- a/cogs/alias.py
+++ b/cogs/alias.py
@@ -83,7 +83,7 @@ class Alias:
             fileIO("data/alias/aliases.json", "save", self.aliases)
         await self.bot.say("Alias '{}' deleted.".format(command))
 
-    @commands.command(pass_context=True)
+    @commands.command(pass_context=True, no_pm=True)
     async def aliaslist(self, ctx):
         server = ctx.message.server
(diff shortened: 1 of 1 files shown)


cogs/alias.py
```diff
@@ -83,7 +83,7 @@ def __init__(self, bot):
             fileIO("data/alias/aliases.json", "save", self.aliases)
         await self.bot.say("Alias '{}' deleted.".format(command))
 
-    @commands.command(pass_context=True)
+    @commands.command(pass_context=True, no_pm=True)
```
irdumbs, line 86: nice catch

tekulvw:
Two things: Line 93 you didn't change `bot.say` to `bot.whisper`, and in #195 I wanted to put `aliaslist` into the `alias` group. Thoughts on that? Because if that's something we agree is wanted then I'll just make these changes in that PR

ddday:
I did change it to bot.whisper, maybe I just did the PR wrong, otherwise it wouldn't work(which it does). Also, sure, I guess that would be fine.

tekulvw:
Added with 7e4c41fe

PR #224, opened 2016-05-02
Added support for multiple dice

BillKing94:


Eslyium:
!help roll (should be a bit more detailed)

Before - !roll [max=100] [nDice=1]
After - !roll (amount) (number of dice)
ex. !roll 100 2

Description
Rolls random numbers with one or more dice.

Other than that works fine.

irdumbs:
could you give a reason why this was closed @Kowlin?

Was it inactivity?

Kowlin:
Inactive. Misses a `))` and the feedback on the multible dice is sadly to say horrible. And beyond that the Github user is inactive.

BillKing94:
Feedback? I'll fix the parens if there's any interest in adding this, but any feedback hasn't made its way to me.

irdumbs:
if you were to add it as a cog, it would get approved much quicker.
You could get more substantial feedback then.
once it's all spiffed up, you may get more traction on having it merged into general.py

PR #227, opened 2016-05-02
!song shows local file name

franc-ist:
Will show the local file name when using !song. e.g. `Playing from local file: 'testing/Crazy in Love (Feat. Leah Kelly)'`


franc-ist:
here as a placeholder, as we are changing audio

PR #237, opened 2016-05-08
[Audio] Working local playlists

tekulvw:


Kowlin:
Works on my end

franc-ist:
@Kowlin even with the new audio?

PR #239, opened 2016-05-09
List of server count starts from 1

Eslyium:
Making servers list start from 1 instead of 0


Eslyium:
I dun fucked up

irdumbs:
almost. i'd change it in the actual print 👍

PR #242, opened 2016-05-11
!set stream [stream] [status]

irdumbs:
stream can be twitch channel name or full url.


irdumbs:
oh boy do I dislike github.
1 moment..

irdumbs:
https://github.com/Twentysix26/Red-DiscordBot/pull/243

PR #244, opened 2016-05-11
irdumb's twitch set

ddday:
Author: irdumb
Adds !set stream, because why not ¯ _( ツ )_ /¯.
#242

### DIFF
--- a/cogs/owner.py
+++ b/cogs/owner.py
@@ -275,6 +275,24 @@ class Owner:
             settings.password = ""
             await self.bot.say("Token set. Restart me.")
             log.debug("Just converted to a bot account.")
+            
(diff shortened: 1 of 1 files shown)


irdumbs:
oh, you know what?
people won't be able to use this unless they use the audio hotfix or multivoice pr
since we're telling them to use the old discord.py

Kowlin:
Don't care :D near release anyhow.

ddday:
_twiddles thumbs_

ddday:
well we're no longer telling them to use the old discord.py sooo

irdumbs:
boop bop a rebop a roobarb pie

irdumbs:
see #300?

PR #246, opened 2016-05-11
Whitelist clear+show & Blacklist clear+show the right way

ddday:
Finally figured out how to create a PR the right way so..
It's the same as last time, the title says it all.
Adds: `!whitelist clear` `!whitelist show` `!blacklist clear` `!blacklist show`

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -216,6 +216,50 @@ class Mod:
         else:
             await self.bot.say("User is not in blacklist.")
 
+    @blacklist.command(name="show", pass_context=True)
+    async def _blacklist_show(self, ctx):
+        """Shows current blacklisted users"""
+        blckUsers = ''
+        unknown = []
+        blckList = fileIO("data/mod/blacklist.json", "load")
+        if blckList == []:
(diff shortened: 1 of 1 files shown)


cogs/mod.py
```diff
@@ -216,6 +216,50 @@ def __init__(self, bot):
...
+    @blacklist.command(name="clear", pass_context=True)
+    async def _blacklist_clear(self, ctx):
+        """Resets the blacklist to a blank state"""
+        blklist = fileIO("data/mod/blacklist.json", "load")
+        await self.bot.say("Are you sure you want to clear the blacklist? Everyone who was previously blacklisted be able to use me that way. (yes/no)")
+        response = await self.bot.wait_for_message(author=ctx.message.author)
+        response = response.content.lower().strip()
+        if response == "yes" or response == "y":
+            for i in blklist:
+                self.blacklist_list.remove(i)
```
irdumbs: instead of looping through all of em, could just
`self.blacklist_list = []`

tekulvw:
resubmit PR to develop for now, may change

PR #280, opened 2016-05-26
Additions modtools

franc-ist:
What's new?
- Softban capability
- Option reasons for banning/kicking a user


franc-ist:
will fix and redo

PR #281, opened 2016-05-26
Mod Additions

franc-ist:
- Reasons for kick/ban
  - PM sent to user informing them of the reason
- Softbanning

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -49,14 +49,23 @@ class Mod:
 
     @commands.command(no_pm=True, pass_context=True)
     @checks.admin_or_permissions(kick_members=True)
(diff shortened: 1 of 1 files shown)


Twentysix26:
Codewise, I don't like various stuff in this PR, but the dealbreaker is that this would make the "days" parameter in the ban function mandatory. When banning, you wouldn't be able to specify a reason without specifying the days too.
I don't have a solution for this

irdumbs:
about the dealbreaker, could take the days + message as one argument and parse it yourself

Twentysix26:
I put in softban already, if reasons are to be implemented I'll have to think about something better.

PR #282, opened 2016-05-30
Do nicer dataIO thing, just feels better

tekulvw:
All of the methods in this class are being used as class methods so we should have them as such.


tekulvw:
Closing in favor of #307

PR #292, opened 2016-06-04
Audio patch

tekulvw:
- Various PEP8 formatting stuff
- Acknowledge and don't attempt to play songs that have duration above maxlegth
- Limit bitrate to 64k


irdumbs:
I assume the 64k limit is temporary?

PR #293, opened 2016-06-07
Urban supports specific definitions

Kowlin:
This is the only way to do this as far as I can find out. The reason being that we call for a string, and ye... Strings are evil!

### DIFF
--- a/cogs/general.py
+++ b/cogs/general.py
@@ -7,7 +7,8 @@ import time
 import aiohttp
 import asyncio
 
-settings = {"POLL_DURATION" : 60}
+settings = {"POLL_DURATION": 60}
+
 
 class General:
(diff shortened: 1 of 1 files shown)


irdumbs:
`self, ctx, search_terms, result_number : int=1` ?
sorry, didn't occur to me until now :3

Kowlin:
And how do we preserve multible spaces between words?

irdumbs:
ah you would be correct in dat. carry on.
tiny change could be to take it in as a list instead, save you the trouble

Twentysix26:
Refactored in https://github.com/Twentysix26/Red-DiscordBot/commit/881a1d3212f9544aaf47532acf681f30cac817f3

PR #294, opened 2016-06-07
Audio patch

tekulvw:
- Queue list

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -234,7 +234,6 @@ class Audio:
     def __init__(self, bot):
         self.bot = bot
         self.queue = {}  # add deque's, repeat
-        self.queue_lock = asyncio.Lock()
         self.downloaders = {}  # sid: object
         self.settings = fileIO("data/audio/settings.json", 'load')
         self.server_specific_setting_keys = ["VOLUME", "QUEUE_MODE",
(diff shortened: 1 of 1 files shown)


tekulvw:
Hold off on this, I'm gonna fix it up a bit tonight

19lmyers:
Is this going to be merged??

irdumbs:
we might make a 
`!songlist`
command instead that lists all the songs including ones in the currently playing playlist

should we have `!queue` and `!playlist` also list their respective songs?

jonyroda97:
do the <code>!songlist</code> and the others too

the <code>!queue</code> it's important so that we can see what was queued

Are you adding the names of who have queued the songs to that list? It would be good so that we can punish who puts "bad" music. 

Is there a way to block people to queue songs?

tekulvw:
I don't like any sort of do-all `songlist` command because we'll be bombarding Youtube for no good reason. `!queue` as it is in this PR will show you the currently playing song as well as up to the next 10

irdumbs:
no need to bombard youtube if we store the info on the 1st pass

tekulvw:
The first pass is when we'd bombard youtube. Plus, where would you store the info? For how long? You start entering into the territory of ridiculous memory growth pretty fast.

In other words, trying to get all of the information for every queued song is unnecessary and I am not going to implement it. My only question is do we show the next 5 or 10 songs?

jonyroda97:
5 by default with possibility to ask for an amount of queued songs? (last, only possible to admins?)

tekulvw:
I'm fairly confident I'm just gonna do 5, the use case for more than that is so small I'm just gonna ignore it

jonyroda97:
ok 👌

PR #296, opened 2016-06-08
!audioset status

irdumbs:
sets bot's status as song's title when bot is only playing in 1 server.
reverts to old status when done.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -936,7 +936,7 @@ class Audio:
             await self.bot.say("If only one server is playing music, songs' titles will now show up as status")
         else:
             await self.bot.say("Songs' titles will no longer show up as status")
-        # doen't update for some reason. fix later. PR dangit
+        # go ahead and update here.
         await self.update_bot_status()
         self.save_settings()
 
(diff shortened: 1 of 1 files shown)


tekulvw:
- Don't make stop async, instead use `self.bot.loop.create_task(...)` inside to call game update
- Rename your functions to start with a leading `_` and move them above the commands in alphabetical order
- Refactor line 1552 `if/elif/else` block into it's own function, `_set_status`
- Put some logging into `update_status`

Why are you changing bot status if `TITLE_STATUS` is disabled?

irdumbs:
I left the leading `_` off to say people can use em. should people not use them?
why change if disabled? to revert to the old status. why not?

The others, I'll get around to it. I'm avoiding spending more time on audio.

irdumbs:
clarification from discussion. changes:
- [x] _ for helpers
- [x] helpers up top
- [x] alphabetized
- [x] create_task inside _stop instead of asyncing it
- [x] more functions
- [x] logging
- [x] disable means don't mess with it
- [ ] test it

PR #300, opened 2016-06-14
!set stream

irdumbs:
couldn't be bothered to figure out how to change anomaly's PR #244


irdumbs:
closed in favor of #310

PR #301, opened 2016-06-14
!audioset status

irdumbs:
git git git git git
trying to rebase
just gonna make a new one.
#296

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -832,7 +832,7 @@ class Audio:
     async def _remove_song_status(self):
         if self._old_game is not False:
             await self.bot.change_status(self._old_game)
-            log.debug('Bot status returned to ' + self._old_game)
+            log.debug('Bot status returned to ' + str(self._old_game))
             self._old_game = False
 
     def _save_playlist(self, server, name, playlist):
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -240,6 +240,13 @@ class Audio:
                                              "VOTE_THRESHOLD"]
         self.cache_path = "data/audio/cache"
         self.local_playlist_path = "data/audio/localtracks"
+        self._old_game = False
+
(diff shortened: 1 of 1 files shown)


irdumbs:
moved checklist here.
clarification from discussion. changes:
- [x] _ for helpers
- [x] helpers up top
- [x] alphabetized
- [x] create_task inside _stop instead of asyncing it
- [x] more functions
- [x] logging
- [x] disable means don't mess with it
- [x] test it

cogs/audio.py
```diff
@@ -906,6 +930,19 @@ def _stop_player(self, server):
             self._kill_player(server)
             del voice_client.audio_player
 
+    # no return. they can check themselves.
+    async def _update_bot_status(self):
+        if self.settings["TITLE_STATUS"]:
+            active_servers = self._get_active_voice_clients()
```
tekulvw, line 936: should really wrap this in a try block

cogs/audio.py
```diff
@@ -1727,6 +1777,8 @@ def has_connect_perm(self, author, server):
                 song = None
             self.queue[server.id]["NOW_PLAYING"] = song
             log.debug("set now_playing for sid {}".format(server.id))
+            await self._update_bot_status()
```
tekulvw, line 1780: I'm gonna suggest `loop.create_task()` here too because this isn't essential to `await`. It'll just slow down the loop

irdumbs:
create_task makes sense.
why try though?
where can it fail?

tekulvw:
if the number of voice clients change while you're iterating over them it'll throw `RuntimeError` iirc

irdumbs:
ty ty
- [x] create_task
- [x] try
- [x] make changes in a new PR cause I can't here anymore :3

irdumbs:
moved to #312

PR #303, opened 2016-06-14
Alias case insensitivity

tekulvw:


Kowlin:
Tested and confirmed working.

PR #307, opened 2016-06-16
DataIO overhaul

tekulvw:
- Make DataIO into an actual usable object

### DIFF
--- a/cogs/utils/dataIO.py
+++ b/cogs/utils/dataIO.py
@@ -3,11 +3,14 @@ import os
 import logging
 from shutil import copy
 
+
 class InvalidFileIO(Exception):
     pass
 
+
 class CorruptedJSON(Exception):
     pass
+
 
 class DataIO():
     def __init__(self):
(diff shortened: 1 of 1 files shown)


Twentysix26:
The idea was to specify json (in save_json, load_json, etc) in case we ever wanted to add other data related functionalities to the class since, well, it's called DataIO. Also for clarity too.

You also have to fix fileIO, you keep passing the filename.

I was also thinking to remove the backup stuff. Json corruption issues were (I'm fairly sure at this point) happening due to the fact that json was being saved incorrectly and we have since fixed it

Oh and this is going to break existing cogs, although I'm not sure if anyone beside me was importing dataio yet

tekulvw:
JSON is obviously gonna be the main use case. I don't see any reason why in the future we can't add `save_csv()` and have `save()` be JSON/dict implicit. I'd like to save those 10 chars on load/save if at all possible.

I was importing dataIO but tbh, those few who used it can make the change, it's not a huge fix

irdumbs:
I like having `save()` be JSON as default.
how about a kwarg to specify a different format instead?

Twentysix26:
I prefer save_json, load_json and is_valid_json, it's more clear that way.  
It's used in multiple parts of the bot now anyway. Add a reference to the class as dataIO, since that's what they import and it should be :ok_hand:

tekulvw:
I really don't wanna add that dataIO legacy stuff. You and I right now are the only one's who've used it afaik and it's not a hard fix.

Twentysix26:
I don't see what's the issue in doing a `dataIO = DataIO`. It will be just a reference and it avoids changing stuff left and right (I haven't used that only in the bot itself, so I'd have even more stuff to change).

PR #310, opened 2016-06-18
I mean !set stream

irdumbs:
new PR from #300
game_name instead of status as per discussion

### DIFF
--- a/cogs/owner.py
+++ b/cogs/owner.py
@@ -284,6 +284,27 @@ class Owner:
             await self.bot.change_status(None)
             log.debug('status cleared by owner')
         await self.bot.say("Done.")
+        
+    @_set.command(pass_context=True)
+    @checks.is_owner()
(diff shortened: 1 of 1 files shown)


cogs/owner.py
```diff
@@ -284,6 +284,27 @@ def __unload(self):
...
         await self.bot.say("Done.")
+        
+    @_set.command(pass_context=True)
+    @checks.is_owner()
+    async def stream(self, ctx, stream_name=None, *, game_name=None):
+        """Sets Red's streaming status
+
+        Leaving both stream_name and game_name empty will clear it."""
+
+        if status:
```
tekulvw: Uh...you might've missed a few references

irdumbs:
annd we're to 
[streamer] [stream_title]
stream_title because discord picks up the stream title, not the game.

jaw3l:
what about YouTube streams? is this possible?

irdumbs:
discord doesn't support that atm

PR #313, opened 2016-06-18
Economy rewrite

Twentysix26:
Server based accounts and settings. Added global leaderboard (could be prettier though)

Bank class has been redone with hooks for 3rd party cogs.
The old ones have been removed, there was no way to keep backwards compatibility.

Since the bank.json format changed, once a user registers a bank account on a server if they're already present in the old format they'll keep their old balance.

Requires a bit of testing. Backup your jsons :^)

### DIFF
--- a/cogs/economy.py
+++ b/cogs/economy.py
@@ -41,7 +41,11 @@ class Bank:
     def create_account(self, user):
         server = user.server
         if not self.account_exists(user):
-            acc = {"name" : user.name, "balance" : 0}
+            if user.id in self.accounts: # Legacy account
+                balance = self.accounts[user.id]["balance"]
+            else:
+                balance = 0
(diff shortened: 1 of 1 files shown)


tekulvw:
Can you change the logger to `"red.economy"`? That way output goes to terminal, red.log, and economy.log. `__name__` would be good too (it'd set itself to `cogs.economy`) and we could set up a cog-only logger so red.log isn't as polluted?

Kowlin:
K I did some testing and I found the following problems.

No commands on `economyset`, a blank use `help`
No whitespace on slots to take compact mode in account.
Admins cannot set user credits. Something the old/current Economy has.
No help text/docstring on Leaderboard

Kowlin:
gave this another 4 AM try, and seems perfect to me

PR #314, opened 2016-06-21
Backups for realsies

Kowlin:


tekulvw:
@Twentysix26 I vote merge

Twentysix26:
This is not an oversight. This function was made to try and tackle the json corruption issues. I didn't know the reason at the time and the only likely cause I found were crashes during saving. So I figured that if the bot  crashed it would never reach the copy part.  

This is all useless. The corruption issues were due to the fact we were saving with dumps instead of dump and I was planning to remove this backup feature since that has been fixed

Kowlin:
Planned and never got around to it huh :P

Twentysix26:
I'll do it after Will's PR. No rush

PR #320, opened 2016-06-28
Added a new cog for rolling multiple dice

BillKing94:
As requested, I've moved the "multiple dice" change into a separate cog.

Usage: !mroll [nSides=100] [nDice=1]

### DIFF
--- a/cogs/mroll.py
+++ b/cogs/mroll.py
@@ -1,0 +1,34 @@
+import discord
+from discord.ext import commands
+from random import randint
+
+class MRoll:
+    """Roll multiple dice"""
+
+    def __init__(self, bot):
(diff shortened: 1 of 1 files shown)


irdumbs:
oh snap I never actually realized Red had built-in dice.
I thought you were trying to add a **new** feature :P
I guess it still stands.

What I meant by adding a cog was adding the cog to your own 3rd party repo, not as a permanent addition to the bot. Here is the link on the subject: http://twentysix26.github.io/Red-Docs/red_cog_how_to_publish/

You're missing out on a lot of the culture and conventions by not communicating with us on the Red server. You might want to consider mingling a bit.

PR #324, opened 2016-07-02
Change predefined prefix to use the bot command prefix

phantium:
Changed predefined ! prefix to use the bot command prefix and removed the prefix from the docstrings.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -1111,7 +1111,7 @@ class Audio:
 
     @commands.command(pass_context=True, no_pm=True)
     async def pause(self, ctx):
-        """Pauses the current song, `!resume` to continue."""
+        """Pauses the current song, `resume` to continue."""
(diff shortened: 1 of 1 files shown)


Twentysix26:
Would be better to use `ctx.prefix`, as it shows the prefix used

phantium:
I only used `self.bot.command_prefix[0]`, because it was used already in the _play function and I wanted to keep it consistent throughout the code.
Do you want me to change that one as well or just this one?

tekulvw:
You don't have ctx in _play iirc so no, you won't be able to make that change, however the first two you removed ! and didn't add the context specific prefix back?

phantium:
Those are docstrings, mind you.

phantium:
I modified the prefix to ctx.prefix and added [p] to the command examples so it's clear they need to specify a prefix.

PR #332, opened 2016-07-15
Added felxibility to Shutdown command

Uzendayo:
Added the possibility to set a shutdown message and the option to toggle
if it will show or not before the shutdown.


Twentysix26:
This just shows a preset message in the channel in which the owner does `[p]shutdown`. Frankly I don't see any case where this could be useful.

Uzendayo:
I'll improve it then, _My_ _Lord!_

PR #340, opened 2016-07-28
Audio: Add voting to skip command and perms to stop

calebj:
If a user isn't a mod, admin or owner, they must vote to skip. The threshold percentage is set by a mod/user/admin, and must be met by non-bot votes before the song is skipped. Additionally, permission to stop is limited to moderators if the user is not alone in the voice channel.


calebj:
I need testing!

Twentysix26:
Can you rebase? alias.py is showing up in changes

calebj:
Sorry about that; rebased and squashed.

PR #341, opened 2016-07-28
Checks for ffmpeg and git in PATH

Twentysix26:
Needs to be tested on various Windows systems and maybe needs a more helpful message in the git check. I'll just leave it here in the meantime.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -461,7 +461,7 @@ class Audio:
 
     # TODO: _enable_controls()
 
-    # returns list of active voice channels 
+    # returns list of active voice channels
     # assuming list does not change during the execution of this function
     # if that happens, blame asyncio.
     def _get_active_voice_clients(self):
@@ -1876,6 +1876,19 @@ class Audio:
                 vc.audio_player.resume()
 
 
+def check_ffmpeg():
+    """Windows only"""
+    files = ("ffmpeg.exe", "ffplay.exe", "ffprobe.exe")
+    if os.name == "nt":
+        if "ffmpeg" in os.environ['PATH']:
+            return True
+        elif all([os.path.isfile(f) for f in files]):
+            return True
+        else:
+            return False
+    return True
+
+
 def check_folders():
     folders = ("data/audio", "data/audio/cache", "data/audio/playlists",
                "data/audio/localtracks", "data/audio/sfx")
@@ -1907,6 +1920,8 @@ def check_files():
 
 
 def setup(bot):
+    support_page = "https://twentysix26.github.io/Red-Docs/"\
+                   "red_guide_troubleshooting/#installation-issues"
     check_folders()
     check_files()
     if youtube_dl is None:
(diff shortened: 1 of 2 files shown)


Kowlin:
Right now I cannot get this to work, Git is installed while Ffmpeg isn't in the directory.

As far as I can see without digging through the code I don't think its even checking for ffmpeg.

cogs/audio.py
```diff
@@ -1876,6 +1876,19 @@ def voice_connected(self, server):
                 vc.audio_player.resume()
 
 
+def check_ffmpeg():
+    """Windows only"""
+    files = ("ffmpeg.exe", "ffplay.exe", "ffprobe.exe")
+    if os.name == "nt":
+        if "ffmpeg" in os.environ['PATH']:
```
Kowlin, line 1883: I rewrote from line `1883` to the following,

``` python
        try:
            call(['git.exe'], stdout=DEVNULL, stderr=DEVNULL)
            return True
        except:
            return False
```

The only thing that needs to be done besides this is importing `DEVNULL` in `Subprocess`

Fr6jDJF:
with C:\ffmpeg\bin ; in PATH Windows failed on me
:white_check_mark: C:>ffmpeg -version
:x: C:>ffplay -version
:x: C:>ffprobe -version

:white_check_mark: C:\ffmpeg\bin\ffplay -version
:white_check_mark: C:\ffmpeg\bin\ffprobe -version

A reboot solved this, but the code gave me true while it was not available for python
 :white_check_mark: https://github.com/Twentysix26/Red-DiscordBot/pull/341 returned True :warning: 
 :white_check_mark: subprocess.call(' ffmpeg -i "{0}" -acodec flac "{1}" ', shell=True)
 :x:subprocess.Popen("ffprobe", some_file, stdout = subprocess.PIPE, stderr = subprocess.STDOUT) ([WinError 2] not found)

When adding ffprobe in some data dir
:white_check_mark: subprocess.Popen("data/freesound/ffprobe", some_file, stdout = subprocess.PIPE, stderr = subprocess.STDOUT) 

The code below confirmed my not available status

``` python
def shutil_ffmpeg():
    """Windows only"""
    files = ("ffmpeg", "ffprobe", "ffplay", "git")
    if os.name == "nt":
        # Return the path to an executable which would be run if the given cmd was called. If no cmd would be called, return None.
        # When no path is specified, the results of os.environ() are used, returning either the “PATH” value or a fallback of os.defpath.
        if not all([shutil.which(f) for f in files]): 
            return False
        else:
            return True
    else:
        return True
```

Kowlin:
Oh look who is being fancy with the check marks :P

ghost:
Is there reason to not do this check on linux as well? if checking on linux is fine, it should simplify down to

``` python
def on_path(*executables):
    return all(shutil.which(f) for f in executables)
```

``` python
    if not on_path('ffmpeg','ffprobe','ffplay'):
        raise RuntimeError("You need ffmpeg to play audio.\n{}"
                           "".format(support_page))
```

``` python
    if not on_path('git'):
        raise RuntimeError("You don't have git in your PATH environment variable\n"
                           "That's necessary for the downloader to run\n")
```

`if not on_path` also reads really cleanly imo

ghost (approved):
Just trying something?

Twentysix26:
This was mostly a proof of concept, will come up with something better.

PR #343, opened 2016-07-30
Prefix search method replacement.

Aioxas:
Replacing a function to get bot prefix with a feature that's already in the bot dependencies.


Kowlin:
This limits to 1 prefix rather then expanding on content.

Closing.

PR #352, opened 2016-08-18
Add autodownload of ffmpeg for Windows users

palmtree5:
This PR adds automatic checking for and downloading of the ffmpeg files needed for audio functionality on Windows. The added function is called only if `sys.platform` is win32. Yes, it does use requests


palmtree5:
Right, closing per 56da31527e46a6f39e746ce75a44cb4c9df0dde7

PR #356, opened 2016-08-21
Add update check on bot startup

palmtree5:
Checks for newer commits on github and messages the bot's owner if there is a mismatch (i.e. a newer version)


red.py
```diff
@@ -97,6 +99,33 @@
...
+async def update_check():
+    if settings.owner == "id_here":
+        print("No owner set!")
+        return
+    owner = discord.utils.get(bot.get_all_members(), settings.owner)
+    try:
+        async with aiohttp.get("https://api.github.com/repos/Twentysix26/Red-DiscordBot/commits") as r:
+            gh_commits = await r.json()
+        latest_sha = gh_commits[0]["sha"][:7]
+        cur_version = os.popen(r'git show -s HEAD --format="%cr|%s|%h"')
```
tekulvw, line 116: Even though this should block for nearly no time at all, it's still bad practice. Check the `version` command in the Owner cog for a better (async) way to do this.

palmtree5:
Referring to https://github.com/Twentysix26/Red-DiscordBot/blob/develop/cogs/owner.py#L526 presumably?

tekulvw:
Yep

palmtree5:
So new commit. Sadly this functionality now uses requests as I was able to get the pulling of data from Github's API to work using requests but got [this](http://i.imgur.com/z93N19S.png) when trying to use aiohttp

tekulvw:
`run_in_executor` apparently makes whatever function call you supply it into a separate thread so no worries on the requests usage

Twentysix26 (changes requested):
I haven't yet decided if I should implement this but requests is a no go

palmtree5:
Ok so got it working with `aiohttp` (although it points to this repo regardless of whether it's a fork or not)

Twentysix26 (approved):

Twentysix26:
Outdated and with various issues. Will implement in some other way if needed

palmtree5:
Lol I meant to close this with a note that this would be superseded by #533

PR #360, opened 2016-08-27
Condense slot result to 1 message

YeahImRose:
As a note I don't know whether or not this will work, I wrote it on my phone. Hopefully it does work though :)

### DIFF
--- a/cogs/economy.py
+++ b/cogs/economy.py
@@ -434,6 +434,7 @@ class Economy:
             slotMsg = "{}{} Nothing! Lost bet. ".format(display_reels, message.author.mention)
             self.bank.withdraw_credits(message.author, bid)
             slotMsg += "\n" + "Credits left: {}".format(self.bank.get_balance(message.author))
+            await self.bot.send_message(message.channel, slotMsg)
             return True
         self.bank.deposit_credits(message.author, bid)
         slotMsg += "\n" + "Current credits: {}".format(self.bank.get_balance(message.author))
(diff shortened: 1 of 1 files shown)


tekulvw:
Looks good, haven't personally tested it though

cogs/economy.py
```diff
@@ -407,31 +407,31 @@ def already_in_list(self, accounts, user):
         display_reels += ">" + reels[0][1] + " " + reels[1][1] + " " + reels[2][1] + "\n"
         display_reels += "  " + reels[0][2] + " " + reels[1][2] + " " + reels[2][2] + "\n"
 
-        slotMsg = ""
+        slotMsg = "\n"
```
Fr6jDJF: I don't know if the blame is on Discord it self, but starting newlines will get stripped/ignored now days so it seems.
Wrapping some random markup around it for example, allows you to start with newlines.
`~~\n~~`

https://github.com/Twentysix26/Red-DiscordBot/pull/236
economyc

cogs/economy.py
```diff
@@ -407,31 +407,31 @@ def already_in_list(self, accounts, user):
         display_reels += ">" + reels[0][1] + " " + reels[1][1] + " " + reels[2][1] + "\n"
```
Canule: economyc

cogs/economy.py
YeahImRose: That's very strange... I'll make sure to keep it in mind in the future

PR #363, opened 2016-08-29
Downloader: smarter updating, track new, updated and deleted cogs

calebj:
The following changes were made:
- Show status of running update
- Show new, updated and deleted cogs
- Use git hash to determine if update is necessary
- Remove deleted cogs from the database

### DIFF
--- a/cogs/downloader.py
+++ b/cogs/downloader.py
@@ -195,8 +195,10 @@ class Downloader:
 
         msg = await self._robust_edit(msg, base_msg + status)
 
+        registry = dataIO.load_json("data/red/cogs.json")
+
         for repo, cog in updated_cogs:
-            if self.repos[repo][cog]['INSTALLED']:
+            if (self.repos[repo][cog]['INSTALLED'] and registry.get('cogs.' + cog, False)):
                 installed_updated_cogs.append((repo, cog))
                 await self.install(repo, cog)
 
--- a/cogs/downloader.py
+++ b/cogs/downloader.py
@@ -365,7 +365,8 @@ class Downloader:
         for cog in new & old:
             self.repos[name][cog].update(valid_cogs[cog])
         for cog in old - new:
-            del self.repos[name][cog]
+            if cog != 'url':
+                del self.repos[name][cog]
 
     def update_repo(self, name):
         if name not in self.repos:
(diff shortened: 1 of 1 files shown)


cogs/downloader.py
```diff
@@ -150,27 +149,73 @@ def save_repos(self):
...
+                for k, l in updates.items():
+                    if k == 'A':
+                        new_cogs.extend([(name, c) for c in l])
+                    elif k == 'D':
+                        deleted_cogs.extend([(name, c) for c in l])
+                    elif k == 'M':
+                        updated_cogs.extend([(name, c) for c in l])
+
+            status = ' %d/%d repos updated' % (tasknum, num_repos)
+            await self.bot.edit_message(msg, base_msg + status)
```
tekulvw: Use a try block to check for message existence, juuust in case

cogs/downloader.py
```diff
@@ -150,27 +149,73 @@ def save_repos(self):
...
                                                  author=ctx.message.author)
         if answer is None:
             await self.bot.say("Ok then, you can reload cogs with"
                                " `{}reload <cog_name>`".format(ctx.prefix))
         elif answer.content.lower().strip() == "yes":
-            for (repo, cog) in installed_user_cogs:
+            update_list = []
+            for (repo, cog) in installed_updated_cogs:
+                update_list.append(cog)
                 self.bot.unload_extension("cogs." + cog)
```
tekulvw: I'd definitely surround lines 215-216 in a try block (and possibly think about calling `reload` from Owner instead of doing this directly)

calebj:
Changes made. Not sure about going through owner.py, will look into it further.

Twentysix26 (changes requested):
`[p]cog update` seem to be blocking. It kills the bot with enough repos.

calebj:
Only one task is run at a time now, instead of using `asyncio.gather()` which calls them all at once. Blocking is the only way to ensure the repos are updated before a) checking for new/deleted/updated cogs, and b) reloading updated cogs. The naive method of forking them off and waiting ten seconds is just that, and won't work with my change detection.

Twentysix26 (approved):

Twentysix26:
https://github.com/Twentysix26/Red-DiscordBot/pull/363

PR #364, opened 2016-08-30
Update streams.py: Added Youtube Checks

mkilp:
Heya,

as @Kowlin told me here is my PR.

Added ability to check for youtube livestreams.

(P=prefix)

P youtube <channelid>
P streamalert youtube <channelid>

before this works enter your google API key with:

P streamset youtubetoken <API KEY>

(Optain via https://developers.google.com/youtube/v3/getting-started)

Example usage:
-youtube UCo2IHxtzEKgI8Wc9xN4irXQ 

enjoy. im open for questions :)


franc-ist:
The command `[p]youtube` is already in use in other cogs. May I suggest `[p]ytstream` instead?

cogs/streams.py
```diff
@@ -55,6 +58,21 @@ def __init__(self, bot):
             await self.bot.say("Error.")
 
     @commands.command()
+    async def youtube(self, channelid: str):
+        """Checks if youtube stream is online"""
+        channelid = escape_mass_mentions(channelid)
+        online = await self.youtube_online(channelid)
+        if online is True:
+            await self.bot.say("https://gaming.youtube.com/channel/{} "
```
franc-ist, line 66: Why only the gaming site? Why not other livestreams?

mkilp: I could easily change it to `https://youtube.com/channel/{}`.

Other names for the command could be [p]ytlive or [p]youtubelive

Kowlin: Personally I think its better to make this into another command. This will avoid most problems that will arise and will be easier to understand for users.

mkilp: So what is the point now? 
Do I need to change anything to make this able to merge? :)

Kowlin: Hmm, Irdumb has suggested a command group. But I think that would be to much work for now, The safest best would be to simply rename the command.

mkilp:
changed the command from [p]youtube to [p]ytlive

is that good now? (:

cogs/streams.py
```diff
@@ -119,6 +137,50 @@ def __init__(self, bot):
...
+    async def youtube_alert(self, ctx, stream: str):
+        """Adds/removes youtube alerts from the current channel"""
+        stream = escape_mass_mentions(stream)
+        channel = ctx.message.channel
+        check = await self.youtube_online(stream)
+        if check is None:
+            await self.bot.say("That stream doesn't exist.")
+            return
+        elif check == "error":
+            await self.bot.say("Couldn't contact Twitch API. Try again later.")
```
Tarkoviski, line 150: Twitch?

caustinbrooks:
Any updates on this? It would be quite useful to have

palmtree5:
Closing due to repo not existing anymore

PR #365, opened 2016-08-30
Add question() to chat_formatting.

tmercswims:
One more that I hadn't thought of before. Used the correct format this time!


Twentysix26:
That doesn't look like valid unicode. Have you tested this?

[Reference](http://unicode.org/emoji/charts/emoji-list.html)

tmercswims:
You are absolutely right. I did a dumb and wrote the wrong thing. It is correct now.

PR #372, opened 2016-09-12
[DataIO] Atomically save json

Twentysix26:
This should put an end to the corruption issues. Requires testing on different platforms


Kowlin:
What platforms do you need testing for?

Twentysix26:
All of them! Tested on two Windows machines so far, seems fine. Trying on Ubuntu soon. Don't really see a reason why it wouldn't work but I want to be extra sure since it's critical

PR #374, opened 2016-09-13
Add an option to specify how long the bot lingers after playing audio

alexwh:
Quick PR for some functionality I've been missing in Red.
I haven't extensively reviewed the codebase, but it works as far as I can see so let me know if anything's wrong. :)


alexwh:
Forgot one place to add the new setting - the value is now initialized properly on new servers.

Twentysix26:
I don't think it's really needed.

PR #377, opened 2016-09-15
More accurate detection of twitch account

4o66:
Slightly more accurate method of determining if a twitch streamer exists.
Uses official API endpoint

### DIFF
--- a/cogs/streams.py
+++ b/cogs/streams.py
@@ -316,17 +316,18 @@ class Streams:
         return "error"
 
     async def twitch_exists(self, stream):
-        url = "https://api.twitch.tv/channels/" + stream
+        url = "https://api.twitch.tv/kraken/channels/" + stream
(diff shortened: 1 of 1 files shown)


cogs/streams.py
```diff
@@ -316,17 +316,18 @@ def __init__(self, bot):
...
-                return False
-            else:
+                data = await r.json()   
+            if len(data["display_name"]) > 0:
                 return True
+            else:
+                return False
         except:
             return "error"
+        return "error" CHECK_DELAY = 60
```
Twentysix26: What happened here? :^)

4o66: It got late, and I got sloppy with copy/paste.
Fixed

Twentysix26:
https://github.com/Twentysix26/Red-DiscordBot/commit/2dcb1c88d10dd1d1713527f9ca3c811b6aa3fb0b

PR #325, opened 2016-07-03
Output to Music channel what song is playing.

DevScyu:
Output to Music channel what song is playing.

Issue #284, opened 2016-06-01
Notify channel when song changes?


19lmyers:
This is somewhat related to the queue list feature.

I was wondering if the bot could announce in a channel when it begins playing a queued song, because I'm not always sure if the song has changed.

irdumbs:
we may make it a toggleable option

DevScyu:
You mean something like this: https://github.com/Twentysix26/Red-DiscordBot/pull/325

tekulvw:
Moving to #623

PR #379, opened 2016-09-16
Send notification to channel when song changes

gizmo385:
This is in response to #284 and similar to the change proposed in #325. However, this change should work no matter what channel the bot is in.

This uses an !audioset option, "notify", to determine whether or not to send the message. By default, this notify option is set to False.

It required storing the text channel that the play commands are initiated so that the bot can send the message to the right channel.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -1175,6 +1175,7 @@ class Audio:
         server = ctx.message.server
         author = ctx.message.author
         voice_channel = author.voice_channel
+        self.playing_text_channel = ctx.message.channel
 
         # Checking already connected, will join if not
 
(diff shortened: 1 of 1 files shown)


ghost:
perhaps False for no notification, True for notification to the channel the song was requested from, and channel ID for a specific channel?

gizmo385:
That's definitely a possible thing that could be added. Anyone else have any thoughts on it?

Eslyium:
So after testing this, And Discord started to freak out or something else occurred it started relaying now playing to all servers. Might be a problem.

gizmo385:
Hmm that's odd. I'll look into it and see if I can replicate the issue. Thanks for the heads up!

cogs/audio.py
```diff
@@ -942,6 +942,24 @@ def _stop_player(self, server):
             else:
                 await self._remove_song_status()
 
+    # no return. they can check themselves.
+    async def _notify_song_name(self):
+        if self.settings.get("NOTIFY"):
```
tekulvw, line 947: This should really be a per server configurable option so all servers can have their song announced on change.

tekulvw (changes requested):

xdtrammell:
Is this branch going to be merged I would really like this feature.

PR #387, opened 2016-09-21
[utils] checks: properly handle empty permissions, add simple decorators

calebj:
`@checks.admin_or_permissions()` will always return true, because `any([]) = True`.

This commit fixes that (returns `False` for empty kwargs) and adds:  
- `@checks.admin()`: allows if user has server admin role (or is bot's owner)
- `@checks.mod()`: allows if user has server mod role (or is bot's owner)
- `@checks.serverowner()`: allows if user is server owner (or is bot's owner)

These new decorators are basically wrappers for their `or_permissions` counterparts, but take no arguments.

### DIFF
--- a/cogs/utils/checks.py
+++ b/cogs/utils/checks.py
@@ -28,6 +28,8 @@ def is_owner():
 def check_permissions(ctx, perms):
     if is_owner_check(ctx):
         return True
+    elif not perms:
+        return False
 
     ch = ctx.message.channel
     author = ctx.message.author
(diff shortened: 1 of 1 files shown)


cogs/utils/checks.py
```diff
@@ -30,6 +30,8 @@ def is_owner():
 def check_permissions(ctx, perms):
     if is_owner_check(ctx):
         return True
+    elif not perms:
```
calebj, line 31: `bool({})` is `False`, but if the dictionary contains any keys, regardless of value, `bool(d)` is `True`

Kowlin (approved):
Looks good on my end, tested and confirmed working

tekulvw:
Let's bump this

Jan200101 (approved):

tekulvw (approved):

PR #388, opened 2016-09-22
[core] add exit code 1 on abnormal stop

calebj:


calebj:
Bump. This exit code allows a daemon manager such as supervisor, systemd, or even a simple `bash` loop to restart the bot only on error conditions.

PR #389, opened 2016-09-22
Fix issue with cleanup user and cleanup text

YeahImRose:
I have tested this and it seems to work very well. the one thing I'm unsure of is what the limit for checking logs should be set to, so I would appreciate if someone gave their thoughts on it.

*these limits are found on lines 164 and 223


YeahImRose:
;C

PR #403, opened 2016-10-05
Legacy FileIO to DataIO migration

Redjumpman:
I've made all cogs that use fileIO now use dataIO, with the exception of Audio. This is only because I don't want to mess with audio, and it's in active development.


Twentysix26:
Fixed some stuff. Settings.py and audio.py should be the only ones left

Redjumpman:
whoops. Sorry about that.

Twentysix26:
https://github.com/Twentysix26/Red-DiscordBot/pull/410

PR #407, opened 2016-10-08
Final Version of Red Supervisor

ccubed:
Twentysix told me people wanted a built in process manager that would restart redbot automatically, I told him that could be done no problem. Thus was the Red Supervisor born. Behold, for it arrives. This uses absolutely 0 external imports, so it will work for anyone with python 3.5+.

Features:
1. Automatically check for git in path, don't start if it's not there
2. Automatically check for ffmpeg in path. On windows we'll download this automatically. 64 bit we have predefined links in this repo, on 32 bit it depends on being able to parse out the link on the ffmpeg builds page.
3. Automatically git pull before starting redbot, so you're always up to date.
4. It does not update discord.py, because that can cause more issues than good if automated.
5. It will automatically start and restart redbot as long as the return code is not 0. 0 indicates a normal return (IE: you used shutdown in discord), so it's assumed that you actually want redbot to die in that case.
6. You can kill this supervisor without killing redbot. They're separate entities entirely, though the error output from redbot will show up here. That can be changed if we want: It could go to a file for instance, or even just stay within redbot.
7. Tested extensively on Linux and tested a little bit on windows. Seems to work pretty well. Though I could do with some more windows testing that I'll do tomorrow. I'm confident enough to release the PR.

### DIFF
--- a/supervisor.py
+++ b/supervisor.py
@@ -90,8 +90,7 @@ if __name__ == "__main__":
 
     # Try to pull
     print("We're going to update your instance of redbot now.")
-    updates = subprocess.run(['git', 'pull'])
-    if updates.wait() != 0:
+    if subprocess.run(['git', 'pull']).returncode:
         print("Error updating Redbot. Please investigate.")
         sys.exit(0)
     else:
--- a/supervisor.py
+++ b/supervisor.py
@@ -6,6 +6,7 @@ import os
 import subprocess
 import sys
 import urllib.request
+from functools import partial
 from html.parser import HTMLParser
 from zipfile import *
 
@@ -50,50 +51,6 @@ def check_for_ffmpeg():
         return False
 
 
-class ProcessManager:
-    """
-    This class manages monitoring and restarting the process when it dies.
-    """
-    def __init__(self):
-        """
-        Initialize the class.
-        :ivar cmd: Holds the executable path for the current python interpreter running this supervisor. This will also be used to run the bot.
-        :ivar process: Holds the current subprocess popen reference
-        """
-        self.cmd = sys.executable
-        self.process = None
-
-    def start(self):
-        """
-        Start the subprocess and separate it to a new process group so they can run independently.
-        :return: True or False
-        """
-        if os.name == 'nt':
-            try:
-                self.process = subprocess.Popen([self.cmd, 'red.py'],
-                                                stdin=subprocess.DEVNULL,
-                                                stdout=subprocess.DEVNULL,
-                                                stderr=subprocess.STDOUT,
-                                                creationflags=subprocess.CREATE_NEW_PROCESS_GROUP)
-                return True
-            except Exception as e:
-                return False
-        else:
-            try:
-                self.process = subprocess.Popen([self.cmd, 'red.py'],
-                                                stdin=subprocess.DEVNULL,
-                                                stdout=subprocess.DEVNULL,
-                                                stderr=subprocess.STDOUT,
-                                                preexec_fn=os.setpgrp)
-                return True
-            except Exception as e:
-                return False
-
-    def wait(self):
-        self.process.wait()
-        return self.process.returncode
-
-
 if __name__ == "__main__":
     # Check for Git
     if not check_for_git():
@@ -105,15 +62,10 @@ if __name__ == "__main__":
         if not check_for_ffmpeg():
             if IS_X8664:
                 print("FFMPEG not found. Downloading.")
-                with urllib.request.urlopen("https://github.com/Twentysix26/Red-DiscordBot/raw/master/ffmpeg.exe") as data:
-                    with open("ffmpeg.exe", "wb") as f:
-                        f.write(data.read())
-                with urllib.request.urlopen("https://github.com/Twentysix26/Red-DiscordBot/raw/master/ffplay.exe") as data:
-                    with open("ffplay.exe", "wb") as f:
-                        f.write(data.read())
-                with urllib.request.urlopen("https://github.com/Twentysix26/Red-DiscordBot/raw/master/ffprobe.exe") as data:
-                    with open("ffprobe.exe", "wb") as f:
-                        f.write(data.read())
+                for item in ['ffmpeg.exe', 'ffprobe.exe', 'ffplay.exe']:
+                    with urllib.request.urlopen("https://github.com/Twentysix26/Red-DiscordBot/raw/master/{}".format(item)) as data:
+                        with open(item, "wb") as f:
+                            f.write(data.read())
                 print("FFMPEG downloaded.")
             else:
                 parser = LinkParser()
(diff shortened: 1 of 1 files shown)


supervisor.py
```diff
@@ -0,0 +1,158 @@
...
+    :return:  True or False
+    """
+    try:
+        subprocess.check_output(['ffmpeg', '-version'])
+        return True
+    except Exception as e:
+        return False
+
+
+class ProcessManager:
```
ghost: Isn't a class a bit overkill for

``` python
def run_redbot():
    if os.name == 'nt':
        return subprocess.run(...
    else:
        return subprocess.run(... # run will wait on the process, and will error on non-zero return
```

``` python
while True:
    try:
        run_redbot()
    except Exception as e:
        break
```

though that does limit the supervisor/bot separation., though `Popen` with a `.wait()` on the returned process also works (in which case, `.wait()` will return the returncode, instead of needing to check that separately)

supervisor.py
```diff
@@ -0,0 +1,158 @@
...
+
+    def handle_starttag(self, tag, attrs):
+        if tag == 'a':
+            for attr in attrs:
+                if attr[0] == 'href':
+                    if 'win32-static.zip' in attr[1]:
+                        self.static_download = attr[1]
+
+
+def check_for_git():
```
ghost, line 17: https://docs.python.org/3/library/shutil.html#shutil.which would simplify this a tad, >3.3

``` python
import shutil
def on_path(*executables):
    return all(shutil.which(executable) for executable in executables)
```

supervisor.py
```diff
@@ -0,0 +1,158 @@
...
+    if not check_for_git():
+        print("Please install GIT to properly use this supervisor.")
+        sys.exit(0)
+
+    # Check for FFMPEG
+    if os.name == 'nt':
+        if not check_for_ffmpeg():
+            if IS_X8664:
+                print("FFMPEG not found. Downloading.")
+                with urllib.request.urlopen("https://github.com/Twentysix26/Red-DiscordBot/raw/master/ffmpeg.exe") as data:
```
ghost: probably worth using a loop here over `("ffmpeg.exe", "ffplay.exe", "ffprobe.exe")`

supervisor.py
ccubed: Popen is being used specifically because we need to separate the process groups. If we don't, then when the supervisor is killed, redbot is going to be killed too. This isn't the desired behavior and with run you cannot use the advanced functionality like preexec_fn or creationflags to ensure that the processes have different groups. As for it being class, it wraps holding the python executable path and the current subprocess. We don't need to do that as a class, but five lines to manage a process looks better than moving all of that out of a class and stuffing it into the main block.

supervisor.py
ccubed: shutil.which only verifies that the executable exists, not that it runs for the current user. Using check output actually attempts to run the command, so we can be sure that if it passes this test it will actually run later on.

supervisor.py
ccubed: I didn't actually look to see if those links were the same besides the programs. Yeah, those could be a loop.

supervisor.py
ghost: > mode is a permission mask passed to os.access(), by default determining if the file exists and executable.

with a default of F (existance) and X (executability) permissions

and for os.access

> Use the real uid/gid to test for access to path.

so it _should_ check the executability as well for the current user, I could be misunderstanding what I'm reading though

supervisor.py
```diff
@@ -138,7 +90,7 @@ def wait(self):
 
     # Try to pull
     print("We're going to update your instance of redbot now.")
-    updates = subprocess.Popen(['git', 'pull'])
+    updates = subprocess.run(['git', 'pull'])
     if updates.wait() != 0:
         print("Error updating Redbot. Please investigate.")
         sys.exit(0)
```
ghost, line 38: could go with

``` python
if subprocess.run(['git', 'pull']).returncode:   # since .run does .wait automatically
    print("Error updating Redbot. Please investigate.")
    sys.exit(1)    # maybe not exit code of zero if its an error?
```

supervisor.py
ccubed: That just checks to make sure that the thing in question has the platforms equivalent of +x. Not necessarily that running the file will complete successfully.

supervisor.py
ccubed: Yeah that's shorter, but exit status should be 0. Git had an error, but the manager itself didn't exit with an error.

Jan200101 (changes requested):
On Devices like the raspberry pi ffmpeg does not exist so it will always result in a please install ffmpeg message.

Also the exes referenced are no longer needed

ccubed:
Belatedly, it's entirely possible to have ffmpeg on embedded devices like a raspberry pi. I made this based off the current batch scripts in Red. Red does at the least require Git because that's how it updates and installs the addons. Voice on Discord.py requires ffmpeg. So I'm not sure what you mean by 'exes referenced are no longer needed.' What exes are no longer needed?

Jan200101:
Red on windows needs these executables to run the audio cog yes but in most cases its up to the user and shouldnt be defined by a script.

Then there is the problem that on line 86 - 89
`    else:
         if not check_for_ffmpeg():
             print("Please install FFMPEG and then run this script again.")
             sys.exit(0)`
It will prompt the user to install FFMPEG for their current device which is also bad practice as it again forces the user to download ffmpeg but it also takes the choice of using ffmpeg or avconv.

ccubed:
If you're intentionally using avconv over ffmpeg then you sound like the kind of user that wouldn't be using this kind of process manager to begin with. Also, stop using avconv.

Jan200101:
So it is then.
Some os even tell the user that ffmpeg is deprecate on that os and to use avconv but it is your choice

> *** THIS PROGRAM IS DEPRECATED ***
>This program is only provided for compatibility and will be removed in a future release. Please use avconv  instead.

none the less avconv is ffmpeg but differen and there is no reason to stop using avconv as some devices just run better with it

ccubed:
By some OS you mean Ubuntu and by deprecated it doesn't mean the program itself is deprecated, it means the command ffmpeg is deprecated because Ubuntu made its own commands because it's Ubuntu.

Fuyukai:
This is only for old Ubuntu.

```
eyes@veriny ~> cat /etc/lsb-release 
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=16.04
DISTRIB_CODENAME=xenial
DISTRIB_DESCRIPTION="Ubuntu 16.04.1 LTS"
eyes@veriny ~> apt show ffmpeg
Package: ffmpeg
Version: 7:2.8.8-0ubuntu0.16.04.1
Breaks: libav-tools (<< 6:12~~), qt-faststart (<< 7:2.7.1-3~)
Replaces: libav-tools (<< 6:12~~), qt-faststart (<< 7:2.7.1-3~)
```

Twentysix26:
Closing in favor of #552  
Thanks for this, as it gave me enough good ideas and inspiration to make the launcher 👍

PR #414, opened 2016-10-14
[Audio] FileIO to dataIO conversion

Aioxas:


Aioxas:
Messed up the first commit's name. Supposed to be the title. But anyways.

PR #415, opened 2016-10-14
[Downloader] Added removal of repo folders on repo removal.

Aioxas:
This should remove the repo folders in data/downloader apart from the original feature of removing it from the repos.json Further testing might be needed in order to ensure that no errors occur due to folder/file permissions.


tekulvw (approved):
Untested but looks good

tekulvw:
Please rebase

PR #417, opened 2016-10-16
Fix gitignore to include playlists properly

louy2:
Edit on Dec 5: Make `git status` show untracked trivia and playlists.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,10 @@
 *.json
 *.pyc
 __pycache__
-data
-!data/trivia/*
-!data/audio/playlists/*
+data/*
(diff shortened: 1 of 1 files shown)


tekulvw:
Is there anything actually wrong with the gitignore or is this just semantics?

louy2:
In `.gitignore`, sub-directories cannot be re-included if their parent directory is ignored, because git would not check the content of that directory at all for performance. The original version didn't implement its intent, of re-including the trivia and playlists directories. New playlists had to be explicitly added to be tracked, and otherwise doesn't show in `git status`.

PR #418, opened 2016-10-16
Freeze dependencies

louy2:
as http://www.kennethreitz.org/essays/a-better-pip-workflow


Twentysix26:
No point in doing that for the few dependencies we have.  
`youtube_dl` needs to be always up to date if you want audio to keep working.

Issue #425, opened 2016-10-26
Audio says it's playing, but it's actually not after a while.


ncodex2:
If the bot restarts, or runs for a long time, audio bugs up for some servers and says that a song is already playing but it's not. I have to do reload audio do fix this. Any idea what this might be?

ncodex2:
http://prntscr.com/czbewh This also seems to be happening on some servers.
This doesn't seem to be resolved once i do reload audio.

PR #430, opened 2016-10-31
[Audio] Fix 'Nonetype has no attribute "server"' issues

Aioxas:
Fixes 2nd part of issue [#425]

Caused by queuing a song using [p]play when the bot is in the voice channel, but the queuer isn't.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -575,7 +575,12 @@ class Audio:
         return self.queue[server.id]["PLAYLIST"]
 
     async def _join_voice_channel(self, channel):
-        server = channel.server
+        try:
+            server = channel.server
+        except AttributeError:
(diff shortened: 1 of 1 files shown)


Aioxas:
Woops, wrong PR closed.

tekulvw:
A check for this should already happen outside of this function because you can't just randomly use `bot.say` outside of a command function (and sometimes even within one).

Aioxas:
Seems like it doesn't. For some strange reason. I checked and rechecked and I was running into that error more than once. Only when I put that try except block, did it stop throwing errors and threw the message instead.

tekulvw:
It does happen, just one scope too far down. Check #440

Twentysix26:
https://github.com/Twentysix26/Red-DiscordBot/pull/440

PR #432, opened 2016-11-01
"Imgur" command changes

Jeddunk:
* Allow greater time window (than "day") if there isn't enough content (at least 3 images)
* Output every image with its title in a seperate message
* GIFV support

### DIFF
--- a/cogs/image.py
+++ b/cogs/image.py
@@ -39,11 +39,23 @@ class Image:
                 else:
                     await self.bot.say("Only top or new is a valid subcommand.")
                     return
-                items = imgurclient.subreddit_gallery(text[0], sort=imgSort, window='day', page=0)
-                if (len(items) < 3):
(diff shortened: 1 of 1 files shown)


Jeddunk:
Also, with these changes it might be easier to expand upon the command if needed (e.g. output more or less content depending on user input, metadata, nsfw filtering,...)

palmtree5:
Looks pretty good to me

Kowlin:
Seems to be broken when I tried it out, can anyone confirm that its broken or working?

palmtree5:
Gimme a bit and I'll test it

palmtree5:
Or instead of testing, I could be grossed out by the old tech in the imgur command and wind up rewriting the imgur command to use subcommands (see #500) :p

Twentysix26:
Closing in favor of https://github.com/Twentysix26/Red-DiscordBot/pull/500

PR #437, opened 2016-11-03
Pagify - shorten on the last page + when no delims

Chovin:
iirc, `shorten_by` was used so that you could do something like
```py
pg_fmt = "```py\n{}```"
for page in pagify(data, ['\n', ' '], shorten_by=len(pg_fmt.format(''))):
    await self.bot.say(pg_fmt.format(page))
```
so that the additional formatting would not exceed `page_length`.

iind, pages could exceed that length when no delims are found or on the last page


Chovin:
this PR does not take into account the fact that `escape` adds more characters and should be fixed also
#438

tekulvw:
Closed because #439

Issue #438, opened 2016-11-03
Pagify - escape param ignores page_length


Chovin:
implications: possible msgs over 2000

Chovin:
^ fixes this issue but now msgs with a bunch of @ here @ everyone will chew into the page length quite a bit

Kowlin:
Technically fixed? or can we wait for #543?

Chovin:
not fixed, the problem was just moved the other direction. wait for #543

Tobotimus:
This has been fixed for V3, are there plans to implement for V2?

tekulvw:
Stale

PR #439, opened 2016-11-03
Fix for #438

tekulvw:


Chovin:
this doesn't account for \\* and such.

Chovin:
shortening by more outside is a quick and dirty fix but what happens when you use pagify on a large document?
now page_length < 0

tekulvw:
I agree with defaulting delims but I see no reason for the catchall positional arg

Twentysix26:
To avoid monstrosities like
`pagify(text, ["\n"], False, 8, 2000)`
What is that bool? What are those 2 last ints? You know it because you made this function yourself but at  first glance it's unclear to say the least.  It forces cleaner code.  
I'd also appreciate if you don't undo changes that I made without first discussing it, thank you very much.

Twentysix26:
Merging. We'll fix what @Chovin reported in a second time

PR #444, opened 2016-11-04
Updated userinfo and serverinfo, added serverinfofull

Jan200101:
added user status to userinfo
added AFK to userinfo
added emoji count to serverinfo
added serverinfofull which displays all more info about the emojis names, role names and the channel names
added pagify to all 3

### DIFF
--- a/cogs/general.py
+++ b/cogs/general.py
@@ -1,10 +1,10 @@
 import discord
 from discord.ext import commands
-from .utils.chat_formatting import *
 from random import randint
 from random import choice as randchoice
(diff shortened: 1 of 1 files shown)


Jan200101:
Should be everything for now. 
If someone finds a bug in it tell me right away

Jan200101:
merged with update

Twentysix26:
I don't think that those commands are needed, they seem to give very redundant informations. I plan to add some more details to !serverinfo but that's it.

PR #445, opened 2016-11-04
[Mod] Delete commands after X seconds

tekulvw:
Configurable timed deleter, won't throw errors when forbidden either

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -59,11 +59,14 @@ class Mod:
             await send_cmd_help(ctx)
             roles = settings.get_server(server).copy()
             _settings = {**self.settings[server.id], **roles}
+            if "delete_delay" not in _settings:
+                _settings["delete_delay"] = -1
             msg = ("Admin role: {ADMIN_ROLE}\n"
                    "Mod role: {MOD_ROLE}\n"
                    "Mod-log: {mod-log}\n"
                    "Delete repeats: {delete_repeats}\n"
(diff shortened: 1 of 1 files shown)


xdtrammell:
How is this used,I updated the mod.py file with your commit just not sure what the command is? Thank you so much.

Edit: Thank you for your help on discord.

On top of this issue is there anyway this could be set up to be cleaned up as well. 
http://i.imgur.com/6YWbN9I.png

GostOmega:
@xdtrammell For the request of audio, I recommend to make a text channel for audio commands only, call it music_requests or music_room for example. In the topic of the channel I'd mention a daily wipe of the chat, so at the end of the day you simply do `[p]cleanup messages 9999` (anyone with permissions to manage messages can use the [p]cleanup command).

I use this method to keep certain commands out of general chat, like a slot_machine room for the economy cog, and music_requests for the audio. The mods and I clear chat at the end of the day in those two room, so it's not much of an hassle.

I also recommend utilizing an alias for the cleanup command, `[p]alias add clear cleanup messages` for example, so the results would be `[p]clear #`

Kowlin:
Works like a charm for me.

Twentysix26 (changes requested):
Docstring needs a fix, shows up weird in the help. The setting also needs to show up when you do [p]modset

Twentysix26 (approved):

PR #454, opened 2016-11-08
[customcom, alias] Selfbot compatibility v2

calebj:
* Customcom:
  - Don't ignore self directly, use `user_allowed()` from main
* Alias:
  - Same as customcom

Moving the `message.author==user.bot` check to user_allowed() allows all cogs that use it to avoid self-triggering now, as well as being triggered by another bot. This only applies to userbots, since bot accounts are matched by the existing `author.bot` check.

The purpose of moving it into red.py is to allow `user_allowed()` to be redefined centrally.

### DIFF
--- a/cogs/alias.py
+++ b/cogs/alias.py
@@ -108,8 +108,7 @@ class Alias:
                 await self.bot.say("There are no aliases on this server.")
 
     async def check_aliases(self, message):
-        if message.author == self.bot.user or \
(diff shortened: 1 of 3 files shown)


calebj:
I'm looking for comments and opinions on these changes

calebj:
Bleh, messed up indentation on red.py line 144. Fixed now.

PR #453, opened 2016-11-08
Empty channel check

palmtree5:
see also https://github.com/Twentysix26/Red-DiscordBot/pull/394

PR #456, opened 2016-11-08
stop playing on empty channel after 5 minutes

SomeFixItDude:
I removed the elif and changed the statement to an "or" comparison, you'll see below. This way if you disconnect reconnect due to network or accident you don't dump your whole queue. Also then moved the clear queue to the section where it preforms the disconnect. So basically if the bot is done playing or no one is in the channel for 5 minutes, its clears queue, stops, disconnects.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -1842,12 +1842,10 @@ class Audio:
                     stop_times[server] = int(time.time())
 
                 if hasattr(vc, 'audio_player'):
-                    if (vc.audio_player.is_done() or len(vc.channel.voice_members) == 1) and \
-                            (server not in stop_times or
-                             stop_times[server] is None):
-                        log.debug("putting sid {} in stop loop".format(
-                            server.id))
-                        stop_times[server] = int(time.time())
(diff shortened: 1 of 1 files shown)


SomeFixItDude:
Improvement from https://github.com/Twentysix26/Red-DiscordBot/pull/453

palmtree5:
Endorsing this over #453. Definitely a much better idea to wait until it's time to disconnect to stop the music rather than immediately on the bot being the only user in the voice channel

Twentysix26 (changes requested):
I assume you haven't tested it, [this](https://github.com/SomeFixItDude/Red-DiscordBot/blob/232b86357ea32cce69b6b8e936f9c7b421f6c2cd/cogs/audio.py#L1851) will equal to true in the next cycle (5 secs later) and the timestamp gets reset everytime.

SomeFixItDude (commented):
Thanks @Twentysix26 - Updated the logic and now will not fall through to the elif where the timer is reset

Twentysix26 (approved):

PR #461, opened 2016-11-10
[Mod] Allow bots to delete their own messages

tekulvw:
For whatever reason, the `mass_purge` endpoint absolutely requires the `manage_messages` permission so let's get around that.


calebj (changes requested):
It will still return an error if the bot doesn't have the `manage_messages` permission. You need to change line 485. Also, your logic forces it to use slow deletion even if it can use mass_purge. Just tweak the logic on line 485 to `if not (has_permissions or self_delete)` and line 505 to `if is_bot and has_permissions`

PR #463, opened 2016-11-11
[Utils] json helper classes

Twentysix26:
Since dealing only with DataIO is a pain in the 🍑 I made a couple of helper classes to streamline the process.  
Upon instancing it, it creates the file or loads it if it already exists. Saves automatically at every `.set()`, creates the intermediate folders of the file if requested, a default value can be set in the Server based class...  
I'm still not convinced on the naming and the structure ( @tekulvw proposed to join the two classes ) and I'm most likely gonna extend it, so suggestions are welcome

### DIFF
--- a/cogs/utils/json_helpers.py
+++ b/cogs/utils/json_helpers.py
@@ -1,21 +1,38 @@
 import os
+import discord
+from collections import defaultdict
 from cogs.utils.dataIO import dataIO
+
+GLOBAL_KEY = '__global__'
+
 
 class DataDB:
     """
     A helper class to streamline the saving of json files
     """
-    def __init__(self, file_path, *, create_dirs=False):
+    def __init__(self, file_path, *, create_dirs=False, default_value=None):
         self.path = file_path
 
-        if create_dirs:
-            self._create_dirs(file_path)
+        path_exists = os.path.isfile(file_path)
 
-        if os.path.isfile(file_path):
+        if create_dirs and not path_exists:
+            path, _ = os.path.split(file_path)
+            if path:
+                try:
+                    os.makedirs(path)
+                except FileExistsError:
+                    pass
+
+        if path_exists:
             self._data = dataIO.load_json(file_path)
         else:
             self._data = {}
             self._save()
+
+        if default_value is not None:
+            def _get_default():
+                return default_value
+            self._data = defaultdict(_get_default, self._data)
 
     def set(self, key, value):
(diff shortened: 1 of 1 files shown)


palmtree5:
So I could do something like:

``` python
from blah import blah
import bla
from .utils.json_helper import DataDB, ServerDB

class BlahBlah():
    def __init__(self, bot):
        self.bot = bot
        self.data = ServerDB("data/blahblah/kowlin.json", create_dirs=True)


def setup(bot):
    bot.add_cog(BlahBlah(bot))
```

That's just a very rough example but I can see how this could be useful

AraHaan:
There is a such thing as an import hook which can allow you to import a json file with it's file name. This could help as well.

Twentysix26:
Should `.get` allow for a default value? In case nothing is passed, it will either throw an exception or what the `defaultdict` is set as
Edit: That could be a thing for ServerDB's get too

cogs/utils/json_helpers.py
```diff
@@ -0,0 +1,132 @@
...
+    def set(self, server, key, value):
+        """Sets a server's entry"""
+        if not isinstance(server, discord.Server):
+            raise TypeError('Can only set server data')
+        if server.id not in self._data:
+            self._data[server.id] = {}
+        self._data[server.id][key] = value
+        self._save()
+
+    def get(self, server, key):
```
Twentysix26, line 92: Should this have a default? Actually, should *all* `.get_*` functions have a passable default argument?

AraHaan: Yes, it should in case no key is provided.

Also should make this for getting channel data as well if ``channel`` is not None. So I suggest doing something like

```py
    def get(self, server=None, channel=None, key=None):
        """Returns a server or channel's entry"""
        if key is not None:
            if channel is not None:
                if not isinstance(channel, discord.Channel):
                    raise TypeError('Can only get channel or server data')
                return self._data[channel.id][key]
            else:
                if server is not None:
                    if not isinstance(server, discord.Server):
                        raise TypeError('Can only get channel or server data')
                    return self._data[server.id][key]
                else:
                    raise TypeError('Expected a channel or a server object.')
        else:
            raise TypeError('Expected a key, got None instead.')
```

So that way people can tell the bot to explicitly do special things per channel on a server if they want to have it do certain things in certain channels on their server(s).

Twentysix26:
Replaced by config in the rewrite

PR #466, opened 2016-11-12
[mod] Add cleanup bot command, remove slow deletion delay

calebj:
also make `[p]cleanup after` bot-only, since it uses an API endpoint only available to bots.

The purpose of `[p]cleanup bot` is so mods and above can clean up bot messages in general, or ones that fit a pattern like payday or leveler.

As for removing the delay, there's no need for it since discord.py handles rate limits and the limit is only for deletions. In the worst case, deletion in other channels might be a few seconds slow.

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -6,6 +6,7 @@ from __main__ import send_cmd_help, settings
 from collections import deque, defaultdict
 from cogs.utils.chat_formatting import escape_mass_mentions, box
 import os
+import re
 import logging
 import asyncio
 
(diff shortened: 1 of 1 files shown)


calebj:
Can some people test this?

Jan200101:
Testing now on both Windows (10) and Linux (Raspberry Pi running Raspbian 4.4.12v7-aufs)

Twentysix26:
Seems good.  
I've seen some requests for a command that also removes the issued commands other than the bot's messages. Perhaps it could be integrated with this one? Thoughts?

calebj:
Implemented cleanup of bot commands and responses

PR #469, opened 2016-11-15
Use embed for user/server info

YeahImRose:
User color is the same as the user's role
Server color is completely random
How it looks

### DIFF
--- a/cogs/general.py
+++ b/cogs/general.py
@@ -182,7 +182,7 @@ class General:
             game = "Streaming: {} ({})".format(str(user.game), user.game.url)
 
         usr = discord.utils.get(server.members, id=self.bot.user.id)
-        if ctx.message.channel.permissions_for(usr).embed_links and self.bot.user.bot:
+        if ctx.message.channel.permissions_for(usr).embed_links:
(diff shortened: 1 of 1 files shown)


Kowlin:
Looks good to me, I have one concern though. Only bots can make the used of Rich Embedding, is there a fallback for non bot accounts?

YeahImRose:
There's no fall back but I can add it in easily

YeahImRose:
Issue
~~Please do not merge this until I have sorted an issue with dates.~~ Thanks @Eslyium for pointing this out c:

EDIT: Fixed now

Eslyium:
Works fine. 👍

skeith:
Roles on userinfo should be sorted descending

YeahImRose:
I was debating whether or not to add some form of sorting to roles, wasn't sure if it should be alphabetical or by hierarchy

skeith:
Probably better by hierarchy just like profile card

YeahImRose:
Good point

Twentysix26:
Er. Selfbots can send rich embeds, no need for fallbacks.

PR #367, opened 2016-09-03
Add setting for credits given on registering account

palmtree5:
Adds in a new command along with the necessary code to allow someone with the appropriate permissions to set the balance to be given on account creation via [p]bank register

PR #472, opened 2016-11-16
[Economy] Add setting for new account initial balance

palmtree5:
Supersedes #367 as that PR is extremely out of date. This brings the changes into the current version of economy, thus using dataIO instead of fileIO.

### DIFF
--- a/cogs/economy.py
+++ b/cogs/economy.py
@@ -510,9 +510,10 @@ class Economy:
         server = ctx.message.server
         if credits >= 0:
             self.settings[server.id]["REGISTER_CREDITS"] = credits
-            await self.bot.say("Registering an account will now give " + str(credits) + " credits.")
         else:
             self.settings[server.id]["REGISTER_CREDITS"] = 0
+            credits = 0
+        await self.bot.say("Registering an account will now give {} credits.".format(credits))
         dataIO.save_json(self.file_path, self.settings)
 
     def display_time(self, seconds, granularity=2):  # What would I ever do without stackoverflow?
--- a/cogs/economy.py
+++ b/cogs/economy.py
@@ -222,7 +222,7 @@ class Economy:
         user = ctx.message.author
         credits = 0
         if ctx.message.server.id in self.settings:
-            credits = self.settings[ctx.message.server.id]["REGISTER_CREDITS"]
+            credits = self.settings[ctx.message.server.id].get("REGISTER_CREDITS", 0)
         try:
             account = self.bank.create_account(user, initial_balance=credits)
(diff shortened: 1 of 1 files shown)


cogs/economy.py
```diff
@@ -220,8 +220,11 @@ def __init__(self, bot):
     async def register(self, ctx):
         """Registers an account at the Twentysix bank"""
         user = ctx.message.author
+        credits = 0
+        if ctx.message.server.id in self.settings:
+            credits = self.settings[ctx.message.server.id]["REGISTER_CREDITS"]
         try:
-            account = self.bank.create_account(user)
+            account = self.bank.create_account(user, credits)
```
tekulvw: This line MUST be `account = self.bank.create_account(user, initial_balance=credits)` because of the damned `*` operator.

cogs/economy.py
```diff
@@ -501,6 +504,14 @@ def already_in_list(self, accounts, user):
         await self.bot.say("Every payday will now give " + str(credits) + " credits.")
         dataIO.save_json(self.file_path, self.settings)
 
+    @economyset.command(pass_context=True)
+    async def registercredits(self, ctx, credits : int):
+        """Credits given on registering an account"""
+        server = ctx.message.server
+        self.settings[server.id]["REGISTER_CREDITS"] = credits
```
tekulvw, line 513: Should probably limit this to 0 through infinity

tekulvw (changes requested):

palmtree5:
Changes have been made

cogs/economy.py
```diff
@@ -508,8 +508,11 @@ def already_in_list(self, accounts, user):
     async def registercredits(self, ctx, credits : int):
         """Credits given on registering an account"""
         server = ctx.message.server
-        self.settings[server.id]["REGISTER_CREDITS"] = credits
-        await self.bot.say("Registering an account will now give " + str(credits) + " credits.")
+        if credits >= 0:
+            self.settings[server.id]["REGISTER_CREDITS"] = credits
+            await self.bot.say("Registering an account will now give " + str(credits) + " credits.")
```
tekulvw: This should be outside of the if/else block

tekulvw (dismissed):

cogs/economy.py
palmtree5: Not sure I fully understand

Twentysix26: If you don't do that, setting credits lower than 0 will save changes without the bot saying anything

palmtree5:
whoops, sorry. Kinda forgot about this pr lol

PR #475, opened 2016-11-18
[Economy] Leaderboard pagify

tekulvw:
And some pep8 stuff


Redjumpman (approved):

Twentysix26 (changes requested):
I gave the priority to #472 since it was older. Solve the conflict and it's good to go 👌

PR #476, opened 2016-11-18
[utils] DataIO file flusher

tekulvw:
Hopefully this should help alleviate some file writing problems. NEEDS LOTS O TESTING

### DIFF
--- a/cogs/owner.py
+++ b/cogs/owner.py
@@ -13,7 +13,6 @@ import threading
 import datetime
 import glob
 import os
-import time
 import aiohttp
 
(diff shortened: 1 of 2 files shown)


Redjumpman:
Currently testing these changes with heist and casino on my develop branch [here](https://github.com/Redjumpman/Jumper-Cogs/tree/develop)

tekulvw:
In my tests with casino it seems to be working as expected. I think that making the file writes in another thread may be even more beneficial.

tekulvw:
Not gonna be used in V2, useless in V3.

PR #479, opened 2016-11-18
[Audio] Fix local song issues with song command

tekulvw:


Jan200101 (approved):
Works like it should

Jan200101:
A small suggestion I had would be using kwargs in line 1161 to make starting a local song not need quotes around it

PR #483, opened 2016-11-20
[Owner] Lock [p]set owner to owner only.

H3xadecimal:
Why was it not set to owner only in the first place.


Twentysix26:
Try to think before opening a pull request. Thanks

PR #485, opened 2016-11-22
[Audio] Stop disconnect countdown when users rejoin voice channel

tekulvw:


tekulvw:
Can someone double check that logic? I think this pr may be useless/incorrect

PR #486, opened 2016-11-22
[Core] Move __main__.settings into bot

tekulvw:
The core bot settings are now available as `bot.settings`. This does maintain backwards compatibility so all your `from __main__ import settings` uses still work though you all should begin to move away from that.

PEP8


Jan200101 (approved):
Works fine.
Makes checking for owner easier andd the backwards compatibility makes it not break with older cogs

PR #491, opened 2016-11-25
[mod] Numerous enhancements:

calebj:
* Case management:
  - Centralize formatting code
  - Store raw action type
  - Centralize pretty-print (emoji) representation of actions
  - Store creation and update timestamps
  - Allow edits from admin or bot owner
  - Enable or disable case creation on a per-action basis
  - Preliminary support for action durations

* Mute:
  - add case creation (on by default for server, not channel)
  - Show channel member was muted in in modlog entry

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -207,10 +207,13 @@ class Mod:
 
     @modset.command(pass_context=True, no_pm=True, name='cases')
     async def set_cases(self, ctx, action: str = None, enabled: bool = None):
-        """Enables or disables case creation for each type of mod action"""
+        """Enables or disables case creation for each type of mod action
+        
+        Enabled can be 'on' or 'off'"""
         server = ctx.message.server
 
         if action == enabled:  # No args given
+            await self.bot.send_cmd_help(ctx)
             msg = "Current settings:\n```py\n"
             maxlen = max(map(lambda x: len(x[0]), ACTIONS_REPR.values()))
             for action, name in ACTIONS_REPR.items():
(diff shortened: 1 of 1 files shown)


aikaterna:
With muting reporting to a mod channel, a server mute will result in an error:

```
[27/11/2016 21:15] ERROR red on_command_error 238: Exception in command 'mute server'
  File "C:\Users\user\AppData\Local\Programs\Python\Python35\lib\site-packages\discord\ext\commands\core.py", line 50, in wrapped
  File "C:\Red-DiscordBot\cogs\mod.py", line 411, in server_mute
    reason=reason)
  File "C:\Red-DiscordBot\cogs\mod.py", line 1237, in new_case
    "channel"      : channel.id,
```

calebj:
Just rebased, please test

calebj:
Bumping this; needs testing...

Twentysix26:
Looks good to go. Anyone else want to give it a test?

cogs/mod.py
```diff
@@ -204,7 +277,7 @@ def __init__(self, bot):
 
     @commands.command(no_pm=True, pass_context=True)
     @checks.admin_or_permissions(ban_members=True)
-    async def ban(self, ctx, user: discord.Member, days: int=0):
+    async def ban(self, ctx, user: discord.Member, days: int=0, *, reason: str = None):
```
Twentysix26: I can't find a good solution for this. As it is, the user would be forced to input the days to enter a reason and I really don't like it. I could parse it manually without relying on the command ext but dunno, it seems to get messy fast between defaults, sanity checks on the days, etc. Any suggestion?

calebj: Check to see if the first word is a number, then have the caller confirm if it's part of the reason or the number of days. If it's only a number, assume the latter.

PR #492, opened 2016-11-25
[Core] Server based prefixes

Twentysix26:
This will break cogs that make use of `bot.command_prefix` and expect a list. It is now a callable that takes `Bot` and `Message`.  
Cogs that will need to be fixed can use that or make use of `bot.settings`, which has now the `get_prefixes` method. Takes a server as argument and return its prefixes if set, otherwise returns the global prefixes.

### DIFF
--- a/cogs/alias.py
+++ b/cogs/alias.py
@@ -57,7 +57,7 @@ class Alias:
             server_aliases = self.aliases[server.id]
             if command in server_aliases:
                 help_cmd = server_aliases[command].split(" ")[0]
-                new_content = self.bot.command_prefix[0]
+                new_content = self.bot.settings.get_prefix(server)[0]
                 new_content += "help "
                 new_content += help_cmd[len(self.get_prefix(help_cmd)):]
                 message = ctx.message
--- a/cogs/alias.py
+++ b/cogs/alias.py
@@ -3,8 +3,9 @@ from .utils.chat_formatting import *
 from .utils.dataIO import dataIO
 from .utils import checks
 from __main__ import user_allowed, send_cmd_help
-import os
 from copy import deepcopy
+import os
+import discord
 
 
 class Alias:
@@ -12,6 +13,7 @@ class Alias:
         self.bot = bot
         self.file_path = "data/alias/aliases.json"
         self.aliases = dataIO.load_json(self.file_path)
+        self.remove_old()
 
     @commands.group(pass_context=True, no_pm=True)
     async def alias(self, ctx):
@@ -36,7 +38,7 @@ class Alias:
             await self.bot.say('I can\'t safely add an alias that starts with '
                                'an existing command or alias. Sry <3')
             return
-        prefix = self.get_prefix(to_execute)
+        prefix = self.get_prefix(server, to_execute)
         if prefix is not None:
             to_execute = to_execute[len(prefix):]
         if server.id not in self.aliases:
@@ -57,9 +59,10 @@ class Alias:
             server_aliases = self.aliases[server.id]
             if command in server_aliases:
                 help_cmd = server_aliases[command].split(" ")[0]
-                new_content = self.bot.settings.get_prefix(server)[0]
+                new_content = self.bot.settings.get_prefixes(server)[0]
                 new_content += "help "
-                new_content += help_cmd[len(self.get_prefix(help_cmd)):]
+                new_content += help_cmd[len(self.get_prefix(server,
+                                        help_cmd)):]
                 message = ctx.message
                 message.content = new_content
                 await self.bot.process_commands(message)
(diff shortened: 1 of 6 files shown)


Twentysix26:
Should be good for testing now

Redjumpman:
Slightly related to testing for this, but I compiled a list of what will break with these changes.
First I would like to say that this does not break any cogs from the community-cogs repo. 👍 
Here is the list of cogs from the approved list that will break on merge:

**26-Cogs**
~~trigger.py~~ 👌

**Squid-Cogs**
~~admin.py~~
~~botinfo.py~~
~~mentiontracker.py~~
~~scheduler.py~~

**Dumb-Cogs**
~~hal.py~~ 👌

**Calebj-Cogs**
galias.py

I only checked cogs from community and the approved list.

Chovin:
can we get a `breaking-change` label around here?

Kowlin:
Seems to work like a charm. For testing see Red - Cog support #testing's pinned messages.

cogs/alias.py
```diff
@@ -158,8 +162,9 @@ def remove_old(self):
     def first_word(self, msg):
         return msg.split(" ")[0]
 
-    def get_prefix(self, msg):
-        for p in self.bot.command_prefix:
+    def get_prefix(self, server, msg):
```
calebj, line 165: Why change the signature here? Just access `msg.server` instead

cogs/utils/settings.py
```diff
@@ -171,6 +175,26 @@ def set_server_mod(self, server, value):
         self.bot_settings[server.id]["MOD_ROLE"] = value
         self.save_settings()
 
+    def get_server_prefixes(self, server):
+        if server is None or server.id not in self.bot_settings:
+            return self.prefixes
```
tekulvw: should be `[]`?

cogs/utils/settings.py
```diff
@@ -171,6 +175,26 @@ def set_server_mod(self, server, value):
...
+        p = self.bot_settings[server.id].get("PREFIXES")
```
tekulvw: and why not just `get("PREFIXES", [])` and return p without extra logic?

red.py
```diff
@@ -154,6 +154,9 @@ def user_allowed(self, message):
         else:
             return True
 
+    def prefix_manager(self, bot, msg):
```
tekulvw: why does this require a `bot` arg? It neither uses nor needs it since `self` is a Bot instance

tekulvw: and let it take server not message that's just dumb

tekulvw: and make server optional

cogs/utils/settings.py
Twentysix26: Not sure what you're referring to

red.py
Twentysix26: As discussed on the server, the callable needs to be exactly like that or it won't work

cogs/utils/settings.py
Twentysix26: I wanted to make it return `None` in case: `PREFIXES` is absent or it's set to `None` or it's `[]`

tekulvw: but why None? imo it should be the basecase (e.g. `[]`) or raise an exception. That way the function never gives a data type that people aren't expecting.

cogs/utils/settings.py
tekulvw: Me either

cogs/utils/settings.py
Twentysix26: I can see your point. Changing it.

cogs/alias.py
Twentysix26: If you take a look at the calls to `get_prefix` in alias you'll see that most of the time there's no  message object to be passed. The call in `on_message` seem to be the only exception. I'll leave it as it is.

Twentysix26:
I'm merging this today, giving a bit more time to cog creators who have to fix their stuff.

@neonobjclash @Lunar-Dust  @Aioxas @AznStevy

cogs/alias.py
Chovin: confused me too.
this `msg` is a str, not a discord.Message
👍

red.py
```diff
@@ -52,7 +52,8 @@ def __init__(self, *args, **kwargs):
         self.uptime = datetime.datetime.now()
         self._message_modifiers = []
         self.settings = Settings()
-        super().__init__(*args, **kwargs)
+        prefix_man = lambda bot, msg: bot.settings.get_prefixes(msg.server)
```
Chovin: no docs?
although we don't really write docs on anything other than commands *shrug

Twentysix26: Might as well start now. Looks more elegant anyway.

PR #497, opened 2016-11-28
[Audio] Config option to not disconnect when just bot in voice

FrostTheFox:
Little too hard to explain in the title, but adds an option in audio settings to enable or disable the check that's basically "if only bot in here, disconnect in 5 minutes"

Python noob, be nice :3

also i didn't know what would be a good variable, if you have a better idea lemme know

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -1867,7 +1867,8 @@ class Audio:
                         if server not in stop_times or stop_times[server] is None:
                             log.debug("putting sid {} in stop loop".format(server.id))
                             stop_times[server] = int(time.time())
-                    if self._noppl_disconnect and len(vc.channel.voice_members) == 1:
+                    noppl_disconnect = self.settings["NOPPL_DIS"]
+                    if noppl_disconnect and len(vc.channel.voice_members) == 1:
                         if server not in stop_times or stop_times[server] is None:
                             log.debug("putting sid {} in stop loop".format(server.id))
                             stop_times[server] = int(time.time())
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -997,8 +997,8 @@ class Audio:
     async def audioset_emptydisconnect(self, ctx):
         """Toggles auto disconnection when everyone leaves the channel"""
         server = ctx.message.server
-        server_set = self.get_server_settings(server_set)
-        noppl_disconnect = server_set.get("NOPPL_DISCONNECT", True)
+        settings = self.get_server_settings(server.id)
+        noppl_disconnect = settings.get("NOPPL_DISCONNECT", True)
         self.set_server_setting(server, "NOPPL_DISCONNECT",
                                 not noppl_disconnect)
         if not noppl_disconnect:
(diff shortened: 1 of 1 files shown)


cogs/audio.py
```diff
@@ -290,6 +290,9 @@ def _cache_max(self):
         setting_max = self.settings["MAX_CACHE"]
         return max([setting_max, self._cache_min()])  # enforcing hard limit
 
+    def _noppl_disconnect(self):
+        setting_nodis = self.settings["NOPPL_DIS"]
```
Chovin: `_noppl_disconnect ` is never called and even if it was, the setting is never put into the settings dict `self.settings` or into the `settings.json` file.

cogs/audio.py
```diff
@@ -1849,10 +1852,15 @@ def currently_downloading(self, server):
                     stop_times[server] = int(time.time())
 
                 if hasattr(vc, 'audio_player'):
-                    if (vc.audio_player.is_done() or len(vc.channel.voice_members) == 1):
+                    if (vc.audio_player.is_done() == 1):
```
Chovin: `== 1` eh?

cogs/audio.py
```diff
@@ -1849,10 +1852,15 @@ def currently_downloading(self, server):
...
                         if server not in stop_times or stop_times[server] is None:
                             log.debug("putting sid {} in stop loop".format(server.id))
                             stop_times[server] = int(time.time())
+                    elif self._noppl_disconnect == False:
```
Chovin: you defined this as a function above. this isn't how you call a function.
`if not a_boolean:` is preferred over `if a_boolean == False:`

Chovin (changes requested):
thanks for your work.
I suggest you look at some simple cogs for examples, there are some pretty useful examples in some of them.

cogs/audio.py
FrostTheFox: whoop i'm an idiot

cogs/audio.py
FrostTheFox: haha understood 👍

cogs/audio.py
FrostTheFox: Wouldn't the addition at line 2060 add it to settings.json if not present? Just tried and it seemed to...

FrostTheFox:
Alright, it's at least working now.

cogs/audio.py
FrostTheFox: Fixed this 👍

cogs/audio.py
FrostTheFox: fixed 👍

cogs/audio.py
tekulvw: Yeah it will, Kowlin pulled some magic out of his ass a while back on load to make this happen

tekulvw:
Can you make this editable by contributors? You've got a good start, there's just a few things I'd like to change and add real quick

FrostTheFox:
Done. Sorry, it became habit on other projects.

tekulvw:
Okay I think we're ready to go now @Twentysix26

Twentysix26:
As it is, it's a global setting. It has no reason to be global.

Twentysix26 (changes requested):
It should be server based

Twentysix26 (approved):

FrostTheFox:
heh, thanks for fixing this up, I wrote this really late at night and I may have been drinking some wine so I was not at peak performance :P

PR #500, opened 2016-11-30
[Image] rewrite imgur command to use subcommands

palmtree5:
Overrides #432. The changes from that PR were worked into a rewrite of the imgur command in images.py such that it uses subcommands. Also some general PEP8 formatting stuff all over in the image cog. Tested in #advanced-testing in discord and it appears to work fine (i.e. it doesn't crash and burn)

### DIFF
--- a/cogs/image.py
+++ b/cogs/image.py
@@ -9,6 +9,7 @@ except:
 
 CLIENT_ID = "1fd3ef04daf8cab"
 CLIENT_SECRET = "f963e574e8e3c17993c933af4f0522e1dc01e230"
+GIPHY_API_KEY = "dc6zaTOxFJmzC"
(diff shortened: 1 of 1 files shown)


Twentysix26:
Did some adjustments but could use more work.  
Also, very few people noticed but this cog has blocking issues. Since we're rewriting we might as well do something about this.

tekulvw:
Specifically, the imgurclient lib uses requests which means you'll need to use `loop.run_in_executor` for each call to the lib

palmtree5:
new commit from @fwinest-JediThug on this branch that moves the giphy api key into a variable

Twentysix26:
Closing for impending rewrite.

PR #502, opened 2016-11-30
[core] allow run through import, add bot init hooks

calebj:
This is so people can write another python script that does, for example:
```py
def bot_constructor(bot):
    # custom bot class goes here
    class myBot(bot):
        def __init__(self, *args, **kwargs):
            #
            # do stuff here
            #

        # override this
        async def say(self, message, **kwargs):
            message = message.upper()
            await super().say(message, **kwargs)

    return myBot

def bot_postinit(bot):
    #
    # do stuff to the bot instance before settings are loaded
    #
    return bot

from red import *
#
# do stuff after stuff is ready to launch
#

# start the bot
main()
```


tekulvw:
I...really dislike the use of `__main__`. Why not refactor the setup stuff in `red.py` into a setup function that can optionally take a class/function to use to create the bot object? And the same thing goes if you want to have a postinit function.

PR #512, opened 2016-12-05
[core] Refactor core to be modular

calebj:
This is take two of #502. There's a few changes but everything functions the same. Essentially, nothing is in the main script's namespace (except for backwards compatibility) and nothing is instantiated outside the `if __name__ == '__main__'` code.

This allows anyone to wrap red.py in their own code, importing most of what they need and implementing everything else.

### DIFF
--- a/red.py
+++ b/red.py
@@ -64,7 +64,11 @@ class Bot(commands.Bot):
         self._message_modifiers = []
         self.settings = Settings()
         self._intro_displayed = False
-        kwargs["self_bot"] = self.settings.self_bot
+        self.logger = set_logger(self)
+        if 'self_bot' in kwargs:
+            self.settings.self_bot = kwargs['self_bot']
+        else:
+            kwargs['self_bot'] = self.settings.self_bot
         super().__init__(*args, command_prefix=prefix_manager, **kwargs)
 
     async def send_message(self, *args, **kwargs):
@@ -140,12 +144,12 @@ class Bot(commands.Bot):
         mod = self.get_cog('Mod')
 
         if mod is not None:
-            if settings.owner == author.id:
+            if self.settings.owner == author.id:
                 return True
             if not message.channel.is_private:
                 server = message.server
-                names = (settings.get_server_admin(
-                    server), settings.get_server_mod(server))
+                names = (self.settings.get_server_admin(
+                    server), self.settings.get_server_mod(server))
                 results = map(
                     lambda name: discord.utils.get(author.roles, name=name),
                     names)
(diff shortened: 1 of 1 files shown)


aikaterna:
Cogs loaded aren't counted properly.

```-- Logging in.. --
Make sure to keep your bot updated by running the file update.bat
Official server: https://discord.me/Red-DiscordBot
------
Joker is now online.
------
Connected to:
1 servers
19 channels
42 users

12/11 active cogs with 89 commands
Prefix: .

------
Use this url to bring your bot to a server:
<removed>
------
[04/12/2016 22:39] ERROR red on_command_error 244: Exception in command 'cogs'
  File "C:\Users\Administrator\AppData\Local\Programs\Python\Python35\lib\site-packages\discord\ext\commands\core.py", line 50, in wrapped
  File "C:\Red-DiscordBot_modular\Red-DiscordBot\cogs\owner.py", line 175, in _show_cogs
    loaded = [c.__module__.split(".")[1] for c in self.bot.cogs.values()]
  File "C:\Red-DiscordBot_modular\Red-DiscordBot\cogs\owner.py", line 175, in <listcomp>
    loaded = [c.__module__.split(".")[1] for c in self.bot.cogs.values()]
IndexError: list index out of range```

**Comment by calebj** [comment]
Okay, I've made these changes a bit less radical and almost as flexible as before. Test please.

**Comment by Twentysix26** [comment]
Works great, tested on both my install and a clean one. Thoughts @tekulvw ?

**Comment by Twentysix26** [comment]
I think there's an issue with the loggers. Will have to take a better look.

**Comment by calebj** [comment]
Fixed the logger thing

File: red.py [hunk 1]
```diff
@@ -187,150 +191,148 @@ def _add_subcommands_to_page(self, max_width, commands):
...
-async def get_oauth_url():
-    try:
-        data = await bot.application_info()
-    except Exception as e:
-        return "Couldn't retrieve invite link.Error: {}".format(e)
-    return discord.utils.oauth_url(data.id)
+    import __main__
+    __main__.send_cmd_help = bot.send_cmd_help  # Backwards
+    __main__.user_allowed = bot.user_allowed    # compatibility
+    __main__.settings = bot.settings            # sucks
```
tekulvw, line 202: Using `global` kw should work here instead of the dumb `__main__` import

red.py
```diff
@@ -520,53 +522,54 @@ def load_cogs():
...
+                bot.settings.token = None
+                bot.settings.email = None
+                bot.settings.password = None
+                bot.settings.save_settings()
     except KeyboardInterrupt:
         loop.run_until_complete(bot.logout())
     except:
         error = True
-        logger.error(traceback.format_exc())
+        bot.logger.error(traceback.format_exc())
```
tekulvw: `logger.exception()` has a use-case, this is it

tekulvw (commented):
Haven't tested yet, first glance comments

red.py
calebj: Are you sure it works when I do `from red import foo`?

red.py
calebj: I'll change it, but I didn't write that bit before. Only made it work with my other changes.

red.py
tekulvw: It should, as long as you put it into whatever your `__main__` is too if you're using red.py to launch. `global` sticks vars into the global namespace for that file

calebj: Nope, doesn't work. I don't want the global namespace for red.py, that isn't where cogs import from. It needs to be `__main__`

PR #513, opened 2016-12-05
[Core] CLI flags, initial selfbot support, revamped setup / main screen, etc

Twentysix26:
**Changes:**
Startup flags: can be started and configured without ever doing the interactive setup
Changed default settings format so that all unset values are None
Removed new cogs prompt
Removed `LOGIN_TYPE` from settings.json. It now defaults to token and fallbacks to email/password
Smarter initial setup: only asks for the settings that are actually missing
Startup flag that allows settings to be memory-only
Initial selfbot support
Only reset login credentials (on confirmation) instead of deleting the whole file in case of login failure
Revamped main screen
Made sure that nothing blows up when you run Red on Windows without `chcp 65001`
Possibility of setting credentials in the environment variables `RED_TOKEN` / `RED_EMAIL` `RED_PASSWORD` 

**Possible breaking changes:**
Cogs that make use of bot.email and expect a token (I don't think there are any, but still)
Cogs that set settings and expect them to save automatically. This is now disabled to allow memory-only settings.
`Settings.login_type` is no longer a thing

**Warning:** Testing this PR will change the format of your settings.json. It won't work anymore if you try to use it again in the current build.

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -1378,7 +1378,7 @@ def setup(bot):
     global logger
     check_folders()
     check_files()
-    logger = logging.getLogger("mod")
+    logger = logging.getLogger("red.mod")
     # Prevents the logger from being loaded again in case of module reload
     if logger.level == 0:
         logger.setLevel(logging.INFO)
--- a/cogs/utils/settings.py
+++ b/cogs/utils/settings.py
@@ -52,14 +52,20 @@ class Settings:
 
     def parse_cmd_arguments(self):
         parser = argparse.ArgumentParser(description="Red - Discord Bot")
-        parser.add_argument("--email")
-        parser.add_argument("--password")
-        parser.add_argument("--token")
-        parser.add_argument("--owner", help="ID of the owner")
+        parser.add_argument("--email", help="Email login. Must provide a "
+                                            "password too. Not using a bot "
+                                            "account with a token is "
+                                            "discouraged")
+        parser.add_argument("--password", help="Password of the email login")
+        parser.add_argument("--token", help="Login token")
+        parser.add_argument("--owner", help="ID of the owner. Only who hosts "
+                                            "Red should be owner, this has "
+                                            "security implications")
         parser.add_argument("--prefix", "-p", action="append",
-                            help="Global prefix. Can be multiple.")
-        parser.add_argument("--admin-role")
-        parser.add_argument("--mod-role")
+                            help="Global prefix. Can be multiple")
+        parser.add_argument("--admin-role", help="Role seen as admin role by "
+                                                 "Red")
+        parser.add_argument("--mod-role", help="Role seen as mod role by Red")
         parser.add_argument("--no-prompt",
                             action="store_true",
                             help="Disables console inputs. Features requiring "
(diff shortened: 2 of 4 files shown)


tekulvw:
Can we get a debug cli flag?

Twentysix26:
@tekulvw desired behaviour? Othen than a `Settings.debug` being set to True, which would be useful for cogs as a debug mode flag

Kowlin:
@Twentysix26 pretty sure we talked about this before. the Debug flag was to enable the debug command, with the debug command defaulting to disabled.

tekulvw:
Uh...well...for starters I'd set the `red` logger to `logging.DEBUG` then keep the flag available in settings for cog loggers to do the same (most of the time they shouldn't as they just inherit the logging level from `red`).

@Kowlin not sure where that came from tbh

edit: Use `red.mod` as the logger name in Mod, also `red.discordpy` for the dpy logger?

Twentysix26:
@tekulvw I assume d.py's logger has to be named `discord`
https://discordpy.readthedocs.io/en/latest/logging.html

red.py
```diff
@@ -129,7 +129,10 @@ def clear_message_modifiers(self):
     def user_allowed(self, message):
         author = message.author
 
-        if author.bot or author == self.user:
+        if author.bot:
+            return False
+
+        if author == self.user and not self.settings.self_bot:
```
calebj: Change this to
```py
if author == self.user:
    return self.settings.self_bot
```

cogs/mod.py
```diff
@@ -1378,7 +1378,7 @@ def setup(bot):
     global logger
     check_folders()
     check_files()
-    logger = logging.getLogger("mod")
+    logger = logging.getLogger("red.mod")
```
tekulvw, line 1381: 👍

cogs/utils/settings.py
```diff
@@ -11,27 +14,96 @@ def __init__(self, path=default_path):
...
+            self.update_old_settings_v1()
+
+        if old_format:
+            self.update_old_settings_v2()
+
+        self.parse_cmd_arguments()
+
+    def parse_cmd_arguments(self):
+        parser = argparse.ArgumentParser(description="Red - Discord Bot")
+        parser.add_argument("--email")
```
tekulvw: Can you add help doc for this? Don't want people using user accounts

cogs/utils/settings.py
```diff
@@ -11,27 +14,96 @@ def __init__(self, path=default_path):
...
+                self.email = args.email
+                self.password = args.password
+                self.login_type = "email"
+            elif arg == "owner":
+                self.owner = args.owner
+            elif arg == "prefix":
+                self.prefixes = sorted(args.prefix, reverse=True)
+            elif arg == "admin_role":
+                self.default_admin = args.admin_role
+            elif arg == "mod_role":
```
tekulvw: Uh...why aren't you just using `args.var_name`? and why the `elif`'s?

tekulvw (commented):
Haven't tested it, just first glance comments

cogs/utils/settings.py
Twentysix26: I swear it made sense when I was looking at how argparse worked. Refactoring :')

Kowlin (commented):
Changes work fine, nothing broke, no buildings collapsed. And no demons are summoned (yet). One traceback when using ``--self-bot``

```
Ignoring exception in on_message
  File "C:\Users\K\AppData\Local\Programs\Python\Python35\lib\site-packages\discord\client.py", line 307, in _run_event
    yield from getattr(self, event)(*args, **kwargs)
  File "red.py", line 261, in on_message
    if user_allowed(message):
  File "red.py", line 152, in user_allowed
    for r in results:
  File "red.py", line 150, in <lambda>
    lambda name: discord.utils.get(author.roles, name=name),
AttributeError: 'User' object has no attribute 'roles'
```

PR #514, opened 2016-12-06
[Mod] Changed logger from "mod" to "red.mod"

Kowlin:


Twentysix26:
Dealing with it in #513

PR #523, opened 2016-12-11
[Audio] add server owner check for [p]stop (fixes #522)

### LINKED ISSUE #522: [Audio] Server owner can't stop music using [p]stop
This appears to be a bug resulting from the fact that server owner isn't even checked for in `def can_instaskip` (line 1740 of audio.py). Propose adding a check for server owner in there

palmtree5:


palmtree5:
FYI, I tested this and it works

PR #526, opened 2016-12-14
[Streams] use embeds for stream info if live

palmtree5:
Adds embed support for Beam, Hitbox, and Twitch (though only Twitch and Beam will have an image with them). Embeds only happen if the channel specified by the user is live on the service

### DIFF
--- a/cogs/streams.py
+++ b/cogs/streams.py
@@ -2,6 +2,7 @@ import discord
 from discord.ext import commands
 from .utils.dataIO import dataIO
 from .utils.chat_formatting import *
+from random import choice as randchoice
 from .utils import checks
 from __main__ import send_cmd_help
 import os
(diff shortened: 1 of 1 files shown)


palmtree5:
Should note that I tested `[p]beam`, `[p]hitbox`, and `[p]twitch` before I pushed this to my fork. Haven't tested it with the notifier but that should be fine

palmtree5:
Put in checks on the twitch commands for where the game and/or title aren't set

palmtree5:
Fixed the issue I was having where alerts weren't happening. Note to self: don't try to .format a dict into a string...

Twentysix26:
https://github.com/Twentysix26/Red-DiscordBot/commit/b69b5b30056c5b4fb213b20ee0970f22de5c0503

PR #532, opened 2016-12-16
[Audio] Better URL checking

### LINKED ISSUE #511: [Audio] . make it see text as a url
To be exact if you use

[p]yt S.O.N.G.

it says its not a valid link but it is not a link.

Seenko:
#511 
Instead of checking for a URL based on a single dot, it actually uses RegEx, this way it should get most URLs correctly, as you know, it's of RegEx nature to fail alot so more experimentation is needed, but it should work. I've tested the basic stuff a user would do and replicated and fixed #511 .


Redjumpman:
I haven't had this fail in some simple tests, but just to throw some extra data at this:
```Python
timeit.timeit('URLMatch.match(url)', setup='''import re; URLMatch=re.compile('@^(https?|ftp)://[^\s/$.?#].[^\s]*$@iS', re.IGNORECASE); url="www.google.com"''', number=1000)
>>>0.00028463814669521526  # Fastest method as proposed above
timeit.timeit('"." in url', setup='url="www.google.com"', number=1000)
>>>4.039624036522582e-05 # Slowest method, currently implemented
timeit.timeit('urlparse(url)', setup='from urllib.parse import urlparse; url="www.google.com"', number=1000)
>>>0.0028327088948572055  # Second fastest, parsing the url into parts to view
```

I thought I would throw in urlparse because looking at the input by verifying the different parts is probably the best way to check for validity. It's possible to just create a separate function to handle that logic. BUT I think the approach used above is a great improvement over the current system.

Seenko:
Actually, the implemented version is the fastest, being 0.00004 (as in 4e-5), don't forget your notations!

So the correct order should be (fastest -> slower):

1. (0.00004) Current implemented version.
2. (0.0002)   Using RegEx
3. (0.002)     Using the urlparse library.

Although in this case, the speed difference wouldn't matter much.

Redjumpman:
Oh god. Yeah you are right. Extra data [here](https://i.gyazo.com/3703fbf26c565508ebaa8e542024beaa.png) for any who cares...

tekulvw:
Drop this in a function `_match_any_url(self, text)` right around [here](https://github.com/Twentysix26/Red-DiscordBot/blob/develop/cogs/audio.py#L727)

Seenko:
@tekulvw Well, this function is to be able to make text searches with symbols without it accidentally detecting the search as an URL, I don't think it would be beneficial to add that there.

ghost (approved):

Twentysix26:
That regular expression doesn't seem to be working.

Seenko:
Weird, my initial tests with that RegEx worked fine. I'll do some research.

Twentysix26:
Not a fan of huge regular expressions, did it differently
https://github.com/Twentysix26/Red-DiscordBot/commit/143d41f2f27fec3d42ec0efaa8cb81e601bb5322

PR #533, opened 2016-12-17
[Owner] display availble update count in [p]version

palmtree5:
This will pull the commit log for the repo from github's api and return an int representing the number of available updates (aka commits that have not yet been pulled) which is then added into the embed that [p]version posts or return None if there is some sort of failure in that check (for example if the repo url is somehow not valid)

### DIFF
--- a/cogs/owner.py
+++ b/cogs/owner.py
@@ -4,7 +4,6 @@ from cogs.utils import checks
 from __main__ import set_cog
 from .utils.dataIO import dataIO
 from .utils.chat_formatting import pagify, box
-from requests.compat import json as complexjson
 
 import importlib
 import traceback
@@ -15,7 +14,6 @@ import datetime
 import glob
 import os
 import aiohttp
-import json
 
 log = logging.getLogger("red.owner")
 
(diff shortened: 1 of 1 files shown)


Jan200101:
testing 👍 
Cant really tell if its working as I am up to date

cogs/owner.py
```diff
@@ -4,6 +4,7 @@
 from __main__ import set_cog
 from .utils.dataIO import dataIO
 from .utils.chat_formatting import pagify, box
+from requests.compat import json as complexjson
```
Rapptz: Unused import

cogs/owner.py
```diff
@@ -800,28 +805,49 @@ def _wait_for_answer(self, author):
             print("The set owner request has been ignored.")
             self.setowner_lock = False
 
+    async def _check_remote_commitlog(self, url, chash):
```
Rapptz, line 806: You can do this differently.

Invoking `git remote update` will update the remote refs (pointed by the red repo) in which you can then use something like `git status` on it. Bypassing the github API completely. See [here](http://stackoverflow.com/a/3278427/1381108) for more info.

cogs/owner.py
```diff
@@ -14,6 +15,7 @@
 import glob
 import os
 import aiohttp
+import json
```
Rapptz: Unused import.

Twentysix26 (changes requested):
What @Rapptz pointed out

Twentysix26:
Might be better suited for the launcher. Anyway, closing.

PR #538, opened 2016-12-21
[Economy] Refactored slot machine, fixed [p]bank register

Twentysix26:
**Changes:**
Made the slot machine code saner
Fixed `[p]bank register`, the "register credits" part was broken in a recent update
Nerfed the jackpot from *5000 to *2500. I still think payouts are quite OP, should probably be redesigned. Suggestions are welcome

### DIFF
--- a/cogs/economy.py
+++ b/cogs/economy.py
@@ -570,7 +570,7 @@ class Economy:
         if not payout:
             # Checks for two-consecutive-symbols special rewards
             payout = PAYOUTS.get((rows[1][0], rows[1][1]),
-                     PAYOUTS.get((rows[1][0], rows[1][1]))
+                     PAYOUTS.get((rows[1][1], rows[1][2]))
                                 )
         if not payout:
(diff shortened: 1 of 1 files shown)


cogs/economy.py
```diff
@@ -451,95 +511,91 @@ def already_in_list(self, accounts, user):
...
+            sign = "  "
+            if i == 1:
+                sign = ">"
+            slot += "{}{} {} {}\n".format(sign, *[c.value for c in row])
+
+        payout = PAYOUTS.get(rows[1])
+        if not payout:
+            # Checks for two-consecutive-symbols special rewards
+            payout = PAYOUTS.get((rows[1][0], rows[1][1]),
+                     PAYOUTS.get((rows[1][0], rows[1][1]))
```
irdumbs: I might just be sleepy, but this was meant to be rows[1][1], rows[1][2]
right?

Twentysix26: Yeah 👍

PR #542, opened 2016-12-22
Travis-Ci status in version command.

Jan200101:


Jan200101:
Right now I dont know if I should keep the code in _get_version or put it back to the version command.
Also the current text it is displaying is a placeholder

Jan200101:
Time to tease #coding

Jan200101:
Right now I dont know if I should keep the code in _get_version or put it back to the version command.
Also the current text it is displaying is a placeholder

palmtree5:
[Image](url)

Might I suggest putting the build status as the thumbnail?

Jan200101:
Good idea.
Pushing it now

Kowlin:
To to be honest. I don't know if using this status is the best thing for version. Due the fact that we have other forks running on that code. Simplest way is to check the git remote and see if it vanilla red

Jan200101:
Im gonna try to add a check if the fork has a travis page and then display its status

Jan200101:
It should work fine with forks now i'm going to test myself with Calebs Selfbot repo but it would be nice if others could test too

I am not able to test for not existing ones as it always gives out a valid image / link even when it doesnt exist

Im looking into how I could get that to work too

Jan200101:
Right now a branch counts as the default one if it was the only one cloned.
But there should be no need to check for that

irdumbs:
I don't see why Travis status should be in the version command. could you humor me and explain it?

Kowlin:
Too be perfectly honest. Neither do I.

palmtree5:
Yeah, I'm not sure it's needed (despite my suggestion above)

Jan200101:
Welp that settles it then, no one has a idea why this is needed or why it should be here.
Closing until I find something better to do with it

PR #543, opened 2016-12-23
[pagify] betterify escaping mass mentions

Chovin:
more sensible page lengths when escaping mass mentions #438
still not completely accurate, but at least it doesn't freeze on `list(pagify(('@'+'everyone')*2000))`


Chovin:
probs should check that `this_page_len` is always positive, but that's outside the scope of this PR.
what would peeps be doing asking for page sizes that small anyway?

Kowlin:
Fixed with #920 (yea, really)

Issue #544, opened 2016-12-24
update.bat doesn't work with python 3.6


rreyv:
It quits with 'requested python version (3.5) not installed.
Changing to update.bat to search for 3.6 instead works though.

Kowlin:
We currently don't support Python 3.6 *(As it just released roughly 16 hours ago from writing this post)* So please note that there are already some issues regarding 3.6 on the bot. One of them being not being able to set the owner from the console. And we're looking into fixing them (Mostly after Christmas)

rreyv:
Thanks for the ridiculously quick response! Have a good Christmas. I'll use
3.5 in the meantime.

Kowlin:
*Poof* update.bat is gone! Python 3.6 should be supported now. The new launcher will update and run Red from now on.

rreyv:
Gracias

PR #545, opened 2016-12-24
Update update.bat

skeith:
Using variable for python version on python command execution so it will work on version higher than 3.5

Have not test it on 3.6 (need tester)
It does not have the ability to stop python 2.x being used but I assume the error from lower python version will make it fail

### DIFF
--- a/update.bat
+++ b/update.bat
@@ -12,15 +12,57 @@ if NOT %errorLevel% == 0 (
     GOTO end  
 )
 
+:ban2x
+FOR /F "tokens=* USEBACKQ" %%V IN (`python --version`) DO (
+SET pt=%%V
+)
+set ptv=%pt:~7,1%
+IF %ptv%== 2 (goto halt) else (goto choose)
+
+
+:halt
+echo Your default python version is lower than the requirements (Python 3.5 or higher)
+echo OR this script is not able to detect it ... yet
+CHOICE /M "You still want to continue?"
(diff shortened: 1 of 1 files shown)


Kowlin:
Fixes #544, should work with 3.6 and 3.5 due to the magic of *math* being involved.

Jan200101 (commented):
Code looks good but Im gonna test it now with Python3.6

Jan200101:
Code looks good but Im gonna test it now with Python3.6

Jan200101 (approved):
echo. doesnt seem to work for some reason and there is a 0 command it says its getting.
But aside of that its good and works.

Will only get a bit complicated if someone got python 3.6 installed but used python3.5 for the bot

skeith:
@Jan200101 echo is commented (my vars echo) , they're there for test purposes. I could make this bat scans for installed pythons and make selection but I assume 3.5 will still be mostly used because the guide says so

Twentysix26:
Implemented a less complex solution
https://github.com/Twentysix26/Red-DiscordBot/commit/607f88cf758f839930379946565fff705fbadc64

Jan200101:
what if someone has python 3.2 installed ?

Twentysix26:
The old version didn't prevent this either since it defaulted to `python` as a last attempt. The only difference in this change is that it allows >3.5 python installations.

PR #550, opened 2016-12-27
[settings] Add constructor parameter to skip argparse

calebj:
My docker run script needs to work with the `Settings` object without argparse causing `unrecognized argument` problems. In general, side effects like this are undesirable.


calebj:
Bump. Kind of need this for my launcher script to work

PR #552, opened 2016-12-28
[WIP] Red's launcher / Downloader reqs

Twentysix26:
**Launcher:**
- A launcher that deals with requirements, updates and maintenance
- Interactive with menus / submenus, etc but can also be headless with CLI flags
- Brings the startRed.bat / startRedLoop.bat functionality on all platforms
- Deals with requirements, installs them locally in the local `lib` folder without the need of root
- Plenty of updates options
- Downloads ffmpeg on Windows and verifies its files' integrity
- Maintenance options: can reset Red, its requiremenents and factory reset
- Generates batch files for fast start on Windows

**Downloader changes:**
- Cog-specific `info.json` can now have a `"REQUIREMENTS"` field in which they specify their requirements
- The requirements will be installed via pip in a thread safe way, just before the cog gets installed. The info.json will be verified after each update for any new reqs
- Since the requirements will be installed inside the local `lib` folder it won't require root
- Failure to install the requirements will make the cog installation fail

**Other changes:**
- A `[p]restart` command has been added. It makes Red quit with exit code 26. Once receiving that code the launcher will start it up again
- In the launcher context, `[p]shutdown` will actually quit Red for good, regardless of the autorestart mode
- `[p]shutdown` and `[p]restart` now both use the new `Bot.shutdown` method

**Breaking changes:**
- In theory, nothing. 🤞

**Issues:**
- ~~Currently not working on Macs. Possibly related to [this](https://stackoverflow.com/questions/24257803/distutilsoptionerror-must-supply-either-home-or-prefix-exec-prefix-not-both)~~ This should be fixed by not using `--target`, and as such the `lib` folder, on Macs. They don't require root privilege with Homebrew anyway


ccubed:
Platform isn't a reliable check for  a 64bit run time.  All that tells you is that they're on a 64 bit platform not that they're running in the 64 bit run time space. See the note here https://docs.python.org/3.6/library/platform.html

You may not necessarily care about this, but it's a distinction.

palmtree5:
Suggest changing
```python
if not PYTHON_OK:
    input("Red needs Python 3.5 or superior. Install the required "
          "version.\nPress enter to continue.")
```
to
```python
if not PYTHON_OK:
    print("Red needs Python 3.5 or superior. Install the required "
          "version.")
    exit(1)
```
Seems pointless to make the user input when we're just exiting anyway

palmtree5:
So, issue:

The launcher hangs after exiting the bot process sometimes

The only situation I've found that causes this so far is playing audio, then stopping audio, then shutting down via the shutdown command

palmtree5:
Also, regarding the WIP note, I presume that the `REQUIREMENTS` field applies to the cog-specific info.json file?

palmtree5:
looks like all the resetting stuff works properly as well and I haven't run into any issues with the menuing. Didn't test cli flags stuff yet

Twentysix26:
@palmtree5 Yes, cog specific

tekulvw:
Why not make use of `requirements.txt` at the root level of a cog repo instead of a key in `info.json`?

Kowlin:
If I can make a guess its because not all cogs require the full package of requirements.txt.

palmtree5:
I'm inclined to agree that we should only install a req if the cog that needs it is installed

red.py
```diff
@@ -174,6 +176,32 @@ def user_allowed(self, message):
         else:
             return True
 
+    async def pip_install(self, name, *, timeout=60):
```
calebj: Having a timeout here is unwise, something like PyNaCl will take at least that long to build.

palmtree5:
Can we implement update notifications into the launcher somehow? Though that would likely only run when the launcher is started so there would need to be some plan for notifications after that

tekulvw:
What about adding in stuff for sharding? Nothing needs to be implemented for it, just have a configurable # of shards in the menus that'll be implemented later

Twentysix26:
Should be done. Needs some testing, especially the `[p]cog update` part

@tekulvw I don't think adding options for something that is not implemented yet is a good idea.
@palmtree5 I thought about implementing that, either before merging or in the future. Makes sense for it to be in the launcher.

red.py
Twentysix26: Do you suggest removing the timeout entirely?

Chovin: could prompt the user when it takes too long. ask them to keep trying or kill it

Chovin:
so um.
is there a list of what needs to be tested and what is already tested?
how do I test this given a short timeframe?

Chovin:
Lovely PR ❤️ 

Tested on Mac and here are some 
### points of discussion from discord:
1. breaks on any `pip` operation
  
2. recommend letting the user know it's not done by saying something like "Updating requirements.." after it says "Red has been updated"

3. recommend saying "local installed python packages" or just take that part out, so that people don't think it uninstalls cogs.  
Maybe even point them to the `[p]cog` command as the source for cog management

4. > irdumb: nice catching smooth exit
irdumb: although, at times I'd like to restart it from discord without needing to have access to the console
26: maybe we should have a custom exit code for reboots. So that you can restart in any case
26: For technical reason, I propose 26

Twentysix26:
@Chovin  2, 3 and 4 are dealt with. Point 1 is the main issue now

Jan200101:
What will happen when a requirement is already installed but not in the lib folder ?

Twentysix26:
@Jan200101 The `lib` folder has just priority over other paths. Nothing changes for existing installations. When they'll update with the launcher, requirements will get installed in the `lib` folder, taking priority over the default ones.

Jan200101:
so we are basicly forced to redownload it into the lib folder even though we got it installed for the python version

Twentysix26:
You are not "forced" to do anything. It's a minor inconvenience that happens the first time you decide to update using the *optional* launcher on existing installations. And a very minor inconvenience at that.

Chovin:
1. still seems to be an issue 
image  


~~4. still seems to be an issue
image~~
[edit] didn't see the commit comment.. cause I don't look there :3 

`[p]restart` works
4 is all good 👍

Chovin:
1. seems to be fixed 👍

PR #556, opened 2016-12-30
[Audio] Put connection on cool down if connection times out to limit blocking time

tekulvw:
Will now:
 * timeout after 5 seconds
 * wait 5 minutes before allowing another connection attempt by the server that just failed

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -290,6 +290,8 @@ class Audio:
 
         self.skip_votes = {}
 
+        self.connect_timers = {}
+
         if player == "ffmpeg":
             self.settings["AVCONV"] = False
         elif player == "avconv":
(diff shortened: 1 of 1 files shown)


cogs/audio.py
```diff
@@ -627,13 +629,21 @@ def _is_queue_playlist(self, server):
...
+        connect_time = self.connect_timers.get(server.id, 0)
+        if time.time() < connect_time:
+            diff = int(connect_time - time.time())
+            raise ConnectTimeout("You are on connect cooldown for another {}"
+                                 " seconds.".format(diff))
         if server.id in self.queue:
             self.queue[server.id]["VOICE_CHANNEL_ID"] = channel.id
         try:
-            await self.bot.join_voice_channel(channel)
+            await asyncio.wait_for(self.bot.join_voice_channel(channel),
```
tekulvw, line 640: This line is guaranteed to throw `asyncio.futures.TimeoutError` if it does in fact time out.

tekulvw:
image

Well it definitely works

Kowlin:
Seems fine with me

PR #582, opened 2017-01-20
Add heists and a slight modification to payday

Naleksuh:
Credit to BlueMageX for the code (I got his permission to pr it)

### DIFF
--- a/cogs/economy.py
+++ b/cogs/economy.py
@@ -1,34 +1,44 @@
 import discord
 from discord.ext import commands
 from cogs.utils.dataIO import dataIO
-from collections import namedtuple, defaultdict, deque
+from collections import namedtuple, defaultdict
 from datetime import datetime
+from random import randint
+from random import choice as randchoice
 from copy import deepcopy
 from .utils import checks
 from cogs.utils.chat_formatting import pagify, box
-from enum import Enum
 from __main__ import send_cmd_help
 import os
 import time
 import logging
-import random
 
-default_settings = {"PAYDAY_TIME": 300, "PAYDAY_CREDITS": 120,
-                    "SLOT_MIN": 5, "SLOT_MAX": 100, "SLOT_TIME": 0,
-                    "REGISTER_CREDITS": 0}
+default_settings = {"PAYDAY_TIME": 180, "PAYDAY_CREDITS": 20,
+                    "SLOT_MIN": 3, "SLOT_MAX": 300, "SLOT_TIME": 60,
+                    "REGISTER_CREDITS": 0,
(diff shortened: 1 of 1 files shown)


Kowlin:
Thanks for putting in the effort to make a PR. However as this change includes files from a 3rd party and include changes that remove other fixes and changes from Economy. And there for I don't think we can accept this PR. However, I'll leave it open for @Twentysix26 to decide. I'll add more conflicting stuff in a review.

cogs/economy.py
```diff
@@ -398,21 +360,8 @@ def __init__(self, bot):
         except NoAccount:
             await self.bot.say("User has no bank account.")
 
-    @_bank.command(pass_context=True, no_pm=True)
```
Kowlin, line 401: Deleting a new command.

cogs/economy.py
```diff
@@ -281,22 +235,31 @@ def __init__(self, argument):
 
 
 class Economy:
-    """Economy
+    """Economy - Blue variant
```
Kowlin, line 238: "Blue variant" has to be removed in order to get this mainstreamed for other bots

cogs/economy.py
```diff
@@ -629,60 +956,159 @@ def already_in_list(self, accounts, user):
...
         server = ctx.message.server
         self.settings[server.id]["PAYDAY_TIME"] = seconds
-        await self.bot.say("Value modified. At least {} seconds must pass "
-                           "between each payday.".format(seconds))
+        await self.bot.say("**Payday cooldown:** " + str(seconds) + " seconds")
         dataIO.save_json(self.file_path, self.settings)
 
     @economyset.command(pass_context=True)
-    async def paydaycredits(self, ctx, credits: int):
+    async def paydaycredits(self, ctx, creds: int):
```
Kowlin, line 1035: Unneeded variable rename(?)

cogs/economy.py
```diff
@@ -629,60 +956,159 @@ def already_in_list(self, accounts, user):
...
+##        """List shop items or buy an item!"""
+##        if ctx.invoked_subcommand is None:
+##            msg = "```"
+##            for k, v in self.shoplist:
+##                msg += "{}: {}\n".format(k, v)
+##            msg += "```"
+##            await self.bot.say(msg)
+
     def display_time(self, seconds, granularity=2):
-        intervals = (  # Source: http://stackoverflow.com/a/24542445
```
Kowlin, line 682: Removing comments and source kills off readability

Kowlin (changes requested):

Twentysix26:
I suggest turning it into a separate cog if you want to share it with the community

cogs/economy.py
Naleksuh: Ah yes I forgot about that

Issue #595, opened 2017-02-02
[suggestion] adding build support for git repos of python modules


Jan200101:
title should give all the info you need

a module repo affected by this is https://github.com/Holiverh/python-valve as it doesnt have a up to date pip (because its not made my the author)

Jan200101:
Right now its throwing this Error when adding a git+[github link] to the requirements field
http://i.imgur.com/PyaOALu.jpg

Jan200101:
Traceback
http://i.imgur.com/p4J61NB.jpg

tekulvw:
Should be fixed with #786 probably won't be implemented in V2

Kowlin:
Fixed with #786 (v3), not implemented in v2.

PR #598, opened 2017-02-03
[Downloader] Add support for git repo modules

Aioxas:
Referencing #595

### DIFF
--- a/cogs/downloader.py
+++ b/cogs/downloader.py
@@ -508,7 +508,7 @@ class Downloader:
         return git_name[:-4]
 
     def is_lib_installed(self, name):
-        if "git+" in name:
+        if name.startswith("git+"):
             name = name.split("/")[-1]
         return bool(find_spec(name))
 
--- a/cogs/downloader.py
+++ b/cogs/downloader.py
@@ -508,6 +508,8 @@ class Downloader:
         return git_name[:-4]
 
     def is_lib_installed(self, name):
+        if "git+" in name:
+            name = name.split("/")[-1]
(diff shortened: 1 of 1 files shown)


Jan200101:
Seems okay I would suggest checking if it starts with git+ and doesnt just contain it.

cogs/downloader.py
```diff
@@ -508,6 +508,8 @@ def get_dir_name(self, url):
         return git_name[:-4]
 
     def is_lib_installed(self, name):
+        if "git+" in name:
```
Jan200101: Would suggest changing it to
`if name.startswith("git+"):`

But otherwise its fine

Jan200101 (approved):

Jan200101:
Finding the name over the git link may be a issue as some modules have other names in their github links then the one that need to be imported

Aioxas:
Can you give me an example of one of those.

Jan200101:
[Python Valve](https://github.com/Holiverh/python-valve) its github link links to python-valve but its imported using `import valve`

Aioxas:
Wait a second, installing using git+ link is one thing. It's up to you to have the correct name of the package you are importing. So there's no need to change anything.

Jan200101:
Im gonna give it a try and test if it works. Could be that a brain fart

Jan200101:
yep brain fart.
It works

Jan200101 (approved):

tekulvw:
I'm gonna nix this because there's no dire need to implement it in V2 and it's pointless in Downloader V3

PR #600, opened 2017-02-04
Update worldcapitals.txt

vayandas:


Twentysix26:
Sadly, our userbase is not only french

PR #603, opened 2017-02-08
[Streams] Add Twitch & Beam Viewers

Aioxas:


Twentysix26:
Like discussed in #coding I didn't show the current viewers because it doesn't make much sense for stream alerts. When they pop up the stream just started, people still have to join and the stat is essentially useless. It could make a better fit for the !twitch !beam !hitbox commands

PR #490, opened 2016-11-25
[Mod] Warning system

calebj:
See #488

PR #613, opened 2017-02-10
[Mod]Warn Command added

sbhadr:
The common scenario when you want to 'warn' someone to make a point rather than kicking them. :)


Twentysix26:
#490 is more complete and still needs work. Yours is purely visual and it's not what I'm looking for in a warning system.

PR #615, opened 2017-02-12
[Core] Added on_error logging

Kowlin:
Added on_error logging to handle errors that don't occur in commands. What this does is logging them to the logging module, that redirects them to the handlers (By default ``data/red/red.log`` and stdout)

One thing that I'm not very clear on is the ``ctx`` argument in the ``on_error`` function though. Could not find any information on it.

### DIFF
--- a/red.py
+++ b/red.py
@@ -375,6 +375,10 @@ def initialize(bot_class=Bot, formatter_class=Formatter):
         else:
             bot.logger.exception(type(error).__name__, exc_info=error)
 
+    @bot.event
+    async def on_error(error, ctx):
+        bot.logger.exception(type(error).__name__, exc_info=error)
+
     return bot
 
 
--- a/red.py
+++ b/red.py
@@ -376,8 +376,8 @@ def initialize(bot_class=Bot, formatter_class=Formatter):
             bot.logger.exception(type(error).__name__, exc_info=error)
 
     @bot.event
-    async def on_error(error, ctx):
-        bot.logger.exception(type(error).__name__, exc_info=error)
+    async def on_error(error_event, *args, **kwargs):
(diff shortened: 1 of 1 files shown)


tekulvw:
I don't think that `on_error(error, ctx)` is the correct definition, see [here](https://discordpy.readthedocs.io/en/latest/api.html#discord.on_error)

Aareon:
`ctx` is the context of a given message. Was that was what was in question?

red.py
```diff
@@ -375,6 +375,10 @@ def initialize(bot_class=Bot, formatter_class=Formatter):
         else:
             bot.logger.exception(type(error).__name__, exc_info=error)
 
+    @bot.event
+    async def on_error(error, ctx):
```
tekulvw: Needs to be changed to `on_error(event_name, *args, **kwargs):`

red.py
```diff
@@ -375,6 +375,10 @@ def initialize(bot_class=Bot, formatter_class=Formatter):
...
+        bot.logger.exception(type(error).__name__, exc_info=error)
```
tekulvw: Needs to be changed to `bot.logger.error("Error in event: {}".format(event_name), exc_info=sys.exc_info())`

tekulvw (changes requested):

tekulvw:
@Aareon nope, see above and linked doc page in the previous post

Aareon:
@tekulvw I see now, thank you. Changes look good.

tekulvw (approved):

calebj (changes requested):
`event_name` should be `error_event`, or vice-versa.

tekulvw:
@Twentysix26 thoughts?

Kowlin:
I have some thoughts, this will work in the same way we had on the logging on the v3 branch, thus. this will see everything across the bot.

PR #640, opened 2017-02-22
[Trivia] Update worldflags.txt

aanguyen:


Twentysix26:
I separated chad and romania, as they are slightly different flags.

PR #646, opened 2017-03-02
Fix pagify bug

Zephyrkul:
Right now it seems pagify() will yield an empty string if it can't find a deliminator, and on looking at the code it looks like it's because it finds the deliminator at the start of the string if it's the second page or higher (since it was just cut off there for the previous page). By replacing the second `rfind` argument with `len(d)`, I think that'll fix that, and _limited_ testing shows that it does.

### DIFF
--- a/cogs/utils/chat_formatting.py
+++ b/cogs/utils/chat_formatting.py
@@ -40,8 +40,9 @@ def pagify(text, delims=["\n"], *, escape=True, shorten_by=8,
         shorten_by += num_mentions
     page_length -= shorten_by
     while len(in_text) > page_length:
-        closest_delim = max([in_text.rfind(d, len(d), page_length)
-                             for d in delims])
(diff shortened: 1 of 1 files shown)


Zephyrkul:
Better solution (committed in ebf449b): make it flat-out ignore any deliminators at the start of the page so it doesn't yield empty strings. (`len(d) if in_text.startswith(d) else 0` for the second argument in `rfind`)

First version would miss multiple-character deliminators if they began at a point in the string less than their length.

Zephyrkul:
Even better solution (617f0e2): Just don't yield empty messages (including whitespace-only messages), ever, while always advancing through the string to prevent an infinite loop!

Previous version (as well as current version) would have some bugs still, like yielding the newline character if there are multiple newlines right at the page length. Instead of going through all the obscure causes, simply do the above! Much simpler, I think.

calebj:
lgtm

Zephyrkul:
Closing this now, since the v3 version has been merged.

PR #663, opened 2017-03-18
make it not raise ChannelUserLimit always

Eternity71529:
make it not raise ChannelUserLimit when the user is on the same Voice Channel as the Bot.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -2011,6 +2011,10 @@ class Audio:
                 is_full = False
             else:
                 is_full = len(channel.voice_members) >= channel.user_limit
+            if server.me.voice_channel == channel:
+                same_vc = True
+            else:
+                same_vc = False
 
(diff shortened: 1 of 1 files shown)


Twentysix26:
It would make way more sense if we avoided calling `has_connect_perm` altogether if the bot is already in the user's voice channel (which is what I assumed it did when I worked on that part)

Eternity71529:
I wanted to add in a part where if the bot was in a `VoiceChannel` with no `Connect` permissions, it would still not raise `UnauthorizedConnect` always too :thinking: , and add a part where it would move itself in the `User`'s `VoiceChannel` if it's already in a `VoiceChannel` and avoid throwing `ClientException`

tekulvw:
Over a year old (oops), V2 is feature locked, suggestion not implemented.

PR #667, opened 2017-03-19
[Downloader] add massadd command for adding repos by category

palmtree5:
This will add all repos in a specific category by getting the list from cogs.red

Valid categories are:

approved
beta
unapproved
all

all will add all repos from all categories. The other 3 will only add repos with that category

Command format is `[p]cog repo massadd <repo_category>`


palmtree5:
Should also mention that I gave this command its own confirmation prompt to note that it is adding multiple repos. This confirmation prompt is separate from the one for adding one repo

skeith:
This might be for cogs.red
It would be awesome if cogs.red adds a custom list feature by checking which repo you wants then generating "something" which identified by "some strings" then `cog repo massadd [THE STRING]`

palmtree5:
@skeith if you read the code, you'll see that this uses cogs.red for getting the lists

skeith:
@palmtree5 yeah I know that, what I meant is it would be fun if cogs.red add a new feature for a custom list of repos. Just like when you download fonts from Google WebFonts

palmtree5:
I just realized this needs rebasing too...

palmtree5:
And this one's been rebased as well

palmtree5:
theoretically this could also be modified so that we could add multiple repos by link simultaneously, though that calls into question how naming should be done (for example, should we just use the repo's name by default?)

Aioxas:
Why not separate by comma. Or something. "Dumb Cogs" github link, Rand-Cogs github link, and something else

palmtree5:
Gonna do something like that. Gonna have the format be name1,url1;name2,url2 etc

Aioxas:
That sounds even better.

palmtree5:
And done. Format is like I posted above. No need to put repo names in quotes either since we're not splitting on spaces and the command is setup so that anything after itself is part of the only arg

Aioxas:
Yep. Your way fixes the quote part. Unless the person themselves wants
quotes

PR #671, opened 2017-03-21
[mod] Add deleteself setting, refactor deletedelay logic

### LINKED ISSUE #664: [Suggestion] Deletion of messages after time
Hello,

can you add an option, that the bots delete the invoking messages and his own messages after 30 seconds? Also it would be nice if you can add that the bot can delete messages, which are older than 14 days. I think you can delete the messages with seperated delete requests instead of an bulk request. 

Best Regards
ModdyLP

calebj:
Fixes #664

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -207,7 +207,7 @@ class Mod:
         await self.bot.say("Cases have been reset.")
 
     @modset.command(pass_context=True, no_pm=True)
-    async def deletedelay(self, ctx, time: int=None):
+    async def deletecmd(self, ctx, time: int=None):
         """Sets the delay until the bot removes the command message.
             Must be between -1 and 60.
 
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -236,6 +236,36 @@ class Mod:
                 else:
                     await self.bot.say("I will not delete command messages.")
 
+    @modset.command(pass_context=True, no_pm=True)
+    async def deleteself(self, ctx, time: int=None):
+        """Sets the delay until the bot removes its own messages.
+            Must be between -1 and 60.
+
+        A delay of -1 means the bot will not remove the message."""
+        server = ctx.message.server
+        if time is not None:
+            time = min(max(time, -1), 60)  # Enforces the time limits
+            self.settings[server.id]["selfdelete_delay"] = time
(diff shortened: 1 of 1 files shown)


Twentysix26:
This will also deletes messages that are supposed to be static, such as polls and announcements made by cogs. I can't think of any good solution for this though. We can't reliably distinguish those in `on_message`.

calebj:
Hmm, then maybe it should be dropped altogether?

Twentysix26:
No, this feature gets requested a lot and it would be good to implement. I posted to see if anyone had any good solution for this issue. If not I'll just merge and let the users deal with this drawback (with due warning)

palmtree5:
The only real solution would be to just have a way for cogs to be able to register commands that shouldn't have their messages deleted and then check those on command. Also one could just use @tekulvw's scheduler cog to periodically have the bot clean up after itself

Zephyrkul:
Is there a way to check what command spawned the message in `on_message`? I don't think so. Also, the scheduler cog would delete a batch of messages every X seconds instead of deleting each message after it's existed for X seconds.

One potential solution is a marker character at the start of messages that should remain, but that would require changing each cog that requires that... `\u00AD` is invisible at the start of messages and is never used there otherwise (I'm hesitant about `\u200B` because I know it's useful for editing an embed into a message and deleting the message's content, but it could still be used as well).

palmtree5:
What I was saying is that the commands whose output should be kept could be on a whitelist and the command would be parsed manually on message

Zephyrkul:
I'm still not understanding, sorry, so just to make sure we're on the same page - I'm talking about the bot's own messages, not the commands that users give to the bot. I don't see any way of reliably checking for the actual command in a separate message (the user's) when checking the bot's own message in `on_message`.

Chovin:
*cough monkeypatch and/or subclass commands and provide an interface to the user to select which commands should have their `send_message`s also schedule a deletion.

sounds messy? probably is.

either way, there doesn't seem to be a good solution that isn't messy and wouldn't break the functionality of most cogs as they are now.

Zephyrkul:
no irdumb

Twentysix26:
This still sounds very problematic to implement properly. Closing.

PR #682, opened 2017-03-30
[Owner] Added out of date message

Jan200101:
Version command shows the commits of the current clone and if its out of date or not


Twentysix26:
I don't see the need for this.

PR #685, opened 2017-03-31
Added shebang to the fast startup script.

emberdex:
Added a shebang to the fast startup script for Linux and macOS-based systems. This forces the use of bash to execute the script, allowing the use of `./<filename>` and guaranteeing functionality. Useful if your system's shell is not bash by default (fish, zsh, etc.)


tekulvw:
Not gonna add for V2, will be (probably) unnecessary in V3

Issue #684, opened 2017-03-31
Poll command issue


palmtree5:
Apparently bots with multi-character prefixes lead to there being a space before the question in the poll. See image:

http://i.imgur.com/GhohgfD.jpg

palmtree5:
Tagging @skeith for noticing the issue

PR #689, opened 2017-04-01
[general] fix poll question slice when length of prefix > 1

flapjax:
Currently, questions/answers are extracted with a hard-coded slice of the message content:
`msg = message.content[6:]`

Polls invoked with a prefix longer than 1 character have unwanted characters bleeding into the question. This PR fixes that, and also happens to fix a related problem I found, where italics markdown would break if answers contained whitespace at the beginning or end.

This is in reference to issue #684 .

### DIFF
--- a/cogs/general.py
+++ b/cogs/general.py
@@ -339,7 +339,7 @@ class General:
             if "@everyone" in check or "@here" in check:
                 await self.bot.say("Nice try.")
                 return
-            p = NewPoll(message, self)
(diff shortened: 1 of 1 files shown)


flapjax:
@palmtree5, @skeith, would you like to review this and confirm it fixes #684? I've tested it pretty extensively on general.py and on my reaction polls, which are basically the same code. I know a poll rewrite is in the works, but I would still propose that we merge this one as a hotfix until then.

PR #699, opened 2017-04-08
Minor details

boswelja:
Show cog name in load/unload/reload commands where applicable
Update hug command
Tell us how many messages are deleted by .cleanup


Kowlin:
I'm sadly not a fan of these changes. You (sadly) remove a lot of readability of the code and needless styling changes. And I don't see the need for this.

Stylite:
I'm a fan of the cog name in reload but the other ones don't seem needed

PR #709, opened 2017-04-13
[Mod] move ignore/unignore of servers and channels to Owner

palmtree5:
Moves the ignore and unignore commands to the owner cog. They don't really belong in mod IMO as they target ignoring channels/servers and aren't really related to moderating users so much as granular control over where the bot works. Related to #700

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -1582,10 +1582,8 @@ def check_folders():
 
 
 def check_files():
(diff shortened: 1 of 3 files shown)


palmtree5:
Ok, this should be golden now that I remembered that core needed changes to make this work

palmtree5:
After further thought, I'm feeling like this doesn't belong in owner

PR #710, opened 2017-04-14
Add core cog

palmtree5:
Per further discussion in #700 and in Discord, I'm opening this PR as a template for work towards moving some things around. As it stands now, the core cog will successfully load but doesn't contain any actual functionality. It has also been setup such that it cannot be unloaded by the unload command and so that the bot will just exit if it is missing when launching

### DIFF
--- a/cogs/core.py
+++ b/cogs/core.py
@@ -1,0 +1,29 @@
+import discord
+from discord.ext import commands
+from .utils.dataIO import dataIO
+from .utils import checks
+from datetime import datetime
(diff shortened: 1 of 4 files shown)


palmtree5:
As for configs relating to this cog, I think it should just share data/red with owner

palmtree5:
Ok, popped server/channel ignore over to core

tekulvw:
Closing as this already exists on the `v3` branch.

PR #716, opened 2017-04-16
[Streams] twitch > v5

palmtree5:
Move twitch api usage to v5

### DIFF
--- a/cogs/streams.py
+++ b/cogs/streams.py
@@ -29,6 +29,10 @@ class InvalidCredentials(StreamsError):
     pass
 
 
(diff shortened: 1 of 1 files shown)


palmtree5:
Tested and it works. 

Names and user ids are now cached in a file called `twitch_users.json` as the user's id should never change

Invalid name? Tells you the name is invalid. 

User changed names? Name gets updated in the cache so the ID is properly mapped to the user's current name

Twentysix26:
Done it differently in #718
Thanks, this saved me some time 👍

PR #718, opened 2017-04-16
[Streams] Twitch API v5

Twentysix26:
This PR implements Twitch's api v5, which makes use of IDs instead of names.
Since names is all we used to store, this attempts to fetch and save all the missing stream IDs on loading (with a single request if they're inferior to 100).
Some testing would be appreciated, especially the migration part.


palmtree5:
looks good to me

PR #719, opened 2017-04-17
[Launcher] add function for generating systemd autostart

palmtree5:
Pretty self explanatory. Ideally it would've just put the file in the place it needed to be in but I was getting permissions errors trying to write directly to that location. This also will only work if the OS uses systemd and doesn't have checks in place for that yet

### DIFF
--- a/launcher.py
+++ b/launcher.py
@@ -150,9 +150,10 @@ def setup_autostart():
         "WantedBy=multi-user.target"
     with open("{}/red.service".format(homedir), "w") as as_out:
         as_out.write(file_contents)
-    print("Created red.service file in your home directory!")
-    print("To use it, do sudo mv ~/red.service /etc/systemd/system/")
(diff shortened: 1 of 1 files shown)


palmtree5:
Ok, it'll now move it to the right location. Now for systemd checks...

palmtree5:
and systemd checks in. Fails if the return code from `which systemd` is nonzero and exits back to the menu in that case

palmtree5:
Gonna work on extending this to include Upstart, OpenRC, and launchd. Maybe I'll look into trying to autogenerate something that will autostart the bot on Windows too

palmtree5:
Upstart checks in. OpenRC in progress

Twentysix26:
As discussed on Discord I don't think this belongs in the launcher

palmtree5:
not to mention this PR was stale anyway :P

PR #726, opened 2017-04-21
Fixes an issue with patch notes not showing

Ryonez:
There was a discovery that the way repos are being added could lead to
multiple path separators being used within the "file" and "folder"
strings. This was brought about by "/" being used statically in some
locations of the downloader and the use of "os.path.join" in others.
However, it while the windows operating system, can handle this, the git
program installed cannot.
Until one path delimiters method is settled on and applied, I've
temporarily resolved the patch note issue by recreating the repo
location, and the cogfile within it using "os.path.join" so git now get
the correct values.

Resolves issue Twentysix26/Red-DiscordBot#725

### DIFF
--- a/cogs/downloader.py
+++ b/cogs/downloader.py
@@ -519,8 +519,24 @@ class Downloader:
         return bool(find_spec(name))
 
     def _do_first_run(self):
-        invalid = []
         save = False
+        repos_copy = deepcopy(self.repos)
+
+        # Issue 725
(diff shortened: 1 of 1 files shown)


Ryonez:
Implemented path changes from @palmtree5, correct some bugs introduced by it
Accounted for systems already affected by prior incorrect paths saved to "repos.json".

Further update to Twentysix26/Red-DiscordBot#725

PR #728, opened 2017-04-21
added async support for io thru aiofiles

srhinos:
also added backwards compatibility so if this is updated and deps aren't nothing will break. I could add in a forceful update of deps but I dont want to control the local environment of the end user. This is very hastily written though so highly advise it to be reviewed and tested

### DIFF
--- a/cogs/utils/dataIO.py
+++ b/cogs/utils/dataIO.py
@@ -10,7 +10,7 @@ class DataIO():
     def __init__(self):
         self.logger = logging.getLogger("red")
         try:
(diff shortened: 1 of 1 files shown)


srhinos (commented):
why did I forget a "

srhinos (commented):
whats up with namespaces, anyway

srhinos (commented):
""s

palmtree5 (changes requested):
"dependancies" -> "dependencies"
"dependancy" -> "dependency"

PR #729, opened 2017-04-21
added async support for io thru aiofiles

srhinos:
also added backwards compatibility so if this is updated and deps aren't nothing will break. I could add in a forceful update of deps but I dont want to control the local environment of the end user. This is very hastily written though so highly advise it to be reviewed and tested

### DIFF
--- a/cogs/utils/dataIO.py
+++ b/cogs/utils/dataIO.py
@@ -9,6 +9,15 @@ class InvalidFileIO(Exception):
 class DataIO():
     def __init__(self):
         self.logger = logging.getLogger("red")
+        try:
+            __import__("aiofiles")
+            import aiofiles
+            self.async_compatible = True
+        except ImportError:
+            self.logger.exception("Attempted to use aiofiles but "
+                                  "dependency wasn\'t installed. "
+                                  "Please update all dependencies!")
+            self.async_compatible = False
 
     def save_json(self, filename, data):
         """Atomically saves json file"""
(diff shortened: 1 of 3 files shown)


cogs/utils/dataIO.py
```diff
@@ -9,6 +9,15 @@ class InvalidFileIO(Exception):
 class DataIO():
     def __init__(self):
         self.logger = logging.getLogger("red")
+        try:
+            __import__("aiofiles")
+            import aiofiles
```
Chovin, line 14: can you explain what the `__import__()` is used for?
I'm curious as to why you used both

cogs/utils/dataIO.py
```diff
@@ -9,6 +9,15 @@ class InvalidFileIO(Exception):
...
+            self.async_compatible = True
+        except ImportError:
+            self.logger.exception("Attempted to use aiofiles but "
```
Chovin, line 17: we have a `bot.pip_install` thing. since it's kinda important for it to be there, should we just use that instead of just logging?

cogs/utils/dataIO.py
srhinos: `__import__` wont actually import, it just checks for the existance of the imported package.

cogs/utils/dataIO.py
srhinos: I wrote it that way knowing from my time watching MusicBot users update their bots by downloading and dragging the bot folder into the main install directory and not actually updating their deps file. This ensures that a deps change wont be the reason the bot breaks and can expect this code to be removed once most are back up to date.

tekulvw:
We're gonna nix this in preparation for V3. We've decided against using `aiofiles` directly because it's just a wrapper around `run_in_executor` and we see no need to include another dependency. Asynchronous file loading/saving already exists as part of the new data storage mechanism in V3.

Issue #713, opened 2017-04-15
Error: "Sharding required.."


ncodex2:
discord.errors.ConnectionClosed: WebSocket connection is closed: code = 4011, reason = Sharding required..

Zephyrkul:
How many servers is your bot connected to?

ncodex2:
Around 2800.

Jan200101:
are you sure ?
you can do
`[p]debug len(bot.servers)`
to find the real number

ncodex2:
Yes, i am sure, my bot is quite large.

Zephyrkul:
Red only supports connecting up to 2500 servers - any more requires [sharding](https://discordapp.com/developers/docs/topics/gateway#sharding), which Red does not support out of the box. This is a requirement of Discord's API.

Kowlin:
Like Zephyr said, bots that have over 2500 servers require sharding. Out of the box we do not support sharding due the number of critical changes we would have to make to Red. 

For Red to reliable support sharding we would need to move away from our current method of saving everything into Json's (What works fine for the majority of cogs and add-ons) into a structured or non-structured database (#517)

We would also have to specifically ask for the number of shards, and thread them into their own red processes. (Not sure on that) And that logic currently doesn't exist in Red. There are probably more challenges for Red to support sharding down the line. 

Why have we not added support for sharding when it was announced? Because we simply don't have a need for it, the majority of users who use Red (A good 99.5%) don't require it. Building the back-end for sharding is by it self a huge task, a task that would only apply for 0.5% of users. And that isn't a good method for spending our time in my opinion.

There however is some good news in this bundle of information; the rewrite of Discord.py will support sharding out of the box. (https://github.com/Rapptz/discord.py/projects/1) And this should make it easier for you to have support for sharding. Does this mean that it will work once the Discord.py rework is out? Absolutely not, we would have to port over Red first.

What could you do for now? You can try to get the bot running in multiple shards and hard lock it down to <2501 servers until we either support sharding. Or get someone to develop it, however sadly we cannot help with sharding as we currently don't have the back-end to support sharding.

ncodex2:
That's okay, but how can i lower my server count to 2500 when i can't even start the bot up?

Twentysix26:
A third party script I assume. I suggest staying up to date with Discord API's news if you want the responsibility of hosting a public bot. Details on sharding have been around since April 2016.
  
Sharding support won't be implemented until the mentioned issues are solved.

PR #730, opened 2017-04-21
[Launcher and Red.py] Red-DiscordBot-Sharded v1.0

Pierce01:
intro
### PR closed. (I made a mistake with disabling shards, go to the repo for the update. "will you finish v2.0?" eh.

# To-do
* Change file system to a database (or change file system that is shard friendly)
## Description
This is just Red-DiscordBot that can be sharded for the users that have big bots using the base.

## Why did you make this?
Since it seemed that others didn't take sharding as a priority, i whipped up something that allows you to shard with a red base.

## Setting up

If you modified your `red.py` you'll need to copy and paste the code i added.

Same with the launcher. Note, when using `Update Red` in the launcher, the new update may overwrite my code, so you'll need to repaste what i added.

### Setting up your first instance

make sure your shards.json file looks something like
```
{
    "SHARD0": "0",
    "SHARD1": "1",
    "SHARDC": "2",
    "SHARDS": "False"
}
```
SHARDC is for the shard count, so thats the amount of shards you have. 

the shard counting system starts from 0, so if SHARDC (shard count) is set to 2, the two shards would be `0,1`

defaultly, `red.py` is set up to run with shards disabled. You can turn them on in the launcher. When enabling, it'll ask you for a shard ID, you need to type 0, this is `SHARD0`.

example

Both `SHARD0` and `SHARD1` have been defined, but enabling will make you set `SHARD0` to a value, which I'll change, my main goal was to make the the user get a hang of the `Add Value` button which I'll be going over.

next you can run the first instance.

### Setting up your second instance

Setting up the second instance is easy. Go to line 65, and change `SHARD0` to `SHARD1`.

Example2

now run the second instance. that was easy right??

### Now here's the fun stuff.

you might be asking, "how do i add a third one??".

you'll need to add a new value to `shard.json`. open the launcher, and open the shards menu.
select `Add new shard Value` and go through the steps of making a value. Since we started at 0, the third shard will be `2`

Example3

Now redo what i did for the second instance, but instead of `SHARD1` it would be `SHARD2`

## FAQ

1. Why don't all my servers show up??
A. Sharding makes your bot login to only a certain amount of servers, so the other shards fill in the rest.

2. Why didnt my prefix update?
~~A. You need to reload/restart your other shards, its like turning off one stove but leaving the other on, you need to turn the other off also.~~ Data corruption

3. What happened to my custom commands?
A. As for v1.0 of this, there is data corruption. My idea for fixing it is either to wait for a different type of database or to make separate shard files. Idea would be the bot would read both files, but only write to the specific shard file, thus loading cc's from another shard, but not over writing it. "is this possible?" I'll have to look into it more, this is all based off of ideas at the moment.

### DIFF
--- a/launcher.py
+++ b/launcher.py
@@ -575,7 +575,7 @@ def shardss_menu():
                 print("Enabled Sharding")
                 wait()
         elif choice == "2":
-            with open("data/data/shard/shard.json", "r+") as settings:
+            with open("data/shard/shard.json", "r+") as settings:
                 settings["SHARDS"] = "False"
                 f.seek(0)
                 f.write(json.dumps(settings, indent=4, sort_keys=True))
--- a/launcher.py
+++ b/launcher.py
@@ -572,7 +572,7 @@ def shardss_menu():
                 f.truncate()
                 print("Enabled Sharding")
         elif choice == "2":
-            with open("data/data/shard/shard.json", "r+") as settings:
+            with open("data/shard/shard.json", "r+") as settings:
                 settings["SHARDS"] = "False"
                 f.seek(0)
                 f.write(json.dumps(settings, indent=4, sort_keys=True))
--- a/launcher.py
+++ b/launcher.py
@@ -529,7 +529,7 @@ def main():
         print("3. Update")
         print("4. Install requirements")
         print("5. Maintenance (repair, reset...)")
-        print("6. Open Shard Menu (Back up Data, this is in Beta)")
+        print("6. Open Shard Menu v1.0 (Back up Data, There's a data issue)")
         print("\n0. Quit")
         choice = user_choice()
         if choice == "1":
--- a/launcher.py
+++ b/launcher.py
@@ -10,6 +10,7 @@ except ImportError:
 import platform
 import webbrowser
 import hashlib
+import json
 import argparse
 import shutil
 import stat
@@ -27,6 +28,10 @@ FFMPEG_BUILDS_URL = "https://ffmpeg.zeranoe.com/builds/"
 
 INTRO = ("==========================\n"
          "Red Discord Bot - Launcher\n"
+         "==========================\n")
+
+SHARD_INTRO = ("==========================\n"
+         "Red Discord Bot - Shard Menu\n"
          "==========================\n")
 
 IS_WINDOWS = os.name == "nt"
@@ -287,6 +292,7 @@ def update_menu():
         print("3. Update requirements")
         print("\nOthers:")
         print("4. Update pip (might require admin privileges)")
+        print("5. Disable Sharding")
         print("\n0. Go back")
         choice = user_choice()
         if choice == "1":
@@ -524,6 +530,7 @@ def main():
         print("3. Update")
         print("4. Install requirements")
         print("5. Maintenance (repair, reset...)")
+        print("6. Open Shard Menu (Back up Data, this is in Beta)")
         print("\n0. Quit")
         choice = user_choice()
         if choice == "1":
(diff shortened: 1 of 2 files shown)


palmtree5:
Should fix #713 if merged

palmtree5:
Is there a reason for not using dataIO?

Pierce01:
no reason. didn't even think about it woops. 😅

tekulvw:
Have you come up with a solution that avoids cog data corruption? This has been a priority, but that is the biggest problem we've yet to completely solve.

Pierce01:
## Solution of not changing
I do think in my own opinion, if shards do change servers, most if not all have saved the certain configs for said server. an example would be the custom commands.

## Solution of changing
to prevent data corruption either a temp backup for the day or a temp file could be made for said server/ over all config. Open to ideas though

tekulvw:
I'm not entirely sure what you're trying to communicate so here's why sharding simply won't work with Red as it is today:

1. Each cog loads a complete copy of it's own data, which means that each cog in each shard will load it's own copy.
2. As cogs start accumulating data, that data will ONLY be stored on the shard that made the change. This means that all other shards fall out of sync.
3. As cogs start *saving* data only the LATEST version will be saved to a file and it will completely overwrite any changes made by other shards, leading to blatant data corruption (that will ONLY be visible once you restart the bot).
4. This PR forces a user to modify source code to start secondary shards??? In all honesty, that's not an acceptable way of doing this, not to mention the data corruption that is going to occur.
5. In order to fix the data corruption issue, a new data storage backend will need to be implemented that allows cogs to save only the data that changed to a global data source (e.g. a database). And then each and EVERY cog writer will need to update their cogs to make use of this system, any cogs that don't get updated WILL have data corruption issues.

Pierce01:
I understand, Thank you for explaining. 

I've only started working with the base as of today, due to being asked to make a way for sharding to work for the base. I wasn't aware of each file resetting or being over written by the bot itself (thus causing the big issue), thought it was added on to. I'll look into solving that using different saving method or ~~opening the file in said command, so it would be updated globally when the command is used, and maybe change the overwrite to just add on~~ (I'll experiment)  As for `4`, that was a "draft" idea, so its going to be changed for the future.

Thank you for your time 👍

Kowlin:
@tekulvw But we don't if we just adapt DataIO to use "Drivers"... I *really* don't see the need for cog creators to rewrite their cogs just to support sharding.

palmtree5:
I agree. If there's a way to do this in which we reduce the workload to support sharding, that would be ideal

Pierce01:
I'm messing around with a new file system to include sharding support for the current redbot version.
Since the d.py rewrite will support auto sharding, I'm going to continue this as a project when i have free time.

## What i did so far. 
I made the cogs save and right to normal file paths, only with shard ID included into their file names, so they will be only writing to those shard files, so the other shards wont break. File system is being converted something like this.
```
        c = "data/downloader/repos{}.json".format(self.bot.shard_id)
        f = c.replace("None", "")
```
So json files would be named `filename0.json` or `filename.json` if there are no shards
should it be different?

## Planned
After i finish this, i plan on making it so the bot reads from different shard folders. Depending on the way i plan on doing it, my current idea is ghetto. So I'll look into it when i get to that point

Ryonez:
Sorry if this is a dumb question, but why is this a pull request? The code is still being worked on, and if it's implemented now, it would break to current working code that is in active use.

Note, I'm still new to git really, so I don't know if there's "method behind the madness", as it were.

Pierce01:
I tried PRing v1.0, however people stated the above mistakes, so I'm going back and fixing them 👌

Stylite:
@ryonez it's to show and for progress reports

Ryonez:
I see, thank you for explaining it >.<

Twentysix26:
Sharding will most likely be implemented in the incoming rewrite, as the new discord.py [seems to allow it effortlessly](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.AutoShardedClient).  
@Maxie01 As it stands this PR doesn't solve the main problem, which is the (massive) data loss that such a thing would introduce. Even the last method you described is very hacky, assuming it would work.  
I'm not looking to implement big changes considering that this release is nearing its end and I'm pretty sure that major ones would be required to make sharding work properly here.

Pierce01:
Seems sensible to wait till the rewrite. Thanks for taking your time to read all those comments. 

tip

PR #736, opened 2017-04-28
[core, launcher] Workaround for Debian pip quirk

calebj:
Checks if the --system parameter exists; if so, use it.
See https://github.com/pypa/pip/issues/3826#issuecomment-232115009


tekulvw:
You can't just blindly assume every user who *can* install to system wants to. And regardless, this bug is fixed in the latest version of pip.

PR #739, opened 2017-04-28
Red - Rewrite

Twentysix26:
A general and incomplete list of what changed and what is going to change with this ongoing rewrite.  
We are in the early stages, each of these points can be discussed, improved and even scrapped.  
Your feedback is very important :+1:

## Packages instead of cogs

This allows for multi file cogs, bringing a much better organization for large cogs.  
Each package can contain a variable amount of cogs and files. The user will now install packages.  
They will be contained in the "cogs" folder (to be renamed?).  
Each package will hold its own data inside its own folder.
The downloader will now install packages, not single cogs.

The juicy details on how the downloader will deal with this are still to be decided and up to discussion

## No more DataIO, aka how we deal with data

The old DataIO was nothing but a thin wrapper around some json functions. It was also not threadsafe, causing blocking issues.
We are now using Will's "config", #770.
This allows us to support a variety of backends (flat files, which will be default, and databases) and makes dealing with data much less of an hassle.

## Removed global admin and mod roles

They are pretty much unused. Every server sets their own roles when the bot gets invited, might as well remove them altogether.

## Removed email / password login support

~~I don't think it's worth supporting these when you can login with a token. A CLI flag is provided to allow user accounts to login.~~ I have been informed that the new discord.py doesn't support email / password at all, so this makes the decision even easier 👌 

## Co Owners

Yes, the common stance is that this is most of the time a pretty awful idea. And it is.  
It also has a legit use case: a hosted bot that multiple people have access to anyway.
I don't think it's fair not to implement this feature just to protect the clueless. There will of course be a lot of warnings.

## Re organization of standard packages

Still haven't thought much about it. I would want to avoid just a simple convertion of the old cogs and I'd like to bring something new to the table

## Both global and local blacklist / whitelist

This is already in and functioning, it just lacks the commands. I may end up extending this idea for complex permissions, like the widely used permissions cog offers

## Sharding

Lots of talk about sharding in recent times. We were ready to support DBs in this rewrite just to deal with the issue. At the end it turned out that discord.py can do [the heavy lifting](https://discordpy.readthedocs.io/en/rewrite/api.html#discord.AutoShardedClient) for us and we can stick to the old ways quite safely.


palmtree5:
I'd suggest adding in a prompt for when the bot joins a server to provide info to the server about the bot, including setting the mod/admin roles.

As for co-owners, perhaps have a "primary owner" who would be able to use things like [p]debug and make co-owners not able to use those sorts of things by default but have a toggle for it? Thinking here is that if someone wants to add a co-owner who doesn't have some sort of access to the computer the bot is running on, the co-owner wouldn't be able to use debug then but if someone was adding a co-owner who did have access, then having access to debug in that case would be fine

core/utils/helpers.py
```diff
@@ -0,0 +1,200 @@
...
+    file_path: str
+      The path of the json file you want to create / access
+    create_dirs: bool=False
+      If True, it will create any missing directory leading to
+      the file you want to create
+    autosave: bool=False
+      If True, any change to the "database" will be queued to the
+      flusher and scheduled for a later write
+    default_value: Optional=None
+      Same behaviour as a defaultdict
```
Kowlin: Possible rewording here, not everyone knows what a ``defaultdict`` was in its current context.

Twentysix26: Yeah it could be clearer

Kowlin:
Just going to (badly) quote Danny/Rapptz on this regarding the Json Flusher so we don't forget,

```
The most frequent cause of data corruption is when you run out of memory mid-write.
The operating system will just kill you at that point with no chance to recover safely.
If you're scheduling writes to be done in bulk (not a bad idea but this is actually not how databases do it, more on this later),
then if you die between the flushes you lost X frames of data instead of the last frame of data.
Databases use something called a transaction
It's also atomic.
You set up changes and it keeps track of it in a transation
Then when you're done with your changes you commit.
You can rollback a transaction too but that's something else
Transactions have to comply with ACID standards
So if you make a system you should ideally make it meet the same ones.
Anyway like I said earlier
A lock literally meets this lol
It's how it's usually implemented
Some RDBS allow control over the lock
Scheduling writes is a bad idea unless you don't care about the data.
So some data loss is acceptable
like spoops does with his presence data writes
he schedules it
and if it crashes it doesn't matter cause the data loss is fine
the presence data will re-do when he reboots(edited)
Also the rewrite automatically reconnects.
On every case.
```

Twentysix26:
If you already started experimenting with cogs watch out, the parameter `relative_path` that I just pushed defaults to True. Adjust your paths accordingly.

palmtree5:
Would like to advocate to have something like `bot.session` which would be an instance of `aiohttp.ClientSession()` that all cogs could then use.

palmtree5:
Also, a way of mass-saving data would be useful (for example, dumping json from a aiohttp request to a file via JsonDB)

palmtree5:
Also (unless I'm missing something) there doesn't appear to be support for getting keys from nested dicts short of accessing the data directly

Kowlin:
A method to bulk add/remove keys into the JsonDB would be something I would like to see.

Chovin:
may not be even remotely the same in v3, but currently we have `[p]load filename` and `[p]help ClassName`
could we set a standard in v3 to let users refer to Cogs or Packages in the same way throughout the bot?

palmtree5:
As for core cog conversion (for anyone trying to help with this):

- [x] Alias - Done in tekulvw/mongoV3_alias
- [ ] Audio - ~Probably~ Definitely being rewritten ~?~ (Will's working on it, but more than willing to hand it off)
- [ ] Customcom - 26 mentioned a complete rewrite ( #769 )
- [x] Downloader - Gonna rewrite it with cogs.red in mind (Will)
- [ ] Economy - #781 - Palm__
- [x] General - Converted ( #757 )
- [ ] Image - Rewritten ( #768 )
- [ ] Mod - In progress - Palm__
- [x] Owner - in progress? `core_commands` and `dev_commands` have some of the commands from v2 owner
- [ ] Streams - 26 mentioned a complete rewrite ( #769 )
- [ ] Trivia - Possible rewrite ( #769 )

Someone with the ability to edit please edit this as things are done or if I'm off base and got something wrong in this

tekulvw:
Do we want to bring Squid Permissions into the main bot? Or the concept at least, because if so I'll get to work on rewriting that as well.

JoshB84:
Hello, 
I would like to suggest multiple choice for trivia questions.
What color is the Sun?
A) White
B) Yellow
C) Purple
D) Blue

tekulvw:
This will be reopened at a later date containing the same information.

CreatorVilius:
k

PR #740, opened 2017-04-28
[owner] Handle nonstandard cog module names

calebj:


tekulvw:
Would you care to explain why this is necessary?

calebj:
The way I implement shared analytics between my cogs causes problems with this command. Some ideas for other shared "utility" cogs I haven't implemented would also break this.

This patch is for until I find a better way to share code.

PR #743, opened 2017-05-01
Changed magic method to builtin in Economy

Redjumpman:
Changed magic method to builtin to make it more Pythonic.


Twentysix26:
Apparently slower performance-wise.

PR #755, opened 2017-05-12
(Re/Un)Loading of multiple cogs

Jan200101:
In this PR I changed the way load, unload and reloaded worked.
The command now converts the strings given to it into a Tuple in which every entry gets tried to be load (or unloaded).

Line 65 is a fallback incase someone is able to pass a non string object into the tuple

### DIFF
--- a/cogs/owner.py
+++ b/cogs/owner.py
@@ -51,61 +51,123 @@ class Owner:
     def __unload(self):
         self.session.close()
 
-    @commands.command()
+    @commands.command(pass_context=True)
     @checks.is_owner()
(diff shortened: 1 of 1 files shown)


Jan200101:
Im split between making every load (unload reload etc. you get it) display what cog was loaded or just print out Cog was Loaded.

For now no change is going to happen but I'am going to experiment with it

Jan200101:
Broke Downloader cog

boswelja:
I merged this into my custom fork, I have fixed the downloader cog there :)

PR #758, opened 2017-05-16
[Core][v3] add method for setting all data, method renames

palmtree5:
Adds in a method to set all data for a particular db, also renames `all` (in JsonDB) to `get_all` and `get_all` in JsonGuildDB to `guild_get_all`


Kowlin (approved):
Seems good enough to not break upon impact.

Twentysix26:
#770

PR #765, opened 2017-05-20
Make audio work with mix playlists

tekulvw:


tekulvw:
Closing because wrong branch

PR #766, opened 2017-05-20
Now actually make audio work with mix playlists

tekulvw:


aikaterna:
Works great with a couple different playlist mixes tested.

PR #769, opened 2017-05-20
[Streams] v3 rewrite

palmtree5:
Here's a rewrite of streams for v3


Twentysix26:
I'm planning a full rewrite of this, not just a simple convertion. Same goes for custom commands and most likely trivia

PR #770, opened 2017-05-21
[WIP] V3 Mongo/Json config drop in replacement

tekulvw:
I will continue to update this as I make changes and flesh out usage once this is much closer to working correctly. If you'd like to help test/develop please contact me on discord


tekulvw:
This is working as of right now. Please check out the Example518 cog on Squid-Plugins for a quick example of *one way* to use these tools. I will soon add documentation for usage of this PR without preregistering default values.

Twentysix26:
This isn't using the available thread safe methods for saving json files. It's going to cause the same blocking issues we have now.

Twentysix26:
There are still references to "Servers". They should be called "Guilds" everywhere

PR #773, opened 2017-05-21
[streams] Add Picarto to streamalert stop

calebj:
Missed a spot.


Twentysix26:
Aaand this is why I'm refactoring this mess on the rewrite.

PR #776, opened 2017-05-25
[alias, customcom] use bot's internal command_prefix

calebj:
This allows any command_prefix to work with the bot, without relying on
Red settings. I also rebuilt part of the help command in alias.


tekulvw:
Don't see a real need to change this for now.

PR #780, opened 2017-05-28
[Alias] V3 Rewrite

tekulvw:


Kowlin:
I only found one thing from my testing, and that is that it pops up a double error message when you're trying to delete a alias that doesn't exist and when their no aliases, either global and on the server.

Twentysix26 (changes requested):
Tests are failing

tekulvw:
So I also fixed this nagging issue of the bot's http session hanging around and exiting without closing

Twentysix26 (approved):

PR #781, opened 2017-05-28
[Economy] [WIP] rewrite

palmtree5:
Conversion of economy for v3.

Changes:

- Bank has been split out to `cogs/economy/bank.py`
  - Not really sure how to handle giving bank a config option whilst having it accessible to any cog that wants to use it though (at the moment, it's being passed Economy's config object which likely isn't the best solution for that)
    - Maybe loading Bank as a cog and then just have any cog that's gonna use it access it via `self.bot.get_cog("Bank")`?
- Leaderboard and subcommands for it are currently commented out as it needs to be reworked due to the new config system
  - bank reset is commented out for the same reason

Still needs tests to be written though


palmtree5:
This is probably ready unless there's functionality missing that should be added. All tests are passing and all commands are working. The package's `__init__.py` file has a check to only load Bank if it isn't loaded already (so that third-party cogs could load Bank without the default Economy cog needing to be loaded), which is a check that third-party cogs using Bank should also use

palmtree5:
Actually, I'd hold off. There's a couple more changes to (potentially) be made

palmtree5:
Branch cleaned up now

tekulvw (dismissed):

palmtree5:
This should be ready once #869 is merged

tekulvw:
Hold off on merging until some tests can be written. I'll start working on those tonight. Can you rebase on top of V3 in the meantime?

palmtree5:
Rebase done @tekulvw (cc @redjumpman)

Kowlin (changes requested):
Command Slotmin and Slotmax can be negative, causing Slots to become unusable
Command slotmin can be below Slotmax, causing slots to be unusable
2 bugs:
https://gist.github.com/Kowlin/60ba65d86371df4146b0c0f319968fe8 - Bank Transfer
https://gist.github.com/Kowlin/21dde013e5ea53cca0f53486f7a394e9 - Bank set

tekulvw:
Just fixed the issue with bank set and transfer. Occurred because of a mutability bug initiated after showing the leaderboard.

tekulvw:
...and that fixes the slotmin/max bugs

tekulvw (approved):

Kowlin (approved):
Issues have been fixed.

Issue #270, opened 2016-05-21
[Audio] Child folders within playlist folder


Dommage:
is there a way to have the bot to play local playlist even if there's folder inside.

Actually if I copy a folder to Red-DiscordBot/data/audio/localtracks it won't work if there's folder inside the folder

Thanks

Kowlin:
As a example, your `bot/data/audio/localtracks` folder should look like this. If the bot cant find the folder. Try reloading audio or restarting the bot.

![](http://i.imgur.com/wenlsgu.png)

jonyroda97:
I think that it would be useful what @Dommage says.
For example: have a folder per type of music and in there per band

Dommage:
I mean if you put a folder inside one of those...

Kowlin:
Currently there is no way to do that.

tekulvw:
it'd be as simple as using walk() instead of listdir() iirc, thoughts @Twentysix26 ?

tekulvw:
Moving to #623

PR #783, opened 2017-05-29
[Audio] Add local music features

EmmaSimon:
Adds a new command in local, `[p] local play [file]` that looks through all local files and queues the first one that matches.

Also fixes issue #270, by walking local playlist dirs.


Tobotimus:
Not going to be added in V2 since it is feature locked, sorry.

PR #785, opened 2017-05-29
[Mod] [WIP] v3 rewrite

palmtree5:
Very much in progress

Changes (will add as I go along):
- ModLog is being split out to modlog.py

Not really sure what to do with mod itself. I feel like it could probably be split up as at the moment, it's over 1500 lines but the question is how to split it up


tekulvw:
What's the status on this?

palmtree5:
In progress

palmtree5:
No tests written for this yet but:

- Filter and Cleanup have been split out into their own cogs that are loaded on loading the mod package
- Mod itself needs a delayed load because it depends on ModLog being loaded
  - At the moment, that's being accomplished via #802 but that could very well change
- ModLog is mostly ready
  - It has the following:
    - `[p]modlogset` which has subcommands:
      - `modlog` (for setting the mod log channel)
      - `cases` (for changing whether or not a case type is enabled or not) (may still need work)
      - `resetcases` (for resetting case settings to defaults) (needs work still)
    - `[p]reason` for setting the reason after a case has been created (calls `update_case` which makes use of `importlib` and `getattr` to actually find the class for the casetype)
  - As noted above, cases are now an object of their action's casetype (for example: `KickCase`). These are all subclasses of `Case` which can be found in `cogs/mod/casetype.py`
    - Cases can be created by creating an instance of the action's casetype class. Cases are dumped into the guild's config by a call to `obj.to_json` (which takes a message as a param in order to add the message into the data) which returns a `dict` which looks like `{"1": {"key": "value"}}` which is then added to the `cases` `dict` via something like `cases[case_id] = new_case[case_id]`
    - Case types are registered one of 2 ways (to be called in the registering cog's `__init__` function):
      - `register_new_casetype` (a coroutine)
        - Takes the class for the case type being registered (not an instance) and a bool representing its default setting (True or False)
          - So something like `self.bot.loop.create_task(self.modlog.register_new_casetype(KickCase, True))`
      - `register_casetypes` (a coroutine) (helper for registering multiple new types)
        - Takes a list of tuples
          - The tuples contain the args to be passed to `register_new_casetype`
            - So `[(KickCase, True), (BanCase, False)]`
        - Called by doing `self.bot.loop.create_task(self.modlog.register_casetypes(new_types)`
      - Note that any cog needing to register case types needs to get the modlog cog by having `self.modlog = self.bot.get_cog("ModLog")` in its `__init__` function BEFORE either function is called

palmtree5:
Added in some voice moderation commands as well as commands that allow searching mod log cases based on case number or the target user's user id. I see I also need to merge changes from the main v3 branch though

palmtree5:
Ok, merge conflicts resolved

palmtree5:
So #857 causes issues specifically with the voice-related mod commands

palmtree5:
Modlog (other than commands) is now in the process of being moved to `core` (it's at `core/modlog.py` and is importable by doing `from core import modlog` but is also incomplete at the moment)

palmtree5:
At this point, assuming everything with the mod log works properly (there's a bug atm due to the switch to async getters that Will's aware of), this just needs tests written and it'll be ready to go. Docs are present for this as well

cogs/mod/common.py
```diff
@@ -0,0 +1,129 @@
...
+        user = obj.author
+    elif isinstance(obj, discord.Member):
+        user = obj
+    elif isinstance(obj, discord.Role):
+        pass
+    else:
+        raise TypeError('Only messages, members or roles may be passed')
+
+    server = obj.guild
+    admin_role = bot.db.guild(server).admin_role()
```
tekulvw: Needs `await`

cogs/mod/common.py
```diff
@@ -0,0 +1,129 @@
...
```
tekulvw: Needs `await`

tekulvw (dismissed):

palmtree5:
There's a bit of starting work on tests present, but they're broken atm

palmtree5:
And the tests I have in there are now passing

palmtree5:
Modlog and mod now both have a `messages.pot` (for translations)

palmtree5:
This should be ready at this point though people may wish to test this before a merge

tekulvw:
```
[01/10/2017 13:33] DEBUG json_io _save_json 29: Saving file /home/will/.local/share/Red-DiscordBot/cogs/Mod/settings.json
Ignoring exception in command
  File "/home/will/.pyenv/versions/Red-3.5.3/lib/python3.5/site-packages/discord/client.py", line 223, in _run_event
    yield from coro(*args, **kwargs)
  File "/home/will/Documents/Projects/Red-DiscordBot/redbot/cogs/mod/mod.py", line 1111, in on_command
    await asyncio.sleep(delay)
  File "/home/will/.pyenv/versions/3.5.3/lib/python3.5/asyncio/tasks.py", line 522, in sleep
    future, result)
  File "/home/will/.pyenv/versions/3.5.3/lib/python3.5/asyncio/base_events.py", line 541, in call_later
    timer = self.call_at(self.time() + delay, callback, *args)
TypeError: unsupported operand type(s) for +: 'float' and 'str'
```

tekulvw:
```
/home/will/Documents/Projects/Red-DiscordBot/redbot/cogs/mod/mod.py:517: RuntimeWarning: coroutine 'Value._get' was never awaited
  if self.settings.guild(guild).reinvite_on_unban():
```

tekulvw:
```
[01/10/2017 13:39] ERROR events on_command_error 96: Exception in command 'hackban'
  File "/home/will/.pyenv/versions/Red-3.5.3/lib/python3.5/site-packages/discord/ext/commands/core.py", line 62, in wrapped
  File "/home/will/Documents/Projects/Red-DiscordBot/redbot/cogs/mod/mod.py", line 373, in hackban
    is_banned = discord.utils.get(ban_list, id=user_id)
  File "/home/will/.pyenv/versions/Red-3.5.3/lib/python3.5/site-packages/discord/utils.py", line 227, in get
    return find(predicate, iterable)
  File "/home/will/.pyenv/versions/Red-3.5.3/lib/python3.5/site-packages/discord/utils.py", line 168, in find
    if predicate(element):
  File "/home/will/.pyenv/versions/Red-3.5.3/lib/python3.5/site-packages/discord/utils.py", line 221, in predicate
    obj = getattr(obj, attribute)
AttributeError: 'BanEntry' object has no attribute 'id'
```

tekulvw:
It also seems like that deletedelay trigger is still running even after setting it to a non -1 number and then resetting it to -1.

Edit: This seems to be the issue.
image

tekulvw:
Softban results in an unban message in modlog.

tekulvw:
Case numbers jumped too in modlog:

image

Issue #516, opened 2016-12-08
[Suggestion] Selective Cog Updates


Kyrislian:
Just a suggestion for the Downloader Cog. Would be really nice to have the ability to pick which cogs to be updated and which to leave be, as sometimes some have been further customized for particular Red instances ^^

Chovin:
I second this for security reasons.
being able to see what can be updated and choose which to update is pretty important if you're pulling from the repos of a bunch of new faces

Kowlin:
This is a issue, not sure how I can test this.

Twentysix26:
@Kowlin It's in the downloader, which is already merged

PR #786, opened 2017-05-29
[Downloader] WIP V3 rewrite

tekulvw:
Goals

- [x] Make downloader in an API accessible format in order to hook cogs.red into it.
- [x] Find some way to support repo-wide "libraries"


orels1:
Like it so far 👍 

Will copy my comment from discord too

so like my process will look like this.

- user calls `redp search redportal`

- user clicks on the download react

- redp cog calls `add_repo` for the repo, `RepoManager` adds it if it's not there and clones it and returns the list of cogs available

(from this point - not implemented yet) 

- redp looks throught the list for the cogs it needs and either drops an error, or calls (for example) `install`

- `Downloader` installs the cog and returns true to redp

- `Downloader` asks if user wants to load the cog

and that's it
looks like a nice process to me

tekulvw:
Do we want to retain the "changelog" that existed in V2, modify it, or drop it entirely?

Twentysix26:
Any plans for #516 ?

tekulvw:
Already implemented

Kowlin:
From my testing this seems to work and can be merged into the alpha. If anyone wants to add on that, feel free.

PR #790, opened 2017-06-02
[Versioning] V3 Add dynamic versioning based on git tag

tekulvw:
For future record, when we tag commits the tag name MUST be in the format: `vMAJOR.MINOR.PATCH`

### DIFF
--- a/core/__init__.py
+++ b/core/__init__.py
@@ -9,10 +9,14 @@ BASE_VERSION = version_info(3, 0, 0)
 
 
 def get_latest_version():
-    p = run(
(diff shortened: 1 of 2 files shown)


Twentysix26:
This will raise a `FileNotFoundError` if git isn't installed. It should be handled.

tekulvw:
I handled it, refresh your page

AraHaan:
It can also raise that exception not only if git is not installed. On Windows if it is installed but not in the path so it can be used with just ``git`` on any command line program like cmd.exe or powershell.

PR #796, opened 2017-06-08
Replace verbose matching w/ simple \w RegEx

JodanJodan:


tekulvw:
Except your regex is too permissive. It allows people to put `/` and `.` in playlist names which is a security problem.

Twentysix26:
Unneeded.

JodanJodan:
@tekulvw `\w` is identical to `[a-zA-Z0-9_]`, so I'm not sure why you believe `/` and `.` are included. This is basic RegEx.

image

tekulvw:
Okay, my bad, I was thinking `\w` was the non-whitespace character code.

Please explain to me how regex is "simpler."

This PR is still going to stay closed because it provides no real benefit and V2 audio is pretty much locked in prep for V3.

PR #802, opened 2017-06-09
[Core] [WIP] V3 Delayed cog loading

tekulvw:
After speaking with @palmtree5 (and experiencing this problem myself) it seems to make sense to have a single, centralized way of delaying cog loading until either the bot is ready or some other cog dependency has loaded first.

### DIFF
--- a/core/bot.py
+++ b/core/bot.py
@@ -115,8 +115,11 @@ class Red(commands.Bot):
                 return False
         return True
 
-    def handle_load_extension(self, _: str):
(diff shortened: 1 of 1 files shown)


Twentysix26:
I have noticed that you added a `wait_until_ready`. Currently, unlike v2, we are loading cogs on `on_ready`

tekulvw:
Trying to do this is actually a lot more complex than it seems on the surface. I'm going to work with @palmtree5 on making this viable but am going to close the PR for now.

PR #813, opened 2017-06-15
[Core V3] Add a sentry logger to only grab stuff that pertains to core cogs

tekulvw:


Twentysix26 (changes requested):
Tests need to be updated

Twentysix26 (approved):

PR #782, opened 2017-05-28
[Streams] [WIP] Rewrite

Twentysix26:
Left to do:

- [x] Mention settings for stream alerts (custom roles too?)
- [ ] Tests

PR #821, opened 2017-06-20
Added subcommand for Streams cog "list"

The-Tasty-Jaffa:
This is a temporary thing (before v3) to allow for users to list all the current stream names in the server the command was called in.

### DIFF
--- a/cogs/streams.py
+++ b/cogs/streams.py
@@ -294,15 +294,14 @@ class Streams:
     @checks.admin()
     async def List_streams(self, ctx):
         """Returns all the names of streams for current server""" #Should also add the channel names (for easy of use... just maybe)
-        streams = (self.twitch_streams, self.hitbox_streams, self.mixer_streams, self.picarto_streams)
+        streams = self.twitch_streams + self.hitbox_streams + self.mixer_streams + self.picarto_streams
         stream_list = []
         
+        def stream_in_channels(stream, channels):
+            return any(c.id in stream["CHANNELS"] for c in channels)
         
-        for channel in set(ctx.message.server.channels):#..
-            for stream in streams: #...
-                for s in stream: #!!!
-                    if channel.id in s['CHANNELS']: #*cry*
-                        stream_list.append(s['NAME'])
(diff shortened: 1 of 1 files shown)


cogs/streams.py
```diff
@@ -291,11 +290,33 @@ def __init__(self, bot):
         if ctx.invoked_subcommand is None:
             await self.bot.send_cmd_help(ctx)
 
+    @streamset.command(pass_context=True, name = "list")
+    @checks.admin()
+    async def List_streams(self, ctx):
+        """Returns all the names of streams for current server""" #Should also add the channel names (for easy of use... just maybe)
+        streams = (self.twitch_streams, self.hitbox_streams, self.mixer_streams, self.picarto_streams)
```
tekulvw: `streams = self.twitch_streams + self.hitbox_streams + self.mixer_streams + self.picarto_streams`

cogs/streams.py
```diff
@@ -291,11 +290,33 @@ def __init__(self, bot):
...
+        stream_list = []
+        
+        
+        for channel in set(ctx.message.server.channels):#..
```
tekulvw: This whole block should just get changed. With the above changes in place:

```py
in_chan_streams = filter(lambda s: ctx.message.channel.id in s["CHANNELS"], streams)
stream_names = [s["NAME"] for s in in_chan_streams]
```

edit (for in server streams):
```py
def stream_in_channels(stream, channels):
    return any(c.id in stream["CHANNELS"] for c in channels)
in_server_streams = filter(lambda s: stream_in_channels(s, ctx.message.server.channels), streams)
stream_names = [s["NAME"] for s in in_server_streams]
```

tekulvw (changes requested):

The-Tasty-Jaffa (commented):
Updated with suggested changes

RSNFreud:
1 month and no update?

Kowlin:
Red v2 (on which this PR is based) is feature locked. I don't know how we will approach old v2 PR's, but please hold on while we get trough the work load in the following weeks.

tekulvw:
Gonna say no because of the feature lock, feel free to comment on/contribute to #782

PR #825, opened 2017-06-23
[Admin V3] Rewrite of Squid-Plugins Admin cog

tekulvw:
Still need to implement:
- [x] Server lock
- [x] tests
- [x] role hierarchy check


Twentysix26:
Something like ServerTools would be a better name for such a cog, as "admin" implies they're all admin related commands.

This also still lacks *user* hierarchy checks, a person with manage roles shouldn't be able to add all the roles they want at will but strictly the ones they are entitled to (which means any role below their highest  one).

tekulvw:
It's not strictly server tools though. I feel like Admin is as fitting a name as any seeing as how all of these commands have something to do with server/bot administration.

I will add an optional ability to enable that user hierarchy check. I am sure that some users will want to be able to override discord hierarchy.

Kowlin:
I don't feel like renaming the cog would have any effect, its simply just a name. Might as well keep it as it currently is. Let alone the people who are familiar with the name and its current toolset, I feel that this will just cause confusion.

redbot/cogs/admin/announcer.py
```diff
@@ -0,0 +1,71 @@
...
+        return channel
+
+    async def announcer(self):
+        guild_list = self.ctx.bot.guilds
+        bot_owner = (await self.ctx.bot.application_info()).owner
+        for g in guild_list:
+            if not self.active:
+                return
+
+            if self.config.guild(g).announce_ignore():
```
palmtree5: needs to be awaited

palmtree5 (changes requested):
"GUILD_MEMBER_REMOVE referencing an unknown guild ID:"
Getting that when adding a bot that's been serverlocked

Adding roles to be available via `[p]selfrole` works but actually doing `[p]selfrole rolename` doesn't

tekulvw:
Fixed selfrole not working. No repro on the serverlock issue, can you provide a traceback?

palmtree5 (approved):

PR #834, opened 2017-06-30
I made red better 

Red-Sucks:
THE BETTER VERSION OF RED


tekulvw:
What kind of fucked up do you need to be to do this?

PR #849, opened 2017-07-06
[Streams] implement support for alerting for twitch community streams…

palmtree5:
… (Twentysix26/Red-DiscordBot/issues/824)


palmtree5:
What this brings:
- turns `streamalert twitch` into a group with subcommands `channel` and `community`
- implements `community` specific counterparts of several methods used for individual channel alerts
- Posts a message for each channel online in a community
  - Probably needs a method of deleting those messages when the channel doesn't appear in the community's streams anymore
- Community check loop is called from `_stream_alerts` after the individual channel check loop
  - Perhaps this should be different?

PR #853, opened 2017-07-10
[V3] Multiple cog paths

tekulvw:
## Warning
This exists as a proof of concept, this PR is currently missing major UI stuff so it's not applicable to anyone but developers at the moment. This PR **does not** modify `sys.path` but it does make use of some `importlib` internals and overrides d.py's `load_extension` and `unload_extension` so we may run into some issues down the line when Danny overhauls his loading system (this is OK and expected).

## How to use
1. Let the bot boot up once and close it
2. Go to cogs/.data/CogManager/settings.py and edit the GLOBAL section
3. You need to add `paths: []` to the GLOBAL section containing any path you wish to access cogs from
4. Use load/unload/reload as normal

Things to add:
- [x] Adding paths command
- [x] Removing paths command
- [x] Re-ordering paths command
- [x] ~~Cogs local path needs to be added at install, not at runtime~~ see comment below
- [x] Add an install path
- [x] tests

### ~~This PR depends on #869~~

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,6 +3,7 @@
 
 # Core
 core/config.py              @tekulvw
+core/cog_manager.py         @tekulvw
 core/drivers/*              @tekulvw
 core/sentry_setup.py        @Kowlin @tekulvw
 
--- a/cogs/downloader/downloader.py
+++ b/cogs/downloader/downloader.py
@@ -21,12 +21,6 @@ from .checks import install_agreement
 
 
 class Downloader:
-
-    COG_PATH = Path.cwd() / "cogs"
-    LIB_PATH = Path.cwd() / "lib"
-    SHAREDLIB_PATH = LIB_PATH / "cog_shared"
-    SHAREDLIB_INIT = SHAREDLIB_PATH / "__init__.py"
-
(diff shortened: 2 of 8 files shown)


tekulvw:
So I realized that the local relative cog path shouldn't be added at install time, but should always be available and determined at runtime. The next few commits will reflect that.

tekulvw:
This is ready for review.

Kowlin (approved):
Works great and not all to confusing to setup either.

tekulvw:
Alright, I've written tests, this is actually ready for full review/testing.

Stylite:
Worked perfectly fine for me. No issues at all.

mikeshardmind:
Had to update a couple of my cogs to be able to test (I had some older ones from pre config rewrite). Loading from multiple paths works (at least on my configuration, details below)
Ubuntu 16.04 server host, Python 3.5.2.
Loading worked even from separate devices as long as I gave it the absolute path.

aikaterna:
Worked with loading from multiple paths on Windows 7 Professional with Python 3.6.2.

tekulvw:
This will definitely break cog loading when danny overhauls his loading mechanisms but we'll deal with the fallout from that when it happens.

PR #855, opened 2017-07-11
[Downloader V3] Find a cog from a command

tekulvw:
For the record, we can't reload downloader right now because the module refresher tries to import every single cog in every repo downloader has cloned. This should be fixed in #853

### DIFF
--- a/cogs/downloader/downloader.py
+++ b/cogs/downloader/downloader.py
@@ -1,6 +1,6 @@
 import os
 import shutil
-from typing import Tuple
+from typing import Tuple, Union
 
 import discord
 from discord.ext import commands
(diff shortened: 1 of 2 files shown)


cogs/downloader/downloader.py
```diff
@@ -318,3 +318,71 @@ def installed_cogs(self) -> Tuple[Installable]:
...
+    def cog_name_from_instance(self, instance: object) -> str:
+        """
+        Determines the cog name that Downloader knows from the cog instance.
+
+        Probably.
+        :param instance:
+        :return:
+        """
+        splitted = instance.__module__.split('.')
+        return splitted[-2]
```
tekulvw, line 365: If #853 gets merged this should become `return splitted[0]`

PR #860, opened 2017-07-18
make is_owner actually check the owner setting

palmtree5:
As it stands now, our `is_owner` function simply checks if the user is a coowner and failing that, calls d.py's `is_owner` function (which checks if the user is the application's owner). Thus it looks to me as if our owner setting is completely ignored in this check (hence the PR fixing that)

### DIFF
--- a/core/bot.py
+++ b/core/bot.py
@@ -51,7 +51,7 @@ class Red(commands.Bot):
         super().__init__(**kwargs)
 
     async def is_owner(self, user):
-        if user.id in self.db.coowners():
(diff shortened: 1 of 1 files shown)


tekulvw (approved):

Twentysix26:
74910a49
This is unneeded. The typo in `[p]set owner` was preventing it from working correctly. D.py's `is_owner` checks the `Bot.owner_id` attribute. If it happens to be None, it fetches the owner ID from the API.

PR #861, opened 2017-07-18
[Audio] Add "[p]cache stats" command

americanhanko:
**This PR:**
- Adds a `!cache maximum` command to give the user the current maximum cache value
- Removes a duplicate `outtmpl` key from `youtube_dl_options` dictionary
- Fixes some PEP8 whitespace and newline issues


tekulvw (approved):

Twentysix26:
Three different commands that show three different stats about the same thing seem a bit silly. Mind turning all these into a single command (`[p]cache stats` would be fine) that shows min, max and current size together?

Twentysix26 (changes requested):
A single `[p]cache stats` command for min, max and current size would be better

Kowlin (commented):
I'm a little confused, will this also remove older files that are over the max cache size? Or was that already build into the system?

americanhanko:
@Twentysix26 Apologies for the delayed response. Having `[p]cache stats` makes far more sense. Working on it now.

@Kowlin This change wouldn't modify any files. It would only report the current values for `cache_minimum`, `cache_maximum` and `cache_size` to the user.

Kowlin:
Ah ye okay, I see. Source digging does wonders from time to time.

Kowlin (approved):

tekulvw (approved):

americanhanko:
Sorry for so many pushes...

Here's what the output looks like as of right now:
1. screenshot 2017-07-25 12 23 47

Some other ideas:
2. screenshot 2017-07-25 12 27 29
3. screenshot 2017-07-25 12 27 36

Twentysix26:
I think the number 3 would look best 👌

bobloy:
Graphical percent of total?

E.G.

Current: 69.97 MB
Min: 60.0 MB  █▒▒▒▒▒▒▒▒▒ Max: 1000.0 MB

Kowlin:
Seems a bit overkill for something as little as data.

Twentysix26 (approved):

Issue #868, opened 2017-07-24
[Config V3] Support nested default values


tekulvw:
The way I'm seeing this work is something like this:

```py
conf = Config.get_conf(...)
conf.register_global("foo.bar.baz"=True)
baz = conf.foo.bar.baz()
await conf.foo.bar.baz.set("thing")
```

Issue #812, opened 2017-06-15
[Config V3] implement get_all for getting the config for all of something


palmtree5:
For example, getting the config for all users would look something like:

```py
all_user_settings = self.config.user(user).get_all()
```
that would return either:

- a dict containing all the <insert thing here> ids as keys and those things' settings as the values
  - Disadvantage to this could be someone trying to modify directly as it wouldn't save
- a tuple of (thing_id, dict) (the dict would be the same as the values in the keys in the first option but has the advantage of hopefully deterring direct modification)

Possible uses for this could be for something like ranking users by some setting (as one example, for some global trivia leaderboard)

Issue #810, opened 2017-06-14
[Config V3] Config member stored IDs need to be in reverse order


tekulvw:
Currently config stores members as follows:

`data["MEMBER"][user_id][server_id]` which should be `data["MEMBER"][server_id][user_id]`

Doing it that way will make it far easier to remove  every member from a given server (which is a lot more common than removing every server from a given user).

This change will corrupt all current config files that have MEMBER data in them. You can delete the config files or just clear the member data manually to get around this.

PR #869, opened 2017-07-24
[Config V3] Rewrite (already?)

tekulvw:
Front facing features shouldn't change a whole lot but the backend of config is getting totally revamped. I'll list new features here as they come to fruition.

### New Features
- Nested defaults!
  * `conf.register_global(foo__bar=True)` and to get that data `conf.foo.bar()`
- `.set()` now does not require the field name
  * `config.set("enabled", True)` is now `config.enabled.set(True)`
  * As a result, if you're trying to do dynamic attribute setting you will need to use a new method: `config.set_attr("some_field", True)`
  * Also, to maintain consistency, `conf.get('foo', False)` is now `conf.get_attr('foo', False)`
- Member features! You now access all the guilds in member data as well as all members in a given guild.
  * All guilds: `conf.member(member_obj).all_guilds()` returns a dict of servers -> members -> data
  * All members same guild: `conf.member(member_obj).all()` returns a dict of members -> data

### To Do
- [x] Implement #812 
- [x] Write some more tests to cover new functionality

### Currently Fixes
- #868 
- #810
- #812


tekulvw:
Okay so for those of you who have used config before, does anyone deem the `.get()` function necessary?

palmtree5:
`.get()` is necessary if you don't force registration (which the first case that comes to mind of not forcing registration is the modlog part of my rewrite of mod because of `[p]modset cases` and having an interface available for adding casetypes which means that its key where its enabled status is located would not necessarily be registered automatically and would instead be added after the fact)

mikeshardmind:
I've used config and migrated about half my repo over to v3. 
Using the alpha (this is pretty stable for an alpha) to manage temp channels, formboxes, and a few other things.  `self.conf.guild(ctx.guild).somekeyhere` already retrieves the key (as long as it exists) , and you provide a way to register defaults already, so I don't see the use case for `.get()`

Edit:
@palmtree5 is right, I've actually just forgotten how I worked around that detail by abstracting everything a level up, storing temporary channnel settings as owned by the server, not the channels themselves, alongside forcing registration for a few things.

tekulvw:
To use your example, `self.conf.guild(ctx.guild).somekeyhere()` works both with and without registering `somekeyhere`. The current functionality forces `somekeyhere` to have a default of `None` if you don't register it and you can pass a different default (such as `self.conf.guild(ctx.guild).somekeyhere(True)`) into the getter function if you wish.

mikeshardmind:
If that's already current functionality (was it before? ....) then getting rid of it doesn't seem awful

tekulvw:
It's always been that way, I was asked to specifically include the `.get()` function as an alternative.

mikeshardmind:
Then .get() is redundant

QingyuChai:
But what if what I'm looking for in the config is a parameter of the command?
Example:
```python
async def test(self, ctx, foo):
    # I can't do self.conf.guild(ctx.guild).foo()
    # In this case I would need get()
    self.conf.guild(ctx.guild).get(foo)

```

Twentysix26:
I'd argue that
`self.conf.guild(ctx.guild).somekeyhere(True)`
is significantly less clear and immediate than
`self.conf.guild(ctx.guild).get("somekeyhere", True)`
to anyone who has not source dived into (or developed) Config.  
Not to mention that it would force the use of `getattr` if the key is dynamic.

tekulvw:
@QingyuChai You bring up a very good point, dynamic attribute access is impossible without `.get()` or using `getattr()` and the use case for that is so large I'm going to add it back in.

@Twentysix26 My argument is that while not using `.get()` may be unclear at someone's first glance, after reading and learning how config works (with enough documentation), using `.get()` becomes more of a hassle because than directly accessing the attribute. However, the point is now moot.

Kowlin (approved):
Code looks good from my end.

tekulvw:
@QingyuChai @Twentysix26 In order to support dynamic attributes and to make things explicitly clear I had to add a new method to set the value of dynamic attributes. You'll have to use `config.set_attr("dynamic_attr", value)` instead of `config.set()` because now `.set()` only takes 1 argument.

tekulvw:
~~Active development is complete, ready for review/testing.~~

~~Making a few changes.~~

Ready again for review.

palmtree5 (approved):
Looks good to me too

Twentysix26 (approved):

PR #875, opened 2017-07-28
[Core] Added [p]contact and [p]dm

Twentysix26:
I added [p]contact, which is unchanged from v2, ~~and a syntax to DM users through Red.~~  
~~The syntax is `dm:<id> <text>`, it's owner-only, only works in DM with Red and you receive the full string to copy paste when you get `[p]contact`ed
I'm actually not sure if I should add a normal command for it instead~~
Let me know what you think 👌 
EDIT: Syntax turned to [p]dm

### DIFF
--- a/core/core_commands.py
+++ b/core/core_commands.py
@@ -2,6 +2,7 @@ from discord.ext import commands
 from core import checks
 from string import ascii_letters, digits
 from random import SystemRandom
+from collections import namedtuple
 import logging
 import importlib
 import os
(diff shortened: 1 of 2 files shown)


tekulvw:
To move my thoughts from discord onto here: I think that having a normal command (dm/pm/whisper whatever you want to call it) that simply takes two args, a user ID and a message, is a way simpler solution to this problem. Especially because contact already gives you the user id.

Twentysix26:
Yeah I'm of the same opinion now, this is just the result of various things I tried and I eventually got too deep into it x)
I'll turn it into a command later

Twentysix26:
Should be good to go

Kowlin (approved):
Looks good from my view.

tekulvw (approved):

PR #876, opened 2017-07-28
Multiplication payouts should not add original bid to winnings

timhargreaves:
Labels for bid winnings don't match the results. Multiplication payouts should not also add the original bid to the winnings.

When you are paid out at 2:1, you should get double your bid back, not triple.


Tobotimus:
I understand where you're coming from here, but I don't think we're willing to change the payouts at this stage.

If anything, the payout message could simply be reworded to make it clearer what is being payed out.

Issue #867, opened 2017-07-24
[Red v3]  Failure to Identify git HEAD


mikeshardmind:
attempting to launch red from a directory other than red's fails. 
it is trying to find it in the current working directory rather than the directory red is in

`$ env/bin/python3.5 Red-v3/main.py `
results in:

`fatal: Not a git repository (or any of the parent directories): .git`
some interaction during setup and then
`Traceback (most recent call last):
  File "Red-v3/main.py", line 99, in <module>
    init_sentry_logging(sentry_log)
  File "/home/ubuntu/git/Rv3/Red-v3/core/sentry_setup.py", line 35, in init_sentry_logging
    release=fetch_git_sha(str(Path.cwd()))
  File "/home/ubuntu/git/Rv3/env/lib/python3.5/site-packages/raven/versioning.py", line 25, in fetch_git_sha
    'Cannot identify HEAD for git repository at %s' % (path,))
raven.exceptions.InvalidGitRepository: Cannot identify HEAD for git repository at /home/ubuntu/git/Rv3
`

tekulvw:
Just like with V2, V3 is still dependent on CWD being the bots folder. This ought to change and we're working on it but for now that's the way it is.

tekulvw:
No longer applicable to V3

Issue #779, opened 2017-05-28
[Red V3] What do you want to see?


Kowlin:
As the title suggests, what do you want to see out of Red v3, nothing here is off-limits.

palmtree5:
To put this here (though I remember seeing @orels1 mention it in discord): Integrate the redportal cog into downloader (example: this would allow for a one-click install from a search)

Stylite:
Be able to order pizza from the bot and make cleverbot core. I feel bad for coding with this issue

pgaetani:
Will's admin should probably be core.

tekulvw:
Permissions core?

palmtree5:
Both of those

orels1:
@palmtree5 not really like that. I feel like wiring cogs.red into the red itself is not the best idea, due to it being an external service supported by someone from outside of main red dev.

My idea was the following:
>Expose some kind of an API for downloader, so we could call something like `downloader.install_cog(<repo/repoURL>,<cog>)` from inside custom cogs (redportal as an example). So we could, for example have a reaction on `redp search` results embed, that will start installation process without the need to call `cog repo add` and `cog install` manually.

pgaetani:
I like orels' idea... could it even be used for something like cog dependencies? (Yes, I'm going there)

SHADOW6:
Modmail for red:
 https://github.com/Dragory/modmailbot

tekulvw:
Downloader needs to get rewritten then. And I want to see people writing tests for core cogs at the least!

Kowlin:
*"Writing tests"*, its like you want us dead 😂.

Eslyium:
Softban to send an invite link to the user in pms

BNegetive:
to play music from streams on youtube.

SHADOW6:
Suggestion for: Red V3,
+ playlists and Qued music can generate a playlist 
1) song 
2) song
3) song
4) song
5) song
[p]remove # would remove select song from a Que or a list.

+ url.mp3 support
+ Podcast support 
+ if its not bot breaking, how about adding YouTube stream support

Its a rework, so throwing the moon at you. If it can be done Great!
That will mean no one has to worry about it later.

AraHaan:
Red to work with discord.py rewrite branch. v1.0.0

tekulvw:
...everything is now clear

boswelja:
[...]

PR #878, opened 2017-07-28
Allow launching red from a seperate directory

mikeshardmind:
Useful for situations where you have multiple virtual enviorments due to multiple bots with conflicting dependencies.

See: #867 , #779

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,12 @@
+# Default
+*                           @Twentysix26
+
+# Core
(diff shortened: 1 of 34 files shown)


tekulvw:
I like the idea, I don't like the implementation, especially with what it's doing to the tests.

tekulvw:
Okay it's not quite working but this is much more along the lines of an acceptable solution. TBH this may be easier to hold off on until we get configurable data paths.

tekulvw:
This will be superseded by #879

PR #879, opened 2017-07-30
[Core V3] Make the bot data path configurable

tekulvw:
Hey guys, this PR is another step towards making the bot pip installable and fully containerizable.

### New features
- Added `from core import data_manager`
  - Useful function is `data_manager.cog_data_path(cog_instance)` which gives you your own data folder to store whatever you want in.
- Bot now requires a config file to start which defines where your data folders should be stored on the file system.

### Dependencies
- ~~#869~~
- ~~#853~~
- ~~Launcher rewrite #850~~
  - Gonna need to add some custom shit into this @palmtree5

PR #880, opened 2017-07-31
[Core V3] Add setup instructions for external data and cog paths

tekulvw:
One more step towards pip installability.

### New Features
- Refactored initial setup
- Add components to initial setup to ask for data path and cog install path for Downloader
- Setup creates a `config.json` file that's now required to start the bot as it contains the data path

### What does this mean?
*We're no longer depended on current working directory.*

### How do I use this?
1. Update your requirements: `pip install -U -r requirements.txt`
2. `python /path/to/main.py`
3. The bot will then generate a file named `config.json` to the current working directory.
4. From now on run the bot using `python /path/to/main.py /path/to/config.json`

### Dependencies
- ~~#853~~
- #879

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -4,6 +4,7 @@
 # Core
 core/config.py              @tekulvw
 core/cog_manager.py         @tekulvw
+core/data_manager.py        @tekulvw
 core/drivers/*              @tekulvw
 core/sentry_setup.py        @Kowlin @tekulvw
 
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,6 +3,7 @@
 
 # Core
 core/config.py              @tekulvw
+core/cog_manager.py         @tekulvw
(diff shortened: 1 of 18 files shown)


mikeshardmind:
It works, but it shouldn't be asking me for a cog data path on each launch, that should be detected as having already been set.

tekulvw:
It doesn't ask you for a data path every launch if you provide the config
file it generates when you run the bot. E.g. `python main.py config.json`

tekulvw:
Added extra instructions to this PR.

mikeshardmind:
Right, but if that config file exists in Red's directory, it should be considered the default config file unless a different one is specified.

tekulvw:
The point of this PR is to get stuff out of the bot folder. I'm not going to add support for that, just use the command line argument.

tekulvw:
Right idea, wrong implementation. New stuff will be coming soon in different PRs.

PR #889, opened 2017-08-06
[Config V3] Fix 'all' handling to match docs

### LINKED ISSUE #888: [Config V3] User all returning wrong level
All is currently returning USER_ID -> data instead of just data

tekulvw:
Fixes #888


tekulvw:
@palmtree5 this may break bank's get_all_accounts

tekulvw:
The latest commit allows users to do stuff like: `_conf.casetypes.get_attr(casetype, resolve=False).case_repr()`

palmtree5 (approved):

PR #907, opened 2017-08-12
[Config] Asynchronous getters

tekulvw:
I tried to get everythng but I'm sure something is still going to break.


palmtree5 (approved):
lgtm

PR #919, opened 2017-08-14
Implement seeking in YouTube links through the t and end URL parameters

alexwh:


cogs/audio.py
```diff
@@ -1382,7 +1393,11 @@ def _valid_playable_url(self, url):
             url = "[SEARCH:]" + url
 
         if "[SEARCH:]" not in url and "youtube" in url:
-            url = url.split("&")[0]  # Temp fix for the &list issue
+            parsed_url = urllib.parse.urlparse(url)
```
tekulvw, line 1480: What is this section doing exactly?

alexwh: Based on the original comment, I assume the original code was there just to cut out any possible &list arguments, so youtube_dl didn't mess with the existing playlist system.
I need youtube_dl to get the &t= and &end= parameters so it can parse them to return start and end_time, so instead of cutting off all parameters, I parsed the URL and only removed list (if any).

Twentysix26 (approved):
Works as expected

tekulvw (changes requested):
Rebase and resolve conflicts

PR #920, opened 2017-08-14
[Core V3] Fix pagify bug

Zephyrkul:
Yes, I'm bringing this up again, since development has moved on from my original PR at #646.

This one also implements the fix in PR #543 by @Chovin. While that issue has been *addressed*, the fix is... lacking, since it counts mass mentions in the entire text and removes that count from each page. Chovin's fix (counting only in the specific page, not in the entire text) should be better, since the page length will be more regular and there won't be any strange bugs of trying to post low-length messages in the rare case that there's a lot of mentions.

For commentary on my fix, which altogether prevents sending 0-length / whitespace-only messages, please see the original PR. (#646)

This has not yet been thoroughly tested, though it did fix the original issue that I had that spawned this PR in the first place.

### DIFF
--- a/core/utils/chat_formatting.py
+++ b/core/utils/chat_formatting.py
@@ -52,7 +52,7 @@ def pagify(text, delims=["\n"], *, escape_mass_mentions=True, shorten_by=8,
             yield to_send
         in_text = in_text[closest_delim:]
 
-    if len(in_text.strip()) > 0
+    if len(in_text.strip()) > 0:
         if escape_mass_mentions:
             yield escape(in_text, mass_mentions=True)
         else:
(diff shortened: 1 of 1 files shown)


core/utils/chat_formatting.py
```diff
@@ -35,25 +35,28 @@ def pagify(text, delims=["\n"], *, escape_mass_mentions=True, shorten_by=8,
...
-        yield to_send
+        if len(to_send.strip()) > 0:
+            yield to_send
         in_text = in_text[closest_delim:]
 
-    if escape_mass_mentions:
-        yield escape(in_text, mass_mentions=True)
-    else:
-        yield in_text
+    if len(in_text.strip()) > 0
```
palmtree5: Missing colon here

palmtree5 (dismissed):

Zephyrkul:
We got that one at the same time, it seems, Palm.

tekulvw:
It looks good to me, as you stated the current issues with pagify are fixed here and any new ones that come up can be dealt with then.

tekulvw (approved):

PR #926, opened 2017-08-15
Quote package name in pip function to prevent RCE

ghost:
Quote the name variable in the pip_install function to prevent people from
running arbitrary shell commands.


Twentysix26:
As discussed on Discord we were unable to reproduce this and the examples you provided are run through an insecure 3rd party cog.

PR #928, opened 2017-08-17
[Dev V3] mockmsg command

Tobotimus:
Since we already have a `mock` command which pretends a command was invoked by another user, another similar but equally useful feature is this `mockmsg` command, which pretends a message was sent by another user.

This would be really useful when developing cogs which interact with multiple users through `on_message` or `wait_for` events, such as Trivia.


tekulvw:
I'm going to approve and merge this because it's added as a dev command.

tekulvw (approved):

Issue #644, opened 2017-02-28
[Audio] Bug and Suggestions


jonyroda97:
### Bug
- Bot owner isn't marked as a listener.
If the autodisconnect on empty channel is enabled and if the owner is the only one on the channel, the bot disconnects after the set time.

### Suggestions
- On [p]song command show the current time that is playing.

- Multi playlists on queue.
Random playing between the playlists queued (but not like 1 from playlist A other from playlist B and then from playlist A again. Really randomised from all the songs).

- Auto Update Youtube Playlists.
I know it can be made with the scheduler cog but would be easier if it is implemented in the audio cog.

- Add a better queued response. 
This is good when you want to know who queued the sound and have the bot command messages delete enabled. Maybe something like: 
"\<user> queued in  \#<queue_position> \<title>
 (\<link>)"
Example: 
jonyroda97 queued in  \#800 **_Post Malone - Congratulations ft. Quavo_**
(https://www.youtube.com/watch?v=SC4xMk98Pdc)

- In #366 a way to export most replayed links to a youtube playlist (only youtube links)

jonyroda97:
More things to the #623 rewrite list

Twentysix26:
I can't seem to reproduce the bug.

jonyroda97:
i will try to find out how it happened

Twentysix26:
Added the suggestions to #623
Feel free to reopen if you have more details about that issue

Issue #623, opened 2017-02-16
[Audio] Suggestions for rewrite


tekulvw:
We're gonna use this as a collection board for any/all current or future audio suggestions.
- [ ] sfx #622 
- [x] better queue control #621 
- [x] local playlist files #620 
- [x] play individual local files
- [x] jump to time #619 
- [x] web radio stuff (which is the same thing as livestream) #548
- [x] event system in place to allow for channel based notifications #284
- [ ] configurable limit on max concurrent audio streams #272 
- [x] better soundcloud/yt playlist support #265
- [x] walk local playlists (sub-folder support) #270
- [x] auto pause if empty? #721 
- [x] current time in [p]song #644
- [x] ability to update (automatically?) saved playlists #644
- [x] allow skipping multiple songs #1845

*kill me now*

tekulvw:
![](https://github.trello.services/images/mini-trello-icon.png) [General Audio rewrite](https://trello.com/c/7G31eLks/1-general-audio-rewrite)

Fr6jDJF:
- #366 
- Downloading status/response so ppl don't re-queue (large tracks x100MB)

Jonny0181:
Maybe some sort of progress on the [p]song command out of the duration of the song, and something to be able to scroll through the queue with reactions when It doesn't show the whole queue?

Br0kenMaverick:
An embed control for a music dedicated text channel, with:
Info:
-Now playing
-Next 5 or 10 queued songs
Reaction buttons:
-play, pause, stop, volume.
-help

Something like this, but well put together :'D https://cdn.discordapp.com/attachments/295226583936008192/295227238423461890/Captura.PNG

PJP2810:
The ability to ADD a local list to the current queue (at least whilst it's playing from local music) e.g. [p]local start Playlist1 then whilst that is playing [p]local add Playlist2 Also if/when this is added it would make sense to be able to play multiple at once with something like [p]local start Playlist1 Playlist2 etc

rtil5:
a couple things if they haven't already been suggested: 

- queue pagination
- ability for anyone to skip/delete their own queued songs
- more information on each song (length, who queued it)
- in general a more verbose bot that lets users know what's playing and when, etc.

nadekobot and erisbot have good examples of these queuing features

PJP2810:
Fix emptydisconnect.

[...]

PR #931, opened 2017-08-17
Show info for songs when they are queued

alexwh:
Was requested in #644 and listed in #623

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -1391,6 +1391,19 @@ class Audio:
         if "[SEARCH:]" not in url and "youtube" in url:
             url = url.split("&")[0]  # Temp fix for the &list issue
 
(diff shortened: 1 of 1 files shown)


Twentysix26:
Works as expected. I took the liberty to adjust the queue info message a bit and to delete the command's message (if possible).
Would you mind extending this behaviour to the first song that gets `[p]play`ed too? I think it would be good for consistency. Currently there's no message at all when the first song gets played.

alexwh:
No worries, didn't put much thought into the formatting admittedly. I just duplicated the code for now, not really confident enough to find a DRYer way of doing it with this codebase.

Tobotimus:
Sorry this didn't make it in, V2 is now feature-locked. Also #1232 seems to have implemented this.

Issue #822, opened 2017-06-20
Add non-shared server bot blacklist and/or whitelist


Aioxas:
As name implies, add the ability for a bot owner to blacklist someone from using the bot, similar to hackban, but said person can stay in the server(s). Not sure if this would be needed on whitelist as well.

This would be done using the user.id. An example could be something like this:

```python

    @blacklist.command(name="hackadd")
    async def _blacklist_hackadd(self, uid):
        """Adds user to Red's global blacklist.
        For use if the owner doesn't share a server with the offending user."""
        if uid not in self.global_ignores["blacklist"]:
            self.global_ignores["blacklist"].append(uid)
            self.save_global_ignores()
            await self.bot.say("User has been blacklisted.")
        else:
            await self.bot.say("User is already blacklisted.")
```

This could be done either for v2 or v3. If the idea is too nonsensical, it can be scrapped out then.

Twentysix26:
My solution (#1068) is slightly different, as in you can't input any number you want and the user has to at least share one server with the bot, but I think it covers the majority of use cases. Waiting for more reviews on it.

Aioxas:
looks good 26, it covers the case of gibberish number adding that I might have missed. I guess we can close this when that gets accepted or something

Twentysix26:
Merged

PR #941, opened 2017-08-22
Add a user to the blacklist using its ID

laggron42:
#822 So I'm answering to this issue by adding the two commands `[p]blacklist hackadd <ID>` and `[p]blacklist hackremove` which add a user to the bot's blacklist using its ID, with a line that checks if the ID is valid. But (I know that's weird asking this in a PR) the command `hackremove` does not works (says the user isn't blacklisted even if it is) and I couldn't get help in #coding so I'm asking there how to correct that problem before the merge.
screenshot

### DIFF
--- a/cogs/owner.py
+++ b/cogs/owner.py
@@ -554,6 +554,38 @@ class Owner:
         else:
             await self.bot.say("User is not blacklisted.")
 
+    @blacklist.command(name="hackadd", pass_context=True)
+    async def _blacklist_hackadd(self, ctx, user):
+        """Adds user to Red's global blacklist using its ID."""
+
+        test = discord.utils.get(ctx.message.server.members, id=user)
+        if test is None:
+            await self.bot.say("User not found. Are you sure you provided the good user ID?")
+            return
+        else:
+            if user not in self.global_ignores["blacklist"]:
+                self.global_ignores["blacklist"].append(user)
+                self.save_global_ignores()
+                await self.bot.say("User has been blacklisted.")
+            else:
+                await self.bot.say("User is already blacklisted.")
+
+    @blacklist.command(name="hackremove", pass_context=True)
+    async def _blacklist_hackremove(self, ctx, user):
+        """Remove user to Red's global blacklist using its ID."""
+
+        test = discord.utils.get(ctx.message.server.members, id=user)
+        if test is None:
+            await self.bot.say("User not found. Are you sure you provided the good user ID?")
+            return
+        else:
+            if user not in self.global_ignores["blacklist"]:
+                self.global_ignores["blacklist"].remove(user)
+                self.save_global_ignores()
+                await self.bot.say("User has been removed from the blacklist.")
+            else:
+                await self.bot.say("User is not blacklisted.")
+
     @blacklist.command(name="list")
     async def _blacklist_list(self):
         """Lists users on the blacklist"""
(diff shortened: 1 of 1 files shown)


cogs/owner.py
```diff
@@ -554,6 +554,38 @@ def check(m):
         else:
             await self.bot.say("User is not blacklisted.")
 
+    @blacklist.command(name="hackadd", pass_context=True)
+    async def _blacklist_hackadd(self, ctx, user):
+        """Adds user to Red's global blacklist using its ID."""
+
+        test = discord.utils.get(ctx.message.server.members, id=user)
```
palmtree5: should use `self.bot.get_all_members()` instead of `ctx.message.server.members`

cogs/owner.py
```diff
@@ -554,6 +554,38 @@ def check(m):
...
+                self.save_global_ignores()
+                await self.bot.say("User has been blacklisted.")
+            else:
+                await self.bot.say("User is already blacklisted.")
+
+    @blacklist.command(name="hackremove", pass_context=True)
+    async def _blacklist_hackremove(self, ctx, user):
+        """Remove user to Red's global blacklist using its ID."""
+
+        test = discord.utils.get(ctx.message.server.members, id=user)
```
palmtree5: `self.bot.get_all_members()` here too

cogs/owner.py
```diff
@@ -554,6 +554,38 @@ def check(m):
...
+        if test is None:
```
palmtree5: TBH there's situations where this could be None but the user's ID could still be in the blacklist. Having this check would prevent removing those IDs without flat out resetting the blacklist (which probably would not be desirable)

cogs/owner.py
```diff
@@ -554,6 +554,38 @@ def check(m):
...
+            await self.bot.say("User not found. Are you sure you provided the good user ID?")
+            return
+        else:
+            if user not in self.global_ignores["blacklist"]:
```
palmtree5: Uhh, why would you want to remove a user that's NOT in the blacklist? I'd drop the `not` on this line and your issue should be fixed

palmtree5 (changes requested):

palmtree5:
See my request for changes

laggron42:
img_1420
Oh kill me I can't click on `Approuve` on mobile... Wait for tomorrow :c

Kowlin:
@retke. You cannot approve your own PR's. Logic does apply on GitHub

laggron42:
@Kowlin nope was trying to approve palm's changes

Twentysix26:
@retke He hasn't done any change, that's a code review

laggron42:
*One day I will not ask stupid questions and don't nuke my PRs, one day...*

laggron42:
So the user won't be able to interact with users who's not in the server, which is stupid for a hack action, but the issue is fixed. Gonna check how I can check if the ID is a user ID (because it won't be clever to have non-ID object in the json, juste imagine all the poop str), without checking in the member's list

laggron42:
I know what I did is shit, because you just can't add a user if the bot does not share a server with him, which is stupid. I would like to commit again to fix this but I deleted the repository. I think it's better to just close this now

laggron42:
If it's still needed, I can make a quick PR, and a working one this time

PR #945, opened 2017-08-24
Add !clearqueue command

EdwardJXLi:
Implement command for internal `_clear_queue` function.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -1798,6 +1798,13 @@ class Audio:
 
         await self.bot.say(msg)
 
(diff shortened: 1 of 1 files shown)


jorivanee (changes requested):
I would change clearqueue to a mod command, idk what permission, but this would definitely lead to trolls

Tobotimus:
It seems as though you've deleted your branch for this PR.

Since V2 is feature locked, I suggest PRing this once we see some more progress on audio for V3. I'd also recommend mentioning this over at #623

PR #950, opened 2017-08-28
[V3] [Audio] queues, mixing, and downloader

flapjax:
I'm going to get this massive boulder rolling... here are a few features I've implemented.

- basic queue system for downloaded songs (use `[p]play`)
- simultaneous/overlayed playing of sounds (nice proof of concept for sfx, use `[p]local`)
- data folders for local sounds and the download cache (using `cog_data_path` of course!)

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -8,12 +8,73 @@ from concurrent.futures import ThreadPoolExecutor
 
 import discord
 import youtube_dl
-from core import Config
+from core import Config, checks
 from core.data_manager import cog_data_path
 from discord import FFmpegPCMAudio, PCMVolumeTransformer
 from discord.ext import commands
 
 log = logging.getLogger("red.audio")
+
+
+class NoVoiceClient(Exception):
+    pass
+
+
+class VoiceManager:
+
+    def __init__(self, ctx):
+        self.guild = ctx.guild
+        self.text_channel = ctx.message.channel
+        self.voice_client = None
+        self.disc_timer = 0
+        self.timeout = 10
+        self.queue = []
+        self.loop = asyncio.get_event_loop()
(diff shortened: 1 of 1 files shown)


cogs/audio/audio.py
```diff
@@ -1,112 +1,367 @@
...
         self.bot = bot
+        self.conf = Config.get_conf(self, identifier=8675309)
+        self.conf.register_global(
+            **self.default_global_settings
+        )
+        self.conf.register_guild(
+            **self.default_guild_settings
+        )
+        self.data_path = cog_data_path(self)
+        self.local_path = os.path.join(self.data_path, 'local')
```
tekulvw: take a look at the python documentation for `pathlib`, it'll make your life easier by turning this `os.path` crap into `self.data_path / 'local'`

cogs/audio/audio.py
```diff
@@ -1,112 +1,367 @@
...
+        self.downloader_path = os.path.join(self.data_path, 'cache')
+        self.downloader = Downloader(self.downloader_path)
+        if not os.path.exists(self.local_path):
+            os.makedirs(self.local_path)
```
tekulvw: `pathlib` will also make this easier, turning it into `self.local_path.mkdir(exists_ok=True)`

flapjax: Done and done! Only problem I ran into with pathlib was that FFmpegPCMAudio did not enjoy being fed the Path object. Using the string representation worked.

tekulvw:
@flapjax do you want to see this merged or do you want to hold off for a bit?

flapjax:
Hold off for now, unless someone else has tested it recently. Imports are likely broken since I opened this before we changed up the folder structure a while back.

tekulvw:
Closing because of staleness and further work on alternative audio solutions.

PR #972, opened 2017-09-09
[Config] Add MongoDB driver

tekulvw:
This code *should* not impact any current uses of V3. The bot as it is in this PR will not actually touch or load any of the mongo code. This PR is simply a stepping stone towards mongo integration.

### DIFF
--- a/docs/framework_config.rst
+++ b/docs/framework_config.rst
@@ -192,5 +192,11 @@ Driver Reference
 
 .. automodule:: redbot.core.drivers
 
-.. autoclass:: red_base.BaseDriver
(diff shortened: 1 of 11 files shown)


tekulvw:
bumping this because it's working now. to test, run mongod and create a new instance of the bot. It will ask you for host, port, and login credentials. If you have not setup login credentials just leave those fields blank.

Kowlin:
From my limited testing I've done it seems to work fine, however if we can get some people on board with more Mongo experience such as @AznStevy to try it out that would be great.

Kowlin (dismissed):

Kowlin (approved):

PR #980, opened 2017-09-12
[V3 Core] Fix reload of core cogs

Tobotimus:
This solves the issue of core cogs not being found when doing `reload`


tekulvw (commented):
I need to actually test this myself tomorrow but it looks good at first glance.

tekulvw (approved):

PR #983, opened 2017-09-13
[V3] Allow Red to be run with custom bot class

Tobotimus:
This opens up opportunities for custom bots (e.g. selfbots) to be run seamlessly through Red's framework.


Tobotimus:
Should we perhaps expand on this whilst the PR is open? It would doubtlessly be beneficial for anyone trying to do their own setup to split the `main` function into `initialise` and `main`/`run`/`login`, but do we have any other ideas for how this is designed? Perhaps we could think up a different solution to this dict abuse 😬

Tobotimus:
Gonna close this for now as it's a bit out of date and not really a priority right now. Might try it again after #1017 is merged, and perhaps other PRs which add a bit more to the bot's initialisation.

PR #830, opened 2017-06-29
[CustomCom] [WIP] V3 Rewrite

QingyuChai:
Hello.

I'll be trying to rewrite this.

PR #1003, opened 2017-10-06
[V3] CustomCom

palmtree5:
I've tested this and it's working. See also #830 for more info on what's included in this


tekulvw (approved):
lgtm

PR #1005, opened 2017-10-09
[V3][Docs] add install guides to the documentation

palmtree5:
This adds install guides for the following operating systems:
- Windows
- OSX
- Ubuntu 16.04
- Debian Stretch
- CentOS 7
- Raspbian Stretch

One note: The Windows and Ubuntu pages seem to be causing issues with the sidebar for a reason that isn't immediately apparent to me

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -7,6 +7,17 @@ Welcome to Red - Discord Bot's documentation!
 =============================================
 
 .. toctree::
+    :maxdepth: 1
+    :caption: Installation Guides:
+
+    install_windows
+    install_mac
+    install_ubuntu
+    install_debian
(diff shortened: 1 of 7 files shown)


palmtree5:
Ok, cannot reproduce the problem I was having. 

System 1:
- Ubuntu 16.04, Python 3.5, in a virtualenv

System 2:
- Windows 10 (Version 1709, build 16299.15), Python 3.6.3, no virtualenv, new install of Python too

So I guess this is ready unless a) someone sees something that's wrong or should be changed/added/removed, or b) someone can reproduce the issue (in which case we'll have to find a fix)

tekulvw (approved):

Issue #850, opened 2017-07-07
[Rewrite] Launcher


palmtree5:
Any suggestions for the rewrite when it comes to the launcher?

boswelja:
Not sure if this was already a thing or not, but what about checking for updates on launcher start?

tekulvw:
It's been a thing for about a year now, it's harder than it sounds to get right.

tekulvw:
If #853 goes through we'll need to have users select a "cog install" path where downloader will put installed cogs. This should default to `~/.config/something` on mac/*nix and `%appdata%/somewhere` on windows.

boswelja:
What about a way to install / uninstall cogs from the launcher as well?

tekulvw:
Right now that's only possible if cogs are loaded prior to launcher start

Kowlin:
Actually I was thinking of a method of adding autostarting systems (like pm2 and supervisord) into the launcher. This seems like a good thing to pick up for it, check what is installed. And provide options accordingly

Jan200101:
pm2 requires root permissions to add a task to autostart that could be a problem

Kowlin:
Yet another reason to use Supervisor 😏

palmtree5:
Could be a good thing though one would need to generate files for that.
I've actually looked into this myself to some extent

Twentysix26:
Considering the variety of process managers, the need for root, the amount of stuff that can go wrong with automating this, I think it's better to leave this to the user. They installed a bot on a Linux vps, I'm sure they can spend 2 additional minutes looking at a 5 steps tutorial

palmtree5:
>I'm sure they can spend 2 additional minutes looking at a 5 steps tutorial

That I've seen multiple people manage to screw up in various methods. Automating at least the creation of such a file could potentially reduce issues caused by people making incorrect changes or failing to change certain things in the files (for example, username/group or python path). Even if we manage to clarify our docs to reduce these issues, someone will probably still manage to screw it up despite that.

[...]

Issue #569, opened 2017-01-10
"WinError 5 Access denied" on Windows


KrisDLuna:
* Affects probably any Windows OS with UAC
* Running in Admin mode doesn't help

After the latest update, Python runs into this error when installing required libs.

> Successfully installed aiohttp-1.0.5 async-timeout-1.1.0 chardet-2.3.0 discord.py-0.16.0 imgurpython-1.1.7 multidict-2.1.4 pip-9.0.1 requests-2.12.4 websockets-3.2 youtube-dl-2017.1.8
> Exception:
> Traceback (most recent call last):
>   File "C:\Users\Joebob6778\AppData\Local\Programs\Python\Python35-32\lib\site-packages\pip\basecommand.py", line 215, in main
>     status = self.run(options, args)
>   File "C:\Users\Joebob6778\AppData\Local\Programs\Python\Python35-32\lib\site-packages\pip\commands\install.py", line 423, in run
>     shutil.rmtree(target_item_dir)
>   File "C:\Users\Joebob6778\AppData\Local\Programs\Python\Python35-32\lib\shutil.py", line 488, in rmtree
>     return _rmtree_unsafe(path, onerror)
>   File "C:\Users\Joebob6778\AppData\Local\Programs\Python\Python35-32\lib\shutil.py", line 383, in _rmtree_unsafe
>     onerror(os.unlink, fullname, sys.exc_info())
>   File "C:\Users\Joebob6778\AppData\Local\Programs\Python\Python35-32\lib\shutil.py", line 381, in _rmtree_unsafe
>     os.unlink(fullname)
> PermissionError: [WinError 5] Access is denied: 'C:\Jason\Red-DiscordBot\lib\aiohttp\_websocket.cp35-win32.pyd'

So I'm assuming 'aiohttp_websocket.cp35-win32.pyd' is locked to read-only.

I was able to bypass this by editing the init.py for aiohttp though, adding this:

> import errno, os, stat, shutil
> 
> def handleRemoveReadonly(func, path, exc):
>   excvalue = exc[1]
>   if func in (os.rmdir, os.remove) and excvalue.errno == errno.EACCES:
>       os.chmod(path, stat.S_IRWXU| stat.S_IRWXG| stat.S_IRWXO)
>       func(path)
>   else:
>       raise
> 
> shutil.rmtree(filename, ignore_errors=False, onerror=handleRemoveReadonly)

Then using the Launcher to reinstall requirements (basic + audio).
[...]

PR #1006, opened 2017-10-10
[V3] Launcher

palmtree5:
This PR adds a launcher for V3.

**Features:**
* Starting Red (with or without autorestart)
  - Lists available instances on the spot and has the user select one to be used
* Updating Red (though only what is on PyPI atm)
* Instance creation (using `redbot-setup` via an import) (requested in #850)
* Some basic debug info (Python version, Red version, OS version/arch, user who ran the bot) (as requested in #850)

Obviously there's some improvements to be made still but this is a starting point for an actual launcher

### DIFF
--- a/launcher.py
+++ b/launcher.py
@@ -7,12 +7,15 @@ import os
 import platform
 import subprocess
 import sys
+import argparse
+import time
 
 import pkg_resources
 from redbot.core.data_manager import config_file
 from redbot.setup import basic_setup
 
 PYTHON_OK = sys.version_info >= (3, 5)
+INTERACTIVE_MODE = not len(sys.argv) > 1  # CLI flags = non-interactive
 
 INTRO = ("==========================\n"
          "Red Discord Bot - Launcher\n"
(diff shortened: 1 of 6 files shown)


palmtree5:
Should also note that I included a restart command in core because I needed it to test the autorestart functionality anyway

palmtree5:
CLI flags are now implemented in this. Also added the ability to update from what's on github (the `V3/develop` branch) and support for the various extras (voice, docs, test, mongo). Should note that I was having issues with updating on Windows 10 (with Python 3.6) using this (specifically an "Access Denied" error when pip attempted to remove the local copy of https://github.com/Rapptz/discord.py/blob/rewrite/discord/bin/libopus-0.x86.dll). Could be a 3.6 on Windows problem or it could've been completely random, but I do know that the user running the launcher has full control on that file as shown in the permissions dialogue

palmtree5:
Ok, that issue is happening with 3.5 on Windows as well (at least for me)

palmtree5:
Issue is NOT occurring on Linux (Ubuntu 16.04, specifically). Which means this is definitely Windows-specific

palmtree5:
Issue is also not occurring if I manually update discord.py on Windows (i.e. `py -3.6 -m pip install -U https://github.com/rapptz/discord.py/tarball/rewrite#egg=discord.py[voice]`)

palmtree5:
Seems this is the same thing that was happening in #569

palmtree5:
Problem solved (somewhat). I had to remove the instance creation stuff from the launcher in order to make the update function work properly

palmtree5:
Ok, problem officially solved, and instance creation support readded

palmtree5:
Ok, current summary of what the launcher has:

* Running Red (with or without autorestart)
* Updating Red (either from PyPI or Github)
* Instance creation (via `redbot-setup`)
* Debug info (Python version, Red version, OS version/arch, user who ran the launcher)
* CLI flags for all of the above except instance creation

Things that could probably still be added:
* support for selecting/providing (depending on whether the user is in interactive mode or not) cli flags to be passed to `redbot`

If anyone can think of anything else, feel free to suggest it. Otherwise, this is probably ready pending testing

palmtree5:
Scrapping this PR because of some issues. Going to just flat out start over on a fresh branch

PR #1009, opened 2017-10-12
begin subclassing context to expose commands

mikeshardmind:
also, while doing this, stuff a quality of
life command into context
cogs can issue `await ctx.show_help()` with this change.


tekulvw:
Closed with #988

PR #1012, opened 2017-10-14
Add issue and pull request template

palmtree5:
Just a starting point, feel free to suggest changes

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,5 +1,3 @@
-Please do **NOT** use the issue tracker for support questions. Any questions or comments of that nature can be answered on our support server located [here](https://discord.gg/red).
-
 Please be sure to read through other issues as well to make sure what you are suggesting/reporting has not already
 been suggested/reported
 
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,22 @@
+Please do **NOT** use the issue tracker for support questions. Any questions or comments of that nature can be answered on our support server located [here](https://discord.gg/red).
+
+Please be sure to read through other issues as well to make sure what you are suggesting/reporting has not already
+been suggested/reported
+
+### Type:
+
+- [ ] Suggestion
+- [ ] Bug
+
+### Brief description of the problem
+
+### Expected behavior
+
+### Actual behavior
+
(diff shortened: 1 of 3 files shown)


tekulvw (dismissed):

.github/ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,22 @@
+Please do **NOT** use the issue tracker for support questions. Any questions or comments of that nature can be answered on our support server located [here](https://discord.gg/red).
```
Kowlin: I strongly disagree with this. A issue tracker can and should also be for support. Its just that nobody wants to do them. However! This is actually one of the best tools we have available to us. People will mostly Google first before going to a chat.

Chovin: I agree, I think we should allow both methods.
The advantage of directing them to our chat as well is that there is a possibility they may become enthusiastic new members of our community, which lives in the Discord anyway.

tekulvw: I agree with Palm. Support questions should either be answered in the docs or asked in the support server. I am fully behind the idea that GH issues should be for code bugs/feature suggestions only. @Twentysix26 your time as tie-breaker has come lol

palmtree5: For reference, I didn't pull that bit about not using issues for support questions out of nowhere. It's literally in [here](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/.github/CONTRIBUTING.md#what-kinds-of-contributions-are-we-looking-for) which has been in for 3 months now

Twentysix26: I think support questions are fine as issues too, that's how it works in most projects. And I agree with @Kowlin , they have the advantage that solutions to common issues *will* show up on Google too. Not to mention that they can be referenced in Discord to avoid repeating the same steps everytime.  
I don't see a reason to limit issues to code topics only, IMO it would do more harm than good.

palmtree5:
Ok, per the vote, the language referenced in @Kowlin's review has been removed from `ISSUE_TEMPLATE.md` and `CONTRIBUTING.md` (where I had pulled it from)

tekulvw (approved):

Kowlin:
Good, now lets pass this onto develop so it has an actual impact on stuff that's going on right now.

PR #1013, opened 2017-10-15
Fix #989: Unable to repeat queue on latest commit

### LINKED ISSUE #989: [V2] Unable to repeat queue on latest commit
Hey, a new error that occurs on the audio cog has appeared recently, blocking the repeat function of the queue. Let me explain what the problem is:
1. I make my queue, using `[p]play` or `[p]queue`
2. I toggle the repeat function 

As of now, the queue show normally, but when the first song will be finished/skipped, the next song will be playing as usual __but__ the queue won't work, showing an error, and it won't repeat. After the last song, the status will be reset (no queue, no current playing song).
Here is the error that shows when typing the `[p]queue` command after the first song was skipped: 
```Py
Exception in command 'queue'
  File "lib/discord/ext/commands/core.py", line 50, in wrapped
  File "/home/laggron/Coding/Red/cogs/audio.py", line 1734, in _queue
    return await self._queue_list(ctx)
  File "/home/laggron/Coding/Red/cogs/audio.py", line 1799, in _queue_list
    queue_song_list = await self._download_all(queued_song_list, channel)
  File "/home/laggron/Coding/Red/cogs/audio.py", line 491, in _download_all
    d = Downloader(queued_song.url)
AttributeError: 'str' object has no attribute 'url'

The above exception was the direct cause of the following exception:

  File "lib/discord/ext/commands/bot.py", line 846, in process_commands
    yield from command.invoke(ctx)
  File "lib/discord/ext/commands/core.py", line 374, in invoke
    yield from injected(*ctx.args, **ctx.kwargs)
  File "lib/discord/ext/commands/core.py", line 54, in wrapped
    raise CommandInvokeError(e) from e
discord.ext.commands.errors.CommandInvokeError: Command raised an exception: AttributeError: 'str' object has no attribute 'url'
```

I talked to the members of the server many times of that bug but no one seems to care of that, and the staff wasn't aware of the situation, so here is the issue.
Note that (as weird as this can be) the bug only occurs on __freshly installed instances__ (as of now this show be for those younger than a month). Don't tell me how, as all of the bots are on the same commits, this don't happen on older bots.

mvanalphen:
Resolves an issue where repeat would break the queue. Turns out I forgot to replace a URL string with a Queued_Song instance (that I had already instantiated).

Fixes https://github.com/Cog-Creators/Red-DiscordBot/issues/989.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -2177,7 +2177,7 @@ class Audio:
                     return
                 if repeat and last_song:
                     queued_last_song = QueuedSong(last_song.webpage_url, last_song_channel)
-                    queue.append(last_song.webpage_url)
+                    queue.append(queued_last_song)
             else:
                 song = None
             self._set_queue_nowplaying(server, song, channel)


aikaterna:
Thank you for taking the time to work on this. Tested and works.

Kowlin (approved):
Looks good with me, but I want more reviews on this before we go ahead and merge this.

tekulvw (approved):

laggron42 (approved):
tested, gonna keep that

MalikAza (approved):

tekulvw:
@Twentysix26 @palmtree5 @irdumbs need a review/approval

Twentysix26 (approved):

PR #1017, opened 2017-10-15
[V3 RPC] Add basic RPC functionality

tekulvw:
* Uses `aiohttp-json-rpc` library
* Need a list of methods to provide to the RPC server


tekulvw:
I'm going to make this enable-able from a startup flag. We shouldn't be opening the bot application up to outside applications without user knowledge.

palmtree5 (approved):

PR #1028, opened 2017-10-16
[Alias V3] Refactored functions

Aioxas:
Features of this PR:
Duplicate code reduced
trailing whitespace removed


Aioxas:
Seems it breaks on trying to run "self.bot.get_command()" but it says that self.bot doesn't have that attribute. Which is weird because it clearly shows it.

tekulvw:
Will revisit in a newer PR.

PR #1032, opened 2017-10-17
[V3] NumPy Docstrings

Tobotimus:
This PR allows our docstrings to be formatted nicely in code, whilst RTD is mostly unaffected. See the [NumPy specification](https://github.com/numpy/numpy/blob/master/doc/HOWTO_DOCUMENT.rst.txt) for more info.

I also (manually 😓) went through and updated almost all of our old docstrings to NumPy. I have skipped over protected methods.

I also set the default role for sphinx to `:any:`. This means that whenever you surround something with backticks without specifying a role (the things in between `:` like `:class:`), the `:any:` role will be implied, meaning it will look for "any" reference to the name within the backticks, and use the first result as a link. This reduces the need to specify the role for every object / section we wish to reference.

Existing PRs which add documentation should conform to this new style (sorry Palm).

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -37,6 +37,7 @@ extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.intersphinx',
     'sphinx.ext.viewcode',
+    'sphinx.ext.napoleon',
     'sphinxcontrib.asyncio'
 ]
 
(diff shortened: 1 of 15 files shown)


tekulvw:
In `redbot/core/drivers/__init__.py` you added three import statements.

We cannot do this because we're not guaranteed that we have the requirements for all of the drivers, hence why the tests failed. Take a look at my mongo PR and you'll see how I handle this.

tekulvw (approved):

PR #1033, opened 2017-10-17
[V3] [Config] Redesign "all_from_XXX" and "clear_all" methods

Tobotimus:
This PR essentially provides an alternative to config's `Group.all_from_kind` method, with the following changes:
 - An arbitrary object (guild, user, whatever) no longer needs to be passed in to get all of its "sibling's" data. As an example, retrieving all bank accounts for a guild-based bank:
    - Before:
        ```py
        accs = await _conf.member(some_member).all_from_kind()
        ```
    - After:
        ```py
        accs = await _conf.all_members(guild)
        ```
 - Snowflakes (guild IDs, user IDs etc) which are keys in the returned dictionary are now of type `int`.

I am willing to remove `all_from_kind` from this branch entirely if there is no need for it, just thought I'd put this solution out there first before making that change.

This also makes me question whether or not `Group.clear_all` should be refactored in a similar manner.

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -244,32 +244,6 @@ class Group(Value):
         defaults.update(await self())
         return defaults
 
-    async def all_from_kind(self) -> dict:
-        """
-        This method allows you to get all data from all entries in a given Kind. It will return a dictionary of Kind
-        ID's -> data.
-
-        .. note::
-
-            Any values that have not been set from the registered defaults will have their default values
-            added to the dictionary that this method returns.
-
-        .. important::
-
-            This method is overridden in :py:meth:`.MemberGroup.all_from_kind` and functions slightly differently.
-
-        :rtype: dict
-        """
-        # noinspection PyTypeChecker
-        all_from_kind = await self._super_group()
-
-        for k, v in all_from_kind.items():
-            defaults = self.defaults
-            defaults.update(v)
-            all_from_kind[k] = defaults
-
-        return all_from_kind
-
     async def set(self, value):
         if not isinstance(value, dict):
             raise ValueError(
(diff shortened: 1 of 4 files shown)


tekulvw:
I see no reason to keep `all_from_kind`.

Tobotimus:
Alright, config's `Group.clear_all` method has been redesigned now in a similar fashion. All uses of `clear_all` and `all_from_kind` have also been replaced within the existing code, however I'm aware of some other PRs which will also need these methods replaced (sorry again @palmtree5 )

I should also mention that the docstrings on these methods are using NumPy style in anticipation of #1032 being merged. This PR is more or less dependent on that PR for now because of that fact, although there will be merge conflicts for me to resolve if/when it is merged.

redbot/core/config.py
```diff
@@ -648,3 +586,265 @@ def member(self, member: discord.Member) -> MemberGroup:
...
+        """
+        if not scopes:
+            group = Group(identifiers=(self.unique_identifier),
+                          defaults={},
+                          spawner=self.spawner)
+        else:
+            group = self._get_base_group(*scopes)
+        await group.set({})
+
+    def clear_all(self):
```
tekulvw: Why are these not coroutines?

Tobotimus:
The `clear_all_xxx` and `all_xxx` methods are now all async functions which return a dict, instead of functions which return a coroutine.

redbot/core/config.py
Tobotimus: Just gonna more or less summarise our conversation from discord for the record: I made them functions which return a coroutine because it's a micro-optimisation which removes one extra `await` statement, so the task gets done slightly faster, however I agree with your sentiment that the clarity of the function (esp. in documentation) is more important than this micro-optimisation, so I've agreed to make the changes.

tekulvw (approved):

PR #1034, opened 2017-10-18
[V3] Mod rewrite

palmtree5:
Supersedes #785 (making a new PR because of a branch rename)


palmtree5:
Ok, we should be good on this now that the tests are passing.

tekulvw (approved):

Issue #1020, opened 2017-10-16
[V3] Debug Command doesn't support over 2k character outputs


Redjumpman:
As the title suggests, the debug command does not support outputs that are greater than 2k through pagification or other methods. 

Steps to reproduce:
Run red with --dev flag
Use [p]debug dir(bot)

tekulvw:
This has been fixed.

PR #1037, opened 2017-10-19
[Dev V3] Fix for Debug

Redjumpman:
Debug will now properly display content over 2k characters.

This fixes issue #1020.

Requires the fix for pagify in PR #1036

### DIFF
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -115,7 +115,26 @@ class Dev:
 
         result = self.sanitize_output(ctx, str(result))
 
-        await ctx.send(box(result, lang="py"))
+        def check(m):
+            return m.content.lower() == "more"
+
+        pages = list(pagify(result, shorten_by=18))
+        for idx, page in enumerate(pages, 1):
+            if idx != 1:
+                last = await ctx.send(_("There are still {} messages. "
+                                        "Type `more` to continue."
+                                        "").format(len(pages) - (idx - 1)))
+
+                try:
+                    msg = await ctx.bot.wait_for('message', check=check, timeout=10)
+                except asyncio.TimeoutError:
+                    await ctx.send("Your msg timed out")
+                else:
+                    await ctx.channel.delete_messages([last, msg])
+                    await ctx.send(box(page, lang="py"))
+                finally:
+                    break
+            await ctx.send(box(page, lang="py"))
 
     @commands.command(name='eval')
     @checks.is_owner()


redbot/core/dev_commands.py
```diff
@@ -115,7 +115,26 @@ def sanitize_output(ctx: commands.Context, input_: str) -> str:
...
         result = self.sanitize_output(ctx, str(result))
 
-        await ctx.send(box(result, lang="py"))
+        def check(m):
+            return m.content.lower() == "more"
+
+        pages = list(pagify(result, shorten_by=18))
+        for idx, page in enumerate(pages, 1):
+            if idx != 1:
+                last = await ctx.send(_("There are still {} messages. "
```
tekulvw, line 124: Can you add this "more" functionality as a method in utils? I have a feeling it will come in handy for a lot of things.

redbot/core/dev_commands.py
```diff
@@ -115,7 +115,26 @@ def sanitize_output(ctx: commands.Context, input_: str) -> str:
 
         result = self.sanitize_output(ctx, str(result))
 
-        await ctx.send(box(result, lang="py"))
+        def check(m):
+            return m.content.lower() == "more"
```
Tobotimus, line 119: Just be aware that this will pass for any message "more" sent by any member in any channel which the bot sees. Remember to check that the channel and author match the context.

redbot/core/dev_commands.py
```diff
@@ -115,7 +115,26 @@ def sanitize_output(ctx: commands.Context, input_: str) -> str:
...
+
+                try:
+                    msg = await ctx.bot.wait_for('message', check=check, timeout=10)
+                except asyncio.TimeoutError:
+                    await ctx.send("Your msg timed out")
+                else:
+                    await ctx.channel.delete_messages([last, msg])
+                    await ctx.send(box(page, lang="py"))
+                finally:
+                    break
```
Tobotimus, line 136: This breaks out of the loop on the first try, as `finally` runs regardless of whether or not an exception was caught, raised, or never occurred at all.

redbot/core/dev_commands.py
```diff
@@ -115,7 +115,26 @@ def sanitize_output(ctx: commands.Context, input_: str) -> str:
...
+        for idx, page in enumerate(pages, 1):
+            if idx != 1:
+                last = await ctx.send(_("There are still {} messages. "
+                                        "Type `more` to continue."
+                                        "").format(len(pages) - (idx - 1)))
+
+                try:
+                    msg = await ctx.bot.wait_for('message', check=check, timeout=10)
+                except asyncio.TimeoutError:
+                    await ctx.send("Your msg timed out")
```
Tobotimus, line 131: I feel like this should instead just delete the "Type `more` to continue." message, since in my opinion that is a good enough indicator that their response timed out, and it doesn't clutter the chat.

redbot/core/dev_commands.py
Tobotimus: I agree, could be expanded on in the future too, reaction buttons would work quite nicely. But another PR for that.

Tobotimus (changes requested):

redbot/core/dev_commands.py
Tobotimus: I agree, could be expanded on in the future too as a reaction-based menu, but I suppose that should be in another PR.

palmtree5 (changes requested):
I just tried to test this and it doesn't give any sort of response when I type "more". Also, when there's only 1 message remaining, it should probably say "There is still 1 message." as opposed to "There are still 1 messages" because grammar

PR #1038, opened 2017-10-19
V3/start screen

Redjumpman:
Created a new screen for Red's start-up. Includes the following information:

- Bot Name
- Prefixes
- Red's Version
- Language
- Servers
- Users
- Cogs
- Commands
- If Sentry is reporting Errors
- If MongoDB is being used.
- Number of shards

image

This PR also adds a required bordered function I made to create the boxes. I stuck it in chat_utils incase someone else wanted to use it.

### DIFF
--- a/redbot/core/events.py
+++ b/redbot/core/events.py
@@ -1,21 +1,30 @@
 import datetime
 import logging
+import pkg_resources
 import traceback
 
 import discord
 from .sentry_setup import should_log
 from discord.ext import commands
 
-from .utils.chat_formatting import inline
+
+from . import data_manager
+from .utils.chat_formatting import inline, bordered
 from .core_commands import find_spec
+from colorama import Fore, Style
 
 log = logging.getLogger("red")
 sentry_log = logging.getLogger("red.sentry")
 
-INTRO = ("{0}===================\n"
-         "{0} Red - Discord Bot \n"
-         "{0}===================\n"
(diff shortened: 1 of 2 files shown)


DannyB5544:
Idea: Have the ticks/crosses light up green or red (of course, green for tick/red for cross)
Should be doable to single characters via escape characters

bobloy:
In this situation, what is Users? The # of users on all servers combined?

DannyB5544:
That's exactly what it is

Redjumpman:
@bobloy Yes that's the total users.
@DannyB5544 It will be hard to do that because of how I wrote my border function. I'll have to look into it.

DannyB5544:
I think it could be done in string. I'll experiment too

tekulvw:
As per @palmtree5's suggestion, can we add d.py version information?

Redjumpman:
Yeah that's a good one. Are there any others you would like for me to add?
I could also add some checks for the optional packages voice and for docs/tests

tekulvw:
That information would be good to add as well. I don't personally have anything else on my mind to add adn I think that after dpy/package information is in this PR we should go ahead and merge.

Redjumpman:
Added Discord.py Version
Added Docs/Test package checking
Added Voice package checking

New screen looks like this: 
image

palmtree5 (approved):

PR #1042, opened 2017-10-22
Fix "Playing..." tag

zsr2531:
This worked for me! Something happened with the wrapper that requires changing the way of updating the "Playing..." tag.

### DIFF
--- a/cogs/owner.py
+++ b/cogs/owner.py
@@ -408,11 +408,11 @@ class Owner:
 
         if game:
             game = game.strip()
-            await self.bot.change_presence(game=discord.Game(name=game),
-                                           status=current_status)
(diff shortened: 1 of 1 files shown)


Twentysix26:
Could you go into details on what this is supposed to fix?

zsr2531:
I saw a message in #announcements in the Discord server about "Playing..." tags not working. And I thought to myself, "Hmmm...That stopped working for me as well for some reason." (on another bot), and I did some research and found out if I change the `change_presence` function like so, it fixes the* issue.

Twentysix26:
The message after that asks to update discord.py to fix the issue. No code change is needed.

zsr2531:
I was using the latest discord.py version when I had the issue with another bot, I thought the issue was the same here. Seems like it wasn't.

PR #1043, opened 2017-10-22
[Docs] fix install commands

palmtree5:
This fixes the `pip install` commands to have `-U --process-dependency-links` in them.


tekulvw (commented):
What about the GH url? Especially since we're talking about doing releases prior to officially releasing people will need to be using the GH url instead of the pypi package.

palmtree5:
Updated with the option of installing from either pypi or github

tekulvw (approved):

PR #1053, opened 2017-10-24
[alias] Change message deepcopy to ordinary copy

calebj:
For extremely large servers (tens of thousands of members), copying the Server object can take several seconds and peg the CPU since this is a necessarily blocking operation. Testing needed to see if this fixes the bug and doesn't cause any others.


calebj:
Bump.

Twentysix26:
Looks good to me. I can't think of any possible side effects.

Twentysix26 (approved):

palmtree5 (approved):

PR #1059, opened 2017-10-25
[V3 Mod] handle RuntimeError from registering unchanged case types

palmtree5:


redbot/cogs/mod/mod.py
```diff
@@ -148,10 +153,12 @@ def __init__(self, bot: Red):
...
-
-        self.bot.loop.create_task(modlog.register_casetypes(casetypes_to_register))
-
-        self.last_case = defaultdict(dict)
+        try:
+            await modlog.register_casetypes(casetypes_to_register)
+        except RuntimeError:
+            pass
+        except:
+            raise
```
tekulvw: This is unnecessary, it will auto raise if not caught

tekulvw (approved):

PR #1063, opened 2017-10-27
[V3 Cog/Data Manager] Bundled Cog Data

tekulvw:
This PR does a few different things:
* Provides a method for cog creators to use to copy their data into the path available from `data_manager.cog_data_path(self)`
* Fixes some of the issues with cog manager not finding core cogs

### DIFF
--- a/redbot/core/data_manager.py
+++ b/redbot/core/data_manager.py
@@ -13,7 +13,8 @@ if TYPE_CHECKING:
     from . import Config
 
 __all__ = ['load_basic_configuration', 'cog_data_path', 'core_data_path',
-           'load_bundled_data', 'storage_details', 'storage_type']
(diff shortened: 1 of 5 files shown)


redbot/core/data_manager.py
```diff
@@ -108,6 +113,115 @@ def core_data_path() -> Path:
...
+        cog.
+
+    Examples
+    --------
+    >>> from redbot.core import data_manager
+    >>>
+    >>> def setup(bot):
+    >>>     cog = MyCog()
+    >>>     data_manager.load_bundled_data(cog, __file__)
+    >>>     bot.load_cog(cog)
```
palmtree5: This should be `bot.add_cog`

palmtree5 (dismissed):

palmtree5 (approved):

PR #1068, opened 2017-10-28
[Core] Target any user with blacklist / whitelist commands

Twentysix26:
This allows for the owner to globally whitelist / blacklist people from any server. The converter I've used could be implemented in other commands too.


Kowlin (approved):
Looks good.

tekulvw (approved):

PR #1073, opened 2017-11-02
Command doesn't work unless you put git+ in url

TurnrDev:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

The installation command on http://red-discordbot.readthedocs.io/en/v3-develop/guide_cog_creation.html didn't work. I adjusted the docs to reflect what I had to do to get it working

### DIFF
--- a/docs/guide_cog_creation.rst
+++ b/docs/guide_cog_creation.rst
@@ -19,7 +19,7 @@ Getting started
 
 To start off, be sure that you have installed Python 3.5 or higher (if you
 are on Windows, stick with 3.5). Open a terminal or command prompt and type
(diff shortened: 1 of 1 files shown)


jorivanee (approved):
Can confirm pip requires this, seeing as https://github.com/.... isn't a package

Kowlin (approved):

Kowlin:
You sure are onto something, I've given the current command and your changed PR a try and you are right that it does not work. However after looking at the docs I wonder if we will need to adjust the commands to install the latest stable/beta instead of the current dev version. But that's something for later.

PR #1089, opened 2017-11-10
[V3 Modlog] Modlog layout and formatting changes

Tobotimus:
### Type

- Bugfix
- Enhancement

I've made a couple changes to the modlog to make it:
- more mobile friendly,
- more readable and
- more logical

Changes are:
- User and moderator positions have been swapped, user's avatar appears instead of moderator. This makes the user more identifiable by the case.
- Emojis are actual emoji chars so they are rendered properly on mobile
- Title is no longer bold so it renders properly on mobile
- Channel field is slightly repositioned
- The *[p]reason <num> <reason>* bit is now formatted inline

### Desktop Screenshots
Old:

New:

### Mobile Screenshots
Old:

New:

Sorry about the massive screenshots.

### DIFF
--- a/redbot/cogs/mod/mod.py
+++ b/redbot/cogs/mod/mod.py
@@ -64,91 +64,91 @@ class Mod:
             {
                 "name": "ban",
                 "default_setting": True,
-                "image": ":hammer:",
+                "image": "\N{HAMMER}",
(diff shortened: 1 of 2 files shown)


palmtree5 (approved):
lgtm

PR #1098, opened 2017-11-14
[V3 Mod] rewrite softban to purge instead of ban/unban

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This rewrites softban so that it just purges messages from the specified user in the specified timeframe in every channel in the server, rather than banning and unbanning. This allows for some flexibility in the time period to remove messages for. As such, the stuff regarding tracking current softbans is removed as it is now unnecessary

This PR also makes a minor change to unban to make it not send an invite on unban if the bot doesn't share any guilds with the target user.

### DIFF
--- a/redbot/cogs/mod/locales/messages.pot
+++ b/redbot/cogs/mod/locales/messages.pot
@@ -5,13 +5,276 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: PACKAGE VERSION\n"
-"POT-Creation-Date: 2017-10-22 16:33-0800\n"
+"POT-Creation-Date: 2017-11-14 13:34-0900\n"
 "PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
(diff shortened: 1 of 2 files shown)


Twentysix26:
I would argue that with this PR you changed the command enough that it could be considered a new one altogether. 
My advice is to leave softban as it was, because:
- Softban is not a term we coined, it's something that is present in other bots as well. The user could have a certain expectation about what it does.
- If I read the command's name (with no previous knowledge of what the command is supposed to do) I expect that at the very least an expulsion will take place. With this PR this is no longer true  

Furthermore I wouldn't really consider this an enhancement: the original command is able to delete messages even in places where the bot has no permissions to, since it uses the ban's purge feature. This command requires "manage messages" permissions in every channel you attempt to delete from.
An actual enhancement would be giving the user the possibility to specify the days (1-7) with the current implementation of softban (leaving the default to 1).

PR #1100, opened 2017-11-17
[V3][Cleanup] change cleanup commands to use channel.purge

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This changes all of the cleanup commands to use [discord.TextChannel.purge](http://discordpy.readthedocs.io/en/rewrite/api.html#discord.TextChannel.purge)

As such, I removed the `mass_purge` and `slow_deletion` functions that I had ported over from v2 when I rewrote cleanup from `redbot.core.utils.mod`.

### DIFF
--- a/redbot/cogs/cleanup/__init__.py
+++ b/redbot/cogs/cleanup/__init__.py
@@ -3,4 +3,4 @@ from redbot.core.bot import Red
 
 
 def setup(bot: Red):
-    bot.add_cog(Cleanup(bot))
+    bot.add_cog(Cleanup())
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -6,7 +6,6 @@ from discord.ext import commands
 
 from redbot.core import checks
 from redbot.core.i18n import CogI18n
-from redbot.core.utils.mod import slow_deletion, mass_purge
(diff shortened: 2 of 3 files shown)


palmtree5:
Ok, so as it turns out, `channel.purge` is really only usable in `cleanup messages` and `cleanup after` (because in both cases, messages are deleted indiscriminately). In `cleanup text`, `cleanup user`, `cleanup bot`, and `cleanup self`, there's a check involved, meaning that the `limit` param passed to `channel.purge` is not necessarily going to be the number of messages deleted. I will have to modify those commands to do something else

palmtree5:
Ok so the four commands that couldn't really use channel.purge have been moved to finding their own list of messages to be deleted and then passing it to a helper function (`redbot.core.utils.mod.check_purge`, which takes the context from the command calling it and the list of messages to delete) which then either uses `channel.delete_messages` (if `ctx.bot.user.bot` is `True`) or loops through and uses `message.delete` on each message (if the user is not a bot). Also added a helper function for getting the minimum time a message should have been sent at to be eligible for deletion, which is utilized in the four commands that use checks

palmtree5:
ehh, what's already in place is working just fine anyway, no real need to change it now that I think about it

PR #1110, opened 2017-11-20
[V3 Launcher] add launcher

palmtree5:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Supersedes #1006

Current summary of what the launcher has:

* Running Red (with or without autorestart)
* Updating Red (either from PyPI or Github)
* Instance creation (via `redbot-setup`)
* Debug info (Python version, Red version, OS version/arch, user who ran the launcher)
* CLI flags for all of the above except instance creation

Things that could probably still be added:
* support for selecting/providing (depending on whether the user is in interactive mode or not) cli flags to be passed to `redbot`

### DIFF
--- a/redbot/launcher.py
+++ b/redbot/launcher.py
@@ -14,6 +14,9 @@ from pathlib import Path
 import pkg_resources
 from redbot.setup import basic_setup
 
+if sys.platform == "linux":
+    import distro
+
 config_dir = Path(appdirs.AppDirs("Red-DiscordBot").user_config_dir)
 config_file = config_dir / 'config.json'
 
@@ -188,13 +191,14 @@ def debug_info():
         os_info = platform.mac_ver()
         osver = "Mac OSX {} {}".format(os_info[0], os_info[2])
     else:
-        os_info = platform.linux_distribution()  # pylint: disable=deprecated-method
-        osver = "{} {} {}".format(os_info[0], os_info[1], platform.machine())
+        os_info = distro.linux_distribution()
+        osver = "{} {}".format(os_info[0], os_info[1]).strip()
     user_who_ran = getpass.getuser()
     info = "Debug Info for Red\n\n" +\
         "Python version: {}\n".format(pyver) +\
         "Red version: {}\n".format(redver) +\
         "OS version: {}\n".format(osver) +\
+        "System arch: {}\n".format(platform.machine()) +\
         "User: {}\n".format(user_who_ran)
     print(info)
     exit(0)
--- a/redbot/launcher.py
+++ b/redbot/launcher.py
@@ -1,0 +1,286 @@
+# pylint: disable=missing-docstring
+from __future__ import print_function
+
+import getpass
+import json
+import os
+import platform
+import subprocess
+import sys
+import argparse
+import appdirs
+from pathlib import Path
+
+import pkg_resources
+from redbot.setup import basic_setup
+
+config_dir = Path(appdirs.AppDirs("Red-DiscordBot").user_config_dir)
+config_file = config_dir / 'config.json'
+
+PYTHON_OK = sys.version_info >= (3, 5)
+INTERACTIVE_MODE = not len(sys.argv) > 1  # CLI flags = non-interactive
+
+INTRO = ("==========================\n"
+         "Red Discord Bot - Launcher\n"
+         "==========================\n")
+
+IS_WINDOWS = os.name == "nt"
+IS_MAC = sys.platform == "darwin"
+
+
+def parse_cli_args():
+    parser = argparse.ArgumentParser(description="Red - Discord Bot's launcher (V3)")
+    with config_file.open("r") as fin:
+        instances = json.loads(fin.read())
+    parser.add_argument("instancename", metavar="instancename", type=str,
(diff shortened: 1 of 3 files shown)


palmtree5:
So this introduces a dependency on [distro](https://github.com/nir0s/distro/) but only if on Linux (this is used for determining the distribution and version). Also the cli flag selecting thing has been finished, with an interactive walkthrough being done if running the launcher in interactive mode and a cli flag to pass in a file name containing a list of cli flags to be passed in (one per line) for non-interactive mode

calebj:
Recommend using `parser.parse_known_args()` and/or supporting the use case of adding CLI flags after a positional escape (--) by capturing extra positionals with `nargs='+'`

palmtree5:
Ok, this is now using `parser.parse_known_args` as well as setting `argparse.ArgumentParser`'s `allow_abbrev` kwarg to `False`. Thus to run the launcher in non-interactive mode with some flags being passed on to `redbot` (for example, `--dev` and `--debug`), one could do `python3 redbot/launcher.py <instance_name> -s --dev --debug` which would result in running `redbot <instance_name> --dev --debug`

tekulvw:
Can you catch `KeyboardInterrupt`?

mikeshardmind:
of course you can catch KeyboardInterupt.
```py
import time

try:
    while True:
        print('..')
        time.sleep(1)
except KeyboardInterrupt:
    print('\ndo cleanup')
```
results in:
```
..
..
..
^C
do cleanup
```
when interrupted

palmtree5:
ok, `KeyboardInterrupt` is now being handled for the launcher itself

redbot/launcher.py
```diff
@@ -0,0 +1,383 @@
...
+    elif resp == "y":
+        flags = []
+        print("Ok, we will now walk through choosing cli flags")
+        print("Would you like to specify an owner? (y/n)")
+        choice = user_choice()
+        if choice == "y":
+            print("Enter the user id for the owner")
+            owner_id = user_choice()
+            flags.append("--owner {}".format(owner_id))
+        print("Would you like to specify any co-owners? (y/n)")
```
Chovin: I think we should make this feature more hidden

redbot/launcher.py
```diff
@@ -0,0 +1,383 @@
...
+            print("Enter the user ids for the co-owners, separated by a space")
+            co_owner_ids = user_choice().split()
+            for co in co_owner_ids:
+                flags.append("--co-owner {}".format(co))
+        print("Would you like to specify any prefixes? (y/n)")
+        choice = user_choice()
+        if choice == "y":
+            print(
+                "Enter the prefixes, separated by a space (please note "
+                "that this will not handle prefixes containing a space)")
```
Chovin: maybe "(please note that prefixes containing spaces will have to be added later through the bot)" instead?

redbot/launcher.py
```diff
@@ -0,0 +1,383 @@
...
+            prefixes = user_choice().split()
+            for p in prefixes:
+                flags.append("-p {}".format(p))
+        print("Would you like to disable console input? (y/n)")
```
Chovin: this is kind of confusing to me.
an option to skip prompts makes sense but phrasing it like this is confusing to me as a user

redbot/launcher.py
```diff
@@ -0,0 +1,383 @@
...
+        choice = user_choice()
+        if choice == "y":
+            flags.append("--no-prompt")
+        print("Would you like to start with no cogs loaded? (y/n)")
+        choice = user_choice()
+        if choice == "y":
+            flags.append("--no-cogs")
+        print("Is this a selfbot? (y/n)")
```
Chovin, line 166: if it's not allowed (but we don't want to shut off the capability), maybe we should make this more hidden as well

redbot/launcher.py
```diff
@@ -0,0 +1,383 @@
...
+        if choice == "y":
+            print("Please note that selfbots are not allowed by Discord. See"
+                  "https://support.discordapp.com/hc/en-us/articles/115002192352-Automated-user-accounts-self-bots-"
+                  "for more information.")
+            flags.append("--self-bot")
+        print("Does this token belong to a user account rather than a bot account? (y/n)")
+        choice = user_choice()
+        if choice == "y":
+            flags.append("--not-bot")
+        print("Do you want to do a dry run? (y/n)")
```
Chovin, line 177: what's this?

redbot/launcher.py
```diff
@@ -0,0 +1,383 @@
...
+        name_num_map[str(counter+1)] = name
+        counter += 1
+    selection = user_choice()
+    try:
+        selection = int(selection)
+    except ValueError:
+        print("Invalid input! Try again.")
+        return None
+    else:
+        if selection not in list(range(1, counter+1)):
```
Chovin, line 229: shouldn't need list()
but it's still fine to merge and refactor later

redbot/launcher.py
Chovin: disregard

Chovin:
looks good to go. 👍 from me unless you guys want to hide those 2 flags before merging

Chovin (approved):
looks good. we can hide the selfbot / coowner flags later if we feel that is better

PR #1121, opened 2017-11-27
[V3 Help] Internationalize docstrings

tekulvw:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

In order for this to work you must add a decorator on top of the commands decorator using a function from the existing i18n object.

E.G.
```py
from discord.ext import commands
from redbot.core.i18n import Cogi18n

_ = CogI18n("SomeName", __file__)
i18n_decorator = _.i18n_decorator

class SomeName:
    @i18n_decorator
    @commands.command()
    async def some_command(self, ctx):
        """
        help text
        """
        await ctx.send(_("some i18n string"))
```

### DIFF
--- a/redbot/core/cog_manager.py
+++ b/redbot/core/cog_manager.py
@@ -291,10 +291,13 @@ class CogManager:
         invalidate_caches()
 
 
-_ = CogI18n("CogManagerUI", __file__)
+locales_dir = Path(__file__).parent / 'locales' / 'cog_manager'
+_ = CogI18n("CogManagerUI", __file__, locales_dir)
+i18n_command = _.i18n_decorator
 
 
 class CogManagerUI:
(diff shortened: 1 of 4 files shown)


mikeshardmind:
How exactly is this more efficient than?

```py
from discord.ext import commands
from redbot.core.i18n import Cogi18n

_ = CogI18n("SomeName", __file__)
some_doc_string = _("help text")

class SomeName:
    @commands.command(help=some_doc_string)
    async def some_command(self, ctx):
        await ctx.send(_("some i18n string"))
```

Which works without these changes by just passing the translated help string to d.py explicitly as the help, rather than d.py falling back to inspecting docstrings.

tekulvw:
Closing in favor of #1143

PR #1125, opened 2017-11-29
[V3 Core] add traceback, invite, leave, and servers commands

palmtree5:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This ports `[p]traceback`, `[p]join`, `[p]leave`, and `[p]servers` from v2. These commands are in `redbot/core/core_commands.py`. The functionality is pretty much the same as from v2


Kowlin (approved):
Looking good, but you lied about the ``join`` command :P, it doesn't exist.

PR #1127, opened 2017-11-29
 Fixed cleanup pulling more logs than required.

renebulous:
Added conditional statement to break loop once number of messages to be deleted has been reached.

**Description of problem identified**
Previously, on command [p] cleanup messages, the bot would pull the full one thousand messages (as indicated by limit=1000) even when number to cleanup = 1. This was because _break_ was not reached since it would only be reached when messages were older than or equal to 14 days. As a result, even if the first IF condition was not fulfilled, there was no way to end the loop until 1000 iterations. From my tests, waiting for 1000 messages to be pulled from the Discord servers could take up to 3 seconds. This is, of course, unnecessary if the user only seeks to cleanup a few messages.

**Solution**
Added a conditional statement to break the loop once len(to_delete) == number.

Hope this helps.

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -60,7 +60,7 @@ class Cleanup:
                 if len(to_delete) - 1 < number and check(message) and\
                         (ctx.message.created_at - message.created_at).days < 14:
                     to_delete.append(message)
-                elif (ctx.message.created_at - message.created_at).days >= 14:
+                elif len(to_delete) == number or (ctx.message.created_at - message.created_at).days >= 14:
                     too_old = True
                     break
                 tmp = message
@@ -109,7 +109,7 @@ class Cleanup:
                 if len(to_delete) - 1 < number and check(message) and\
                         (ctx.message.created_at - message.created_at).days < 14:
                     to_delete.append(message)
-                elif (ctx.message.created_at - message.created_at).days >= 14:
+                elif len(to_delete) == number or (ctx.message.created_at - message.created_at).days >= 14:
                     too_old = True
                     break
                 tmp = message
@@ -192,7 +192,7 @@ class Cleanup:
                 if len(to_delete) - 1 < number and \
                         (ctx.message.created_at - message.created_at).days < 14:
                     to_delete.append(message)
-                elif (ctx.message.created_at - message.created_at).days >= 14:
+                elif len(to_delete) == number or (ctx.message.created_at - message.created_at).days >= 14:
                     done = True
                     break
                 tmp = message
@@ -249,7 +249,7 @@ class Cleanup:
                 if len(to_delete) - 1 < number and check(message) and\
                                 (ctx.message.created_at - message.created_at).days < 14:
                     to_delete.append(message)
-                elif (ctx.message.created_at - message.created_at).days >= 14:
+                elif len(to_delete) == number or (ctx.message.created_at - message.created_at).days >= 14:
                     too_old = True
                     break
                 tmp = message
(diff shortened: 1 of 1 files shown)


redbot/cogs/cleanup/cleanup.py
```diff
@@ -1,345 +1,345 @@
...
+        too_old = False
+        tmp = ctx.message
+
+        while not too_old and len(to_delete) - 1 < number:
+            async for message in channel.history(limit=1000,
+                                                 before=tmp):
+                if len(to_delete) - 1 < number and check(message) and\
+                        (ctx.message.created_at - message.created_at).days < 14:
+                    to_delete.append(message)
+                elif len(to_delete) == number or (ctx.message.created_at - message.created_at).days >= 14:
```
calebj, line 112: Add another elif clause that doesn't set `too_old = True` and uses `len(to_delete) >= number` to be safe.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -1,345 +1,345 @@
...
+        tmp = ctx.message
+
+        done = False
+
+        while len(to_delete) - 1 < number and not done:
+            async for message in channel.history(limit=1000, before=tmp):
+                if len(to_delete) - 1 < number and \
+                        (ctx.message.created_at - message.created_at).days < 14:
+                    to_delete.append(message)
+                elif len(to_delete) == number or (ctx.message.created_at - message.created_at).days >= 14:
```
calebj, line 195: Same as above.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -1,345 +1,345 @@
...
+        to_delete = [ctx.message]
+        too_old = False
+        tmp = ctx.message
+
+        while not too_old and len(to_delete) - 1 < number:
+            async for message in channel.history(limit=1000, before=tmp):
+                if len(to_delete) - 1 < number and check(message) and\
+                                (ctx.message.created_at - message.created_at).days < 14:
+                    to_delete.append(message)
+                elif len(to_delete) == number or (ctx.message.created_at - message.created_at).days >= 14:
```
calebj, line 252: Ditto.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -1,345 +1,345 @@
...
+            to_delete.append(ctx.message)
+            number += 1
+        too_old = False
+        tmp = ctx.message
+        while not too_old and len(to_delete) < number:
+            async for message in channel.history(limit=1000, before=tmp):
+                if len(to_delete) < number and check(message) and\
+                        (ctx.message.created_at - message.created_at).days < 14:
+                    to_delete.append(message)
+                elif len(to_delete) == number or (ctx.message.created_at - message.created_at).days >= 14:
```
calebj, line 325: ...

calebj (changes requested):
Can you recreate the commit without changing the line ending scheme and doesn't patch the entire file, then rebase your patch from master so it doesn't include the merge commit to your own branch?

Once that's done, consider implementing the changes I commented with in the patch. Then, force push to your own branch to overwrite the old history with a clean commit.

renebulous:
Alright, I did a clean commit and avoided borking the line ending scheme.

Thanks for your help calebj.

renebulous:
Ok it should be up properly for real now.

calebj:
Yep, looks good... but I think you may have orphaned this PR by accident.

renebulous:
Yup... "to" and "from", "pull" and "push" can be quite ambiguous. Sometimes it gets confusing.

PR #1131, opened 2017-12-01
[Config] Retrieve/save values with async context manager

Tobotimus:
### Type

- Enhancement

Allows the use of an async context manager to get and set config values.

This works like so:
- Entering the context manager gets the value
- Exiting the context manager sets it.

This can be classed as both a stylistic enhancement as well as a functional one - the functional enhancement being inherent from the behaviour of context managers - if any exception occurs from within the context manager, or a the function returns, config will attempt to set the value as per any modifications which had been made.

Also important to note is that this will only work with mutable types such as `list` or `dict`. The context manager can only keep track of modifications to the value by referencing its address in memory, so keep that in mind. For this reason the context manager will raise a `TypeError` if a non-list or -dict is returned by the value getter.

### Examples

Appending a value to a list:
```py
async with self.config.list_of_foo() as list_of_foo:
    list_of_foo.append("foo")
```

The following will raise a `TypeError` upon entering the context manager, and not save any changes:
```py
async with self.config.my_int() as my_int:
    my_int += 1
```

The following code will *not* save the modifications to `list_of_foo` since its reference has been changed, and it was not modified in place:
```py
async with self.config.list_of_foo() as list_of_foo:
    list_of_foo = list_of_foo + ["foo"]
```

### DIFF
--- a/docs/framework_config.rst
+++ b/docs/framework_config.rst
@@ -29,7 +29,7 @@ Basic Usage
 
         @commands.command()
         async def return_some_data(self, ctx):
-            await ctx.send(config.foo())
+            await ctx.send(await config.foo())
 
 ********
 Tutorial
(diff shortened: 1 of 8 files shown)


Tobotimus:
I've provided documentation for this, and have also implemented it in existing modules/cogs where it could be used. I've tested those modules/cogs and they are working as they were before, the code is just a bit cleaner.

palmtree5 (approved):

PR #1133, opened 2017-12-02
[V3 JSON Driver] Don't save json if it exists

Tobotimus:
### Type

- Enhancement

### Issue
JSON files are currently being saved every time its driver is instantiated. When the JSON file already exists, this does nothing; it just loads the JSON and saves it again.

This PR makes sure the JSON is only saved if it does not already exist. Perhaps someone just missed an indent.


Kowlin (approved):

Kowlin (approved):
LGTM

PR #1135, opened 2017-12-02
[V3 Streams] Fix for #1132

### LINKED ISSUE #1132: [V3 Beta 3 Streams] 'Task' Object has no attribute 'append' when
### Type:

Bug

`!streamalert twitch channel <channel_name>` gives error `'Task' object has no attribute 'append'` 

### Expected behavior

Add a stream to the list of channels to be alerted for

### Actual behavior

Error...

### Steps to reproduce

1. Execute command `!streamalert twitch channel` with any (valid) stream name

### Observations

Line 44 of `cogs/streams/main.py`'s assignment to `self.streams` from `self.bot.loop.create_task` returns a `Task` object instead of a `list` or similar. (No other object assignments found).

VicenteRD:
### Type

Bugfix

Fixes #1132 : `self.streams` being a `Task` object instead of a list. 
Also replicated for `self.communities`, since the issue is the same 
(although untested, it can be inferred).


palmtree5:
Just checked, definitely does happen with communities too, so good call on fixing it for that too

cominixo (approved):

redbot/cogs/streams/main.py
```diff
@@ -41,8 +41,9 @@ def __init__(self, bot: Red):
 
         self.bot = bot
 
-        self.streams = self.bot.loop.create_task(self.load_streams())
-        self.communities = self.bot.loop.create_task(self.load_communities())
+        self.bot.loop.create_task(self.load_streams())
```
Tobotimus: `self.streams` and `self.communities` must be initialised as an iterable before `_stream_alerts()` is called, otherwise it will raise an `AttributeError` or `type is not iterable` error. This is possible in the current scenario because these three tasks are scheduled to run asynchronously, and not linearly.

You could do this just by initialising them as empty lists before scheduling their `load_X` functions, or otherwise.

Tobotimus (changes requested):
Other than this one change, everything looks good to me.

redbot/cogs/streams/main.py
VicenteRD: I thought about an empty pre-initialization too, but I didn't think it'd be needed...

Now that I look at `_stream_alerts`, you might want the 2 previous tasks to complete before that happens though (So the first run of that method is done with a complete list of streams and communities).
So the 2 possible solutions are empty pre-init, or waiting for the tasks. You tell me :)

Tobotimus: ~~I don't think waiting for tasks is a good idea here, as it will have blocking effects.~~ You could create another async function which awaits the first two tasks before creating the `self._stream_alerts()` task, so you can guarantee linearity. Then you only have to use `create_task` once in the `__init__`.

Edit: I think I misunderstood what you meant by "waiting for the tasks".

redbot/cogs/streams/main.py
```diff
@@ -41,9 +41,12 @@ def __init__(self, bot: Red):
...
         self.bot = bot
 
-        self.bot.loop.create_task(self.load_streams())
-        self.bot.loop.create_task(self.load_communities())
-        
+        self.bot.loop.create_task(self._initialize_lists())
+
+    async def _initialize_lists(self):
+        await self.bot.loop.create_task(self.load_streams())
+        await self.bot.loop.create_task(self.load_communities())
```
Tobotimus: This will fail since `create_task` isn't a coroutine function. You only need to await the coroutines themselves, i.e. `await self.load_streams()` and `await self.load_communities()`.

Tobotimus (changes requested):

redbot/cogs/streams/main.py
VicenteRD: How am I so stupid >.< Sorry, I'll fix tomorrow morning

Tobotimus (approved):
Tested and it works :+1:

tekulvw (approved):

tekulvw:
Please DM me on Discord to receive your "Contributor" role!

PR #1143, opened 2017-12-06
[V3 i18n] Internationalise help for commands and cogs

Tobotimus:
### Type

- Enhancement

This PR is an alternative to #1121.

### Breaking changes
 - `redbot.core.RedContext` has been renamed and relocated to `redbot.core.commands.Context`.
 - `redbot.core.i18n.CogI18n` has been renamed to `redbot.core.i18n.Translator`.

- The `redbot.core.context` module has been renamed to `redbot.core.commands` and made into a package which acts as a substitute for `discord.ext.commands`.
- The `Command` and `Group` classes from `discord.ext.commands` are subclassed in `redbot.core.commands` to change the behaviour of the `help` attribute. To instantiate commands/groups with these classes, the `command()` and `group()` decorators have also been provided in this module.
- The `CogI18n` class at `redbot.core.i18n` has been renamed to `Translator`, to avoid confusion with a new decorator, and because it doesn't just apply to cogs.
- A new class decorator is introduced, `cog_i18n()`, located at `redbot.core.i18n`. It takes one argument - the `Translator` object for that cog.

### Example Usage
```py

from redbot.core.i18n import cog_i18n, Translator

_ = Translator("MyCog", __file__)

@cog_i18n(_)
class MyCog:

    @commands.command()
    async def foo(self, ctx: commands.Context):
        """Translateable Help"""
       ...

    ...

```

~~There are some cases where both the `redbot.core.commands` and `discord.ext.commands` modules will want to be imported. For this, you should import at least one of the modules as an alias, e.g.~~ ***Update:*** `redbot.core.commands` now contains references to all non-modified `discord.ext.commands` attributes as well.
```py

...

class MyCog:

    @commands.command()
    @commands.guild_only()
    async def foo(self, ctx: commands.Context):
        """Translatable Help"""
       ...

    ...
```

All existing core cogs and modules have been updated to include this change.

I should also mention that `pygettext -D` currently has a bug when it comes to extracting docstrings, which makes internationalising help a bit harder. For the CogManagerUI I extracted the strings manually, however I'll look for a patch/alternative to pygettext in the near future to remedy this (***Update***: python/cpython#4745 has been merged and will be backported to the next release of 3.6 and 3.7)


Tobotimus:
I'm actually going to make a few changes to this PR. In retrospect, perhaps naming the `redbot.core.commands` wasn't the best idea, since `discord.ext.commands` would need to be imported as an alias as outlined in the PR description.

Tobotimus:
Some changes have been made here, and my decision to rename `redbot.core.commands` has somewhat changed; instead, I've made the module into a package, and it can now be imported entirely in place of `discord.ext.commands`.

Tobotimus:
Alright, so obviously this PR is a pretty massive change. Some people, naturally, won't like it due to the way it mocks up the discord.py commands extension. So, I'm going to lay out some reasons why it's a positive and necessary change, since that seems to be missing from this PR conversation.

### Abstract
Changing locale at runtime is not a problem for strings passed into their `_()` function, as long as said function is called every time that string is needed. However, due to the discord.py convention of having command and cog help being located in docstrings (a convention which should not be broken), translating these strings at runtime becomes a more complex issue.

In order to solve the problem with commands, we need to have a better understanding of how discord.py manages command help after the command is instantiated, and how the help formatter retrieves it. The default `discord.ext.commands.Command` object has just one relevant attribute here: `help`, which is simply assigned the value of the `help` kwarg if it exists, else it is assigned a *cleaned up* original async function's docstring. The help formatter then simply uses this `help` attribute in the embed which is sent to the user.

From here, I believe we have two options: translate the command help from the help formatter and leave the `Command` object as is, or subclass the `Command` object so that the translation occurs internally.

### Help formatter method
Doing translation within the help formatter would mean very minimal changes would need to be made to cogs. Perhaps an extra import and an extra decorator would be all it takes to make command help translatable at runtime. However, the help formatter itself would need to do some funky things to be able to do the translation. The original docstring of the callback would need to be accessed and the `help` attribute of the `Command` would be ignored. The help formatter would also need to find the `_()` function for this cog. This sort of operation, in my opinion, should be done internally by the `Command` object.. Not only this, but it complicates the writing of custom help formatters for Red. Also, should the help formatter have the responsibility of translation? It is a *formatter* after all.

### Subclassing method
This method is the one implemented in this PR. In most cases, updating a cog to command help translatable would be as simple as changing the line `from discord.ext import commands` to `from redbot.core import commands`, and adding a single decorator to the cog class. The `help` attribute has become a `@property`, and this is where the translation is done.

However I didn't just see this method as a solution, but as an opportunity. Structuring the imports and core `commands` package like this makes future subclassing and overriding of `discord.ext.commands` attributes more than trivial. The class can simply be added as an attribute to `redbot.core.commands`, and existing cogs needn't even be modified to be using the new object.

### Overriding `discord.ext.commands`
So, as this PR states, the new `redbot.core.commands` module acts as a complete substitute for importing `discord.ext.commands`. The `redbot.core.commands` module simply imports everything attached to `discord.ext.commands`, then overrides any imports which we have subclassed. I think this is what's really making people apprehensive about this PR, however it would be nice to get some more specific thoughts. :)

Kowlin (approved):

PR #1152, opened 2017-12-09
[V2] Remove playlist extend

Zephyrkul:
### Type

- [X] Enhancement

Since V2 is feature-frozen at this point may as well remove the unimplemented `playlist extend` and leave it for V3.


Tobotimus (approved):
Good point, no reason for this to be in here, even if there were/are plans to implement it down the line.

PR #1157, opened 2017-12-10
[Core] Deal with duplicate command names by renaming them

Twentysix26:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Instead of making the cog's loading fail, conflicting command names (and aliases) will be renamed to "command_name-1", "command_name-2" and such.  
Command name conflicts will inevitably go up as the number of cogs grows, this is the best way I could think of to fix the problem and it may very well be used in v3 too with due modifications.

### DIFF
--- a/red.py
+++ b/red.py
@@ -7,6 +7,7 @@ import logging.handlers
 import traceback
 import datetime
 import subprocess
+import importlib
 
 try:
     from discord.ext import commands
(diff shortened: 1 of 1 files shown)


laggron42:
So, as I see, it load the cog a second time and all of its commands will be named `command[i]`. Wouldn't renaming only the conflicting command be better (means that this is automatic)? I imagine something like
1. You load `MyCog` with command `twentysex`
2. You load `MyOtherCog` also with the command `twentysex`
3. A message appears to tell that there are conflicting commands and `twentysex` from `MyOtherCog` will be renamed `twentysex1`

DannyB5544:
It's probably easier. Besides, you could use aliases.

PR #1160, opened 2017-12-11
[V3 Core] Add blacklist/whitelist commands

tekulvw:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

The framework for using the blacklist and whitelist already existed...we just didn't have commands for manipulating it.


skeith:
A quick test has been done. One caveat that blacklisting yourself is accepted as a valid command while it should not.

Tobotimus (approved):

PR #1164, opened 2017-12-11
[V3 Utils] add embed utils with a random_colour function

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This adds a place for embed-related stuff in `redbot.core.utils`. At the moment, the only thing in the new file is a random color generator that would allow for the following:
```py
from redbot.core.utils.embed import random_colour 

class SomeCog:
    
    @commands.command()
    async def test(self, ctx):
        await ctx.send(embed=discord.Embed(title="Test", color=random_colour()))
```

This new function is also aliased to `random_color` (much like `discord.Color` being an alias for `discord.Colour`)

### DIFF
--- a/redbot/core/utils/embed.py
+++ b/redbot/core/utils/embed.py
@@ -19,11 +19,7 @@ def randomize_colour(embed: discord.Embed) -> discord.Embed:
         The embed with the color set to a random color
 
     """
-    embed.colour = discord.Color(
-        value=int(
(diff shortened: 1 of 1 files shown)


Tobotimus (commented):
Could you also include the module in docs/framework_utils?

tekulvw (approved):

tekulvw:
We'll get the utils added to the docs later

Issue #1095, opened 2017-11-13
[V3 Beta 2] [p]ignore doesn't ignore


aikaterna:
### Type:

- [ ] Suggestion
- [X] Bug

### Brief description of the problem
[p]ignore does not ignore channels or servers
customcom will also need to be changed when this is written 
https://github.com/Cog-Creators/Red-DiscordBot/blob/46e37d5ff58c9bbdff37b19c2ced346572e9255a/redbot/cogs/customcom/customcom.py#L336

### Expected behavior
-Bot ignores channel or server 
-Red v3 would have something like bot.user_allowed in v2 red.py

### Actual behavior
Bot does not ignore anyone.

### Steps to reproduce
1. Use [p]ignore server or [p]ignore channel
2. Alternate user tries interacting with the bot
3. Channel or server is not ignored, bot responds

Tobotimus:
I'm not getting the issues with `[p]ignore`, this is from my attempt to reproduce:
```
Tobotimus - Today at 6:11 PM
!ignore channel
TARS.DevBOT - Today at 6:11 PM
Channel added to ignore list.
Tobotimus - Today at 6:11 PM
!ping
TARS.DevBOT - Today at 6:11 PM
Pong.
ToboTest5 - Today at 6:11 PM
!ping
TARS.DevBOT - Today at 6:11 PM
:no_entry: You are not authorized to issue that command.
Tobotimus - Today at 6:11 PM
!unignore channel
TARS.DevBOT - Today at 6:11 PM
Channel removed from ignore list.
ToboTest5 - Today at 6:11 PM
!ping
TARS.DevBOT - Today at 6:11 PM
Pong.
```

I understand the bot responds with the "unauthorized" message when a command is sent in an ignored channel. This is simply what the bot responds with whenever a check fails on a command. Perhaps we should remove that and make the bot just not respond if a check fails.

tekulvw:
Fixed in #1166

PR #1166, opened 2017-12-13
[V3 Mod] fix some issues relating to ignore

palmtree5:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

First, this fixes `[ignore|unignore] guild` being unusable without the manage guild perm

Second, this fixes the actual check that happens with ignores so that it doesn't automatically permit if the admin role isn't set (because is_admin was returning True before if the admin role wasn't set, and the same was true with is_mod but with the mod role in play there)


tekulvw (approved):

tekulvw:
Fixes #1095

PR #1169, opened 2017-12-13
[V3 Sentry] Fix sentry logging

tekulvw:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

So when we updating to pip packaging we screwed up our namespacing for sentry filtering. This *should* fix that but may be a bit too permissive on the filter. We'll see.


Tobotimus (approved):
Indeed, we will see

PR #1171, opened 2017-12-13
[V3 Config] Reserved keys and some group data validation

### LINKED ISSUE #1167: [V3 Config] Restricted Keys
Please be sure to read through other issues as well to make sure what you are suggesting/reporting has not already
been suggested/reported

### Type:

- [ ] Suggestion
- [x] Bug

*So the fun's over, the wild west days of Config are quickly coming to an end.*

As it turns out, allowing you to use *anything* as a key for Config has issues, namely if you try to use a key that is an attribute of either the "Group" or "Value" classes and as such we will need to start throwing errors when keys of those names are used. If we don't throw errors, you still get errors, they're just a lot less helpful.

Tobotimus:
### Type

- Bugfix

Throws a `ValueError` when registering or setting values for reserved keys.

Fixes #1167


Tobotimus:
So, I've increased the scope of this PR a little bit whilst implementing detection of reserved keys when a `Group` is set directly.

I realised doing this is also a workaround for `force_registration`, and also allows nested groups to be set as something other than a dict. So, I've also added validation for these two situations.

Lastly, one of my tests was failing when checking for an error upon setting an un-registered key in a nested group. This is because the `Group.defaults` attribute is mutable since it only returns a shallow copy of defaults - this must be a deepcopy when there can be nested mutable objects. So, I changed it to a deepcopy.

redbot/core/config.py
```diff
@@ -321,11 +323,50 @@ def get_attr(self, item: str, default=None, resolve=True):
...
 
     async def set(self, value):
-        if not isinstance(value, dict):
-            raise ValueError(
-                "You may only set the value of a group to be a dict."
-            )
-        await super().set(value)
+        try:
+            self.validate_group_data(value)
+        except (TypeError, ValueError):
```
tekulvw: This try/except is doing nothing, no matter what if `validate_group_data` throws an exception it will be raised.

redbot/core/config.py
```diff
@@ -321,11 +323,50 @@ def get_attr(self, item: str, default=None, resolve=True):
...
+            If ``group_data`` is not a `dict`.
+        ValueError
+            If there are un-registered or reserved keys in ``group_data``.
+
+        """
+        if not isinstance(group_data, dict):
+            raise TypeError("You may only set the value of a group to be a dict.")
+        for item in group_data:
+            # Check for reserved key
+            if item in dir(self):
```
tekulvw: You should probably reference the `RESERVED_KEYS` constant here.

redbot/core/config.py
```diff
@@ -345,6 +386,8 @@ def get_attr(self, item: str, default=None, resolve=True):
             The raw data value to set the attribute as.
 
         """
+        if item in dir(self):
```
tekulvw: Same thing here.

tekulvw (dismissed):

redbot/core/config.py
Tobotimus: I tend to add these to make it clearer what exceptions a method might raise. However if you don't deem it necessary, I'm happy to remove it

tekulvw:
I honestly don't know if this is necessary anymore. Sure, you wouldn't be able to set data using the normal `.set` method, but you can using the `.set_raw` method recently introduced.

Tobotimus:
Closing since this was not really the solution we were looking for

PR #1173, opened 2017-12-14
[V3] Warning system

palmtree5:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This adds a warning system to the core bot (the cog is Warnings and is that name because of the built-in Warning)

- Allows for adding actions to be taken at various point levels
  - The action list is checked when issuing a warning for a user and the highest point action will be taken on a user (so if there's actions of 10, 20, and 30 points and a 40 point warning is issued, only the 30 point action will be taken)
- Allows for adding predefined reasons for a warning
  - Custom reasons can be used when warning a user if enabled

This is a work in progress


palmtree5:
This won't be ready for beta 5

Tobotimus (dismissed):
Just one simple change I'm gonna request is to rename the package and main module to `warnings`, just so it's consistent with the cog name, and I personally think it's more intuitive.

palmtree5:
This won't be ready for beta 6. I am still working on this, but it's not ready to be merged at this time

palmtree5:
Not ready for b7 either. I actually haven't done much with this in the past 2 weeks. Will try to work more on it for it to hopefully be ready for b8

palmtree5:
So at this point, this should be ready for some kind of testing. All of the needed commands to setup actions/reasons and warn/unwarn users are there as well as commands to list actions, reasons, and warnings.

The action creation allows for specifying any command (that's not restricted to the bot owner) the server owner (or someone with the administrator permission) wants to have run at a specific point value by having them enter the command in a specific format (namely, exactly as if you were trying to run the command but without the prefix and with any user/member args being replaced with {user}). This command will be run without regard to checks/cooldowns as it's assumed that since the server owner (or someone with the admin perm) is authorizing usage of the command for the specific usage of taking an automated action.

Thanks to @mikeshardmind for the help with getting the automated action stuff sorted out

tekulvw (approved):

PR #1184, opened 2017-12-17
Fixed Quote arrangement

sbhadr:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Showed bright red bar on Sublime Text 3, thought I'd take the time to fix it so no one gets confused that it's an actual "life-threatening" error. When the file is executed, no error is shown since it's perfectly legal. It's more of an aesthetic change that was annoying me. : )


Twentysix26:
Looks fine on my end. Consider using MagicPython for Sublime Text.

Issue #1118, opened 2017-11-23
[V3 Alias] using symbols for aliases


Aioxas:
### Type:

- [ ] Suggestion
- [X] Bug

### Brief description of the problem
in V2 you could use symbols such as ?, !, etc to make aliases, now it's not allowed. Any explanation for this change?
### Expected behavior
Alias adds an alias with a symbol.
### Actual behavior
Doesn't allow the symbol alias to be added.
### Steps to reproduce

1. Try to make an alias with a symbol such as [p]alias add ? profile na
2. Get denied
3. ????
4. Profit

Kowlin:
So far I cannot find out if this is intended or a bug, its related to the ``isidentifier()``

https://github.com/Cog-Creators/Red-DiscordBot/blob/064e9b6bd0b0152f162c555e01bcfa5947f04a36/redbot/cogs/alias/alias.py#L78

Tobotimus:
Fixed in eb10508.

PR #1201, opened 2017-12-19
[V3 Alias] Allow wider range of aliases

Aioxas:
Possible fix for #1118

Makes it so that the only non-valid names are the ones that can't be printed by something.

### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Possible fix for #1118


Aioxas:
Since there is no name that results in false, Tests break. Unless we don't want spaced aliases, then we can change it to "return not str.isspace()"

Tobotimus:
I don't think we want spaced aliases, in fact I don't think they'd work anyway, since the alias parser only checks the first word in the command.

I think a check for containing whitespace would be sufficient. In my opinion this should be done with regex, since it covers additional whitespace chars which don't fall under ascii.

Aioxas:
Just read about it and using isspace() was wrong since it only applies to strings consisting only of whitespace. 

Would `return not bool(re.search(r'/s', str)) ` since it checks for what you asked?

 Got to cast it to bool since it would return a non-boolean object otherwise. As well as invert it due to the nature of the regex.

Tobotimus:
@Aioxas Sounds good to me.

Although I'd keep the `isprintable` check in there too.

Aioxas:
Weirdly enough, attempting to create an alias that has " anywhere in the alias name results in non-execution of the command due to lack of or incorrect args.

Tobotimus:
Nice.

The `"` issue probably comes from the way quoted arguments are passed into commands. I'd be content if we simply didn't allow them to be in an alias name.

Aioxas:
The alias subcommand won't execute at all if there's a `"` on the name, so there's no worry right there

However, it will work if you have it on both sides of the alias name string, but it won't be part of the alias execution command. so `"meow"` will be saved as `meow` and show as `meow`

Tobotimus:
I see what you mean now. That's not a worry then. From a bit of testing with V2's alias, I've found that you can actually have a `"` in your alias by escaping it with a backslash, but that doesn't really matter, since it works as an alias after that.

Tobotimus (approved):

Aioxas:
Aight cool. Thanks for the help with the PR. Hopefully no further changes for this PR is required.

palmtree5 (approved):

PR #1209, opened 2017-12-23
[V3 Config] Add ability to reset individual values

### LINKED ISSUE #951: [V3 Config] Method to delete individual key:value pairs
I came across a situation when I was working on v3 cogs where I wanted to delete a single key:value pair from a user's settings, rather than clear all of the user's settings. I think it would be useful to have a method for this! The workaround I have been using is to set the value to `None`, and just deal with that situation in my code, but deleting the key entirely has the benefit of reverting back to the default setting.

Tobotimus:
### Type

- Enhancement

Adds a `clear()` method to `Value` objects which resets them to their default value.

Should resolve #951


Redjumpman:
Does this clear the data out or does it simply change the value in the real database to the default? If it is the latter, then perhaps changing the method name to something other than clear would provide more clarity.

tekulvw:
Replaced by #1344

PR #1214, opened 2017-12-26
[V3 Sentry] Allow toggling of Sentry logging

### LINKED ISSUE #1213: [V3] Add ability to toggle Sentry logging
### Type:

- Enhancement

We currently don't have a way to enable or disable Sentry Logging after instance creation. Some users are starting to modifying their Red installations, and we're receiving a lot of errors from them, so there needs to be a way for users to disable it.

Tobotimus:
### Type

- Enhancement

This allows sentry logging to be enabled or disabled via the `set` command. It involves a bit of a refactor to how sentry logging is managed, with the main interface now being part of the bot class itself.

Resolves #1213


Kowlin (approved):
So far its looking good, might want to hold off on merging until Will has a look at it.

PR #1232, opened 2018-01-02
[Audio] Song notifs / autodc, queue list improvements

aikaterna:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

Credit goes to Stevy for the queue and song changes.

- audioset notify - a toggle for the bot sending "Now Playing" messages on song change
- audioset notifychannel - a command to specify the channel for the bot to announce in
- Queue changes - add song duration, number of songs left in the queue, queue draws now playing duration for currently playing song, increase song details to 10 songs, make queue an embed
- Song changes - add "np" alias for Now Playing, song information now in embed with duration
- audioset timerdisconnect was added to facilitate audio add-on cogs. When this is toggled to False (True is the default, and previous audio behavior), the bot will not disconnect when other audio-based cogs are in use. I admit this is sort of a selfish part of the PR, I have a radio add-on cog that Paddo originally created that I've been adding to. It will be able to interact with the audio cog and play icecast/http streams, but the bot was disconnecting after 300 seconds because of this loop. There didn't seem to be a way to cancel the Task outside of the audio cog.

One possible improvment/change: embed color could use the bot's top role color instead of a randomly generated color.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -19,6 +19,7 @@ import time
 import inspect
 import subprocess
 import urllib.parse
+import datetime
 from enum import Enum
 
 __author__ = "tekulvw"
@@ -148,6 +149,10 @@ class Song:
         self.duration = kwargs.pop('duration', 60)
         self.start_time = kwargs.pop('start_time', None)
         self.end_time = kwargs.pop('end_time', None)
+        self.thumbnail = kwargs.pop('thumbnail', None)
+        self.view_count = kwargs.pop('view_count', None)
+        self.rating = kwargs.pop('average_rating', None)
+        self.song_start_time = None
 
 class QueuedSong:
     def __init__(self, url, channel):
@@ -310,8 +315,10 @@ class Audio:
         self.queue = {}  # add deque's, repeat
         self.downloaders = {}  # sid: object
         self.settings = dataIO.load_json("data/audio/settings.json")
+        self.settings_path = "data/audio/settings.json"
         self.server_specific_setting_keys = ["VOLUME", "VOTE_ENABLED",
-                                             "VOTE_THRESHOLD", "NOPPL_DISCONNECT"]
+                                             "VOTE_THRESHOLD", "NOPPL_DISCONNECT",
+                                             "NOTIFY", "NOTIFY_CHANNEL", "TIMER_DISCONNECT"]
         self.cache_path = "data/audio/cache"
         self.local_playlist_path = "data/audio/localtracks"
         self._old_game = False
@@ -892,6 +899,7 @@ class Audio:
             except FileNotFoundError:
                 raise
 
+        song.song_start_time = datetime.datetime.now()
         voice_client = await self._create_ffmpeg_player(server, song.id,
                                                         local=local,
                                                         start_time=song.start_time,
@@ -1177,6 +1185,34 @@ class Audio:
         await self.bot.say("Maximum length is now {} seconds.".format(length))
         self.save_settings()
 
+    @checks.mod_or_permissions(manage_messages=True)
+    @audioset.command(name="notifychannel", pass_context=True)
+    async def audioset_notifychannel(self, ctx, channel: discord.Channel):
+        """Sets the channel for the now playing announcement"""
+        server = ctx.message.server
+        if not server.me.permissions_in(channel).send_messages:
+            await self.bot.say("No permissions to speak in that channel.")
+            return
+        self.set_server_setting(server, "NOTIFY_CHANNEL", channel.id)
+        dataIO.save_json(self.settings_path, self.settings)
+        await self.bot.send_message(channel, "I will now announce new songs here.")
+
+    @audioset.command(name="notify", pass_context=True)
+    @checks.mod_or_permissions(manage_messages=True)
+    async def audioset_notify(self, ctx):
+        """Sends a notification to the channel when the song changes"""
+        server = ctx.message.server
+        settings = self.get_server_settings(server.id)
+        notify = settings.get("NOTIFY", True)
+        self.set_server_setting(server, "NOTIFY", not notify)
+        if self.get_server_settings(server)["NOTIFY_CHANNEL"] is None:
+            self.set_server_setting(server, "NOTIFY_CHANNEL", ctx.message.channel.id)
+            dataIO.save_json(self.settings_path, self.settings)
+        if not notify:
+            await self.bot.say("Now notifying when a new track plays.")
+        else:
+            await self.bot.say("No longer notifying when a new track plays.")
+        self.save_settings()
     @audioset.command(name="player")
     @checks.is_owner()
     async def audioset_player(self):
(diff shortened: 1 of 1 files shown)


TheWyn:
Been testing this before it went up as pr and haven't encountered any issues under heavy use. Come approve I know how much you love audio @Kowlin

Stylite:
Did some tests and it looks good. I hope your audio dreams come true @Kowlin

Bakersbakebread:
**Question**: How many testers does it take to change a light bulb?
**Answer**: None. Testers do not fix problems; they just find them.

Haven't found any. Light bulb works fine.

skeith:
I tested this way back then when v2 is still active (not feature locked) and works wonder. The only issue I could find back then is why aika haven't PRd it to v2 audio

Electromaster232:
seems good!

cogs/audio.py
```diff
@@ -2210,6 +2322,36 @@ def has_connect_perm(self, author, server):
                     message = escape(message, mass_mentions=True)
                     await self.bot.send_message(next_channel, message)
 
+    async def disp_now_playing(self, server, song, notifychan:int):
```
Twentysix26: `disp_now_playing` to `display_now_playing`
`notifychan` to `notify_channel`

clarity / naming convention > sparing a few characters

cogs/audio.py
```diff
@@ -2210,6 +2322,36 @@ def has_connect_perm(self, author, server):
...
+        channel = discord.utils.get(server.channels, id=notifychan)
+        if channel != None:
```
Twentysix26: I would short circuit (aka return early) this `if` and some of the following ones to reduce the nested indentation level. Also remember that with `True`, `False` and `None` `is` / `is not` must be used instead.

Twentysix26 (changes requested):

Twentysix26 (approved):

jorams:
I would like to suggest two small additions to this PR:

1. The queue command (without arguments) normally displays now playing information as well as what's in the queue. As soon as the queue is empty, it displays nothing. I think it would be more useful if it would still display now playing info. I did it like this:

    ```diff
    diff --git a/cogs/audio.py b/cogs/audio.py
    index 637150c1..da749e79 100644
    --- a/cogs/audio.py
    +++ b/cogs/audio.py
    @@ -1827,19 +1827,15 @@ class Audio:
             """Not a command, use `queue` with no args to call this."""
             server = ctx.message.server
             channel = ctx.message.channel
    -        if server.id not in self.queue:
    +        now_playing = self._get_queue_nowplaying(server)
    +        if server.id not in self.queue and now_playing is None:
                 await self.bot.say("Nothing playing on this server!")
                 return
    -        elif len(self.queue[server.id][QueueKey.QUEUE]) == 0:
    -            await self.bot.say("Nothing queued on this server.")
    -            return
     
             colour = ''.join([choice('0123456789ABCDEF') for x in range(6)])
             em = discord.Embed(description="", colour=int(colour, 16))
             msg = ""
     
    -        now_playing = self._get_queue_nowplaying(server)
    -
             if now_playing is not None:
                 msg += "\n***Currently playing:***\n{} ({})\n".format(now_playing.title,
                     str(datetime.timedelta(seconds=now_playing.duration)))
    @@ -1879,7 +1875,9 @@ class Audio:
                 except AttributeError:
                     song_info.append("**[{}]** {.webpage_url} ({})".format(num, song, str_duration))
     
    -        msg += "\n***Next up:***\n" + "\n".join(song_info)
    +        if song_info:
    +            msg += "\n***Next up:***\n" + "\n".join(song_info)
    +
             em.description = msg.replace('None', '-')
             more_songs = len(self.queue[server.id][QueueKey.QUEUE]) - 10
             if more_songs > 0:
    ```

2. The now playing information currently always displays :arrow_forward:, even if the music is paused. It would be slightly nicer for this to actually reflect whether or not it is playing. I added it like this, though I'm not too fond of the three-deep `if`:

    ```diff
    diff --git a/cogs/audio.py b/cogs/audio.py
    index 637150c1..77ef3dc1 100644
    --- a/cogs/audio.py
    +++ b/cogs/audio.py
    @@ -1843,7 +1843,7 @@ class Audio:
             if now_playing is not None:
                 msg += "\n***Currently playing:***\n{} ({})\n".format(now_playing.title,
                     str(datetime.timedelta(seconds=now_playing.duration)))
    -            msg += self._draw_play(now_playing) + "\n"  # draw play thing
    +            msg += self._draw_play(now_playing, server) + "\n"  # draw play thing
                 if now_playing.thumbnail is None:
                     now_playing.thumbnail = (self.bot.user.avatar_url).replace('webp', 'png')
                 em.set_thumbnail(url=now_playing.thumbnail)
    @@ -1886,7 +1886,7 @@ class Audio:
                 em.set_footer(text="And {} more songs...".format(more_songs))
             await self.bot.say(embed=em)
     
    -    def _draw_play(self, song):
    +    def _draw_play(self, song, server):
             song_start_time = song.song_start_time
             total_time = datetime.timedelta(seconds=song.duration)
             current_time = datetime.datetime.now()
    @@ -1896,7 +1896,15 @@ class Audio:
     
             bar_char = '━'
             seek_char = ':radio_button:'
    -        play_char = ':arrow_forward:'
    +        play_char = ':play_pause:'
    +
    +        if self.voice_connected(server):
    +            voice_client = self.voice_client(server)
    +            if hasattr(voice_client, 'audio_player'):
    +                if voice_client.audio_player.is_playing():
    +                    play_char = ':arrow_forward:'
    +                else:
    +                    play_char = ':pause_button:'
     
             msg = "\n" + play_char + " "
     
    @@ -2082,7 +2090,7 @@ class Audio:
                         "**Duration:** `{}`\n**Rating: **`{:.2f}`\n**Views:** `{}`".format(
                         song.creator, song.uploader, str(datetime.timedelta(seconds=song.duration)), song.rating,
                         song.view_count))
    -            msg += self._draw_play(song) + "\n"
    +            msg += self._draw_play(song, server) + "\n"
                 colour = ''.join([choice('0123456789ABCDEF') for x in range(6)])
                 em = discord.Embed(description="", colour=int(colour, 16))
                 if 'http' not in song.webpage_url:
    ```

cogs/audio.py
```diff
@@ -1795,21 +1858,56 @@ def _clean_url(self, url):
...
+        song_start_time = song.song_start_time
+        total_time = datetime.timedelta(seconds=song.duration)
+        current_time = datetime.datetime.now()
+        elapsed_time = current_time - song_start_time
+        sections = 12
+        loc_time = round((elapsed_time/total_time) * sections)  # 10 sections
+
+        bar_char = '━'
+        seek_char = ':radio_button:'
+        play_char = ':arrow_forward:'
```
Tobotimus: Personally I prefer all emoji characters to be represented in code with `\N{}` notation, like this:
```py
bar_char = '\N{BOX DRAWINGS HEAVY HORIZONTAL}'
seek_char = '\N{RADIO BUTTON}'
play_char = '\N{BLACK RIGHT-POINTING TRIANGLE}'
```

Using raw emoji characters in code can lead to problems for people using a font in their text editor which doesn't support the character. As for the colon notation, I've found it can lead to inconsistencies when used in various scenarios.

AmusingThrone:
Tested this out, and noticed that on rare cases the bot sends an empty embed (with just the color) when the `queue` command is called. I am unsure of why this happens, as line 1855 and 1866 return data and are not empty.

laggron42:
Tested it, absolutely no bug

Tobotimus (approved):

cogs/audio.py
```diff
@@ -1975,14 +2063,19 @@ def can_instaskip(self, member):
...
-            await self.bot.say(msg.replace("**Author:** None\n", "")
-                                  .replace("**Views:** None\n", "")
-                                  .replace("**Uploader:** None\n", "")
-                                  .replace("**Duration:** None\n", ""))
+
+            msg = ("**Author:** `{}`\n**Uploader:** `{}`\n"
+                    "**Duration:** `{}`\n**Rating: **`{:.2f}`\n**Views:** `{}`".format(
+                    song.creator, song.uploader, str(datetime.timedelta(seconds=song.duration)), song.rating,
+                    song.view_count))
+            msg += self._draw_play(song) + "\n"
```
AmusingThrone: This is missing the server argument. 

Should be:

`msg += self._draw_play(song, server) + "\n"`

AmusingThrone (changes requested):

AmusingThrone:
Whoops sorry for the bump, I never actually submitted this, and I accidentally submitted it now.

cogs/audio.py
aikaterna: Thanks, I was wondering what caused your issue as I couldn't reproduce it.

AmusingThrone: Hi, sorry for the late reply. I think I messaged you on Discord about the JSON unpacking, that is all I could find. I don't really use Red, but I can do some testing this weekend if you want.

Kowlin (approved):

PR #1233, opened 2018-01-04
[Mod] [p]filter: Performance improvements

Twentysix26:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This should drastically reduce the memory usage of the message internal cache, as it now stores only the ID of the author and the content of the messages instead of the objects. Furthermore, the cache is limited to 100k unique users to prevent it from increasing indefinitely.

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -1518,10 +1518,10 @@ class Mod:
                 return False
             if author.id not in self.cache:
                 self.cache[author.id] = deque(maxlen=3)
-            if len(self.cache) > 100000:
+            self.cache.move_to_end(author.id)
+            while len(self.cache) > 100000:
                 self.cache.popitem(last=False) # the oldest gets discarded
             self.cache[author.id].append(message.content)
-            self.cache.move_to_end(author.id)
             msgs = self.cache[author.id]
             if len(msgs) == 3 and msgs[0] == msgs[1] == msgs[2]:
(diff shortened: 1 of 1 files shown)


cogs/mod.py
```diff
@@ -1516,10 +1516,14 @@ def format_case_msg(self, case):
         if self.settings[server.id]["delete_repeats"]:
             if not message.content:
                 return False
-            self.cache[author].append(message)
-            msgs = self.cache[author]
-            if len(msgs) == 3 and \
-                    msgs[0].content == msgs[1].content == msgs[2].content:
+            if author.id not in self.cache:
+                self.cache[author.id] = deque(maxlen=3)
+            if len(self.cache) > 100000:
```
Tobotimus: I suppose this could be a `while` instead of `if`, in case the length of the cache somehow becomes greater than 100001, however I don't see that happening unless it's changed outside the cog, so it's not that big of a deal

Tobotimus (approved):
Worked in my testing, and overall looks good

cogs/mod.py
Twentysix26: Yeah I had the same train of thought. I'll swap it with `while` anyway and move the `move_to_end` before the deletion to be extra-safe.

Kowlin (approved):
Looking good

PR #1234, opened 2018-01-06
[V3 Bank] Fix can_spend and withdraw_credits

Redjumpman:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Changed the equality operator for can_spend ~~and withdraw_credits~~ function from > to >=. In the current release you cannot use an amount that equals the member's balance.

**Steps to reproduce:**
set your balance to 100

make a bet in slots for 100.
[p]slot 100

EDIT: Withdraw_credits is fine.

### DIFF
--- a/redbot/core/bank.py
+++ b/redbot/core/bank.py
@@ -143,7 +143,7 @@ async def can_spend(member: discord.Member, amount: int) -> bool:
     """
     if _invalid_amount(amount):
         return False
-    return await get_balance(member) > amount
+    return await get_balance(member) >= amount
 
 
 async def set_balance(member: discord.Member, amount: int) -> int:
--- a/redbot/core/bank.py
+++ b/redbot/core/bank.py
@@ -215,7 +215,7 @@ async def withdraw_credits(member: discord.Member, amount: int) -> int:
         raise ValueError("Invalid withdrawal amount {} <= 0".format(amount))
 
     bal = await get_balance(member)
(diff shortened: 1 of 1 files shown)


redbot/core/bank.py
```diff
@@ -215,7 +215,7 @@ def _invalid_amount(amount: int) -> bool:
         raise ValueError("Invalid withdrawal amount {} <= 0".format(amount))
 
     bal = await get_balance(member)
-    if amount > bal:
+    if amount >= bal:
```
Tobotimus: This one was right originally, `withdraw_credits` shouldn't raise if the amount being withdrawn is equal to the balance.

Tobotimus (dismissed):

redbot/core/bank.py
Redjumpman: I'm confused as to what you are talking about. withdraw_credits will raise if the amount is equal to the balance.
Both can_spend and withdraw_credits are using greater than instead of greater than or equal to. I can write a mock cog for you to test both functions if you like. I'm unsure if you are in agreement or not lol.

Redjumpman: Nvm I was looking at that particular spot wrong. Sorry about that.

palmtree5 (approved):
Looks good to me

Kowlin (approved):
Looks fine.

PR #1238, opened 2018-01-07
[V3 Streams/Core] Remove aiohttp.ClientSession assignment

### LINKED ISSUE #1195: [V3 Streams/Core] Use of aiohttp.ClientSession assignment
### Type:

- Bug

In various places throughout the bot (namely in streams and core_commands), `aiohttp.ClientSession()` is used like this:
```py
session = aiohttp.ClientSession()
async with session.request(url) as r:
    # Do stuff with response
await session.close()
```

The session should be used with an async context manager, so that it closes successfully even if an error is raised, or the function returns, whilst it is being used:
```py
async with aiohttp.ClientSession() as session:
    async with session.request(url) as r:
        # Do stuff with response
```

Tobotimus:
### Type

- Bugfix

This reduces the risk of `aiohttp.ClientSession` objects being left unclosed due to exceptions being raised.

Personally I think less `aiohttp.ClientSession` objects should be used overall, as it leverages the use of connection pooling, reusage and keep-alives, [as mentioned in aiohttp's documentation](https://aiohttp.readthedocs.io/en/stable/client.html#make-a-request). However, one step at a time.

Resolves #1195.

### DIFF
--- a/redbot/cogs/streams/streamtypes.py
+++ b/redbot/cogs/streams/streamtypes.py
@@ -25,7 +25,6 @@ class TwitchCommunity:
         self.type = self.__class__.__name__
 
     async def get_community_id(self):
-        session = aiohttp.ClientSession()
         headers = {
             "Accept": "application/vnd.twitchtv.v5+json",
             "Client-ID": str(self._token)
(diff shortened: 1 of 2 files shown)


palmtree5:
There's actually also an instance of this in https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/image/image.py#L24 though in that case, it's actually one session for the whole cog

However, I agree with the idea of reducing the number of `ClientSession` objects, but I'm not quite sure as to how atm.

palmtree5 (approved):
For now, this is good. I think we should still take a look at some way of reducing the number of `ClientSession` objects in the future, but we can discuss that later

PR #1241, opened 2018-01-08
[Trivia] Update general.txt list

Mathster0598:
Added 36 new questions

### Type

-  Enhancement

Added more questions to the general category

### DIFF
--- a/data/trivia/general.txt
+++ b/data/trivia/general.txt
@@ -700,7 +700,7 @@ Who wrote the Vampire Chronicles, which include the novels Armand, Blood and Gol
 How tall would a double elephant folio book be in inches?`50`50 inches
 Who wrote the contemporary children's books about mermaids set on the coast of Cornwall?`Helen Dunmore
 How old is the world's oldest dictionary?`2300 BC
-How many times was the Men's Tennis Singles at Wimbledon won by Bjorn Borg?`Five
+How many times was the Men's Tennis Singles at Wimbledon won by Bjorn Borg?`Five`5
 In 2011, which country hosted a Formula 1 race for the first time?`India
 Name the game played on a lawn called a 'crown green'.`Bowls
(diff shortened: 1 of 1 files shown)


MiniJennJenn:
Definitely needs more quality control on the questions. 
Please consider either making the questions more specific so there is only one correct answer or having the questions accept all potentially correct answers (with this, make sure the first answer includes all acceptable answers for purposes of presentation when people get it wrong).

Examples: 

- Who invented the telephone?`Bell (Should also accept Alexander Graham Bell)
- What temperature does water boil at?`100C  (Should either specify in Celcius or should also accept 212F. If it specifies Celcius, 100 should also be a correct answer) 
- Where would you find the Sea of Tranquility?`The Moon (Also, Moon)
- What is another word for lexicon?`Dictionary (Too vague, vocabulary is also another word for lexicon)

MiniJennJenn:
This PR should be closed soon as well. 

You took my comments too literally. I'm asking you to use some critical thinking and quality control to look over ALL the questions. Not only taking my examples and adopting them as the new answers. 

Examples:

- "What temperature does water boil at?`100C`100`212F"
Rewriting the question to specify Celsius instead of just also taking 212F. What if the person answered 212. Would that have been wrong? No. Bad question.
- "Who was the legendary Benedictine monk who invented champagne?`Dom Perignon" 
This is false. A quick Google search will tell you this. 
- "Name the largest freshwater lake in the world?`Lake Superior"
This is only true by area. It is otherwise the 3rd largest by volume. 
- "What is another word for lexicon?`Dictionary`Vocabulary`Terminology"
I feel like this question just got worse. Honestly, just toss it. There's just too many synonyms. 

Please. Employ critical thinking and revise this. Otherwise, it's not really worth the headache. It's costing me way to much time to fact check let alone revise.

Edit: Fun fact, pimento is a type of cheese. Pimenta is another name for allspice.

PR #1247, opened 2018-01-12
[V3 Config] Fixed raised Attribute Error Message  

Redjumpman:
Added space in raised Attribute Error message

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Added a space in raised Attribute Error message

**Before**
foo is not a valid registered Groupor Value.

**After**
foo is not a valid registered Group or Value.


Tobotimus (approved):
Thank you for this

PR #1256, opened 2018-01-15
[V3] Mentionable prefix

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Adds the ability to use mentioning the bot as a prefix.

Can be activated through CLI flags, is not yet a persistent setting, can't be toggled during runtime or per-server.


Tobotimus (approved):
Looks good to me. I'd like to see this as a persistent/toggleable setting (perhaps per-server as well) at a later date.

Kowlin (approved):
Looks good, but the best case would be a toggle command in ``set``

PR #1259, opened 2018-01-16
[V3/Economy] Fix for commands that require guild

Redjumpman:
Added guild_only_check decorator to _set, transfer, and balance

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Added the guild_only_check decorator to balance, transfer, and _set. All of these functions require ctx.guild, and therefore **without** the decorator raises a Runtime error for the user. Adding the decorator to these functions sends the `You are not authorized to issue that command` when used without a guild (only when bank is not global. Global still functions as intended). 

----------
**Steps to Reproduce**
Make sure bank is not global
Start a DM with the bot
type `[p]bank balance`

**Solution**
Added decorator now sends the "You are not authorized to issue that command." 
message.

### DIFF
--- a/redbot/cogs/economy/economy.py
+++ b/redbot/cogs/economy/economy.py
@@ -144,6 +144,7 @@ class Economy:
             await ctx.send_help()
 
     @_bank.command()
+    @guild_only_check()
     async def balance(self, ctx: commands.Context, user: discord.Member = None):
         """Shows balance of user.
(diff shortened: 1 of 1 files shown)


Redjumpman:
Closing and resubmitting because of too many conflicts to deal with. Will resubmit at a later time.

Issue #1278, opened 2018-01-29
[V3] Custom error handling support.


Redjumpman:
Please be sure to read through other issues as well to make sure what you are suggesting/reporting has not already
been suggested/reported

### Type:

- [x] Suggestion
- [ ] Bug

Adding this here for visibility based on the discussion in discord. 

Add a check for complex cogs using the special methods added in the [discord.py rewrite](https://discordpy.readthedocs.io/en/rewrite/migrating.html#cog-changes) via `__errror()`. For example, this will allow a cog like Audio to have an error message tailored to the problem. This could be integrated with the new issue templates proposed in #1275 . In addition, this could also allow cog-creators to develop their own custom error messages, such as directing people to their channel in the Cog Support server or their own GH issues.

PR #1279, opened 2018-01-29
[V3 Events] Make the vague error message a fallback

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

modify CommandInvokeError handler in `on_command_error` to use the vague message as a fallback if a cog doesn't provide its own `__error` function
See #1278


Kowlin (approved):
Looks good.

PR #1305, opened 2018-02-13
[V3 Core] Add timeout exceptions to [p]servers

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Add timeout exceptions to self.bot.wait_for("message") in the servers command, or it will throw an asyncio.TimeoutError on both waiting for a server number from the user, or waiting for the confirmation message on server leave.


palmtree5 (approved):
Good catch

PR #1308, opened 2018-02-13
[V3 Config] Fix async context manager bug and add test

### LINKED ISSUE #1294: [V3 Config] config items defaulted with a list all share the same list until reloaded from file
It appears that name changes are logged ans saved to every user in the same guild. Error somewhere in ``on_member_update`` logic.

tekulvw:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes the issue where the default mutable value would be shared between all entries in that category.

Fixes #1294

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -171,7 +171,7 @@ class Group(Value):
 
     @property
     def defaults(self):
-        return self._defaults.copy()
+        return deepcopy(self._defaults)
 
     # noinspection PyTypeChecker
     def __getattr__(self, item: str) -> Union["Group", Value]:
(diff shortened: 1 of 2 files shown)


Redjumpman:
These changes fix the bug observed in #1294 for me.

palmtree5 (approved):
Tested, confirmed I'm not seeing the issue with this change

mikeshardmind:
No longer experiencing problematic behaviors from issue #1294 with these changes.

Kowlin (approved):
Seems to be fixed.

PR #1315, opened 2018-02-18
[V3 Config] Driver code initial cleanup

tekulvw:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Based on the feedback we received through our survey, I'm beginning to pursue a better version of config that allow easier storage and access of complex data that doesn't necessarily "fit" into one of the existing config categories. This cleanup of driver handling within config is the first step in that process. Documentation, examples, and further building out driver code will help make the transition from V2 to V3 easier for all parties involved.

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -419,7 +419,7 @@ class Config:
         self.cog_name = cog_name
         self.unique_identifier = unique_identifier
 
-        self.driver = driver
+        self._driver = driver
         self.force_registration = force_registration
         self._defaults = defaults or {}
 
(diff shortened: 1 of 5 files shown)


Redjumpman:
This is working for me. If anyone else wants to try this out, you can use the following code:

```Python
data_structure = {"Foo": {"Bar": {"Baz": None}}}
key1, key2 = "Bar", "Baz"

mem_group = self.config.member(member)
path = ("B", key1, key2)
await self.config.driver.set(*mem_group.identifiers, *path, value=3)
```
Which is the same thing as:
```Python
data_structure = {"Foo": {"Bar": {"Baz": None}}}

my_key1, my_key2 = "Bar", "Baz"
await self.config.member(member).Foo.get_attr(my_key1, resolve=False).set_attr(my_key2, "Test")

tekulvw:
Even with these changes the above code still isn't as clear or clean as we'd like it to be. I am working on adding methods to config to make this far simpler.

palmtree5 (approved):

PR #1344, opened 2018-02-24
[V3 Config] Implement a true clear method

### LINKED ISSUE #951: [V3 Config] Method to delete individual key:value pairs

tekulvw:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Clears data from config.

Supersedes #1209, fixes #951


Redjumpman:
This is working for me. I am able to clear a value such that it is reverted to the default value, and does not interfere with other data in the group.

palmtree5 (approved):

Issue #1332, opened 2018-02-21
[V3] Cleanup returns trace


TheWyn:
Using Red 3.0.0b8 cleanup returns trace:

```
Exception in command 'cleanup messages'
Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/discord/ext/commands/core.py", line 62, in wrapped
    ret = yield from coro(*args, **kwargs)
  File "/usr/lib/python3.6/site-packages/redbot/cogs/cleanup/cleanup.py", line 233, in messages
    async for message in channel.history(limit=1000, before=tmp):
UnboundLocalError: local variable 'tmp' referenced before assignment

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/lib/python3.6/site-packages/discord/ext/commands/bot.py", line 886, in invoke
    yield from ctx.command.invoke(ctx)
  File "/usr/lib/python3.6/site-packages/discord/ext/commands/core.py", line 904, in invoke
    yield from ctx.invoked_subcommand.invoke(ctx)
  File "/usr/lib/python3.6/site-packages/discord/ext/commands/core.py", line 498, in invoke
    yield from injected(*ctx.args, **ctx.kwargs)
  File "/usr/lib/python3.6/site-packages/discord/ext/commands/core.py", line 71, in wrapped
    raise CommandInvokeError(e) from e
discord.ext.commands.errors.CommandInvokeError: Command raised an exception: UnboundLocalError: local variable 'tmp' referenced before assignment
```

Command used to cause trace: `[p]cleanup messages 1`

tekulvw:
This was fixed in #1334

PR #1347, opened 2018-02-25
[V3 Cleanup] code consolidation

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This consolidates some stuff in `Cleanup`

The primary things here are:
- The code that asks the user if they still want to continue when attempting to delete more than 100 messages is now in its own function that returns True if they want to continue and False if not
- The code to get the list of messages to be deleted is now in its own function, which takes the command's context, the channel and the number of messages to get for deletion as required parameters, and a check (defaulting to `lambda x: True`, which is good for functions where we don't care about checking something about the message) as an optional parameter used by the function itself. The other params (limit, before, and after) are all passed to `channel.history`, and all default to the documented default values for those params.

These functions could also probably be used by third-party cogs, so if we'd rather move these to `utils` somewhere, we could

This will need testing though to ensure that all `cleanup` commands still function as intended after these changes


Redjumpman:
This is working for me, and fixes in the issue in #1332 . I noticed when deleting more than 100 messages, the message for "Continuing" is deleted before it can be read. May need asynchronous wait if you want it read, or just remove it altogether. 

Some other QoL things I noticed:

- Period missing in docstring for `after` command.
- Period missing in the docstring for the `cleanup_bot` command
- Periods are missing in the reasons.

tekulvw (approved):

PR #1351, opened 2018-02-26
[V3] Add listlocales command

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Adds a `listlocales` command to `Core`. Also remove a hanging two-letter translation file (`redbot/cogs/downloader/locales/de.po`) that was missed when updating translations earlier today (note that there's actually no loss of translations here, it's just an old file that was around before the switch to using the current format (example: `de-DE.po`)


tekulvw:
I sorted the locales list and added a bit of documentation on the `set locale` command

tekulvw (approved):

PR #1365, opened 2018-02-28
[V3]Encoding issue fix

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

specify opens as happening with utf-8 encoding

### DIFF
--- a/redbot/cogs/mod/mod.py
+++ b/redbot/cogs/mod/mod.py
@@ -322,7 +322,7 @@ class Mod:
         except discord.errors.Forbidden:
             await ctx.send(_("I'm not allowed to do that."))
         except Exception as e:
-            print(str(e).encode('utf-8', errors='ignore'))
+            print(str(e, encoding='utf-8', errors='replace'))
         else:
             await ctx.send(_("Done. That felt good."))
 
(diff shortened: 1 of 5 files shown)


tekulvw:
You're also gonna want to do this for the json config driver.

mikeshardmind:
json driver already had the encoding specified. (There were other `open()` statements that already had a specified encoding too)

Tobotimus (dismissed):

redbot/cogs/mod/mod.py
```diff
@@ -322,7 +322,7 @@ def __unload(self):
         except discord.errors.Forbidden:
             await ctx.send(_("I'm not allowed to do that."))
         except Exception as e:
-            print(e)
+            print(str(e).encode('utf-8', errors='ignore'))
```
Tobotimus: I don't think this does what you think it does. You're going to be printing a `bytes` object. There must be a better way to solve #1366

Tobotimus (dismissed):
All of the `.encode` stuff needs to be removed

redbot/cogs/mod/mod.py
mikeshardmind: Forgot to attatch decodes as well, but the alternative solution is just to set environment vars in the launcher (`PYTHONIOENCODING='UTF8'`) either that or ignore it because this is fixed in python3.6 https://www.python.org/dev/peps/pep-0528/

mikeshardmind: Latest commit removes the `.encode` and uses the encoding param in `str()` (rather than encode + decode)

tekulvw (approved):

tekulvw:
@Tobotimus thoughts on recent changes?

PR #1375, opened 2018-03-03
[V3] Fix help_formatter to obey embed length limits

Sitryk:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Help didn't account for docstrings passing embed field length limits which I noticed a while ago and as noticed again when Palm forgot a dual newline in a command docstring.

This PR sees to fix this by enforcing length limits on description, field names and field values


Tobotimus (approved):
Could be a more elegant solution but this works for now

PR #1380, opened 2018-03-03
Add os import in dev commands

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

When using dev commands, this is a bit annoying to import every time os, as this is a module needed very often (especially in `[p]debug` where you need to do `__import__('os')` instead of just `os` if already imported).
That module was already default imported on v2 and that was much easier when debugging.

### DIFF
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -3,6 +3,7 @@ import inspect
 import io
 import textwrap
 import traceback
+import os
 from contextlib import redirect_stdout
 from copy import copy
 


palmtree5:
`That module was already default imported on v2`

It was imported in V2's owner cog because it was needed for things, not just to be there to make things easier for `[p]debug`. Also if you're trying to do something with dev commands in V3 where you're needing to use `__import__`, you could probably use `[p]eval` which accepts code blocks

As an example:

[p]eval
```py
import os

await ctx.send(os.getcwd())
```

tekulvw:
Agreed, use eval or repl

PR #1383, opened 2018-03-04
Downloading Lavalink Jar

TheWyn:
Tell FredBoat's build server we're guests and to download jar.


tekulvw:
Changes made in #1384

Issue #1371, opened 2018-03-01
Economy output overhaul


Stradinius:
### Type:

- [X] Suggestion
- [ ] Bug

This suggestion specifically refers to the payday command. I'd like to have payday show the following lines all within one message:

Here, take some credits. Enjoy! (+ {(Credits gained)} credits)
You currently have {(Current balance)} credits.
You're currently #{(Position on the leaderboard, only if the user is actually on the leaderboard)} on the leaderboard!  Congrats!

I imagine the one that would present the most issue is the leaderboard position, though I'm not sure.

Having the user use payday and then making them use [p]bank balance after the fact to see how many credits they have feels like it could get annoying.

Also, a question.  If I were to make these changes locally, considering they're part of a core cog, would an update to the economy cog on the developer's side overwrite my changes?  And if so, is there any way to prevent this?

(Moved from Red-Docs, because I was stupid and didn't realize I was posting it there.  When I brought this up in #coding, I was told to make an issue here.)

tekulvw:
Suggestions added in #1386

PR #1386, opened 2018-03-04
[V3 Economy] Expand payday output

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

See #1371 for the original suggestion. This implements display of the user's current balance and (if applicable) leaderboard position when getting a payday.

### DIFF
--- a/redbot/cogs/economy/economy.py
+++ b/redbot/cogs/economy/economy.py
@@ -293,11 +293,9 @@ class Economy:
             pos = pos
             poswidth = 2
             name = acc[1]["name"]
-            namewidth = 25
-            if len(name) > namewidth:
-                name = name[:namewidth - 6] + "..."
+            namewidth = 35 - len(name)
             balance = acc[1]["balance"]
-            balwidth = len(str(balance)) + 5
+            balwidth = 2
             highscore += "{pos: <{poswidth}} {name: <{namewidth}s} {balance: >{balwidth}}\n".format(
                 pos=pos, poswidth=poswidth, name=name, namewidth=namewidth,
                 balance=balance, balwidth=balwidth
--- a/redbot/cogs/economy/economy.py
+++ b/redbot/cogs/economy/economy.py
@@ -257,22 +257,14 @@ class Economy:
                         str((await bank.get_balance(author))), credits_name
                     )
                 )
-                if await bank.is_global():
-                    bank_sorted = sorted(await bank.get_global_accounts(),
-                                         key=lambda x: x.balance, reverse=True)
-                else:
-                    bank_sorted = sorted(await bank.get_guild_accounts(guild),
-                                         key=lambda x: x.balance, reverse=True)
-                topten = bank_sorted[:10]
-                pos = 1
-                for acc in topten:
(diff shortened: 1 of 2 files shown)


redbot/cogs/economy/economy.py
```diff
@@ -252,6 +252,27 @@ def __init__(self, bot: Red):
                         credits_name
                     )
                 )
+                await ctx.send(
+                    _("You currently have {} {}.".format(
```
tekulvw: The i18n object requires  `_("...").format()`

tekulvw (dismissed):

redbot/cogs/economy/economy.py
palmtree5: whoops, I'll fix that now

Stradinius:
Thanks for committing this, palmtree!

redbot/cogs/economy/economy.py
```diff
@@ -252,6 +252,27 @@ def __init__(self, bot: Red):
                         credits_name
                     )
                 )
+                await ctx.send(
+                    _("You currently have {} {}.").format(
+                        str((await bank.get_balance(author))), credits_name
+                    )
+                )
+                if await bank.is_global():
```
tekulvw: You're doing a `bank.is_global()` check inside of a `bank.is_global()` check here.

redbot/cogs/economy/economy.py
```diff
@@ -252,6 +252,27 @@ def __init__(self, bot: Red):
...
+                    bank_sorted = sorted(await bank.get_global_accounts(),
+                                         key=lambda x: x.balance, reverse=True)
+                else:
+                    bank_sorted = sorted(await bank.get_guild_accounts(guild),
+                                         key=lambda x: x.balance, reverse=True)
+                topten = bank_sorted[:10]
+                pos = 1
+                for acc in topten:
```
tekulvw: Might be easier to do something like:

```py
names = [acc.name for acc in topten]
try:
    pos = names.index(author.display_name)
except ValueError:
    pass
else:
    await ctx.send(...)
```

tekulvw (dismissed):

redbot/cogs/economy/economy.py
palmtree5: good catch lol

palmtree5:
So the leaderboard changes from #1435 have been implemented, but with some stuff written for this PR.

We now have two functions relating to a bank leaderboard

`get_leaderboard` which gets the leaderboard
and
`get_leaderboard_position` which gets the user's position on the leaderboard

palmtree5:
This is tested and working fine as well

Stradinius:
Not sure if this is already a topic covered in V3 or not, but when a user bets with [p] slot, and wins, it does not take the initial bet before paying out.  So a 4* win actually becomes a 5* win.  Since we're already covering the economy cog here, I figured this would be the best place to note that.

tekulvw:
We've seen that discussion in #876 and decided to keep the existing functionality.

tekulvw (approved):

Issue #1367, opened 2018-02-28
[Core] Update checking


Kowlin:
Since we're on pypi. We can check if the bot has a new version by checking the current version against pypi. I think we can achieve this with the ``pip`` module it self even.

tekulvw:
Added with #1388

PR #1388, opened 2018-03-04
[V3 Core] add update check

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

#1367 

This adds in an update check that is done:
- on startup (if out of date, this is stated when the startup screen is printed)
- in `[p]info` (via a field in the embed)

ATM this doesn't DM the bot owner, though that could certainly be added in if desired


Tobotimus (approved):
Could perhaps make the DM to the owner an opt-in in the future, but for now I like this

PR #1392, opened 2018-03-05
Revert "[V3]Encoding issue fix"

mikeshardmind:
Reverts Cog-Creators/Red-DiscordBot#1365

Changes don't actually solve the issue, and testing environment on my part was not equivalent to the one experiencing the issue.


mikeshardmind:
This reverts the encoding changes in print which did not solve anything, but keeps the commit which fixed something (guaranteeing `open()` as utf-8)

TheWyn:
#1365 broke #1248 
capture

Tobotimus (approved):

PR #1417, opened 2018-03-10
[Core] Don't show 'cog unloaded' message for non existing cogs

kat-kta:
If the cog doesn't exist, the bot still replies with "The cog has been unloaded." The following change fixes this.

### DIFF
--- a/cogs/owner.py
+++ b/cogs/owner.py
@@ -96,17 +96,17 @@ class Owner:
                                " this isn't supposed to happen.")
         else:
             set_cog(module, False)
-        try:  # No matter what we should try to unload it
-            self._unload_cog(module)
-        except OwnerUnloadWithoutReloadError:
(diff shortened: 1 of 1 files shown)


tekulvw:
This functionality is intended.

Consider that a user has deleted a cog file while that cog was loaded. With your proposed changes, they'd be unable to unload that cog.

kat-kta:
Ah nevermind then. Forgot about that. Still weird when it tells you it unloaded your non-existing cog tho.

Twentysix26:
This is because `Bot.unload_extension` just returns None if the "cog" does not exist and the command itself doesn't check anything before invoking it. It could be fixed by adding more checks (`Bot.extensions` ?)

tekulvw:
Decided to keep existing functionality, thanks for the PR.

Issue #1413, opened 2018-03-08
[V3] Can't export backup on redbot-setup


aikaterna:
### Type:

- [ ] Suggestion
- [x] Bug

### Brief description of the problem
While using `redbot-setup --delete` to remove an instance of Red, making a backup of it via the prompts provided will error out (Windows)

### Expected behavior
A compressed backup file in the user directory listed.

### Actual behavior
```
Would you like to make a backup of the data for this instance (y/n)?
> y
Backing up the instance's data...
Traceback (most recent call last):
  File "C:\Program Files (x86)\Python36-32\Scripts\redbot-setup-script.py", line 11, in <module>
    load_entry_point('Red-DiscordBot==3.0.0b9', 'console_scripts', 'redbot-setup')()
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\redbot\setup.py", line 218, in main
    remove_instance()
  File "C:\Program Files (x86)\Python36-32\lib\site-packages\redbot\setup.py", line 185, in remove_instance
    with tarfile.open(str(backup_file), "w:gz") as tar:
  File "C:\Program Files (x86)\Python36-32\lib\tarfile.py", line 1586, in open
    return func(name, filemode, fileobj, **kwargs)
  File "C:\Program Files (x86)\Python36-32\lib\tarfile.py", line 1633, in gzopen
    fileobj = gzip.GzipFile(name, mode + "b", compresslevel, fileobj)
  File "C:\Program Files (x86)\Python36-32\lib\gzip.py", line 163, in __init__
    fileobj = self.myfileobj = builtins.open(filename, mode or 'rb')
OSError: [Errno 22] Invalid argument: 'C:\\Users\\Win7\\redv3-Sass9000-2018-02-27 19:14:19.tar.gz'
```

### Steps to reproduce

1. Have an instance of Red to delete.
2. redbot-setup --delete
3. Provide instance name
4. When asked if you want a backup, respond y

palmtree5:
yeah, let me just choose a different character to separate the components of the time because apparently Windows can't deal with : as a character in file names

tekulvw:
Fixed via #1419

PR #1419, opened 2018-03-10
[V3 Instance setup] Change separator for time in backups

palmtree5:
This moves to using - to separate time components instead of : because Windows doesn't allow : to be used in filenames


aikaterna:
Works great, fixes #1413

tekulvw (approved):

tekulvw (approved):

PR #1421, opened 2018-03-11
[V3 Instance Setup] start work on storage swapping

### LINKED ISSUE #1416: [V3 Data management] We need a way to convert data from JSON to mongoDB, and revert
### Type:

- [x] Suggestion
- [ ] Bug

Right now, we can't switch from JSON to mongoDB (or in the other way), if we do, we lose all data. Creating a data convertor wouldn't be really hard but that'll be really helpful (especially with those glitched JSONs #1415 )

palmtree5:
This will fix #1416 when done. At the moment, this supports a one-way conversion from JSON to MongoDB. I could use some help with the MongoDB -> JSON conversion though. This also adds instance editing functionality

### DIFF
--- a/redbot/setup.py
+++ b/redbot/setup.py
@@ -233,7 +233,8 @@ async def edit_instance():
                     collection = m.get_collection()
                     await collection.update_one(
                         {'_id': m.unique_cog_identifier},
-                        update={"$set": core_data["0"]}
+                        update={"$set": core_data["0"]},
+                        upsert=True
                     )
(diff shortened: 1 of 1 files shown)


palmtree5:
Ok, I've written something that should work for Mongo -> JSON. Both directions need some testing though

palmtree5:
After some further work, I've gotten this working in both directions (JSON to MongoDB and MongoDB to JSON)

tekulvw (approved):

PR #1424, opened 2018-03-12
[Bugfix] Issue 1363

BaIthamel:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

[p]cogs and [p]repo now print alphabetically

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -248,6 +248,7 @@ class Downloader:
         Lists all installed repos.
         """
         repos = self._repo_manager.get_all_repo_names()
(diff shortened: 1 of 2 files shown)


tekulvw:
Why do you have `key=str.lower` in each of the `sorted()` calls? It seems unnecessary

BaIthamel:
Without that entries with capitalised letters will sort together before lowercase entries.
eg.
(Alpha, Zulu , bravo) without it
(Alpha, bravo, Zulu) with it
 It's more a just in case rather than needed.

tekulvw (approved):

tekulvw:
Thanks for the contribution!

PR #1425, opened 2018-03-12
[V3] Set token command

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

The `[p]set token` command from V2 is missing, and we are obligated to manually edit the `settings.json` file. That PR adds this command. This still needs testing (as I'm not home to do it) but, following the code from this file, this should work. I made it the same way as you made your commands in that file.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -580,8 +580,18 @@ class Core:
     async def token(self, ctx, token: str):
         """Change bot token."""
         if not not isinstance(ctx.channel, discord.DMChannel):
-            await ctx.send("Please use that command in DM.")
+            
+            try:
+                await ctx.message.delete()
+            except discord.errors.Forbidden:
+                pass
+            
+            await ctx.send(_("Please use that command in DM.n"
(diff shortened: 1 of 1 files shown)


tekulvw:
This should be DM only and the `if` check is unnecessary, the token argument isn't optional.

laggron42:
when I say I did it the same way you made your commands...
```Py
    @_set.command(aliases=["prefixes"])
    @checks.is_owner()
    async def prefix(self, ctx, *prefixes):
        """Sets Red's global prefix(es)"""
        if not prefixes:
            await ctx.send_help()
            return
        prefixes = sorted(prefixes, reverse=True)
        await ctx.bot.db.prefix.set(prefixes)
        await ctx.send(_("Prefix set."))
```
Anyway, gonna edit this!

tekulvw:
If a user tries to use this command outside of DM the bot should attempt to delete the message and tell the user that they should change their token (since they just released it publicly) before telling them to re-run the command in DM.

tekulvw (approved):

laggron42:
Sorry for not updating, I got to go

Issue #1196, opened 2017-12-18
[V3 Streams] Twitch Communities are broken


Tobotimus:
### Type:

- Bug

### Brief description of the problem
Twitch communities simply don't work in their current state. This is largely due to missed refactors, missing awaits, and general oversights on design.

I've discovered there is no framework set up to prevent duplicate alerts when streams go online for a community, meaning that alerts will be sent every 60 seconds as long as a stream is online.

tekulvw:
Fixed with #1426

PR #1426, opened 2018-03-12
[V3 Streams] fix community alerts

palmtree5:
see #1196. Community alerts were broken due to a number of issues. This should help towards fixing those issues

### DIFF
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -477,7 +477,7 @@ class Streams:
         for stream in streams:
             tw_id = str(stream["channel"]["_id"])
             for alert in self.streams:
-                if alert.id == tw_id:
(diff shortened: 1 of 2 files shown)


palmtree5:
After a bit of testing, this is having issues where it's sending duplicate messages, but only for some channels streaming to a community

palmtree5:
Community alerts have been moved to one embed per alert

The logic for determining what streams appear in the embed is as follows:

1. Get the list of streams currently live on a community (100 of them)
2. Any streams with a stream alert in the current channel (i.e. ones created with `[p]streamalert twitch channel`) are automatically excluded from the possible selections as those streams will already have an alert posted
3. `random.sample` is used to get 10 to display in the embed

At this time, communities are checked every 60 seconds, which will result in the embed updating every 60 seconds. It may be worth discussing reducing the check frequency on communities though

tekulvw (approved):

PR #1435, opened 2018-03-14
[V3 Economy/Bank] modify leaderboard command to allow displaying only guild members when bank is global

palmtree5:
THIS INTRODUCES A BREAKING CHANGE!!!!

~~The breaking change is that both `bank.get_guild_accounts` and `bank.get_global_accounts` both take a bot object so they can use that to get a user and return a list of tuples of (user_or_member, account)~~

UPDATE: The breaking change is now that both `bank.get_guild_accounts` and `bank.get_global_accounts` have been removed as they were only used for `[p]leaderboard` in core and the functionality has been implemented directly rather than passing a bot object off

The leaderboard command now supports (and defaults to) showing only guild members in the leaderboard when the bank is global


redbot/core/bank.py
```diff
@@ -312,19 +315,27 @@ def _invalid_amount(amount: int) -> bool:
     ret = []
     accs = await _conf.all_members(guild)
     for user_id, acc in accs.items():
+        member = guild.get_member(user_id)
+        if member is None:
+            member = await bot.get_user_info(user_id)
```
mikeshardmind: isn't doing this counterproductive if this is intended to just get that guild's members?

Wouldn't
```py
if member is None:
    continue
```
be more appropriate here?

palmtree5: Given that this is the guild-specific leaderboard, I'm thinking that the way I've done it there makes more sense rather than flat-out skipping over people simply because they happened to not be in the guild at the time the command was run (not to mention that the leaderboard didn't skip over people for not being in the guild before, why should it do it now?)

redbot/cogs/economy/economy.py
```diff
@@ -278,24 +277,30 @@ def __init__(self, bot: Red):
...
         if len(bank_sorted) < top:
             top = len(bank_sorted)
         topten = bank_sorted[:top]
         highscore = ""
         place = 1
         for acc in topten:
-            dname = str(acc.name)
-            if len(dname) >= 23 - len(str(acc.balance)):
-                dname = dname[:(23 - len(str(acc.balance))) - 3]
+            dname = str(acc[0].display_name)
```
mikeshardmind: This should be probably be `acc[0].name` (or include a comment as to why),  `display_name` is misleading to anyone reading the code later as this will be a user object any time the leaderboard is global, and the behavior of this on a user object is identical to name, even if there is a nickname available as nickname is not a thing for user objects

palmtree5: changed this now

palmtree5:
Closing in favor of #1386

PR #1441, opened 2018-03-19
[V3] Allow load, unload and reload to deal with multiple packages

Sitryk:
This PR allows multiple packages to be loaded, unloaded or reloaded with the one command, the packages are delimited by the space character as suggested by Will
This is just the functionality, I'm sure the code could be better

### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature


tekulvw:
Can you refactor this so that you don't have duplicate code in load/unload/reload? Extract some of that code into helper functions and call those instead.

palmtree5 (approved):

Kowlin (approved):

PR #1450, opened 2018-03-23
[Mod] DM Kick, Ban, and Softban Reason

200percentmicky:
Added the ability for Red to DM the member their reason for being kicked, banned, or softbanned (even though the bot DMs you anyway) from a server, along with the server's name and the Moderator responsible for the action. Also changed the bot's responses when using moderation commands, by mentioning the member being affected.

### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New Feature

Forgive me, this is my first time managing a repository.
Also, many of the responses that changed are for personal reasons. It's only specific to this branch.

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -331,7 +331,7 @@ class Mod:
                 await self.bot.say(":white_check_mark::boot: `Successfully kicked {}#{}` Begone ***T H O T !***".format(
                     user.name, user.discriminator))
             except discord.errors.Forbidden:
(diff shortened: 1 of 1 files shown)


Twentysix26:
This has been suggested before but I don't see it being a particularly useful feature. The responses don't need to be changed either.  
Also, it's good practice not to make multiple unrelated edits in a single PR, split them into smaller ones if needed.

PR #1459, opened 2018-03-24
[V3 Core commands] Message when loaded cog was not found

laggron42:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

The line that says "No module named {} was found in any cog path" was somehow as a comment in the load command. While keeping that, the bot won't say anything if the cog you are trying to load doesn't exist


tekulvw:
Closing in favor of #1460

Issue #1463, opened 2018-03-24
[V3 Core] Optional embed support discussion


palmtree5:
Ok, for implementing embed setting support (#1436) in `redbot.core`:

```
core_commands.py: info, contact, dm
cog_manager.py: cogs
help_formatter.py: help
modlog.py: Case.message_content
```
Those are all the places currently using embeds. I think the obvious skip here is `[p]info`. Everything else should be able to be modified to support text only.

palmtree5:
Done in #1467

PR #1467, opened 2018-03-24
[V3 Core] embed toggling support

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Implements #1463 
Still WIP as only `core_commands` and `cog_manager` are touched by this.


palmtree5:
Implemented into modlog as well. Please note that as a result of having to add a param for the bot to both `create_case` and `Case.edit` (so we can check the embed setting), this is a BREAKING CHANGE for anyone using either of those functions

palmtree5:
Implemented for `[p]help` now as well. For this one, I basically check the setting inside the help command and if the user is set to not use embeds, it'll fall back to discord.py's default help formatter (so basically what `[p]help` looked like in v2

redbot/core/core_commands.py
```diff
@@ -881,25 +895,36 @@ def check(m):
...
         fake_message = namedtuple('Message', 'guild')
         prefixes = await ctx.bot.command_prefix(ctx.bot, fake_message(guild=None))
         prefix = prefixes[0]
-        e.set_footer(text=_("You can reply to this message with %scontact"
-                            "") % prefix)
-        if ctx.bot.user.avatar_url:
-            e.set_author(name=description, icon_url=ctx.bot.user.avatar_url)
-        else:
-            e.set_author(name=description)
+        description = _("Owner of %s") % ctx.bot.user
```
tekulvw: Should really be using `.format()` here, if we're gonna use it everywhere else we should standardize on it.

tekulvw (dismissed):

tekulvw (approved):

PR #1470, opened 2018-03-25
[V3 Instance setup] implement backup support for instances using Mongo

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Implement backup support for MongoDB instances. Previously not implemented but as a result of my working on storage backend switching, I now know what I needed to do to export to JSON for backup purposes.

### DIFF
--- a/redbot/setup.py
+++ b/redbot/setup.py
@@ -16,6 +16,8 @@ from redbot.core.cli import confirm
 from redbot.core.data_manager import basic_config_default
 from redbot.core.json_io import JsonIO
 from redbot.core.utils import safe_delete
+from redbot.core.drivers.red_mongo import Mongo, get_config_details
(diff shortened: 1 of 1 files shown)


redbot/setup.py
```diff
@@ -16,6 +16,8 @@
 from redbot.core.data_manager import basic_config_default
 from redbot.core.json_io import JsonIO
 from redbot.core.utils import safe_delete
+from redbot.core.drivers.red_mongo import Mongo, get_config_details
```
tekulvw: You can't import these here as the bot will fail if the user doesn't have `motor` installed. Both of these imports should be in the `X_to_Y` functions.

tekulvw (changes requested):

tekulvw (approved):

PR #1471, opened 2018-03-25
[V3 Core] add [p]backup

palmtree5:
This will allow for creating backups while the bot is running

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -23,9 +23,6 @@ from redbot.core import checks
 from redbot.core import i18n
 from redbot.core import rpc
 from redbot.core.context import RedContext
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -21,6 +23,9 @@
 from redbot.core import i18n
 from redbot.core import rpc
 from redbot.core.context import RedContext
+from redbot.core.data_manager import basic_config, instance_name
+from redbot.core.drivers.red_mongo import Mongo
```
tekulvw: Gotta move these imports down to right before they're used.

tekulvw (changes requested):

tekulvw (approved):

PR #1472, opened 2018-03-27
[V3] Docker Compose 

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

Docker compose solution with persistent storage through volumes and safe handling of tokens with secrets. Documentation not here yet. Needs a token in a file at
`docker/settings/TOKEN` 
and a prefix at 
`docker/settings/PREFIX`


docker/settings/TOKEN.example
```diff
@@ -0,0 +1 @@
+notactuallyatokenyoushallnotpassblargh
```
tekulvw, line 1: Why can't this go into `docker/.env`?

mikeshardmind: becuase if we deploy an image, anything in env will show up in inspect iimage (including any tokens used for testing.

mikeshardmind:
Switched the base image from python:3.6-jessie to alpine:3.7 , made other modifications to match that resulting in a drop in image size from ~960MB to ~170MB

mikeshardmind:
Still more to do here including stronger documentation, and possible changes pending feedback.
Current testing steps for running without docker-compose
```
docker build . -t redbot
docker run --env-file .env -e UID=$(id -u) -v `pwd`/data:/data -t -d redbot
```
contents of .env file (template)
```
RED_TOKEN=token here
PREFIX=prefix here
```
compose solution still works with this.

mikeshardmind:
This needs enough changes that I'm closing this. Will re-open when the details have been discussed more in the discord, and a solution fitting that has been made.

PR #1481, opened 2018-03-31
[audio] don't remove active .part files

### LINKED ISSUE #1297: [V2] Audio.py - Cache manager appears to remove youtubedl partial downloads.
### Type:

- [ ] Suggestion
- [X] Bug

Looks like we might have a possible race condition between youtube-dl and the cache manager in audio.py
Youtube-dl downloads to a .part then renames, cache manager seems to be dumping .part files (`_cache_required_files()` doesn't seem to include these)

https://github.com/Cog-Creators/Red-DiscordBot/blob/a3bd757120ee99429a59893db7e07dcdb04e9671/cogs/audio.py#L538-L568
There's the cache dump

Here's where the file's required are defined:
https://github.com/Cog-Creators/Red-DiscordBot/blob/a3bd757120ee99429a59893db7e07dcdb04e9671/cogs/audio.py#L374-L383

I'm not 100% sure I've properly tracked the issue back though, and would definitely appreciate a second set of eyes on this before I make a PR of any sort to audio.
### Expected behavior
Things still needed in the cache shouldn't get removed until they are no longer needed.

### Actual behavior
(copied from the support channel in discord)
```
Feb 10 19:06:11 inxa-bot python3.5[25727]: ERROR: unable to rename file: [Errno 2] No such file or directory: 'data/audio/cache/xvJKMyMzjos.part' -> 'data/audio/cache/xvJKMyMzjos'
Feb 10 19:06:12 inxa-bot python3.5[25727]: data/audio/cache/xvJKMyMzjos: No such file or directory
```

calebj:
### Type

- [x] Bugfix

When downloading particularly large videos that cause the cache cleanup function to activate, partially downloaded files are deleted. The `_cache_desired_files()` function is supposed to avoid this, but does not account for the `.part` extension added by the library. This PR fixes that, as well as speeds up the filename checks a bit.

Fixes #1297

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -360,14 +360,15 @@ class Audio:
         self.queue[server.id][QueueKey.QUEUE].appendleft(queued_song)
 
     def _cache_desired_files(self):
-        filelist = []
+        filelist = set()
+
         for server in self.downloaders:
             song = self.downloaders[server].song
             try:
-                filelist.append(song.id)
+                filelist.add(song.id)
             except AttributeError:
                 pass
-        shuffle(filelist)
+
         return filelist
 
     def _cache_max(self):
@@ -380,13 +381,15 @@ class Audio:
 
     def _cache_required_files(self):
         queue = copy.deepcopy(self.queue)
-        filelist = []
+        filelist = set()
+
         for server in queue:
             now_playing = queue[server].get(QueueKey.NOW_PLAYING)
             try:
-                filelist.append(now_playing.id)
+                filelist.add(now_playing.id)
             except AttributeError:
                 pass
+
         return filelist
 
     def _cache_size(self):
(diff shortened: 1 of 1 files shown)


calebj:
Bumping this.

Tobotimus (approved):

aikaterna:
Leaving my approval on this, for what it's worth. Tested it for a few weeks with a constantly-playing bot with a small cache size.

Jackenmen:
I'm testing it for some time (almost 2 months I think) and I would say that it generally works, but I think there's still a problem somewhere as sometimes I still get now downloading file removed from cache. It almost always happens, when I didn't use the audio cog for few hours. I have to type [p]play command second time then and then it doesn't remove file from cache then.

PR #1488, opened 2018-04-01
[V3 Economy] Add per-role payouts to Economy with [p]economyset rolepaydayamount

skmendez:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

This is only possible if the bank isn't global. When `payday` is ran, the maximum of all of the payouts (default of 0 for every role) for the roles the Member has and the default guild payout is paid to the user.


tekulvw:
Why is this limited to global?

palmtree5:
@tekulvw it's not lol. Role-based payday payouts are restricted to the bank NOT being global

redbot/cogs/economy/economy.py
```diff
@@ -255,7 +260,12 @@ def __init__(self, bot: Red):
         else:
             next_payday = await self.config.member(author).next_payday()
             if cur_time >= next_payday:
-                await bank.deposit_credits(author, await self.config.guild(guild).PAYDAY_CREDITS())
+                credit_amount = await self.config.guild(guild).PAYDAY_CREDITS()
+                for role in author.roles:
```
palmtree5, line 264: why iterate through all of a user's roles to find a payday amount rather than just getting the payday amount based on the user's top role?

skmendez: All roles don't necessarily have a corresponding pay-out, so while their top role might be Admin, their highest pay-out role might be Gold/Platinum or the like.

palmtree5 (approved):

PR #1495, opened 2018-04-03
[V3 Mod] Denying add_reactions permission to muted users

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Since some of my members like spamming reactions when they get muted, I think that's a good idea to deny them the permission of adding reactions.


redbot/cogs/mod/mod.py
```diff
@@ -886,6 +886,7 @@ def __unload(self):
 
         perms_cache[str(channel.id)] = overwrites.send_messages
```
palmtree5: Let's actually make this track both the original setting of the `send_messages` overwrite and the `add_reactions` overwrite so that we can just set it back to the old value on unmute

palmtree5 (changes requested):

palmtree5 (approved):

Issue #1462, opened 2018-03-24
[V3 Lavalink] Infinite booting loop on macOS


laggron42:
### Type:

- [ ] Suggestion
- [x] Bug

### Brief description of the problem

When loading the audio cog (or when starting the bot if the cog is already default loaded), the bot will try to connect, fail, and try again...
```
Lavalink jar started. PID: 30762
Lavalink jar shutdown, restarting.
Lavalink jar started. PID: 30763
Lavalink jar shutdown, restarting.
Lavalink jar started. PID: 30764
Lavalink jar shutdown, restarting.
Lavalink jar started. PID: 30765
Lavalink jar shutdown, restarting.
Lavalink jar started. PID: 30766
Lavalink jar shutdown, restarting.
Lavalink jar started. PID: 30767
Lavalink jar shutdown, restarting.
...
```
Java is installed from oracle (openJDK isn't available) and up-to-date
```
MBP-de-Laggron:Audio laggron$ java -version
java version "1.8.0_144"
Java(TM) SE Runtime Environment (build 1.8.0_144-b01)
Java HotSpot(TM) 64-Bit Server VM (build 25.144-b01, mixed mode)
```
here's what shows when I try to run Lavalink from shell and under the `~/Red-DiscordBot/cogs/Audio` folder
```
MBP-de-Laggron:Audio laggron$ java -jar Lavalink.jar
       .   _                  _ _       _    __ _ _
      /\\ | | __ ___   ____ _| (_)_ __ | | __\ \ \ \
     ( ( )| |/ _` \ \ / / _` | | | '_ \| |/ / \ \ \ \
      \\/ | | (_| |\ V / (_| | | | | | |   <   ) ) ) )
       '  |_|\__,_| \_/ \__,_|_|_|_| |_|_|\_\ / / / /
    =========================================/_/_/_/

    :: Lavalink    ::
    :: Spring Boot ::        (v1.5.10.RELEASE)

[21:21:27] [ INFO ] [Launcher] Starting Launcher on MBP-de-Laggron with PID 30710 (/Users/laggron/Library/Application Support/Red-DiscordBot/cogs/Audio/Lavalink.jar started by laggron in /Users/laggron/Library/Application Support/Red-DiscordBot/cogs/Audio)
[21:21:28] [ INFO ] [Launcher] No active profile set, falling back to default profiles: default
[...]

PR #1498, opened 2018-04-05
[V3 Audio] Harden Lavalink boot sequence

tekulvw:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This PR does a lot...
* Ensures that java 1.8+ is installed
  * If 1.8, ensures a compatibility flag is passed to the lavalink launch command. This should fix some lingering issues with lavalink boot looping e.g. #1462.
* Downloads the lavalink jar from our releases page (sorry)
  * This has to happen so we can keep the lavalink version static to ensure compatibility.
* Completely redoes version handling so we only have to modify it in one place `redbot/core/__init__.py`


redbot/core/__init__.py
```diff
@@ -1,11 +1,40 @@
...
+        self.serial = serial
+
+    def __lt__(self, other):
+        my_index = self._levels.index(self.releaselevel)
+        other_index = self._levels.index(other.releaselevel)
+        return self.major < other.major or \
+            self.minor < other.minor or \
+            self.micro < other.micro or \
+            my_index < other_index or \
+            self.serial < other.serial
```
mikeshardmind: This should compare as 
```py
return (self.major, self.minor, self.micro, my_index, self.serial) < (other.major, other.minor, other.micro, other_index, other.serial)
```
The way you have it here with nested or would have 3.0.1 be < 3.1.0

mikeshardmind (commented):
Seems to have an issue with version info comparison

redbot/core/__init__.py
tekulvw: 3.0.1 is less than 3.1.0?

mikeshardmind: my mistake, I meant that both of these version comparisons evaluate true by the above:

3.0.1 < 3.1.0
3.1.0 < 3.0.1

palmtree5 (approved):

PR #1505, opened 2018-04-07
[V3 Docs] Systemd guide + extras

pgaetani:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Added links to each distro's user add functions
Added systemd service guide
Added --user operator to pip install commands

Why the `--user` tag? To eliminate the need to use `sudo` for pip operations, keeping system python clean and avoiding having to teach users to use `virtualenv` or `pyenv`.

I'm not 100% confident in my redaction, so any suggestions/changes are welcome.


mikeshardmind:
Should still be using venv's on the basis of having version conflict avoidance....

pgaetani:
Is venv the final approach you guys want users to take once out of beta?

If so, I could adjust this with some virtualenv instructions, instead.

Kowlin (approved):

PR #1507, opened 2018-04-07
Allow admins to stop polls (update general.py)

ghost:
Add a check to endpoll that checks whether the author of the `[p]poll stop` message has administrator perms. That way, admins can also stop a poll, instead of only the creator of the poll.

### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/cogs/general.py
+++ b/cogs/general.py
@@ -351,8 +351,9 @@ class General:
 
     async def endpoll(self, message):
         if self.getPollByChannel(message):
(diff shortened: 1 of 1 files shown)


ghost:
One note: I think `p == a` would also work instead of `p == a.id`, but as the older code also used `p.author == message.author.id`, I kept it that way.

Tobotimus:
Not going to be added in V2, also branch seems to be deleted

PR #1508, opened 2018-04-07
[V3] supress pynacl warning

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Supresses an irrelevant warning about a lack of pynacl from discord.py


tekulvw:
This is likely a better fit for `redbot/__init__.py`

palmtree5 (dismissed):
I'm still seeing the message 
`PyNaCl is not installed, voice will NOT be supported`
on startup when running this branch

mikeshardmind:
We didn't have a module import in `__main__.py` which meant the version checking didn't happen in there either when started from one of the console scripts. Changes are back in `__init__.py` as requested, with a module import added to `__main__.py`

mikeshardmind:
I've put the changes back in bot.py

Reasoning:
It may appear to be a global flag, but it isn't used as one (and doesn't make sense to be used as one),  discord.py sets this to false immediately after warning if it is True. This shouldn't ever be expected by anything to be True past the creation of the client.

This also avoids the behavior of a duplicate import warning when the bot is launched using `python -m redbot`

palmtree5 (approved):

PR #1509, opened 2018-04-07
[V3 Customcom] Removed spaces in the link

laggron42:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

The link wasn't clickable because there was spaces in it. Now it is with a hyperlink, since embeds supports it.

### DIFF
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -179,7 +179,7 @@ class CustomCommands:
                      ctx: commands.Context):
         """
         CCs can be enhanced with arguments:
-        https: // twentysix26.github.io / Red - Docs / red_guide_command_args/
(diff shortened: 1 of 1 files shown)


Kowlin:
This should be remade into the new docs format hosted on RTD.

laggron42:
what do you mean

pgaetani:
I think he means that the custom command args should be on RTD, and then the link should point there, instead.

laggron42:
yeah I'm aware of that, but for now, there are spaces in the middles of the link (wtf) so you need to copy paste then remove all spaces to access the *not outdated yet* guide

pgaetani:
I believe the point is more that all V3 related things should no longer point to the old docs, outdated or no. But that's not for me to decide, so I'll wait for more core dev feedback.

tekulvw:
Agreed, this needs remake on RTD seeing as the docs *are* outdated for V3 commands.

PR #1458, opened 2018-03-24
[V3 Config] Fix unloading and implement singleton driver

tekulvw:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

### Description of the changes
There was a bug with the unload command that wouldn't allow unloaded cogs to be fully released. This PR fixes that issue (as seen in #1457) and implements the singleton driver referenced in #1406.

PR #1514, opened 2018-04-08
[V3 Instance setup] fix an issue getting credentials to convert Mongo to JSON

palmtree5:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Was originally getting a nonexistent key instead of the actual storage credentials. This fixes that


palmtree5:
Actually, #1458 broke the storage backend switching feature completely (both directions)

tekulvw (approved):

Issue #1511, opened 2018-04-07
[V3] Config / driver singletons


mikeshardmind:
### Type:

- [ ] Suggestion
- [X] Bug

### Brief description of the problem

Driver isn't a singleton, new instances fail to write properly

### Actual behavior

I'm not 100% sure on the internals, but I'll include enough steps below that this can be tracked back quickly. This is also a newer issue somewhat

### Steps to reproduce
`````
[p]eval
```py
from redbot.core.config import Config
from redbot.core import modlog

# how modlog._conf is defined:
_modlog_type = type("ModLog", (object,), {})
x = Config.get_conf(_modlog_type, 1354799444)

y = modlog._conf

assert x.driver == y.driver
```
`````

tekulvw:
This is actually an issue with handling the fake type object. If the `cog_name` kwarg is used config acts as expected.

tekulvw:
Fixed with #1518

PR #1518, opened 2018-04-09
Remove old usage of type tuples with config. 

mikeshardmind:
Update dataconverter to work with that.

### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

Also resolves the weirdness with #1511 (which was only occurring due to the old type tuple syntax)


tekulvw:
Need to get multiple confirmations that this PR does not effect existing data. I don't believe it will, but we need to be pretty damn certain.

mikeshardmind:
My V3 instance isnt handling modlog on any of my servers yet, so I cant test for that. I can test the bank change included though. I also already verified that dataconverter works as expected after these changes.

tekulvw (approved):

PR #1524, opened 2018-04-13
Splitted deployment stage into their own stages.

Kowlin:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This identical build config will ensure that the build doesn't quit out once theres an error on a deployment scheme, where before if a deployment failed it would kill the whole build.

Also added some new files to the code owners.


palmtree5 (approved):
¯\_(ツ)_/¯ Kowlin says it should work, so I guess we'll find out if it did at deployment time

PR #1529, opened 2018-04-16
[V3] guild->server (And customcom recognizes {server} )

bobloy:
Update redbot/cogs/ to guild vs. server best practice
Mostly string renames
customcom now recognized {guild} and {server}

### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Mostly user-facing text fixes of guild -> server 
Allows {server} to be recognized by custom commands

### DIFF
--- a/redbot/cogs/admin/announcer.py
+++ b/redbot/cogs/admin/announcer.py
@@ -65,7 +65,7 @@ class Announcer:
             try:
                 await channel.send(self.message)
             except discord.Forbidden:
-                await bot_owner.send("I could not announce to guild: {}".format(
+                await bot_owner.send("I could not announce to server: {}".format(
                                          g.id
                                      ))
             await asyncio.sleep(0.5)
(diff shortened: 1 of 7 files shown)


shikhir-arora:
`V3` uses Discord.py rewrite, so I think the goal here is to be consistent. From the docs for the rewrite:


> The official API documentation calls the “Server” concept a “Guild” instead. In order to be more consistent with the API documentation when necessary, the model has been renamed to Guild and all instances referring to it has been changed as well.

Bakersbakebread:
I think it was decided that guild was to be used in the back end, and server in front.

palmtree5:
Discord calls them servers in their own UI while the API docs refer to them as guilds. Thus, I think it makes sense that they should be referred to as servers with regards to user-facing output and as guilds otherwise

palmtree5 (approved):

PR #1536, opened 2018-04-16
[V3 Launcher] Reinstall Red option

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This PR adds a new button in the launcher: `Reinstall Red`. This is similar to the v2 option, it can reinstall requirements and discard code changes, reset data, or both. A lit of things were edited in `launcher.py` and `setup.py` to make it (mostly some features moved to a function so they can be used twice without repeating).
**That is not tested yet** because of the packages conflict with `websockets` which makes the update glitching. Please test it and push any changes if you're a maintainer so we can merge this before beta 12.


Kowlin (dismissed):
Changes required since this is removing functionality from another PR.

redbot/setup.py
```diff
@@ -287,7 +283,7 @@ def basic_setup():
                 if confirm("Would you like to import your data? (y/n) "):
                     await json_to_mongo(current_data_dir, storage_details)
         else:
-            storage_details = instance_data["STORAGE_DETAILS"]
+            storage_details = default_dirs["STORAGE_DETAILS"]
```
palmtree5: NO NO NO. This should NOT have been changed. `storage_details` is set to the storage details from the instance settings for a very, very good reason (the actual storage details are needed if the user wants to import from MongoDB to JSON)

palmtree5 (dismissed):

redbot/setup.py
laggron42: That wasn't intended. Seems that's from another PR. Removing that...

laggron42:
Fixed wrong changes. Needs testing

laggron42:
I recorded a video showing the features on the last commit
https://youtu.be/L4rf_ejv4xE

Kowlin (approved):

Issue #902, opened 2017-08-11
[Permissions] Move current checks into something that permissions can manage


tekulvw:

Tobotimus:
It would be great if cogs had access to not only check decorators, but their predicates as well. This is something we lack with the current setup with checks, that we can easily check for admin roles with a decorator, but checking it any other way requires us to write the predicate ourselves from scratch.

PR #1547, opened 2018-04-18
Make checks.py manageable from a permissions cog

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This sets up the framework for a Permissions cog to manage checks.py (#902)
This does not change the behavior of existing checks until a Permissions cog is set up
to use this.

This also moves the predicate logic of those checks outside of the checks themselves, making them accessible to cog developers


mikeshardmind:
Oh, right, forgot to post my reasoning for hooking a cog in this part in checks rather than have this be accessible more similarly to mod_role/ admin_role ( and not directly attached to #1548 )

This allows someone to load a 3rd party permission cog of their own (or someone else's) creation as an alternative, while still providing everything needed in core red.

Tobotimus (approved):

PR #1548, opened 2018-04-18
[V3] Permissions

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

Sets up an extended, higher level permission model


mikeshardmind:
I've added a way to handle specific checks that may be desirable to use a different resolution order than default in here as well. This is still a WIP.

mikeshardmind:
This was turning into an interactive hell. This will be manageable by uploaded yaml file (with modes for both update and set) in the initial iteration. There will be a template provided, and a method to show the current yaml equivalent and to clear all. 

This will be extensively documented, and anyone can add a more user friendly interface at a later point if they feel they would fare better on designing it.

mikeshardmind:
This is ready for review. None of the check logic has any reason to change at this point unless a flaw is found in it. There's an example template in there for the expected input format as well. I'll update with real documentation later, but it won't be today.

This also needs a couple user utility functions to get the IDs which correspond to various objects.
At a later date, I'd like to see this manageable via webpanel, but the level of interactivity needed to do this as discord commands, even with emoji in play was getting overly complex.

mikeshardmind:
Fixed the issues with the permission cog which were introduced earlier:

On my todo list:
    extra commands for individually adding items to existing rule sets
    extra command to generate table of rolenames -> roleids for use with yaml
    docs for end users

On my todo list, but not a priority:
    documentation for what is required to properly override the default permission cog

palmtree5:
Ok, to put these thoughts here:

A few requests:
1. Documentation on the yaml for adding a bunch of rules in one go
2. Some way of setting defaults outside of yaml
3. A method of wiping all configured server/global rules at once

Otherwise, the docs look good so far and I'm not finding any major issues in testing this

mikeshardmind:
@palmtree5 I've added the requested extra commands. I'll update the docs with the yaml usage a little later as well.

Ideally, after RPC is implemented, this could pave the way for a nicely designed webpage managing this. As it is however, the yaml syntax is easy enough for anyone wanting to design a page that could generate a yaml file for use from a guild (would require OAuth scopes)

mikeshardmind:
All user facing commands and normal operation permission logic has been tested and confirmed to be working.

mikeshardmind:
This is ready for merge as far as I'm concerned. There's an argument to be made for additional documentation beyond what is provided regarding the additional check system, but the core permissions cog works as intended now (as do the hooks for additional rules registered by 3rd party cogs)

Those hooks have documentation in docstrings, but if we want a full usage guide for them for developers, I can write that tommorow.

Kowlin (approved):

PR #1552, opened 2018-04-19
[V3] Remove audio relative path for rare error

bobloy:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Removed relative path in audio
Helps clarity and a rare audio error

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -8,7 +8,7 @@ import redbot.core
 from discord.ext import commands
 from redbot.core import Config, checks, bank
 
-from audio.manager import shutdown_lavalink_server
+from . import manager
 
 __version__ = "0.0.5"
 __author__ = ["aikaterna", "billy/bollo/ati"]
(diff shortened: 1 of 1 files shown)


mikeshardmind:
~~snip~~ Was being dumb

tekulvw:
Neither of the two imports I'm seeing here will *ever* work. First, @bobloy, there is no top level package `audio`. Python package resolution does not work that way. Also @mikeshardmind there is no `redbot.audio` package so I'm not really sure what your comment proves?

tekulvw:
What even is the issue that this PR is attempting to solve?

mikeshardmind:
@tekulvw my mistake, I had intended for that to be `import redbot.cogs.audio.manager` , even so we've had one person with an issue (in #v3-support channel), and I'm not convinced (at all) that this PR would (could, or should work here too) fix their issue.

bobloy:
Having this user try `import redbot.cogs.audio.manager` did succeed FYI

bobloy:
Edited import scheme, see commit

bobloy:
Root error was not identified, gonna close for further debugging

PR #1596, opened 2018-05-05
[V3 cleanup] Respect pinned messages by default

### LINKED ISSUE #1589: [Cleanup] Respect pinned mesages
Ohhh ye, we're adding more upon Cleanup's logic like it already wasn't hard enough! 😏 Cleanup should respect pinned messages when cleaning up in any setting used by not removing them when their detected in the list.

mikeshardmind:
resolves #1589

### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

modifies `get_messages_for_deletion()` to only give back pinned messages if explicitly included


palmtree5 (changes requested):
Shouldn't there be some way for the user to actually delete pinned messages using the commands? As it stands, one could `[p]eval` it to get a list and then delete it in the same eval, but otherwise this is just sitting there basically being the same all the time

palmtree5 (approved):

Issue #697, opened 2017-04-06
fuzzysearch recommendations for !help


Chovin:
suggestion:

!help playing

non-fuzzy example:
```
No command called "playing" found. Did you mean..

1   [p]audioset thumb       Toggles displaying song thumbnail where applicable
2   [p]audiostat servers    Number of servers currently playing.
3   [p]queue                Queues a song to play next. Extended functionality in `[p]help`
4   [p]set game             Sets Red's playing status
5   [p]stop                 Stops a currently playing song or playlist. CLEARS QUEUE.
```

additionally, for my lazy self:

4 Sublime Text

```
Done.
```
(invoked [p]set game Sublime Text)

Kowlin:
I would like this but yyee... then we will have the problem of our "we don't have help in Red" and I kinda like that.

Chovin:
I agree..
Maybe as a separate cog...

calebj:
If you want to do this for commands normally, you can use `on_command_error` and checking if it is `commands.CommandNotFound`, I think. The additional cog would need to pull in dependencies like [fuzzywuzzy](https://github.com/seatgeek/fuzzywuzzy) for comparison of keys and docstrings loaded from `bot.commands` and subcommands for command groups.

To replace `[p]help`, you can unload the command. I figured out the code to reload the regular command from discord.py, which can be put in this hypothetical cog's `__unload` method.

Kowlin:
Judging from the above example I think this is meant to be only run on the ``help`` command.

Kowlin:
To be done in V3's new help formatter.

PR #1600, opened 2018-05-06
[V3] Start work on fuzzy command search

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This PR will introduce a fuzzy command search both in the command error handler (for `CommandNotFound` errors), as well as in `[p]help` (for example, if someone does `[p]help playing`)

See also: #697


palmtree5:
Note that this introduces `fuzzywuzzy[speedup]` as a dependency to support this functionality.

That said, this is mostly ready for testing. It works for running non-existent commands by handling `CommandNotFound` errors and works for `help` by checking based on the part after `help` (so if I did `[p]help twi`, it would use `twi` to find possible commands). There could be some wonky behavior in the search though (such as different output in help when using embeds)

Kowlin (approved):

PR #1618, opened 2018-05-08
[V3] Ping changes

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

* Moves ping to core commands (works without general loaded)

* defaults ping behavior to reacting with a ping pong paddle with ball (less noisy verification bot is up)

~~* adds an optional boolean flag to ping to get the avg latency from the bot (strikes a middle ground with intended behavior from developer standpoint, and how users want it, still only makes 1 API call)~~


bobloy:
While you're implementation of latency seems harmless, last I heard it was useless.
image

mikeshardmind:
not ready for b14

mikeshardmind:
Well, considering what people expect `ping` to do (by name, convention of other bots, etc.), I think this makes sense to add, possibly adding another command filling the stated purpose from that quote if something like that is really needed.

palmtree5 (approved):

PR #1622, opened 2018-05-09
[V3 Docs] Removed py 3.6 warning for Windows

ScarletRav3n:
### Type

- [?] Bugfix
- [ ] Enhancement
- [ ] New feature

I believe we've fixed this a while ago.


pgaetani:
Would it not be a good idea to add a warning about not using 3.7 for now?

ScarletRav3n:
I didn't know 3.7 was even out. It looks like it's still in beta.

mikeshardmind:
Would make more sense to put the supported version info on pypi and let pip handle it from there. (#1623 has that now)

Tobotimus (approved):

Issue #1636, opened 2018-05-13
[V3] Red's Development Workflow


Tobotimus:
Ahem.
There's been a lot of discussion from us devs lately about Red's development workflow and how we can improve it. This issue aims to give a summary of that discussion.

# Introduction
Developing for Red in its current state can involve jumping through a lot of hoops and addressing many caveats in order to produce quality code and contributions. Part of this has to do with how we run tests, and another part is how we set up our development environment.

# Development Workflow Outline
Let's picture a developer first starting out contributing to Red. They needs to go through a couple of basic steps, as is the case with many python projects:
1. Fork and clone the repository
2. Install dependencies to a local virtual environment
3. Have some way of running his local development version of Red in said environment
4. Test their changes to verify they haven't broken everything
5. Open a pull request, subject to approval of the following conditions:
    5.1. The developers agree with your design and implementation
    5.2. A code review from the developers passes
    5.3. The Travis build succeeds

### So what's the matter with this?

Step (1) is all good.

Step (2) could be improved currently. One example for how we might do it right now are as follows:

    pip install git+https://github.com/Rapptz/discord.py#egg=discord.py[voice]
    pip install -e .[docs,test,voice,mongo]

This didn't actually work with Red-Trivia<1.1, but luckily that will be history very soon thanks to #1635. In addition, upgrading dependencies involves repeating the same commands again with upgrade flags. I don't like big links.

Step (3) is pretty easy, simply run the bot like you normally would.

Step (4) involves installing pytest and running, but it's not just the usage of this which is lacking right now, but the actual coverage and extensiveness of our testing. For changes to docs, they also need to be built, which we don't really have a guide for. Moreover, style isn't enforced in any way.

Step (5) is all good, except that we should really be testing the stuff I described in step (4) on Travis too. This would help the devs a lot with code reviews.

Lastly, I'd like to note that although this is how I would recommend setting up a good environment to develop and test Red, this isn't the way Travis does it. Personally, I think that local testing and Travis should be standardised and unified.

# Making adjustments

#### Installation

~~Virtual environment tools can be handled by the developer in my opinion. End users will need hand holding when it comes to virtual environments, but it's not too hard for a developer to learn. Tools like pipenv, pyenv, virtualenv all work fine.~~

~~We can do less commands for installation by creating a 2nd pip requirements file, I'll call it `dev_install.txt`, so all we have to do to achieve step (2) is the following:~~

    pip install -r dev_install.txt

[...]

PR #1638, opened 2018-05-13
[V3] Add requirements file for setting up dev environment

Tobotimus:
Update: Superseded by #1642 

### Type

- Meta enhancement

This adds the `dev_install.txt` requirements file I've discussed in #1636. It makes setting up a development environment for Red as simple as `pip install -r dev_install.txt`.

I've also modified the Travis config to use this new requirements file for the installation phase.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,9 +8,7 @@ python:
 - 3.5.3
 - 3.6.1
 install:
-- echo "git+https://github.com/Rapptz/discord.py.git@rewrite#egg=discord.py[voice]" >> requirements.txt
-- pip install -r requirements.txt
-- pip install .[test]
+- pip install -r dev_install.txt
 script:
 - python -m compileall ./redbot/cogs
 - python -m pytest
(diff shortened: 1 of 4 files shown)


tekulvw:
There is a relatively new tool out there called "Pipenv". This tool is the officially recommended packaging tool according to the Python packaging authority (the guys who maintain PyPI) and has a few benefits over simply declaring the development libraries in a "dev_install.txt" file.

First, Pipenv allows us to declare dependencies for both production and development use, using the new modifiers introduced in PEP 508. Pipenv also allows us to "lock" those dependencies to a particular version and which means we can make lists of dependencies that are known to work. Once we have that list it's extremely easy to deterministically reproduce development or production environments that are guaranteed to work.

Pipenv also layers on top of Python virtual environments. So, when you install a list of dependencies using Pipenv, the tool also automatically creates a virtual environment from a recommended Python version. Activating the virtual environment is as simple as running `pipenv shell` in the project directory.

It is important to note the difference between a strict "packaging" tool and a "package manager". Pipenv is a packaging tool where pip is  package manager. Pipenv does not have the ability to install the bot project from PyPI. Pipenv is instead used to recreate a development environment for users working on developing core bot code. For example, to use pipenv, you would first clone the bot project from GitHub, `cd` into the project directory, and run `pipenv install --dev`. Pipenv would then set up a clean virtual environment for you that has all of the dependencies necessary to perform bot development.

Tobotimus:
Superseded by #1642

PR #1641, opened 2018-05-13
[V3] Add tox

Tobotimus:
At the moment, these are the environments tox is running:

- pytest, for python 3.5 and 3.6
- a docs build, with warnings being raised as errors
- a dev install, to make sure pip installing with the `-e` flag works

Travis should also be configured to run these test environments as 4 builds in parallel.


Tobotimus:
This'll be on hold until I drop support #1638 in favour of #1642

palmtree5 (dismissed):

palmtree5 (approved):

PR #1664, opened 2018-05-15
[V3 Warnings] Send DM option

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This option will make the bot send a DM with the warn reason when the command is used. I know you guys don't like the DMed reasons but this is something a lot of servers are using, so I though this would be cool to add.

### DIFF
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -294,7 +294,7 @@ class Warnings:
         else:
             if not await is_admin_or_superior(self.bot, ctx.author):
                 await ctx.send(
-                    warning(_("You are not allowed to check " "warnings for other users!"))
+                    warning(_("You are not allowed to check " "warnings for other users!")))
                 return
             else:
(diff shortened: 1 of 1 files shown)


redbot/cogs/warnings/warnings.py
```diff
@@ -271,8 +294,7 @@ def same_author_check(m):
         else:
             if not await is_admin_or_superior(self.bot, ctx.author):
                 await ctx.send(
-                    warning(_("You are not allowed to check " "warnings for other users!"))
-                )
```
tekulvw: You need to add this back...your code as-is won't even compile.

tekulvw (changes requested):

redbot/cogs/warnings/warnings.py
laggron42: Don't see why it got removed, I'll add it back

laggron42:
Actually I have better plans. I'll close this and make something better. May be ready for next beta.

PR #1665, opened 2018-05-15
Add a makefile for helping with style checking and reformatting

palmtree5:


mikeshardmind:
I have no issues with providing a makefile and a make.bat, but in this implementation, we're duplicating args passed to black, and also hard coding paths to check, while also including files which aren't python files.

This could be changed so that the makefile (and a make.bat) both point to `tox -e style` (only runs the style check) and have tox run a python file in that section which runs black against just the python files which are tracked by git.

Tobotimus:
As for hard-coded paths, that can easily be fixed by substituting those paths with `git ls-files "*.py"`. Also, no matter what is passed as arguments, black will ignore all files except for `.py` and `.pyi` files, so including other files isn't a problem.

`tox -e style` is significantly slower than directly calling black from the makefile. Moreover, there's really not much we can change about black's arguments to warrant such concern for duplicating them, and I certainly don't think this needs yet another script.

mikeshardmind:
Then at the very least, doesn't it make more sense to provide the style check as a single cross platform python script and have tox's style check call that? 

I'm envisioning how easy it would be to forget that this is used in 3 distinct locations 3 months down the line when some file not currently included in hardcoded paths is used, or if arguments to black are changed for some reason.

tekulvw:
I really don't think it's that big of a deal. If we forget to add something down the line then Travis will tell us. However I really don't envision a scenario where we get swamped by new python files that magically appear outside the redbot and tests folders.

Simple is good, as-is this is simple and clear what it does. I don't think we need to over engineer this.

Tobotimus:
[RE: mikeshardmind] Does it make more sense? I don't think so. I think it makes more sense to invoke command line applications from a Makefile or batch script than a python script.

As for what you're envisioning, I can't say I share the same view. This minuscule amount of information is duplicated across 2 files, not 3, and they're both makefiles, so it's intuitive to know that modifying one means modifying the other. I've already addressed the hardcoded paths in my previous comment.

mikeshardmind:
The 3rd being tox.ini right now.

I also don't think programatically detecting which python files belong to the repository is over-engineering it. I think it's an easy way to avoid future mistakes.

Tobotimus:
Avoiding the duplication in tox.ini is trivial, we can just call make from tox.

mikeshardmind:
Okay, so if calling make from tox, then it's a makefile and make.bat, which at least are clearly connected.

However, I still strongly disagree with @tekulvw on 

> I really don't think it's that big of a deal. If we forget to add something down the line then Travis will tell us. However I really don't envision a scenario where we get swamped by new pythpn files that magically appear outside the redbot and tests folders.
> 
> Simple is good, as-is this is simple and clear what it does. I don't think we need to over engineer this.

```
reformat:
	black -l 99 `git ls-files '*.py'`
stylecheck:
	black --check -l 99 `git ls-files '*.py'`
```
is no less clear as what is going on, remains simple, and does not run the risk of missing a tracked file.

tekulvw (approved):

PR #1666, opened 2018-05-15
[V3] Optimize the backup command

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This optimizes the file generated when doing `[p]backup` to exclude:
- The Lavalink jar file
- All `__pycache__` folders
- `RepoManager`'s `repos` folder
- `Downloader`'s `lib` folder
- `CogManager`'s `cogs` folder

However, this also generates a repo list that will be used in a separate, yet-to-be-written restore function and in conjunction with `Downloader`'s settings (which contain the list of installed cogs) to allow reinstalling cogs when restoring.

I still need to implement this into backups done from `redbot-setup`'s functions though, so this is not ready for merge quite yet. I'll comment when I feel it's ready for merge


mikeshardmind:
Would it make sense for this to prompt to send it via DM if the archive size is under file size limitation for bots? (8mb)

palmtree5:
Oh, and another minor detail: you'll be able to specify a path for where to put the file (though the filename will remain in the same format)

palmtree5:
ok, this should be ready for testing at this point

Kowlin (approved):

PR #1667, opened 2018-05-16
Unembed and send message on ping

Stylite:
### Type

- [ X] Bugfix
- [ ] Enhancement
- [ ] New feature

This removes the embed from ping along with sending a reaction and message when the command is run.


tekulvw:
Unless we see some massive community backlash over the reaction that falls back to a message, the functionality is staying as is.

PR #1685, opened 2018-05-21
[V3 Docs] Move the install docs to install Python 3.6

palmtree5:
As a result of #1684 the docs need to be updated to have users install Python 3.6


palmtree5:
Note: This also throws in a page for Ubuntu 18.04.

This should also be ready to go, though someone with a Raspberry Pi may want to check that the instructions there work (they're basically the same as the ones for Debian, just on Raspbian)

Kowlin (approved):

Issue #1605, opened 2018-05-07
[V3 Help] Extra features


Bakersbakebread:
This issue is created to discuss some extra features for help command :

- [X] helpset command (#1694)
- [x] allow users to add 'tagline' to suffix help command (#1705)
- [ ] default to DM on large message but set to channel toggle

tekulvw:
The third feature can be accomplished by doing `!helpset maxpages 0` with the changes in #1694

PR #1705, opened 2018-05-23
[V3 Help] add tagline support

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Implements another feature requested in #1605 

`[p]helpset tagline` allows for adding additional text onto the current footer message (this one)
```
Type !!!help <command> for more info on a command. You can also type !!!help <category> for more info on a category.
```

The command: 
- grabs the old tagline
  - If a tagline already exists, sets the tagline in config to an empty string 
- grabs formatted help for itself using the embed help formatter
- grabs the footer out of the first embed in the list 
- length calculations (split out based on the value of the result)
  - default footer + 2 newlines + the new tagline > 2048 
    - a message is sent indicating the tagline is too long
      - Contains max character count allowed for it (2046 - the length of the default footer)
    - old tagline (if there was one) is placed back in config. 
  - default footer + 2 newlines + the new tagline <= 2048 
    - the user's input is set as the tagline in config
    - a message is output indicating that it has been set

NOTE: The reason for formatting help and extracting from the embed instead of just calling the function containing the text of the default footer (`redbot.core.help_formatter.Help.get_ending_note`) directly is an error message I was getting trying to do it that way

```py
ERROR events on_command_error 203: Exception in command 'helpset tagline'
  File "/home/palm/.local/share/virtualenvs/Red-DiscordBot-zFqGc7tM/src/discord.py/discord/ext/commands/core.py", line 62, in wrapped
  File "/home/palm/code/Red-DiscordBot/redbot/core/core_commands.py", line 886, in helpset_tagline
    end_note_length = len(ctx.bot.formatter.get_ending_note())
  File "/home/palm/code/Red-DiscordBot/redbot/core/help_formatter.py", line 116, in get_ending_note
    self.clean_prefix
  File "/home/palm/.local/share/virtualenvs/Red-DiscordBot-zFqGc7tM/src/discord.py/discord/ext/commands/formatter.py", line 184, in clean_prefix
    user = self.context.bot.user
AttributeError: 'NoneType' object has no attribute 'bot'
```

EDIT: formatting of the command logic description


palmtree5:
On second thought, I've actually changed this to be so that the user has full control over the content of the footer in the help embeds. They can always reset to the default too (by running `[p]helpset tagline` with no args).

The thought with putting the tagline in the footer is that the footer is already consistent across all cases of help command usage, whereas something like the description isn't. 

Consider the following:

- Command
  - Footer: `Type ???help <command> for more info on a command. You can also type ???help <category> for more info on a category.`
  - Description: `"Syntax: {0.clean_prefix}{1.signature}".format(self, command)`
- Cog
  - Footer: `Type ???help <command> for more info on a command. You can also type ???help <category> for more info on a category.`
  - Description: class docstring, possibly translated into another language
- Bot
  - Footer: `Type ???help <command> for more info on a command. You can also type ???help <category> for more info on a category.`
  - Description: *Red - Version \<insert version number here\> - Page \<number\> of \<total\>*

It's much easier to just put this into the footer of the embed rather than trying to put it into the description, which is of variable length (especially when considering class docstrings are used and are likely to be different in length per-cog). We're also not really losing anything giving the user control over the content of the footer as the default is literally a somewhat concise version of `[p]help help`.

Kowlin (approved):

PR #1706, opened 2018-05-24
[V3 Context] use bot's color if it has one (Ux Consistency with help formatter)

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

adds `ctx.color` which will return the top role color a bot has (unless all roles are colorless, or no roles in the context) in which case it will use the bot's set color

`ctx.maybe_send_embed` will use ctx.color as the embed color if it sends an embed


mikeshardmind:
Modified this to leverage changes to be introduced in #1707 As such, this should only be merged after that.

palmtree5 (approved):

PR #1712, opened 2018-05-24
[V3 ping] fixes reaction on message deletedelay of zero 

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Put's the ping command back to only responding with "Pong." per @tekulvw 's decision in discord.


tekulvw:
At this point it may be easier to revert to the initial implementation. I was originally OK with the reaction because it could be considered a "spam" prevention measure since the bot didn't need to send a message to communicate responsiveness. However, now I just don't think the added complexity is worth it especially considering now we've moved away from the "shortest code path possible" to getting a response from the bot.

My vote is to revert.

mikeshardmind:
I've removed the reaction logic per comment, but I do think keeping the maybe embed portion is worth it for Ux consistency.

Stylite:
I don't think keeping the maybe embed portion is worth it because the word pong doesn't need to be inside an embed, it can just be plain text.

mikeshardmind:
I think all core code should be respecting the settings we offer. Arguments for exceptions can be made in exceptional cases (i.e. commands which are only available when loaded with `--dev` and are not intended to be exposed to end users) I think everything which is in core should be respecting this setting or there's no point in actually offering the setting.

Kowlin (approved):

Issue #1716, opened 2018-05-26
[V3] Fire an event on modlog case creation


palmtree5:
# Feature request

#### Select the type of feature you are requesting:

- [ ] Cog
- [ ] Command
- [x] API functionality

#### Describe your requested feature

This would add an event that gets fired when a modlog case is created. Some cog could then handle that in a function `on_modlog_case_create` or something like that to do something when a modlog case is created (for example, possibly notifying the affected user via DM when a case is created that they are the subject of)

Tagging @retke because this idea spawned out of the example I mentioned above

laggron42:
Maybe it would be good to edit mod.py so it creates the modlog **before** performing the action, so the 3rd party cogs can do their things before losing access to the member.

PR #1717, opened 2018-05-26
[V3 Modlog] add events for modlog cases

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

WIP (Milestoned as beta 16 because this probably won't be ready for beta 15 release)

The idea is to add an event for modlog case creation, but this may also end up containing an event for modlog case edits as well. There will likely be refactoring as well to move message send/edit logic out of the create/edit functions and into event handlers
See also #1716


palmtree5:
Ok, this should actually be ready for testing (and it absolutely needs to be tested before merge as it refactors some logic surrounding modlog case message sending/editing by moving it to event handlers)

palmtree5:
Ok, did a bit of testing and found that both events are now working

Tested:
- Case creation (via `[p]eval`)
- Case edits (via `[p]reason` albeit with a bit of `[p]mock` thrown in)

Kowlin (approved):

Issue #1741, opened 2018-05-29
[V3 Core] Error with set if used in DM


laggron42:
# Command bugs

#### Command name

set

#### What cog is this command from?

Core commands

#### What were you expecting to happen?

show the help list

#### What actually happened?

```Py
Exception in command 'set'
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/discord/ext/commands/core.py", line 62, in wrapped
    ret = yield from coro(*args, **kwargs)
  File "/usr/local/lib/python3.6/dist-packages/redbot/core/core_commands.py", line 518, in _set
    admin_role_id = await ctx.bot.db.guild(ctx.guild).admin_role()
  File "/usr/local/lib/python3.6/dist-packages/redbot/core/config.py", line 760, in guild
    return self._get_base_group(self.GUILD, guild.id)
AttributeError: 'NoneType' object has no attribute 'id'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/dist-packages/discord/ext/commands/bot.py", line 886, in invoke
    yield from ctx.command.invoke(ctx)
  File "/usr/local/lib/python3.6/dist-packages/discord/ext/commands/core.py", line 923, in invoke
    yield from injected(*ctx.args, **ctx.kwargs)
  File "/usr/local/lib/python3.6/dist-packages/discord/ext/commands/core.py", line 71, in wrapped
    raise CommandInvokeError(e) from e
discord.ext.commands.errors.CommandInvokeError: Command raised an exception: AttributeError: 'NoneType' object has no attribute 'id'
```

#### How can we reproduce this issue?

Try the `[p]set` command in DM

Tobotimus:
Thanks for submitting this issue!

Fixed in #1748

PR #1748, opened 2018-05-30
issue #1741 fix

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Checks that we're in a guild before attempting to print guild settings in set

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -515,30 +515,31 @@ class Core:
     async def _set(self, ctx):
         """Changes Red's settings"""
         if ctx.invoked_subcommand is None:
-            admin_role_id = await ctx.bot.db.guild(ctx.guild).admin_role()
-            admin_role = discord.utils.get(ctx.guild.roles, id=admin_role_id)
-            mod_role_id = await ctx.bot.db.guild(ctx.guild).mod_role()
-            mod_role = discord.utils.get(ctx.guild.roles, id=mod_role_id)
-            prefixes = await ctx.bot.db.guild(ctx.guild).prefix()
-            if not prefixes:
-                prefixes = await ctx.bot.db.prefix()
-            locale = await ctx.bot.db.locale()
+            if ctx.guild:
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -515,30 +515,31 @@ def maybe_reload(new_name):
...
-                    mod_role.name if mod_role else "Not set",
-                    locale,
+            if ctx.guild:
+                admin_role_id = await ctx.bot.db.guild(ctx.guild).admin_role()
+                admin_role = discord.utils.get(ctx.guild.roles, id=admin_role_id)
+                mod_role_id = await ctx.bot.db.guild(ctx.guild).mod_role()
+                mod_role = discord.utils.get(ctx.guild.roles, id=mod_role_id)
+                prefixes = await ctx.bot.db.guild(ctx.guild).prefix()
+                if not prefixes:
+                    prefixes = await ctx.bot.db.prefix()
```
Tobotimus: Perhaps we should be using `bot.get_prefix(ctx.message)` instead of reading directly from Config? I know it's not exactly the scope of this PR but I'm happy to have it done here as a minuscule change.

Tobotimus (commented):
Thanks for this bugfix.

I think it would be useful to still display some global information in DM, such as global prefixes and the locale.

redbot/core/core_commands.py
mikeshardmind: Probably not. bot.get_prefix(ctx.message) will include both mention forms if the bot is mentionable, neither of which will display properly in the box formatting, and including both may be seen as confusing to users. I will be updating this with the changes allowing the global information to still be displayed.

mikeshardmind:
@Tobotimus I've modified this to grab available settings. I did not switch to using `bot.get_prefix` due to the way mentions are rendered in code block. I can add the mention manually to the code block based on the setting if this is within scope.

Tobotimus (approved):
Thanks, I agree RE: get_prefix. If really required the mention spec can be added in another PR.

Issue #1734, opened 2018-05-28
[V3 Docs] User guides


laggron42:
The Red ReadTheDocs page currently has installation informations and API/coding guides. I think it would be good to also include some user guides, like "getting started" or detailed cog guides, as an alternative to the help command.

I can write these docs, I just need to know what you think about it and what would be the format. I already wrote detailed docs for my repo's cogs and I'm planning to do it the same way for core cogs. 

Please take a look at the cog guides section of my [ReadTheDocs page](https://laggrons-dumb-cogs.readthedocs.io) and see if it the format is ok for core cogs.

mikeshardmind:
I feel like most of the commands in the bot do not need this level of explanation. However, I do think there are certain features in the bot which could be better documented. I think this is overall a good idea, but leave off the installation section since it should be covered by a page explaining how to load cogs, and I think it may make sense to only build this level of detailed documentation for cogs and commands which are not expected to be receiving modifications for a while to prevent it from becoming an extra step all the time (Not to never do it, but to wait on these until closer to release) (for instance, commands in the general cog are unlikely to change much currently, but things like mod and permissions are still getting improvements right now as needed)

laggron42:
I can remove the installation section and just write a getting started guide, like on V2

laundmo:
as someone looking into using red it would be nice to have at least a copy paste of the help messsage somewhere. almost every bot has a command list, and i think the standard cogs for this should too.

i dont want to have to install it just to figure out the exact commands it has. this wouldnt be a problem with a hosted public bot, but since this is selfhosted i would have to install just to test. i need to know if its worth it before that.

palmtree5:
Here's the current status of this issue:

| Cog | Done | In Progress | Not Started | 
|---|---|---|---|
| Admin | x |   |   |
| Alias | x |   |   |
| Audio |   | x |   |
| Bank | x |   |   |
| Cleanup | x |  |  |
| CogManagerUI | x |  |  |
| Core | x |  |  |
| CustomCommands | x |  |  |
| Dev | x |  |  |
| Downloader | x |  |  |
| Economy | x |  |  |
| Filter | x |  |  |
| General | x |  |  |
| Image | x |  |  |
| Mod | x |  |  |
| ModLog | x |  |  |
| Mutes | x |  |  |
| Permissions | x |  |  |
| Reports | x |  |  |
| Streams | x |  |  |
| Trivia | x |  |  |
| Warnings | x |  |  |

[...]

PR #1757, opened 2018-06-01
[V3] User documentation

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This PR will add user docs for red usage, such as detailed cog guides and more... We already discussed about that on #1734 

**This PR is not ready yet, I will commit new docs in the future.** I'm opening it now so you can see how it's going, just tell what you think about what I'm doing.

I created a quick ReadTheDocs page so you can directly see the result instead of looking at the source code. https://red-docspr.readthedocs.io/en/latest


mikeshardmind:
Overall, I think this is actually a really good start. I understand this is a work in progress, and subject to change.

A couple things I noticed:

* I'd advise against linking to external links and guides in the docs as much as possible as we can't guarantee they stay up to date or accurate. The Digital Ocean one is unlikely to be an issue as their documentation is generally kept up to date and is well written, but linking to a gist in docs is probably not the best idea.

* I would not be recommending Ubuntu16.04 as the default as the latest LTS release of Ubuntu is 18.04 which also includes python3.6 by default. Recommending any specific distribution over another is not necessarily something we should do as it starts being less a guide and more editorial, but if we are going to, we should be recommending the latest available long term supported version.

* I'd also advise against calling any linux distro "easy to understand". While they may be from a developer standpoint, this is a guide which targets a lot of users who have never used a CLI before, and while it may be easy for some people, it isn't at first for everyone, and this is likely to come across as off-putting to anyone who doesn't find it easy.

* There are a few grammatical errors which are quickly noticeable reading through which should be handled before merge, but I understand this is still a work in progress.

laggron42:
I'll comment what you said

- you're right with the gist, maybe I should include that into a hidden .rst (not in `index.rst`)

- Ubuntu 16 is not the latest but the most stable. I got issue and missing tools on 18, downgrading to 16 fixed all my problems. I think it's better to stay on a really stable release than going with the latest one.

- I defined Ubuntu as an easy-to-understand distro because of the pre-installed tools *and maybe because of bad memories with Arch*. This is the distro I usually recommand in #support, but I can change this to something else if needed, or just remove the "easy to understand" mention.

- I'm not native english, so I'll let you find and correct the errors. I'm trying to do my best, I promise D:

And thanks for the review, this is what I need

laggron42:
@tekulvw Made admin cog docs, as its author, what do you think about it?

Edit: ReadTheDocs didn't build, lemme fix that

laggron42:
ReadTheDocs built successfully

.gitignore
```diff
@@ -132,3 +132,7 @@ ENV/
 
 # pytest
 .pytest_cache/
+
+# docs build
+.build
+_build
```
Tobotimus: This is already added on line 122, so you can remove these lines. If you'd like to build docs for Red into an ignored folder, you can simply do it in the `docs` sub-directory, and the Makefile is provided for your convenience to build into this folder.

Tobotimus (commented):
I know it can be a pain, but for the checks to pass you must resolve all sphinx build warnings, even if the docs still build successfully otherwise.

.gitignore
laggron42: so instead of running `sphinx`, I need to run the makefile, right?

laggron42:
Added alias docs
https://red-docspr.readthedocs.io/en/latest/alias.html

laggron42:
screen shot 2018-06-03 at 18 27 56

Should the message `command locked to...` be shown as an admonition? (such as note, tip, warning...)
> Admonitions list http://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#directives

.gitignore
Tobotimus: You don't have to if you don't want to. You can do `sphinx-build . _build/html -bhtml` instead of `make html`.

Tobotimus: Also note that tox tests builds the docs using `sphinx-build` with the `-W` flag, which treats warnings as errors - hence why the travis build fails if there are any build warnings.

laggron42: I know that Travis is constantly failing because of missing references, I'll add them when they will exist, at least I won't forgot to add them.

However, this doesn't explain why the build shouldn't be in the gitignore file...

tekulvw: The build is in the gitignore file, you're just building the docs in the wrong folder. Please remove these lines.

tekulvw (dismissed):

tekulvw:
We appreciate what you're doing here Laggron and I realize that English is not your first language but in order for us to even consider changes to end user facing documentation, the spelling and grammar must be correct. If you're not going to work on that yourself then you need to find others to help you. If you don't, this PR is likely going to sit here for a very long time.

laggron42:
Don't worry, the work is in progress. I'll try to find all of these errors (I was going to finish every guide, then check for errors, but I'll search for them now). Thanks for the review!

laggron42:
Travis is not passing for weird reasons. At least tox pass locally
screen shot 2018-06-19 at 16 15 51

Zephyrkul:
If I may note here - a user-focused RST guide for the CustomCommands cog is included in #2051 since that PR introduces a major feature change that I couldn't document in docstrings alone. You can see that guide [here (clicky)](https://github.com/zephyrkul/Red-DiscordBot/blob/patch-1/docs/cog_customcom.rst).

I'd recommend you leave any docs on that cog out of this PR and instead contribute feedback regarding that cog's docs to PR #2051. That way they won't conflict and the docs will include the new feature.

Should my PR be denied the RST from there should still be useful for this PR, after stripping out the docs on the new feature.

laggron42:
The first RC is coming soon, and this have to be finished quickly. For those who want to help me, here is how I write cog guides.

```rst
.. _cog:
========
cog name
========

.. _cog-usage:
-----
Usage
-----

.. Here you should describe how the cog works
   If the cog is complex, the usage of screenshots and flowcharts is recommanded

.. _cog-commands:
--------
Commands
--------

.. In this category, you will explain how each command work in details.

.. _cog-command-command1:
^^^^^^^^
command1
^^^^^^^^

**Syntax**

.. code-block:: none

    .. here you will place syntax (or synopsis on unix) of the command, just paste what the help command provides
    [p]command1 <positional argument> [optional argument...]

**Description**

.. Here you will give a detailled description of the command.

**Arguments**

.. Explain all arguments of the command.

*    ``<posotional argument``: Some description...

*    ``[optional argument...]``: More description...

.. for group commands, do the following

.. _cog-command-command2:
^^^^^^^^
command2
^^^^^^^^

.. Fill the same description as in the above example

.. _cog-command-command2-yoursubcommand:
"""""""""""""""""""""""
command2 yoursubcommand
"""""""""""""""""""""""

.. Fill the same description as in the above example
```

If you want to attach images (which is recommanded), place them in the `.ressources` dir, and access them with `.. image:: ../.ressources/your-file.png`

You should check `audio.rst` as an example.

laggron42:
Due  to the amount of things to do, I'm gonna split this PR as suggested before on Discord. I'm gonna open one pull request right now for the "Getting started" part, and open one PR for each cog guide if you're ok with that.

Issue #1758, opened 2018-06-01
[V3 Permissions] 3rd party checks required modifications.


mikeshardmind:
With the current model for hooking third party checks into the permissions cog, we need permissions to be loaded first and not be reloaded. This was a design oversight on my part. 

I'll be postponing my other WIP stuff to resolve this in a satisfactory manner prior to b16, but this will require changes to how other checks which want to integrate into the permissions cog rather than operate separately need to function.

For anyone currently designing a 3rd party cog which uses these hooks: 
After the intended changes, any function which is an attribute of your cog should be possible to use as an integrated check. Each cog will get 1 before and 1 after check with expected names.

`__red_permissions_before` and `__red_permissions_after`
These can call your already existing functions if needed, and will receive the same arguments the current which current checks already do.

mikeshardmind:
After these changes, cogs can have the following methods (async optional) 

```py
async def __red_permissions_before(ctx: commands.Context, *, level: str)
```

```py
async def __red_permissions_after(ctx: commands.Context, *, level: str)
```
These methods should return `None` if not interfering with the normal check operation, `True` if it should run, `False` if it should not.

More detailed documentation for this will be available at a later date, but the original API was also undocumented.

tekulvw:
Implemented with #1760

PR #1760, opened 2018-06-01
[V3 Permissions] Don't rely on load order to be consistent

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Modifies permissions re #1758

This is technically a breaking change for any 3rd party cogs which were hooking into permissions. As this API was not fully documented yet, I doubt the use is widespread of it, but there should be some level of informing on this. This is however, a relatively easy change to accommodate as a cog creator if anyone was using it.

This is tested and working as intended.

### DIFF
--- a/redbot/cogs/permissions/resolvers.py
+++ b/redbot/cogs/permissions/resolvers.py
@@ -12,20 +12,10 @@ async def val_if_check_is_valid(*, ctx: commands.Context, check: object, level: 
     Returns the value from a check if it is valid
     """
 
-    # Non staticmethods should not be run without their parent
-    # class, even if the parent class did not deregister them
-    if check.__module__ is None:
-        pass
-    elif isinstance(check, types.FunctionType):
-        if (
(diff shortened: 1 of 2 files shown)


redbot/cogs/permissions/resolvers.py
```diff
@@ -12,20 +12,10 @@
...
-            next(filter(lambda x: check.__module__ == x.__module__, ctx.bot.cogs.values()), None)
-            is None
-        ):
-            return None
-
     val = None
     # let's not spam the console with improperly made 3rd party checks
     try:
+        # pretty sure this is redunant, but if it is, it will short circuit at or
         if asyncio.iscoroutine(check) or asyncio.iscoroutinefunction(check):
```
tekulvw: This should just be `iscoroutinefunction()`.

`iscoroutine()` checks if an object is a coroutine object.

tekulvw (changes requested):

tekulvw (approved):

Issue #1765, opened 2018-06-02
[V3] Setup error with invalid token


Eslyium:
# Other bugs

#### What were you trying to do?
Reset token during setup when token is invalid

#### What were you expecting to happen?
Be able to reset the token and enter a new one

#### What actually happened?
sys:1: RuntimeWarning: coroutine 'Value._get' was never awaited and then kicks you out of setup.

![image](https://user-images.githubusercontent.com/18387809/40869498-32289df0-65ea-11e8-8384-f1ff9f7eab3a.png)

#### How can we reproduce this issue?
Ubuntu 16.04
redbot-setup and put an invalid token it will ask if you want to reset it, respond with y.

tekulvw:
Fixed with #1771

PR #1771, opened 2018-06-02
[V3] fixes issue preventing token reset from setup

mikeshardmind:
### Type

- [X] Bugfix

fixes issue preventing token reset from setup see #1765

Also removes a faulty assumption about not needing cleanup tasks

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -139,7 +139,6 @@ def main():
         sys.exit(0)
     if tmp_data["enable_sentry"]:
         red.enable_sentry()
-    cleanup_tasks = True
     try:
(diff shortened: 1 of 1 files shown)


redbot/__main__.py
```diff
@@ -143,14 +143,13 @@ def main():
     try:
         loop.run_until_complete(red.start(token, bot=not cli_flags.not_bot))
     except discord.LoginFailure:
-        cleanup_tasks = False  # No login happened, no need for this
```
Tobotimus, line 146: This will now always be true, so it's a redundant condition - you can remove it

tekulvw (approved):

PR #1774, opened 2018-06-02
[V3 Admin] Fix permission problems

laggron42:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Any owner could interact with an other guild without permissions using [p]announce ignore. Now it checks if the member has permissions on the selected guild.

This is not tested but should work.

### DIFF
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -273,7 +273,7 @@ class Admin:
             guild = ctx.guild
         
         member = guild.get_member(ctx.author.id)    
-        if not (member.guild_permissions.administartor or \
+        if not (member.guild_permissions.administrator or \
             self.bot.is_owner(ctx.author)):
             await ctx.send("You are not allowed to do that on this guild!")
             return
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -271,6 +271,12 @@ class Admin:
         """
         if guild is None:
             guild = ctx.guild
+        
+        member = guild.get_member(ctx.author.id)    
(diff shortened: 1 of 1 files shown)


mikeshardmind:
There's no permission issue here. Something that should be changed is removing the guild argument though. There's no guild converter, so this will never resolve to a guild not the current one.

laggron42:
The `discord.Guild` argument is used. Currently, you just have to create a new server, invite the bot (so you're considered as a guild owner) and use the command while specifying a different guild (with name or ID; what the converter supports)

mikeshardmind:
Just because an argument is type-hinted, doesn't mean there's a supported converter. At the entry to the command callback, guild will **always** be `None` currently

You can test this behavior with the below snippet if you'd like:

`````
[p]eval
```py
from redbot.core import checks
class NoGuildConverter:
    
    @checks.is_owner()
    @commands.command()
    async def guildid(self, ctx, *, g: discord.Guild = None):
        if not g:
            return await ctx.send("No Guild")
        await ctx.send(f"{g.id}: {g.name}")

bot.add_cog(NoGuildConverter())
```
`````
then use `[p]guildid [ID, name, whatever you wanna try here]`, or you can just look and see that there isn't a converter for guilds.

laggron42:
Even if that's not supported atm, what says that discord.py won't bring an update for that? It's just dangerous to keep that in the code. Always `None` for now? Anyway there's never enough security check.

Twentysix26:
"Administrator" is misspelled in the code

tekulvw:
Closing in favor of #1781

PR #1779, opened 2018-06-02
[V3 Alias] Removed useless condition

laggron42:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

The condition `isinstance(ctx.invoked_subcommand, commands.Group)` is always `False`, so when a user type `[p]alias global`, nothing will be posted by the bot (= no help message).

### DIFF
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -182,7 +182,7 @@ class Alias:
         """
         Manage global aliases.
         """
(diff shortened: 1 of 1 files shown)


mikeshardmind:
This isn't a useless condition. Removing this breaks the help for subgroups which are not invoked with a subcommand.

laggron42:
Anyway, it's not working atm, no help message

mikeshardmind:
I'm aware, but these changes do not fix that. There's still no help using that either.

laggron42:
Fixed by #1783

PR #1780, opened 2018-06-02
[V3 RPC] Swap back to initial library and hook into core commands

tekulvw:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

RPC is now available with the `--rpc` flag and runs on `127.0.0.1:6133`. Parameters to methods must be passed through a json rpc client in the following format:

```py
params = {
    "args": [],
    "kwargs": {}
}
```

#### To add an RPC handler
In the setup function for your cog, call `bot.register_rpc_handler(method)`:

```py
def setup(bot):
    c = Cog(bot)
    bot.register_rpc_handler(c.some_method)
```

Keep in mind that all RPC handlers must only take JSON serializable parameters and must only return JSON serializable objects!


Kowlin (approved):

jchang358:
The current implementation doesn't delete rpc_handlers correctly. the bot will throw an error on cog unload and the rpc function will remain active. The problem is simple the handler name is always upper case but was overlooked in core/bot.py line 252. replace:

```python
del self.rpc_handlers[cogname]
```
with:

```python
del self.rpc_handlers[cogname.upper()]
```

PR #1783, opened 2018-06-02
[V3] subgroup help fix

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This fixes help for command subgroups which use an `isinstance` to determine they are the last thing in the chain. A more robust fix for this would require more extensive modifications to `redbot.core.commands`

### DIFF
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -182,7 +182,9 @@ class Alias:
         """
         Manage global aliases.
         """
-        if ctx.invoked_subcommand is None or isinstance(ctx.invoked_subcommand, commands.Group):
+        if ctx.invoked_subcommand is None or isinstance(
+            ctx.invoked_subcommand, discord.ext.commands.Group
+        ):
             await ctx.send_help()
 
     @checks.mod_or_permissions(manage_guild=True)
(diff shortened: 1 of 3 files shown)


mikeshardmind:
Alt to #1779 with correct fix for issue.

redbot/cogs/alias/alias.py
```diff
@@ -182,7 +182,9 @@ def get_extra_args_from_alias(
         """
         Manage global aliases.
         """
-        if ctx.invoked_subcommand is None or isinstance(ctx.invoked_subcommand, commands.Group):
+        if ctx.invoked_subcommand is None or isinstance(
+            ctx.invoked_subcommand, discord.ext.commands.Group
+        ):
```
Tobotimus, line 187: Instead of using `isinstance`, why not simply do an equality check with the group object itself? This technique works with multi-level subgroups, and I'd like to see it used more :smiley:

Tobotimus:
A true fix to this would be to add the decorator methods to `redbot.core.commands.Group()`, since in its current state, a `group.command()`/`group.group()` decorator won't return our subclassed objects.

Actually, I might open a new issue for that.

mikeshardmind:
Closing in favor of fixing as @tobotimus mentioned

Issue #1736, opened 2018-05-28
[V3 Fuzzy help] Option to disable


laggron42:
# Feature request

#### Select the type of feature you are requesting:

- [ ] Cog
- [ ] Command
- [ ] API functionality

#### Describe your requested feature

The fuzzy search was recently added. While it is actually a cool feature, there should be a way to disable it. It is pretty annoying for me on my server (like with random members who spam it).

If I say `...` in the chat (my prefix is `.`), the bot will send the fuzzy help. If I use a custom command, the bot also send the fuzzy help. Ok that one is a bug, but I still find it annoying on some servers, and we should be able to disable it, instead of editing and commenting a line...

laggron42:
Closing this since it is removed now.

Tobotimus:
@retke We keep issues updated in parallel with what's on V3/develop (or develop for V2). On V3/develop, fuzzy search is implemented but not toggleable. The toggle is included in #1788, however this issue shouldn't be closed until that is merged

laggron42:
I though the fuzzy help was going to be removed. I guess there's another issue to open now.

palmtree5:
nah, a release was done without fuzzy because of the problems it was causing (which wound up causing other issues, but that's a different story). It was never removed from V3/develop though as the intention was to fix it up for beta 16

palmtree5:
At any rate, this has been implemented in #1788

PR #1788, opened 2018-06-03
[V3 Fuzzy search] fix several issues with this feature

palmtree5:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

This PR fixes several issues with the fuzzy command search functionality released in beta 15
Namely:
- It can now be toggled (for the server by the server owner, and for DMs by the bot owner) (as requested in #1736)
- It now doesn't try to suggest anything when a valid custom command or alias is run
- It now checks that:
  - The user can run the suggested command and all parent commands of it
  - That the suggested command and all of its parents are not hidden
  - That the score for the match is >= 90
Non-alphanumeric characters were already being replaced with whitespace by the default processor, which was leading to results that had a score of 0. With the score requirement, these won't display anymore.

The one thing that still needs to be done is something about the `speedup` extra, which is technically optional but I feel should be used if at all possible. I still need to look into how I'd want to handle this part of it, so don't merge this quite yet

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -4,6 +4,7 @@ from distutils.errors import CCompilerError
 from pathlib import Path
 import re
 import importlib
+import tempfile
 
 import os
 import sys
@@ -26,8 +27,6 @@ def get_package_list():
 
 def check_compiler_available():
     m = ccompiler.new_compiler()
-    if os.name == "nt":
-        m.initialize()
 
     with tempfile.TemporaryDirectory() as tdir:
         with tempfile.NamedTemporaryFile(prefix="dummy", suffix=".c", dir=tdir) as tfile:
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,6 @@
 from distutils.core import setup
 from distutils import ccompiler
+from distutils.errors import CCompilerError
 from pathlib import Path
 import re
 import importlib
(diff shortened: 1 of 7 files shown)


palmtree5:
I believe I have figured out a way of detecting compiler availability, however this will need testing on all platforms (to the point of literally `pip install`ing this PR's branch).

Eslyium:
Messed around with it for a good 15/20 minutes on a Ubuntu 16.04 server, I can say this is a whole lot better than what it was before in terms of searching and matching up with a command the user meant no issues so far.

mikeshardmind:
This seems to meet the stated goals, and I have not found any issues with it currently

setup.py
```diff
@@ -21,6 +23,28 @@ def get_package_list():
...
+def check_compiler_available():
+    env_dirs = os.environ["PATH"]
+    comp_name = ccompiler.get_default_compiler()
+    comp_info = ccompiler.compiler_class[comp_name]
+
+    mod_name = ".".join(["distutils", comp_info[0]])
+    mod = importlib.import_module(mod_name)
+
+    comp_class = getattr(mod, comp_info[1])
+
```
mikeshardmind: This needs a check that the dict isn't empty

palmtree5:
So we've now changed how we're handling compiler detection by using `tempfile` to create a file called `dummy.c` which is then compiled using `distutils`. If we can't compile that file for some reason, we'd drop `fuzzywuzzy`'s extra

Kowlin (approved):

PR #1789, opened 2018-06-03
[V3 Core] Fix display of whitelist and blacklist members

rngesus-wept:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

`[p]whitelist list` and `[p]blacklist list` fail because strings don't have `.append`.


Tobotimus (approved):
Thanks very much for your contribution!

Issue #1785, opened 2018-06-03
[V3] Remove "ctx.invoked_subcommand is None" boilerplate


Tobotimus:
The vast majority of command groups send the help message when no subcommands are given - they do this with a statement like:
```py
if ctx.invoked_subcommand is None:
    await ctx.send_help()
```

This is really common and honestly, it's unnecessary boilerplate. It gets worse when we have sub-groups, with people often using `isinstance` checks to send help there. 

@mikeshardmind came up with the suggestion to add this to the `group()` decorator instead, but optionally disable it with a flag in case automatically sending help on no invoked subcommand wasn't desired.

We realised this would introduce a somewhat breaking change - all of a sudden, 3rd party cogs who hadn't removed the call to `ctx.send_help()` would be getting duplicate help being sent. For this reason, we've proposed a grace period where the flag will default to `False`, and will be set to `True` for beta 17.

tekulvw:
Fixed with #1790

PR #1790, opened 2018-06-03
[V3 Core] Command group help

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

This puts all the boilerplate for command help in the group decorator. This targets b16. 

This adds `autohelp` as a kwarg for the group decorator. On b16 this defaults to False to give 3rd party cogs time to adjust. For b17 this will be switched to defaulting to `True` as it is the more used case. (exactly one exception in current core code)

This flag is incompatible with `invoke_without_command` 

This is not a breaking change until the additions targeting b17

### DIFF
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -157,13 +157,12 @@ class Admin:
         else:
             await self.complain(ctx, USER_HIERARCHY_ISSUE)
 
-    @commands.group()
+    @commands.group(autohelp=True)
     @commands.guild_only()
     @checks.admin_or_permissions(manage_roles=True)
     async def editrole(self, ctx: commands.Context):
         """Edits roles settings"""
-        if ctx.invoked_subcommand is None:
-            await ctx.send_help()
+        pass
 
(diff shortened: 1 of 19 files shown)


mikeshardmind:
(discussion from #1785 )

tekulvw:
When the default becomes `True`, will this conflict with any existing help checks?

mikeshardmind:
Nope, still only gets called if it normally would, this just gives cog creators time to have a chance to decide if they want to opt out.

mikeshardmind:
I think I may have misunderstood your question. Using this will not conflict with someone sending help from a group command, but it may end up sending it twice if they are using `auohelp=True` on b16 (or after, and not using `autohelp=False`) and have not removed help calls from that group.

What this does is when autohelp is specified to a group decorator:
if there is no invoked subcommand, or the invoked subcommand is that group, and `invoke_without_command` is not used, send help.

This is a relatively common pattern in cogs, and even ones that do other things like display settings on no command can still work using this.

tekulvw (approved):

Kowlin (approved):

PR #1803, opened 2018-06-05
[V3] Root user warning

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This adds a clear warning if the user started Red as a sudoer or a root user. This was directly added to the `__init__.py` file.

### DIFF
--- a/redbot/__init__.py
+++ b/redbot/__init__.py
@@ -21,3 +21,12 @@ if sys.version_info < (3, 6, 0):
         " Please update your environment to Python 3.6 as soon as possible to avoid"
         " any interruptions after the beta 16 release."
     )
+    
+if os.geteuid() == 0:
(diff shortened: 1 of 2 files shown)


laggron42:
Why do you fail Travis, black says it's good for me
screen shot 2018-06-05 at 21 43 12

mikeshardmind:
We already mention the risks of running as root in the docs, and there are reasons to do this, including setting up docker containers which run red and wouldn't have any reason to be on an isolated user with proper container config. I'd prefer if this were in redbot-setup, instead of in init so it's only shown the once if we are really going to have this added in.

tekulvw:
I'm gonna agree, I don't think this is needed as it's fully explained in the docs. You shouldn't run as root but some people do have their reasons (like docker).

PR #1810, opened 2018-06-06
[V3 Launcher] Show version in main menu

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This will show a new line with the current version of Red in the launcher, under the INTRO.


tekulvw:
Why?

redbot/launcher.py
```diff
@@ -16,6 +16,7 @@
     create_backup,
     save_config,
 )
+from redbot.core import __verison__
```
Kowlin: Version is misspelled.

Kowlin (dismissed):

redbot/launcher.py
laggron42: Fixed. I tested my changes locally, then added the changes manually with github's online editor. I'm going to test my changes in a virtualenv, that'll be better.

laggron42:
Sounds good

screen shot 2018-06-09 at 11 19 06

And sorry @tekulvw I forgot to add a more detailed description to my PR: That feature is useful for users who want to update their bot, so they know what's the current version and if update is needed or not.

Maybe I can add something so it tells if the bot is outdated, like with the `[p]info` command. What do you think?

laggron42:
Okay added that feature. If you like it, tell me so I can commit it.

screen shot 2018-06-09 at 11 43 11

laggron42:
Well, nobody told me if I should commit it, so I'll do it since beta 16 is coming...

Tobotimus (approved):
Nice, thanks for your contribution!

If I were to give any feedback, it'd be that we don't really need to use aiohttp here since we're not in an asynchronous program - we could just use urllib. However I don't mind this as it is :smiley:

I tested this on Windows CMD, MinGW and bash console in Ubuntu.

PR #1817, opened 2018-06-07
[V3] Send meaningful responses on conversion failure

### LINKED ISSUE #967: [V3] Meaningful responses to bad command arguments
Right now, when a "Bad Argument" is passed into a command (i.e. when a `commands.Converter` fails), the bot simply responds with the help message. This doesn't really tell the user why the command failed, and I've seen many users be stumped by this.

I have two ideas for solutions:
1. Respond saying exactly what conversion failed, e.g. if a `commands.MemberConverter` fails, it responds with _Member "[user]" not found_
2. Make the help message contain better documentation for command arguments.

Disadvantages of each solution:
1. Relies on a meaningful message being passed into `commands.BadArgument` when a conversion fails, which brings up potential issues with custom converters (there are no guidelines on how `commands.BadArgument` should be used).
2. A more meaningful help message is still fairly ambiguous. The point of error messages is for the user to know exactly what went wrong, not just throwing a bunch of help docs in their face.

Tobotimus:
With this change, the bot can send a more meaningful message when converting an argument fails. Converters can do this by raising `commands.BadArgument` with its first argument being the message they would like to send to the user.

If a message isn't passed into the exception, the bot will simply send the help message. If the exception raised from the converter doesn't inherit from `commands.BadArgument`, the command will error out as usual, so we don't suppress any bugs.

Also note that this message will only be sent if the command is an instance of `redbot.core.commands.Command`.

P.S. sorry for so many changed files, but I went through and updated existing converters and also made sure all `discord.ext.commands` imports are actually now `redbot.core.commands`, since this was sometimes causing issues and inconsistencies with this change.

Resolves #967.

### DIFF
--- a/redbot/cogs/downloader/repo_manager.py
+++ b/redbot/cogs/downloader/repo_manager.py
@@ -2,17 +2,13 @@ import asyncio
 import functools
 import os
 import pkgutil
-import shutil
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from subprocess import run as sp_run, PIPE
 from sys import executable
 from typing import Tuple, MutableMapping, Union
 
-from discord.ext import commands
-
-from redbot.core import Config
-from redbot.core import data_manager
+from redbot.core import data_manager, commands
 from redbot.core.utils import safe_delete
 from .errors import *
 from .installable import Installable, InstallableType
@@ -232,7 +228,7 @@ class Repo(RepoJSONMixin):
         ----------
         branch : `str`, optional
             Override for repo's branch attribute.
-            
+
         Returns
         -------
         str
@@ -381,7 +377,7 @@ class Repo(RepoJSONMixin):
             Directory to install shared libraries to.
         libraries : `tuple` of `Installable`
             A subset of available libraries.
- 
+
         Returns
         -------
         bool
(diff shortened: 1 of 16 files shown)


redbot/cogs/downloader/repo_manager.py
```diff
@@ -2,17 +2,13 @@
...
-import shutil
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from subprocess import run as sp_run, PIPE
 from sys import executable
 from typing import Tuple, MutableMapping, Union
 
-from discord.ext import commands
-
-from redbot.core import Config
```
Tobotimus, line 14: JSYK the `Config` import here was unused

redbot/cogs/downloader/repo_manager.py
```diff
@@ -2,17 +2,13 @@
 import functools
 import os
 import pkgutil
-import shutil
```
Tobotimus, line 5: This import was unused

redbot/cogs/downloader/repo_manager.py
tekulvw: The commands import is used for type hinting, did you account for that?

Tobotimus: The commands import is replaced by `from redbot.core import commands`

tekulvw (approved):

PR #1818, opened 2018-06-07
[Commands] Refactor command and group decorators

Tobotimus:
Curiously, these decorators were previously placed in the `Command` class. Whilst it hadn't caused any issues yet, there's no reason for them to be there. This puts them in a new overridden `GroupMixin` class, so that the bot also inherits these decorators.

I also added some tests as a sanity check, and also to make sure this shit doesn't break in the future.


redbot/core/bot.py
```diff
@@ -24,7 +22,7 @@
 from .sentry import SentryManager
 
 
-class RedBase(BotBase, RPCMixin):
+class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):
```
tekulvw: Why call our own reference to `BotBase`? I don't see a whole lot of sense in doing this

Tobotimus: We can change it to `discord.ext.commands.bot.BotBase` if you'd like, it wouldn't make a difference

Tobotimus: I could instead subclass `discord.ext.commands.bot.BotBase` and add the `GroupMixin` there. It wouldn't be any different to the current setup but would simplify `RedBase`'s direct inheritance

Tobotimus: See [here](https://github.com/Cog-Creators/Red-DiscordBot/compare/V3/develop...Tobotimus:V3/refactor_cmd_decorators_2?expand=1#diff-b4e8893536462f5cd5d7e291288c4d5eR25) for an example of what I mean, I don't really think its necessary but all of this inheritance stuff is pretty confusing to some, perhaps this makes it a bit simpler? I'm not sure.

Tobotimus: Sorry, [this](https://github.com/Tobotimus/Red-DiscordBot/compare/V3/refactor_cmd_decorators...Tobotimus:V3/refactor_cmd_decorators_2?expand=1#diff-b4e8893536462f5cd5d7e291288c4d5eR25) is probably a better way to show the example

Kowlin:
Due to my inexperience I'm not gonna touch this in Beta 17. Todo for Beta 18

mikeshardmind (changes requested):
Everything in the bot appears to be fine with these changes, but I do think a tweak to the mro here should be made.

Here's the mro for commands.group:

```py
(<class 'redbot.core.commands.commands.Group'>, 
<class 'redbot.core.commands.commands.Command'>,
 <class 'redbot.core.commands.commands.GroupMixin'>, 

<class 'discord.ext.commands.core.Group'>,
<class 'discord.ext.commands.core.GroupMixin'>,
<class 'discord.ext.commands.core.Command'>,

<class 'object'>)
```

I'm very hesitant to have our explicit overrides have a different ordering here than discord.py's bases without extremely good reason for doing so, and that reason being documented for future maintaining. While it doesn't seem to matter currently, it seems a recipe for potential subtle, hard to debug breakage later.

calebj:
I've rebased and fixed the MRO in https://github.com/calebj/Red-DiscordBot/tree/v3/refactor_cmd_decorators

Tobotimus:
Thanks @calebj and @mikeshardmind, I've tweaked the MRO accordingly

calebj (approved):
I haven't noticed any issues so far while running this. Approving now that the MRO fix has been applied.

mikeshardmind (approved):

Kowlin (approved):

PR #1821, opened 2018-06-07
[V3 Downloader] Don't do 3rd party agreement without command args

### LINKED ISSUE #1819: [V3 Downloader] [p]repo add does 3rd party check without arguments
# Command bugs

#### Command name

[p]repo add

Downloader

The command should just show the help message, I shouldn't have to agree unless I'm actually adding a repository

#### What actually happened?

screenshot of output

Tobotimus:
### Type

- Bugfix

Resolves #1819.

This functionality is no longer contained in a check decorator because otherwise it gets called when it shouldn't.

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -15,7 +15,7 @@ from redbot.core.utils.chat_formatting import box, pagify
 from redbot.core import commands
 
 from redbot.core.bot import Red
-from .checks import install_agreement
+from .checks import do_install_agreement
 from .converters import InstalledCog
 from .errors import CloningError, ExistingGitRepo
 from .installable import Installable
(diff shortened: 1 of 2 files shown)


redbot/cogs/downloader/downloader.py
```diff
@@ -214,14 +214,16 @@ def __init__(self, bot: Red):
...
     async def _repo_add(self, ctx, name: str, repo_url: str, branch: str = None):
         """
         Add a new repo to Downloader.
 
         Name can only contain characters A-z, numbers and underscore
         Branch will default to master if not specified
         """
+        agreed = await do_install_agreement(ctx)
+        if not agreed:
+            return
```
tekulvw, line 226: Should probably give a notice here that the timeout period has been exceeded.

tekulvw (changes requested):

redbot/cogs/downloader/downloader.py
tekulvw: I'm an idiot, timeout notice happens elsewhere

tekulvw (approved):

PR #1823, opened 2018-06-07
[V3 Help formatter] better prefix handling

### LINKED ISSUE #1801: [V3 Help] Displaying the bot's mention as a prefix
@Kowlin has given a good example of how this is currently displayed here:
screenshot

Obviously this is ugly and undesired - I think in this circumstance the help formatter should detect that the prefix was a mention, and replace it with `@BotName` instead.

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Previously, we only properly escaped normal mentions, not nickname mentions which are handled differently 
This fixes #1801

### DIFF
--- a/redbot/core/help_formatter.py
+++ b/redbot/core/help_formatter.py
@@ -64,10 +64,7 @@ class Help(formatter.HelpFormatter):
     def clean_prefix(self):
         maybe_member = self.context.guild.me if self.context.guild else self.context.bot.user
         pretty = f"@{maybe_member.display_name}"
-        clean = str(self.context.prefix)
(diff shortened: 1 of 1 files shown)


mikeshardmind:
I've also made a PR to the upstream reason we don't handle that currently Rapptz/discord.py#1336 , but since we already are subclassing the helpformatter, we can use this until such time as there is an upstream fix.

tekulvw:
~~You added in a new property, where is it being used?~~ Ignore me.

tekulvw (approved):

PR #1829, opened 2018-06-09
[V3 Mod] QoL change for Userinfo

Kowlin:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

QoL change to userinfo. Hides the roles tab if none are found


mikeshardmind:
I like the changes, but I'd like to see one possible tweak to the docstring with this adding a note that some fields may not be shown if not applicable now that we have both this and the voice channel optional one.

Tobotimus:
Hmm, I think we will need to come to an agreement here. I personally prefer being explicit about the roles, i.e. how it was before.

Kowlin:
I do agree with Sinbad here, my PR should have included help text that explains that roles and names are hidden unless there's actually to show something.

Redjumpman:
I'm bumping this since, there hasn't been any activity or discussion about it since Jun 9th. Have we come to an agreement on what this should look like?

Tobotimus (approved):
I'm gonna merge this after adding clarification about omitted fields in the help docstring.

PR #1835, opened 2018-06-09
[V3] Tox using pipenv

Tobotimus:
there was once a tool called tox;
lots of people said that "it rocks!"
but then came pipenv,
who said "I'm a maglev"
and then I realised I can't write limericks.

So, this makes a few changes:
 - `tox` no longer installs stuff with `pip`, but when creating test environments, they simply inherit the site packages of whichever environment the user is in. Basically, the expectation is that the user has installed dev requirements to their venv with `pipenv` before running tox. This is the case on Travis, so we don't need to change anything there.
 - The `mongo` and `voice` extras are now included when installing via the `Pipfile` or `Pipfile.lock`.
 - ~~When building the `docs` test environment, Travis now caches the doctrees between builds. This should speed up subsequent `docs` builds a bit, since that seemed to always be the slow job.~~
 - ~~The `docs`, `test` and `style` extras have been removed from `setup.py`. Pipenv basically replaces any need for these, and we also no longer need to align the versions in here with the Pipfile.~~


tekulvw:
Can you further update Pipfile and Pipefile.lock?

Tobotimus:
I've reverted a few things I did in this PR:
 - I won't remove the `[docs,test,style]` extras yet, as they are used internally - this can be removed in another PR
 - I'm not caching doctrees between travis builds. Comparing test runs on Travis, it had a marginal speedup at best, as it probably takes just as long for sphinx to regenerate those files than it does for Travis to download them.

Tobotimus:
I'm kinda apprehensive on this PR. Neither of tox and pipenv are really designed to work with each other, this PR actually just removes the install stage from tox because it assumes pipenv has already installed all of the packages. Which kind of defeats some of the purpose of tox, to create isolated virtualenvs with dependencies used in deployment.

Tobotimus:
Not happy. Might reopen at a later date.

PR #1847, opened 2018-06-10
[V3 Core] More features for the bot invite URL

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This PR adds a new `[p]inviteset` group command that allows 3 things:

- Make the `[p]invite` command available for everyone, so they can invite the bot without asking its owner. However, if the "Public bot" option is not ticked in the token settings, the command will block the action and tell the user to modify it.

- Modify the permission level of the bot. The permission is given with the permission calculator (too complicated to use discord's method and iterate every permission) and it will be added to the bot's invite.

- The ability to add a redirect URL. It will be added to the bot's invite too, and will redirect to a website when the bot is invited.

**Warning**

This PR is using a new feature that I made, which allows the bot to use the `bot_public` endpoint (to know if the "public bot" option is enabled or not). A PR is open on discord.py's repo, so please wait for it to be merged before merging this.

The PR : https://github.com/Rapptz/discord.py/pull/1341

If you want to test that feature, do this in a virtualenv:

```
pip3 install -U --process-dependency-links git+https://github.com/retke/Red-DiscordBot@invite-additions
pip3 install -U git+https://github.com/retke/discord.py@patch-1
pip3 install yarl==0.18.0
pip3 install async-timeout==2.0.1
```


laggron42:
My pull request was just merged (actually rebased, but that's the same). This PR is not blocked anymore by discord.py.

redbot/core/core_commands.py
```diff
@@ -235,8 +236,18 @@ def _get_package_strings(self, packages: list, fmt: str, other: tuple = None):
...
         """
-        app_info = await self.bot.application_info()
-        return discord.utils.oauth_url(app_info.id)
+        if self.bot.user.bot:
+            app_info = await self.bot.application_info()
+            perms_int = await self.bot.db.invite_perm()
+            redirect = await self.bot.db.invite_redirect()
+            permissions = discord.Permissions(perms_int)
+            return discord.utils.oauth_url(app_info.id, permissions, redirect_uri=redirect)
+        return "Not a bot account!"
```
mikeshardmind: This special casing for userbots should be removed. We dont support user bots, they are against discord's TOS.

mikeshardmind (changes requested):

Kowlin:
Has conflict, please fix. Also, is the stuff in the rewrite branch of D.py yet?

laggron42:
@Kowlin yep, PR is rebased since august. I'll fix the conflicts.

redbot/core/core_commands.py
```diff
@@ -454,11 +464,89 @@ def get_bot_uptime(self, *, brief: bool = False):
...
+    async def perms(self, ctx, level: int):
+        """
+        Make the bot create its own role with permissions on join.
+
+        The bot will create its own role with the desired permissions\
+        when he join a new server. This is a special role that can't be\
+        deleted or removed from the bot.
+
+        For that, you need to give a valid permissions level.
+        You can generate one [here](https://discordapi.com/permissions.html).
```
mikeshardmind: I don't think this is appropriate to use embed specific formatting, as we specifically have a setting to disable embed use.

redbot/core/core_commands.py
```diff
@@ -454,11 +464,89 @@ def get_bot_uptime(self, *, brief: bool = False):
...
+
+        Please note that you might need the two factor authentification for\
+        some permissions.
+        """
+        await self.bot.db.invite_perm.set(level)
+        await ctx.send("The new permissions level has been set.")
+
+    @inviteset.command()
```
mikeshardmind: OAuth Redirects should not be user facing in the settings here in my opinion, as Red itself does nothing to support this, and this requires further settings properly configured on the developer panel to function.

Any developer who needs this, should have their code remove the core invite command and replace with their own. If this is beyond their capability, this setting is useless to them anyway.

mikeshardmind (changes requested):

redbot/core/core_commands.py
laggron42: This is something that can be useful for users and telling them to edit the bot is not a solution (especially with how hard it is to edit pip packages and keep them up-to-date).
I noticed that even when the bot is created for only one server, admins like when it creates his own role, as the other bots do, so they can directly customize permissions for each bot and drag them at the right place. (yes, that can be done with other roles, but they won't have the special status, making it assigned to the bot)

When I made this PR, the two main goals were making that command public and making it possible to edit the permissions integer, mostly what I had in mind. The `[p]inviteset` command is made to customize the invite link with all Discord's features, and the permissions interger is the main feature of an invite link, it would be weird to remove that...

mikeshardmind: I'm not saying to remove the permissions integer, but to remove the redirect. Without extra work, this will just appear to "break" the invite link for most end users. Even if you do go through half of that work, without handling the remaining portions, you get things like inviting the bot, followed by being sent to a link like:

https://www.google.com/search?q=how+do+I+use+discord?code=sdfgbhbgvcxcvgbgvfcdx&guild_id=42424242&permissions=0

(not a real code or ID)

Beyond this, nothing in red is enabled to use this, so it would have to be coded themselves. 

And even beyond that, the point about pip packages is moot, you do not have to edit the pip package to edit commands included with the bot. We have paths to load from, and the ability to remove and replace commands.

I think this is asking for a headache telling users over and over that they need a site designed to use this just to say "hey, our command does everything an invite can", without considering if it should.

mikeshardmind: Revisiting this: If you add a warning when attempting to add a redirect that this may break the invite if used improperly, it would be sufficient for "well, you didn't listen to the warning"

redbot/core/core_commands.py
```diff
@@ -457,11 +465,69 @@ def get_bot_uptime(self, *, brief: bool = False):
             await ctx.send("No exception has occurred yet")
 
     @commands.command()
-    @checks.is_owner()
-    async def invite(self, ctx: commands.Context):
+    @commands.check(CoreLogic._can_get_invite_url)
```
mikeshardmind, line 468: This isn't a valid for use as a check. Will result in a `TypeError` at runtime. (d.py only passes context, and there's no source of `self` for the instance method)

You can write a check which uses `ctx.cog._can_get_invite_url`, or refactor this to not be a check.

mikeshardmind (changes requested):

redbot/core/core_commands.py
```diff
@@ -457,11 +466,68 @@ def get_bot_uptime(self, *, brief: bool = False):
...
+        """Setup the bot's invite"""
+        pass
+
+    @inviteset.command()
+    async def public(self, ctx, confirm: bool = False):
+        """
+        Define if the command should be accessible\
+        for the average users.
+        """
+        if not self.bot.user.bot:
```
mikeshardmind: Red doesn't have selfbot support, this condition should be removed.

laggron42: Should the condition on line 268 also be removed? It was already there when I made the branch I think.

mikeshardmind: You should probably update your PR branch so you can more accurately view differences. That section isn't there on current code, and was removed prior to 3.0.0

mikeshardmind (approved):

PR #1849, opened 2018-06-10
[V3 Crowdin] dont upload tox/tests/etc

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/crowdin.yml
+++ b/crowdin.yml
@@ -1,5 +1,5 @@
 api_key_env: CROWDIN_API_KEY
 project_identifier_env: CROWDIN_PROJECT_ID
 files:
-  - source: /**/*.pot
+  - source: /redbot/**/*.pot
     translation: /%original_path%/%locale%.po


Kowlin (dismissed):

tekulvw:
I thought we were going to `cd` into the redbot directory? If we don't do that we'll have to coerce the existing translations we have on crowdin to work with the new folder structure.

mikeshardmind:
Maybe I'm misunderstanding you, but this shouldn't change the structure of the files at all, so unless something I can't see here is going on, this should have zero effect other than not touching files outside the redbot folder while retaining the existing structure.

Tobotimus:
This needs to make it into beta 17, @tekulvw are you able to provide a response just so I know it's ok?

Kowlin:
Gonna leave this till B18, need to do a bit more research into it.

mikeshardmind:
This was intended for b17, and we're now approaching b19. It's a small change which only changes the scope of included files to crowdin so that we are only updating thins we use in red, not the test. If this is no longer something we want to restrict despite earlier discussion about this, just close it.

Tobotimus (approved):
From what I can tell this will work fine, so I'm going to merge

PR #1855, opened 2018-06-11
[V3 JSON] Drivers deepcopy input/output data

### LINKED ISSUE #1853: [V3 JSON] Direct reference to mutable datastore being returned
# Other bugs
The JSON driver is returning direct references to the driver's cache. This is should not be happening as mutating the returned value will cause the mutation to also occur in the cache. This can be shown by the following:
```py
>>> dict1 = await self.conf.dict()
>>> dict1
{}
>>> dict1["somekey"] = "somevalue"
>>> await self.conf.dict()
{"somekey": "somevalue"}
```

IMO it's the driver's responsibility to return a copy from its `get()` method, not Config's.

Tobotimus:
### Type

- Bugfix

Resolves #1853 

I've added two tests in `tests/test_config.py`

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -335,7 +335,7 @@ class Group(Value):
                 default = poss_default
 
         try:
-            return deepcopy(await self.driver.get(*self.identifiers, *path))
+            return await self.driver.get(*self.identifiers, *path)
         except KeyError:
             if default is not ...:
                 return default
@@ -365,7 +365,7 @@ class Group(Value):
 
         """
         if not defaults:
-            defaults = deepcopy(self.defaults)
+            defaults = self.defaults
 
         for key, value in current.items():
             if isinstance(value, collections.Mapping):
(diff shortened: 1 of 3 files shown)


redbot/core/config.py
```diff
@@ -335,7 +335,7 @@ def get_attr(self, item: str):
                 default = poss_default
 
         try:
-            return deepcopy(await self.driver.get(*self.identifiers, *path))
+            return await self.driver.get(*self.identifiers, *path)
```
Tobotimus, line 338: Removed the deepcopy here because the driver should return a deepcopy

redbot/core/config.py
```diff
@@ -365,7 +365,7 @@ def nested_update(self, current, defaults=None):
 
         """
         if not defaults:
-            defaults = deepcopy(self.defaults)
+            defaults = self.defaults
```
Tobotimus, line 368: `Group.defaults` already returns deepcopy, so this was also redundant

tekulvw (approved):

PR #1861, opened 2018-06-12
[V3] Use our own checks instead of discord.py's

Tobotimus:
### Type

- Enhancement

### Description
The `@commands.has_permissions` checks don't integrate well with our permissions model, and they actually deny the owner to some commands as well.

I also removed any uses of `@commands.bot_has_permissions`, since these checks just ambiguously disable commands without telling the user why. Allowing the user to run the command, then explicitly requesting the permissions, is more user-friendly.


mikeshardmind:
Wouldn't these checks integrate just fine if we raised specific user-facing output for specific check failures, rather than blanket dropping user output on check failures currently?

Ed: I was referring to  `bot_has_permissions` , but the other the issue is clear.

I think we should consider just keeping the bot_has_permissions one and raise that specific check failure, rather than supressing it in our error handler. This avoids putting all the "I need x permission to do that" all over the code base where we want this behavior.

Kowlin (approved):

Tobotimus:
@mikeshardmind Sorry for not responding earlier. Yeah I have plans to implement that feature, I just wanted to remove the current behaviour for now.

PR #1865, opened 2018-06-12
[V3 Mod] Userinfo past nicks/names

mikeshardmind:
Prevents trying to do a string replace on a NoneType

### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/cogs/mod/mod.py
+++ b/redbot/cogs/mod/mod.py
@@ -1602,23 +1602,28 @@ class Mod:
             if entry.target == target:
                 return entry
 
-    async def on_member_update(self, before, after):
+    async def on_member_update(self, before: discord.Member, after: discord.Member):
         if before.name != after.name:
             async with self.settings.user(before).past_names() as name_list:
-                if after.nick in name_list:
+                while None in name_list:  # clean out null entries from a bug
+                    name_list.remove(None)
+                if after.name in name_list:
                     # Ensure order is maintained without duplicates occuring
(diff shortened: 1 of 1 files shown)


Twentysix26:
I was curious to see how a None would even end up in there, so I checked how the past names / nicknames tracking works. It seems that the past names assignation is wrong (it's adding the nickname instead of the name) and with this PR you're attempting to fix the symptom rather than the root issue
https://github.com/Cog-Creators/Red-DiscordBot/blob/0298b538031f1c922bf00fd29d8c12ee43faf1d2/redbot/cogs/mod/mod.py#L1606-L1613

palmtree5:
https://github.com/Cog-Creators/Red-DiscordBot/blob/1e60d1c2652ea432cfc645c93bcc35fa08afb441/redbot/cogs/mod/mod.py#L1362-1380
Really weird, it was fine at this commit which means there must've been a regression somewhere

mikeshardmind:
I agree fixing the cause is a good thing too, but the fact is we have people with `None` entries in there with this erroring for them. I'll update fixing that as well.

palmtree5:
Regression appears to have occurred in https://github.com/Cog-Creators/Red-DiscordBot/commit/15ea5440a3d51397b83697fcfcbc69fb4c3dc95d#diff-4bbe83517d8a95d2e6e0e46fe9f7b809

Kowlin (approved):

PR #1868, opened 2018-06-13
[V2 General] fixed lmgtfy when search term contains a plus

ToxicFrazzles:
A plus in an equation is interpreted as a space by https://lmgtfy.com which breaks the equation. Fixed by escaping plus characters in the search terms.

### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Before spaces are replaced with pluses replace pluses with "%2B" (the URL escaped plus) to allow search terms containing a plus such as equations to be properly processed by the web page linked to.

### DIFF
--- a/cogs/general.py
+++ b/cogs/general.py
@@ -154,7 +154,7 @@ class General:
     @commands.command()
     async def lmgtfy(self, *, search_terms : str):
         """Creates a lmgtfy link"""
-        search_terms = escape_mass_mentions(search_terms.replace(" ", "+"))
+        search_terms = escape_mass_mentions(search_terms.replace("+","%2B").replace(" ", "+"))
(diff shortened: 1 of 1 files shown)


CDWimmer:
Ran into this problem myself and this works to fix it, would recommend including for sure
Beats copying into notepad and ctrl-H replacing all the + with %2B before running the command

Redjumpman:
This fix does work, the same bug exists in the V3 version of lmgtfy. I'm going to open a PR for this in V3 too.

Redjumpman (approved):

Kowlin (approved):

Tobotimus (approved):

Tobotimus:
Thanks for the patch @ToxicFrazzles, we're sorry it took so long!

PR #1869, opened 2018-06-13
[V3 Downloader] Allow use of prefix in install message

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

With that feature, cog creators will be able to use the context prefix in their installation message, so they can use `{prefix}` which will become `!` or anything depending on the prefix used. It's better than using `[p]` anyway.

### DIFF
--- a/docs/framework_downloader.rst
+++ b/docs/framework_downloader.rst
@@ -22,7 +22,7 @@ Keys common to both repo and cog info.json (case sensitive)
   is installed or a repo is added
   
   .. tip:: You can use the ``{prefix}`` key in your string to use the prefix
-    used for installing.
+      used for installing.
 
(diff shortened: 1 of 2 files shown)


Tobotimus (changes requested):
You must remember whenever you do `str.format()` on a string which you don't know the contents of, it has the potential to raise a `KeyError` should the string contain anything else surrounded by `{}`.

It's always safer to use `str.replace()` for formatting user strings. I also think it's fine to use `[p]` instead of `{prefix}` - it's something we already replace in command help, and it's also being used by cogs currently, so it enhances the user experience without cog creators needing to make any changes :smiley:

Tobotimus (approved):
Thanks for your contribution :tada:

PR #1871, opened 2018-06-15
[V3 General] Changed urban dictionary for my embed version

lionirdeadman:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

Replaced the old urban dictionary command to an embed version I made which adds:
- The author name 
- Thumb ups
- Thumb downs
- Reaction menu to see the different definitions

### DIFF
--- a/redbot/cogs/general/general.py
+++ b/redbot/cogs/general/general.py
@@ -240,35 +240,53 @@ class General:
                 _(("No Urban dictionary entries were found or there was an error in the process"))
             )
 
-        if data.get("error") != 404 and data.get("result_type") == "exact":
+        if data.get("error") != 404:
 
-            # a list of embeds
-            embeds = []
-            for ud in data["list"]:
-                embed = discord.Embed()
-                embed.title = _("{} by {}".format(ud["word"].capitalize(), ud["author"]))
-                embed.url = ud["permalink"]
+            if await ctx.embed_requested():
+                # a list of embeds
+                embeds = []
+                for ud in data["list"]:
+                    embed = discord.Embed()
+                    embed.title = _("{} by {}".format(ud["word"].capitalize(), ud["author"]))
+                    embed.url = ud["permalink"]
 
-                description = "{} \n \n **Example : ** {}".format(
-                    ud["definition"], ud.get("example", "N/A")
-                )
-                if len(description) > 2048:
-                    description = "{}...".format(description[:2045])
-                embed.description = _(description)
+                    description = "{} \n \n **Example : ** {}".format(
+                        ud["definition"], ud.get("example", "N/A")
+                    )
+                    if len(description) > 2048:
+                        description = "{}...".format(description[:2045])
+                    embed.description = _(description)
 
-                embed.set_footer(
-                    text=_(
-                        "{} Down / {} Up , Powered by urban dictionary".format(
-                            ud["thumbs_down"], ud["thumbs_up"]
+                    embed.set_footer(
+                        text=_(
+                            "{} Down / {} Up , Powered by urban dictionary".format(
+                                ud["thumbs_down"], ud["thumbs_up"]
+                            )
                         )
                     )
-                )
-                embeds.append(embed)
+                    embeds.append(embed)
 
-            if embeds is not None and len(embeds) > 0:
-                await menu(
-                    ctx, pages=embeds, controls=DEFAULT_CONTROLS, message=None, page=0, timeout=30
-                )
+                if embeds is not None and len(embeds) > 0:
+                    await menu(
+                        ctx, pages=embeds, controls=DEFAULT_CONTROLS, message=None, page=0, timeout=30
+                    )
+            else:
+                messages = []
+                for ud in data["list"]:
+                    description = "{} \n \n **Example : ** {}".format(
+                        ud["definition"], ud.get("example", "N/A")
+                    )
+                    if len(description) > 2048:
+                        description = "{}...".format(description[:2045])
+                    description = _(description)
+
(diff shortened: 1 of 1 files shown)


lionirdeadman:
I'll update my branch (to include the changes made to V3/develop) after the code is reviewed if you don't mind. 

I don't really want to update this branch all the time while I wait for review, hope you understand.

redbot/cogs/general/general.py
```diff
@@ -224,49 +223,45 @@ def __init__(self):
             await ctx.send(_("I need the `Embed links` permission to send this."))
 
     @commands.command()
-    async def urban(self, ctx, *, search_terms: str, definition_number: int = 1):
-        """Urban Dictionary search
+    async def urban(self, ctx, *, word):
+        """Searches for urban dictionary entries using the unofficial urbandictionary api"""
```
Tobotimus: This help string is a bit long. Try to make the first line shorter than 72 characters, but the user should still know exactly what the command does from that line. Any further detail should be expanded after two line-breaks.

redbot/cogs/general/general.py
```diff
@@ -224,49 +223,45 @@ def __init__(self):
...
 
-        Definition number must be between 1 and 10"""
+        data = None
```
Tobotimus: Early declaration of variable here is unnecessary

redbot/cogs/general/general.py
```diff
@@ -224,49 +223,45 @@ def __init__(self):
...
-            else:
-                await ctx.send(_("Your search terms gave no results."))
-        except IndexError:
-            await ctx.send(_("There is no definition #{}").format(pos + 1))
+                async with session.get(url, headers=headers) as response:
+                    data = await response.json()
+
         except:
-            await ctx.send(_("Error."))
+            await ctx.send("No Urban dictionary were found or there was an error in the process")
```
Tobotimus: You'll want to return here

redbot/cogs/general/general.py
```diff
@@ -224,49 +223,45 @@ def __init__(self):
...
+
+            # a list of embeds
+            embeds = []
+            for ud in data["list"]:
+                embed = discord.Embed()
+                embed.title = ud["word"].capitalize() + " by " + ud["author"]
+                embed.url = ud["permalink"]
+                embed.description = (
+                    ud["definition"] + "\n \n **Example : **" + ud.get("example", "N/A")
+                )
```
Tobotimus: Please use `"".format()` instead of concatenating strings, and also add back in the i18n call (which is surrounding the string with `_()`).

redbot/cogs/general/general.py
Tobotimus: Also, this sentence isn't quite grammatically correct, and needs to be wrapped in `_()`

redbot/cogs/general/general.py
```diff
@@ -224,49 +223,45 @@ def __init__(self):
...
+                embed.set_footer(
+                    text=str(ud["thumbs_down"])
+                    + " Down / "
+                    + str(ud["thumbs_up"])
+                    + " Up , Powered by urban dictionary"
+                )
```
Tobotimus: Please use `"".format()` here too

Tobotimus (changes requested):
Hi, thanks for your contribution! :tada:

I like the new menu interface, much better than the old one! However there are a couple things which need to be changed in your code before we'll accept this :smiley:
 
Once you've made the changes, please post a comment to let us know.

lionirdeadman:
I think I've changed everything you asked for. Thanks for the review. :)

Could it be in Beta 17 or is it feature-freezed? :thinking:

lionirdeadman:
Had to come up with a fix for the 2048 character limit of descriptions, I hadn't thought of it.

Everything should be good now though.

lionirdeadman:
Just commenting again since this is supposed to be for B18 according to the tag so I'd like a review again. Thank you.

Kowlin:
I'm going to place this on hold to review with other core developers. See what they think.

mikeshardmind:
This is sleek, and I like it a lot. The only problem I have with it is that it does not respect the core embed requested setting. It might be better to break this out into 2 separate functions, (old code and new) and have the appropriate one based on that setting called when the command is invoked.

lionirdeadman:
I don't know how to respect the embed setting, if you could give me pointers, I'd be glad to do it, @mikeshardmind. 

Feel free to PM me in Discord instead of commenting here so we can do this faster! :smile:

mikeshardmind:
There's a helper method in [context ](https://github.com/Cog-Creators/Red-DiscordBot/blob/e3bff7e87cfca23dd36bf92507a99636d7033171/redbot/core/commands/context.py#L160)for this since this particular command always should respond in the same location it was called from.

```py
if await ctx.embed_requested():
    #  output version using an embed
else:
    #  output version without an embed
```

lionirdeadman:
I think I did it properly so this should be done now! (sorry for the wait, I was a bit busy)

mikeshardmind:
Looks solid. :+1:

Tobotimus (approved):
Alright, thanks @lionirdeadman for the PR and thanks @mikeshardmind for reviewing. This has really jazzed up the urban command.

A small word of advice @lionirdeadman for any future PR's, try not to change *too much* code unnecessarily. This command has essentially been rewritten from the ground up when looking at the diff - but in retrospect it only aimed to change the output format (not any HTTP request / backend stuff). Changing more than necessary gives the reviewers more work than necessary, hence why this has taken almost 2 months to be merged! However I'll let it slide and I hope you're glad this has finally had its green button pressed :)

lionirdeadman:
I had already made that code in a cog, I simply modified it slightly for this PR. I'll try not to do that again though if that makes things harder for you guys.

And yeah, I'm pretty happy it got merged. ^^

PR #1872, opened 2018-06-16
[V3] --token and --no-instance flags

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

This PR brings two new flags to `redbot`:

- The `--token` flag that will run Red with the given token instead of the saved one.
- The `--no-instance` flag that will run Red without any instance. The data will be saved under a `tmp` dir and removed on next restart.

These flags can be combined for testing on Travis CI for example : 

```
python3 -m redbot --no-instance --prefix ! --token $MY_PREFIX
```
That command will run Red without any setup, useful for online automated testing !

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -116,7 +116,7 @@ def main():
             "\033[0m"
         )
         cli_flags.instance_name = "temporary_red"
-        save_default_config()
+        create_temp_config()
     load_basic_configuration(cli_flags.instance_name)
     log, sentry_log = init_loggers(cli_flags)
     red = Red(cli_flags=cli_flags, description=description, pm_help=None)
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -6,7 +6,7 @@ import sys
 import discord
 from redbot.core.bot import Red, ExitCodes
 from redbot.core.cog_manager import CogManagerUI
-from redbot.core.data_manager import save_default_config, load_basic_configuration, config_file
+from redbot.core.data_manager import create_temp_config, load_basic_configuration, config_file
 from redbot.core.json_io import JsonIO
 from redbot.core.global_checks import init_global_checks
 from redbot.core.events import init_events
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -112,7 +112,7 @@ def main():
     if cli_flags.no_instance:
         print(
             "\033[1m"
-            "Warning: The data will be placed in a temporary folder and removed on next reboot."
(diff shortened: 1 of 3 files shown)


redbot/core/data_manager.py
```diff
@@ -39,6 +41,23 @@
 config_file = config_dir / "config.json"
 
 
+def save_default_config():
```
Tobotimus: I think this should be named `create_temp_config`, to be clearer about what it does.

redbot/core/cli.py
```diff
@@ -138,6 +138,16 @@ def parse_cli_flags(args):
...
         help="Enables the built-in RPC server. Please read the docs prior to enabling this!",
     )
+    parser.add_argument("--token", type=str, help="Run Red with the given token.")
+    parser.add_argument(
+        "--no-instance",
+        action="store_true",
+        help=(
+            "Run Red without any existing instance. "
+            "The data will be saved under a temporary folder "
+            "and deleted on next restart."
```
Tobotimus: "deleted on next restart" could be interpreted as the next bot restart. I would specify "system restart" to be clearer.

redbot/__main__.py
```diff
@@ -106,9 +106,17 @@ def main():
...
         print(description)
         sys.exit(0)
-    elif not cli_flags.instance_name:
+    elif not cli_flags.instance_name and not cli_flags.no_instance:
         print("Error: No instance name was provided!")
         sys.exit(1)
+    if cli_flags.no_instance:
+        print(
+            "\033[1m"
+            "Warning: The data will be placed in a temporary folder and removed on next reboot."
```
Tobotimus: Specify "system reboot" here too

Tobotimus (changes requested):

redbot/__main__.py
```diff
@@ -106,9 +106,17 @@ def main():
...
         print("Error: No instance name was provided!")
         sys.exit(1)
+    if cli_flags.no_instance:
+        print(
+            "\033[1m"
+            "Warning: The data will be placed in a temporary folder and removed on next system reboot."
+            "\033[0m"
+        )
+        cli_flags.instance_name = "temporary_red"
+        save_default_config()
```
Tobotimus: `save_default_config` call needs to be updated

Tobotimus (changes requested):

laggron42:
Fixed, sorry for that dumb error.

Tobotimus (approved):

Issue #1881, opened 2018-06-21
[V3 Help] Restricted command groups display commands


aikaterna:
#### Command name
Group commands that are core with checks for permissions, examples are [p]economyset, [p]cog, [p]repo

#### What cog is this command from?
Downloader and Economy

#### What were you expecting to happen?
Users that don't have permissions for these commands should not receive help when using the commands

#### What actually happened?
A user with no special permissions used [p]repo and the command group help appeared in chat, which shouldn't have been displayed for them. They aren't able to execute the group's subcommands that are shown, as expected.

#### How can we reproduce this issue?
Try !cog, !repo, or !economyset in Red's testing channel

mikeshardmind:
Issue introduced in #1790 working on a fix.

Tobotimus:
Resolved by #1882.

PR #1882, opened 2018-06-21
[V3] fix not verifying checks first

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

While commands would never be invoked, help was being sent for command groups prior to where checks were being processed. (see #1881 )


Tobotimus (approved):
Nice, verified it fixed the issue for me

PR #1885, opened 2018-06-22
[V3/permissions] Performance improvements

mikeshardmind:
### Type
- [X] Enhancement

This has a relatively basic caching layer for permissions. This should help alleviate some of the responsiveness issues with the help command, as well as a small logic improvement for the handling of the guild model in DMs

Things this needs to have implemented before merge

- [X] Basic Caching
- [X] Intelligent cache invalidation
- [x] Method for gathering a full set of commands which have rules set in permissions for a given context (Help formatter related)


mikeshardmind:
Updated this to have an upper size limit on the cache. This limit can probably be tuned upwards more than it is (more testing required still) but don't want this exploding on large servers.

mikeshardmind:
This is ready for testing by others. Does work, anything ambiguous in nature is commented, and there's a preamble in the mass resolver file explaining the purpose of it, and its caveats.

I'll be making an additional PR to have the help formatter use the new method if I have time too soon.

mikeshardmind:
Found an issue with the LRUDict, I'll be fixing it, but this isn't ready.

Kowlin (approved):

Issue #1888, opened 2018-06-23
[V3 Audio] Creating a playlist from queue can generate duplicate playlists


Crossedfall:
# Command bugs

#### Command name
`[p]playlist queue`

#### What cog is this command from?

Audio

#### What were you expecting to happen?

A single new playlist of <name> to be generated from the current queue.

#### What actually happened?

On_message wait allows for duplicate playlist generation if the `[p]playlist queue` command is ran again

#### How can we reproduce this issue?

1. Load music into the queue
2. Run `[p]playlist queue` without a name
3. Run `[p]playlist queue <name>` after receiving the, "Please enter a name for this playlist" message.

![playlist queue](https://user-images.githubusercontent.com/26130695/41805123-af515252-7668-11e8-8ac8-68aa60fce831.gif)

PR #1890, opened 2018-06-24
[V3 Audio] Fix for playlist queue duplicates

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fix for #1888 and playlist names are limited to one word.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -708,7 +708,7 @@ class Audio:
                 return await self._embed_msg(
                     ctx, "Playlist name already exists, try again with a different name."
                 )
-        playlist_name = playlist_name.split(" ")[0].strip("\"")
+        playlist_name = playlist_name.split(" ")[0].strip('"')
         playlist_list = self._to_json(ctx, None, None)
         playlists[playlist_name] = playlist_list
         await self._embed_msg(ctx, "Empty playlist {} created.".format(playlist_name))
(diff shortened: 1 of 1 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -794,11 +795,13 @@ def check(r, u):
...
 
             def check(m):
-                return m.author == ctx.author
+                return m.author == ctx.author and not m.content.startswith(ctx.prefix)
 
             try:
                 playlist_name_msg = await ctx.bot.wait_for("message", timeout=15.0, check=check)
-                playlist_name = str(playlist_name_msg.content)
+                playlist_name = re.sub(
+                    "[^a-zA-Z0-9]", "", str(playlist_name_msg.content.split(" ")[0])
```
Crossedfall: This will terminate the line at the first space even when contained in double quotes. This doesn't match up with how you can have a playlist name containing spaces if you don't utilize the `wait` feature.

e.g.

`[p]playlist queue "Test Playlist"`
Saves a playlist called, "Test Playlist"

`[p]playlist queue`
_prompt for the user to enter in a name_
`Test Playlist` or `"Test Playlist"`
New playlist called, "Test"

Tobotimus: ~~Why don't we get rid of the double quotes rule and make `playlist_name` consume rest?~~ Never mind me, that's not really related to your point

Kowlin (approved):

PR #1899, opened 2018-06-28
[V3] help formatter improvements

mikeshardmind:
### Type

- [X] Enhancement

This looks to process checks in the help formatter significantly more efficiently.

This depends on changes in #1885, #1963 

This is NOT ready and has several things which need changing since 1963

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -3,6 +3,7 @@ import contextlib
 import asyncio
 import discord
 from redbot.core import commands
+from redbot.core.bot import Red
 from redbot.core import checks
 from redbot.core.config import Config
 from redbot.core.i18n import Translator, cog_i18n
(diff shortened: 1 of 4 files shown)


mikeshardmind:
This isn't fully working. I'll fix it when I can, but there's a bug with which things get displayed to a user.

Kowlin:
Pushing this to Beta 18 then.

mikeshardmind:
Any remaining work on this will depend on a fix for #1931 being merged, depending on the route used to fix this, nothing more will need changing. As such, this is on hold.

mikeshardmind:
This likely is no longer needed as of #1885 I'd like to see how the help formatter is performing after #1963 as well just to be certain, but this increases the complexity of the help formatter significantly where it may no longer be needed.

Edit:

Still needed due to the special handling required to process permissions checks being dependent on a view. 1963 has the details on that.

mikeshardmind:
I won't be finishing this. This particular one most likely isn't the right approach to this problem, but there are various things in this which may be useful to someone else solving this later. Can be closed, but I'll leave it open for someone else to decide that.

mikeshardmind:
Solved alternatively with #2149

PR #1913, opened 2018-07-01
[V3 reports] Display user discrim

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This ensures less chance of accidentally mixing up users, still showing the nickname as well if there is one

### DIFF
--- a/redbot/cogs/reports/reports.py
+++ b/redbot/cogs/reports/reports.py
@@ -105,7 +105,7 @@ class Reports:
         *,
         mod: bool = False,
         permissions: Union[discord.Permissions, dict] = None,
-        prompt: str = ""
+        prompt: str = "",
     ):
         """
         discovers which of shared guilds between the bot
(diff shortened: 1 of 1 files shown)


redbot/cogs/reports/reports.py
```diff
@@ -175,7 +175,10 @@ def pred(m):
         if await self.bot.embed_requested(channel, author):
             em = discord.Embed(description=report)
             em.set_author(
-                name=_("Report from {0.display_name}").format(author), icon_url=author.avatar_url
+                name=_("Report from {author}{maybe_nick}").format(
```
Kowlin, line 178: A bit of a questionable formatting job here?

If I'm reading this code correctly this would make a username appear as the following:
``KowlinNickname``

mikeshardmind: Wiith how maybe_nick is passed to that, you'll either just get:

`Username#discrimantor` if no nickname, or `Username#discriminator (Nickname)` if they have one (making it more clear on nicknames, since those are more likely to change)

Kowlin: Ah right, this relied on me thinking that the author object would only return the username and not the full user-discrim.

Kowlin (approved):

PR #1922, opened 2018-07-08
[V2] Audio voice fix for channels with user limit

### LINKED ISSUE #1643: [V2] Audio voice
# Other bugs

Trying to play music while bot is idle in a voice channel will eror out if that voice channel is full with the bot in the channel. 

For the bot to keep playing.

#### What actually happened?

Bot errors out with "Your voice channel is full."

1. Create a voice channel with 2 slots.
2. Join voice channel and [p]play rickroll
3. After bot has joined and song has finished or you can do [p]stop try to
4. Try to play again while it is you and bot in the VC and channel is 2/2
5. Error

Jackenmen:
# Bugfix request

When the bot is idle in full channel with user limit, when you were trying to play music with command `[p]play (name of song)`, it was saying "Your voice channel is full."
Fixes #1643

#### Anything we need to know about this fix?
No, all I did was just adding one if (well changing place of it technically). You used such solution in another function already even, it looks like you forgot to do this here as well.

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -1491,25 +1491,24 @@ class Audio:
 
         # Checking already connected, will join if not
 
-        try:
-            self.has_connect_perm(author, server)
-        except AuthorNotConnected:
-            await self.bot.say("You must join a voice channel before I can"
-                               " play anything.")
-            return
-        except UnauthorizedConnect:
-            await self.bot.say("I don't have permissions to join your"
-                               " voice channel.")
-            return
-        except UnauthorizedSpeak:
-            await self.bot.say("I don't have permissions to speak in your"
-                               " voice channel.")
-            return
-        except ChannelUserLimit:
-            await self.bot.say("Your voice channel is full.")
-            return
-
         if not self.voice_connected(server):
+            try:
+                self.has_connect_perm(author, server)
+            except AuthorNotConnected:
+                await self.bot.say("You must join a voice channel before I can"
+                                   " play anything.")
+                return
+            except UnauthorizedConnect:
+                await self.bot.say("I don't have permissions to join your"
+                                   " voice channel.")
+                return
+            except UnauthorizedSpeak:
+                await self.bot.say("I don't have permissions to speak in your"
+                                   " voice channel.")
+                return
+            except ChannelUserLimit:
+                await self.bot.say("Your voice channel is full.")
+                return
             await self._join_voice_channel(voice_channel)
         else:  # We are connected but not to the right channel
             if self.voice_client(server).channel != voice_channel:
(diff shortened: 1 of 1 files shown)


cogs/audio.py
```diff
@@ -1491,26 +1491,26 @@ def _clean_url(self, url):
...
-            return
-        except UnauthorizedSpeak:
-            await self.bot.say("I don't have permissions to speak in your"
-                               " voice channel.")
-            return
-        except ChannelUserLimit:
-            await self.bot.say("Your voice channel is full.")
-            return
-
         if not self.voice_connected(server):
```
mikeshardmind, line 1494: This is not the correct condition, or at least not without additional conditions, as the else condition prior existing that you left in place would also pass this condition.

cogs/audio.py
```diff
@@ -1491,26 +1491,26 @@ def _clean_url(self, url):
...
-            await self._join_voice_channel(voice_channel)
+            try:
+                self.has_connect_perm(author, server)
```
mikeshardmind, line 1496: This line right here (well, specifically the full channel comparison in the function call (https://github.com/Cog-Creators/Red-DiscordBot/blob/97b70c35ee540710d7172e4fa7f6ac3663a979ea/cogs/audio.py#L2246) is the root cause of the issue, as there is no offset if the bot is in the channel, nor an additional pass condition of the bot is in the channel already with speak perms.

cogs/audio.py
```diff
@@ -1491,26 +1491,26 @@ def _clean_url(self, url):
...
+            except UnauthorizedSpeak:
+                await self.bot.say("I don't have permissions to speak in your"
+                                   " voice channel.")
+                return
+            except ChannelUserLimit:
+                await self.bot.say("Your voice channel is full.")
+                return
+            else:
+                await self._join_voice_channel(voice_channel)
         else:  # We are connected but not to the right channel
```
mikeshardmind, line 1514: Having the else block handled like this means the above permission checking never gets called if you want to bot to switch channels.

mikeshardmind (changes requested):
I think it makes sense to either modify `has_connect_perms` to account for the case of already being in the channel with voice perms (As an admin even join after the bot surpassing the channel limit, making calculating an offest wrong too)

or to modify handling of 

```
except ChannelUserLimit:
```
and handle it there.

I've detailed why the current proposed solution is not quite enough in comments attached.

Jackenmen:
You're right, though I didn't change what was already there. I just changed place of the code - try clause is now in if, like in other places of this code.

Of course this can be fixed in another pull request, I don't think I should do changes not relevant to the issue I mentioned.

I can try to help with that though, because you're right, that this should check for connecting perms in else clause as well.

PR #1924, opened 2018-07-08
[V2 Audio] Fixes channel full error while bot is in channel

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Alt to #1922 for fixing #1643

### DIFF
--- a/cogs/audio.py
+++ b/cogs/audio.py
@@ -2239,19 +2239,20 @@ class Audio:
         channel = author.voice_channel
 
         if channel:
+	    in_channel = bool(server.me in channel.voice_members)
             is_admin = channel.permissions_for(server.me).administrator
             if channel.user_limit == 0:
                 is_full = False
             else:
                 is_full = len(channel.voice_members) >= channel.user_limit
-
+	
         if channel is None:
             raise AuthorNotConnected
-        elif channel.permissions_for(server.me).connect is False:
+        elif channel.permissions_for(server.me).connect is False and not in_channel:
             raise UnauthorizedConnect
         elif channel.permissions_for(server.me).speak is False:
             raise UnauthorizedSpeak
-        elif is_full and not is_admin and server.me not in channel.voice_members:
+        elif is_full and not is_admin and not in_channel:
             raise ChannelUserLimit
         else:
             return True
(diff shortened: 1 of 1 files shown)


Jackenmen (commented):
This is generally good, but check line 1404 and 1739 for example - we're checking twice for the same thing. The only "problematic one" is this in line 1495, when we don't do that. This function is supposed to check, if bot can join the channel, there's no point in checking it, when you're already connected to channel.

PR #1922 also fixes another thing - before, we were checking for permission to connect to channel, even if bot was connected to right channel already. So with that PR,  when administrator moves the bot to channel, to which normally bot can't join, it doesn't check, if it has permissions for connecting, because it's already connected to that channel.

Another problem is that when we use has_connect_perm only, when bot is not connected to channel, we don't check, if bot can speak in channel, if he is already there.

So all codes, the original, your and mine for checking permissions aren't exactly good, each has downsides, we should change more in all those fragments and fix lack of necessary checks.

mikeshardmind:
> This is generally good, but check line 1404 and 1739 for example - we're checking twice for the same thing. The only "problematic one" is this in line 1495, when we don't do that. This function is supposed to check, if bot can join the channel, there's no point in checking it, when you're already connected to channel.
>
> PR #1922 also fixes another thing - before, we were checking for permission to connect to channel, even if bot was connected to right channel already. So with that PR, when administrator moves the bot to channel, to which normally bot can't join, it doesn't check, if it has permissions for connecting, because it's already connected to that channel.
>
> Another problem is that when we use has_connect_perm only, when bot is not connected to channel, we don't check, if bot can speak in channel, if he is already there.
>
>So all codes, the original, your and mine for checking permissions aren't exactly good, each has downsides, we should change more in all those fragments and fix lack of necessary checks.

If you'd like to argue for your fix, feel free. At the end of the day, this both fixes the observed problem, and the underlying design one, rather than trying to work around it in the command which uses it.

You are correct that by my logic, the same should apply for the connection check, so I have added it there as well.

Kowlin (approved):
Should work. If it doesn't, hunt me down and blame me.

PR #1934, opened 2018-07-14
[v2 bat] Attempt to use py -3.5 rather than py -3

mikeshardmind:
### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

This really shouldn't attempt latest 3.x since on v2 of red, we only support 3.5.x on windows.


Tobotimus:
This will force all users who are happily using python 3.6 on late Windows 10 versions to downgrade to 3.5. Not worth it IMO, even if we recommend 3.5 in the docs

PR #1935, opened 2018-07-14
[V3] use uvloop if available

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

uvloop is a significantly faster event loop, and we should be able to just use it where available. Being a drop in replacement, simply checking availability, and using it if is available seems safe.

Note:
I've been using this particular modification on my own bots without issue prior to making this PR

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -19,6 +19,14 @@ import logging.handlers
 import logging
 import os
 
+# Let's not force this dependency, but damn is uvloop fast.
+try:
+    import uvloop
+except ImportError:
+    pass
+else:
+    asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
+
 
 #
 #               Red - Discord Bot v3
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -19,13 +19,14 @@ import logging.handlers
 import logging
 import os
 
-# Let's not force this dependency, but damn is uvloop fast.
-try:
-    import uvloop
-except ImportError:
-    pass
-else:
-    asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
+# Let's not force this dependency, uvloop is much faster on cpython
+if sys.implementation.name == "cpython":
+    try:
+        import uvloop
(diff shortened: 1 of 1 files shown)


mikeshardmind:
Checks are failing on docs here for something not modified by this PR

https://travis-ci.org/Cog-Creators/Red-DiscordBot/jobs/403994298#L771

```
Warning, treated as error:
/home/travis/build/Cog-Creators/Red-DiscordBot/docs/install_windows.rst.rst:20:broken link: https://java.com/en/download/manual.jsp (403 Client Error: Forbidden for url: https://java.com/en/download/manual.jsp)
```

calebj:
I recommend only doing this automatically when the cython interpreter is detected, since uvloop is slow and sometimes problematic on pypy. Granted, pypy hasn't caught up with 3.6 yet, but it's only a matter of time.

mikeshardmind:
Red doesn't run on pypy as you've already pointed out due to compatibility issues. I also don't feel this is a needed step as this will only work if the uvloop package is installed. If someone is specifically opting for pypy despite our docs making no mention of it, I generally assume they also have enough knowledge to not install uvloop on it, and uvloop is not being provided in any extras.

If even with this consideration, you still think this is an issue, I can add the check, but I'd prefer not to. There have been discussions of uvloop being a CFFI extension instead of C specifically for cases like pypy (while retaining the C one for CPython if it were faster) over at magicstack's issue tracker. While nothing has come of this yet, it would become one more thing for the development team to continue tracking to see if the check should be modified.

calebj:
True. I've subbed the issue in question, but I'm still leaning towards putting in the check. My reasoning is that, because pypy shares the site package directories with cpython (on my system's build, anyway), there would be no way to exclude uvloop from being picked up if/when pypy 3.6 is released. It's better to err on the side of vanilla speed than to risk a potential slowdown--at least, until a CFFI implementation is released.

mikeshardmind:
Fair enough. I'll alter the PR to include this.

Kowlin (approved):

Issue #1941, opened 2018-07-16
[V3 Trivia] Users can mention `@here` and `@everyone` through a trivia session.


FixedThink:
# Command bugs

#### Command name

`[p]trivia` (found it with this one, but applies to other commands as well)

#### What cog is this command from?

Trivia (found it with this one, but applies to other commands as well)

#### What were you expecting to happen?

When a user answers a question correctly, either their username is mentioned, or their displayed name with mass mentioned escaped.

#### What actually happened?

The bot pings `@here` and `@everyone` when someone plays trivia, has one of those handles in their name, and answers a question correctly.

[Here is an image showing what happens.](https://i.imgur.com/KLsjnJk.png)

#### How can we reproduce this issue?

1.  Change your nickname to something containing `@here` or `@everyone`
2. Start a trivia session (e.g. `[p]trivia 2016`)
3. Answer a question correctly.

The line in question that causes the issue for trivia: https://github.com/Cog-Creators/Red-DiscordBot/blame/V3/develop/redbot/cogs/trivia/session.py#L194

And this commit (https://github.com/Cog-Creators/Red-DiscordBot/commit/b88b5a2601f56bda985729352d24842f087a8ade) shows that there might be more commands with this issue (although most are from `admin.py`.

mikeshardmind:
*sigh* I'm inclined to say that people should just pair this with the filter commands and auto-rename people with `@here` or `@everyone` in their names. That said, I'll look into a suitable way to fix this as this will literally impact any command that just drops a user's display name into output.

Tobotimus:
Resolved by #1942. Thanks for opening this issue @FixedThink!

PR #1942, opened 2018-07-17
[V3] ctx.send modifications and other output sanitization

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This modifies `ctx.send` to accept a filter, and provide a default filter which removes mass mentions

This also adds username sanitization to modlog's cases and mod's userinfo.

Also adds a convienience feature as `bot.send_filtered` (but more granular application of filtering is still advised) 

See #1941 and the #mod-log channel on discord around August 4th for real world cases on this. 

### Alternatives I've considered:

* Manually scrub most output
* modifications to `user.display_name` to automatically include such a replacement
* don't fix, and point people towards the display name filter in the builtin filter cog

### What this doesn't fix:

* Sending to a channel or user directly is not fixed here, I'm not patching `discord.abc.Messageable` unless discussion about overriding more of discord.py happens to prompt this.

(manually sanitizing using the added filters works for this)


mikeshardmind:
I've tagged this with a warning on not merging yet as well, this is very experimental, and while I'll run it on my own bot, I'm not confident in this being the best solution yet.

mikeshardmind:
I like the way this functions in practice enough that I've gone ahead and PR'd it upstream more robustly. I'd say using this here is safe under the same considerations mentioned in that PR (i.e. don't use this to redact accidental token leaks in things like eval, as that should be done in other ways to prevent user input guess & check)

We can use this solution safely until something like that is (if) merged upstream if we want

officiallyemma:
bump, I've added these changes in manually to my install but the majority of servers right now are still vulnerable. In a server of 10k people this really isn't something we can risk, and I see a lot of other server owners feeling the same way.

@Twentysix26 and other maintainers, please look into this.

Twentysix26:
@qwazwsx This PR is being considered, however, I'd argue that in a 10k people server the bot shouldn't have permissions to mass mention to begin with :)

mikeshardmind:
I strongly disagree that a 10k person bot shouldn't be able to have mass mention perms. There may be announcements which actually should require it in any community. While the opportunity for them is few and far between in most well-run communities, whether a bot having those permissions makes sense is ultimately up to those communities to decide.

Twentysix26:
I'm aware of legitimate uses for mass mentions on big servers, however I was attempting to point out that there are plenty of ways to prevent unwanted mass mentions, even without touching any code.  
For announcements, for example, is possible to give the bot mass mention permissions in one single channel (announcement channels are common), avoiding the risk of users triggering a mass mention in others.
@qwazwsx 's message sounded urgent, so I provided a ready solution.

mikeshardmind:
Not ready for merge. I'll be adding a regexes set in `redbot.core.utils.regexes` add a method named "send_filtered" to the bot, and modify several core things such as modlog to use it. (See incident in #mod-log channel for reasons as to why)

mikeshardmind:
I've done another pass over the various places we take in user input in any form, specifically for anything which could mass mention unintentionally. Searching for places we potentially take in links is a lot more intensive since even in embedded content, invite links need sanitization or discord will render an invite.  The good news is, mentions inside of embeds don't actually ping users. 

I consider the mass mentions the more important nuisance to kill (both would be ideal.) I feel this is ready enough for public use at this point as a solution, but I'm not sure I've covered everywhere this needs handling when it comes to filtering (specifically invite links from) embeds.

This can be 100% covered by automatically re-constructing all embed objects, but I felt this was too much a performance drain in practice on even a 200 user server that tried to help me break it. What it will come down to is applying selective filtering to content before placing it in an embed if from an untrusted source.

mikeshardmind:
A note on behavior in embeds. It seems at some point after the August 4th raiding incident, discord Changed the behavior of invites to not be rendered in embeds. They are still rendered when in code blocks, and in a modlog entry we probably shouldn't be advertising servers just because someone having a mod action done to them was named as an invite. This leaves verifying all of the other locations for this behavior at a much lower priority, but still able to be changed as needed at a later date.

Tobotimus (approved):

PR #1944, opened 2018-07-17
[V3 dev commands] doc string correction

### LINKED ISSUE #1884: [V3]  Dev commands and discord.ext.commands
Right now, debug/eval (and repl, maybe?) expose `discord.ext.commands` as `commands` Given that we subclass various things in this for our use in `redbot.core.commands`, shouldn't we be exposing our version? In either case, the other is still importable, but it seems our default should be to use our version.

mikeshardmind:
### Type

- [X] Bugfix

Corrects a documentation detail about what is passed to debug/eval

Resolves #1884 which is in itself somewhat wrong, as the desired behavior was already here, the docstring just wasn't changed to suit.

### DIFF
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -84,7 +84,7 @@ class Dev:
             author   - command author's member object
             message  - the command's message object
             discord  - discord.py library
-            commands - discord.py commands extension
+            commands - redbot.core.commands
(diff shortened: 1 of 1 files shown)


Kowlin (approved):

Kowlin:
Administrative merge: Travis CI failed due to docs issue, see #1957

PR #1956, opened 2018-07-24
[v2, multi] remove from __main__ import ... statements

calebj:
### Type

- [x] Bugfix

Removed problematic `from __main__ import ...` statements from core cogs and checks.py.

red.py and all applicable core cogs have been refactored to function without any reference to `__main__`. Cogs which rely on importing `user_allowed`, `send_cmd_help`, `set_cog` or `settings` from `__main__` will still continue to function.

### Rationale
The `from __main__` import in [checks.py](cogs/utils/checks.py) causes issues with [ProcessPoolExecutor](https://docs.python.org/3/library/concurrent.futures.html#processpoolexecutor)s configured to use the [`spawn` start method](https://docs.python.org/3/library/multiprocessing.html#contexts-and-start-methods), which is the only option on Windows since `fork` is only present on UNIX.

Because the spawned child processes are sent a pickled representation of the cog's module as a part of their task(s), the child process must re-execute the cog's module before the task can be invoked. Because the spawned processes don't have red.py as their base module, `__main__` is undefined, and the top-level import of `cogs.utils.checks` fails.

### DIFF
--- a/cogs/alias.py
+++ b/cogs/alias.py
@@ -2,7 +2,6 @@ from discord.ext import commands
 from .utils.chat_formatting import box
 from .utils.dataIO import dataIO
 from .utils import checks
-from __main__ import user_allowed, send_cmd_help
 from copy import copy
 import os
 import discord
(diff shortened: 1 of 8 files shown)


Tobotimus (approved):

calebj:
Bump!

Redjumpman (approved):

Issue #1931, opened 2018-07-12
[V3] Permissions nested group issues.


mikeshardmind:
Well, I hate finding I didn't properly account for this, but at least it was designed to fail securely.

The allow override for a command in a group does not implicitly grant an allowance of that group. This is a problem with how the permission model is advertised.

2 (ED: 1) possible routes for fixing:

1. Only commands can have `admin_or_permissions` style checks, not groups. (Ed: this breaks invokable group checks...)
2. Modify check handling done by the bot to account for our permission model.

Things which will not satisfactorily fix this:

1. Setting a subcommand implicitly allows the group.
  This is not okay, as we want the granularity, without needing to define each command individually, and would be insecure if any commands were added to the group after configuration.

2. Abusing reinvoke:
  This leaves the help-formatter in the dust (coincidentally, tracking weirdness on #1899 down is part of how I found this)

3. Moving checks from groups to their subcommands during add_cog:

    This breaks checks on invokable groups.

Additional notes:

Because of the ordering and current priority, the worst impact of this currently is blocking commands to someone you intended to allow them to. This will never incur the opposite.

mikeshardmind:
Fixed in #2149

Issue #1880, opened 2018-06-20
[V3 commands] commands.parents


mikeshardmind:
Currently, both `[p]permissions canrun` as well as command fuzzy search assume all command objects our ours. The logic for command parents needs to be moved somewhere else, or we need to find a way to enforce using our classes, as neither of these work properly when interacting with 3rd party cogs not using our commands override.

Tobotimus:
Enforcing the use of our command objects is gonna be the way to go here, just less headaches in general. Throwing an exception in `bot.add_command` is one way we could do it.

Tobotimus:
Resolved by #1971.

PR #1961, opened 2018-07-26
[V3 WIP] create and use a fork of discord.py

mikeshardmind:
### WIP

This is still a work in progress, but given the scope of changes proposed, putting it forward for discussion prior to completion provides more time for discussion.

### Why?

Several behaviors of discord.py are undesirable to how we use it, and discussion with the maintainer of that library of bringing changes which would not majorly change the overall behavior but help in the specific edge cases we were experiencing was not promising. While this was initially partially solved by using a custom commands module, there are several places we could benefit from underlying changes to this which would already require enforcing our overwritten commands module. In doing this via fork, we can maintain our changes in a much cleaner manner.

This significantly reduces the complexity of much of the existing code in the bot, as well as give us sane avenues for handling several currently open issues.

This also gives us a greater level of control over keeping discord.py compatible with the bot given the original lib's maintainer's stance on breaking changes.

### Plan to maintain the fork

The fork should track upstream changes manually to ensure we don't break anything in the process.
We should tag releases on this fork which correspond to the bot version it was tested for, and specifically use that tagged version. We should periodically update the bot version when the lib adds anything new which when we are able to accommodate it and verify it does not break expectations we have for behavior.

### What does this actually improve?

We'll get performance improvements on autohelp.

We'll be able to collect checks to run when a view is consumed, rather than run checks incrementally, allowing us to discard duplicate checks, and ensure full compatibility with our permission model, without requiring view modification from a check #1931

We won't need to reimplement certain things like Command.parents in a not OOP manner or enforce our command usage through type checking, or class modification to allow use in things like fuzzy search (example #1880)

We gain the ability to implement things in a cleaner fashion where otherwise we are butting heads with the lib in use in some cases.


laggron42:
I like the idea of using our fork of discord.py, especially because they're not responsive with their pull requests. 

I made a PR which adds some infos to the application info but it's not getting merged, while everything is correct (I did everything Danny asked to do, but, even with the recent breaking changes pushed, it's still pending...)
https://github.com/Rapptz/discord.py/pull/1341

Btw, this PR is needed for #1847

mikeshardmind:
Given the discussion about this in discord, I will leave this open until a final decision, but it's likely to be fixed by a combination of 3 other things instead, and won't receive any more work towards this solution unless that changes.

mikeshardmind:
Closing in favor of more suitable fix #1963

PR #1963, opened 2018-07-27
Fixes #1931 (Permissions not able to override subcommands)

### LINKED ISSUE #1931: [V3] Permissions nested group issues.

mikeshardmind:
Fixes #1931 (Permissions not able to override subcommands)
Supercedes: #1961

Additionally, #1899 needs a new implementation and inclusion for the help formatter to correctly hide things.

### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This adds an attribute to context allowing persistent storage of a fully consumed view object. This is needed because discord.py does not consume the view entirely at the point in time checks are run, and out model relies on knowing the command the user wants to run, not just each group, subgroup, and command in question individually.

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -96,6 +96,36 @@ class Permissions:
         if await ctx.bot.is_owner(ctx.author):
             return True
 
+        # This block below is a work around for d.py behavior
+        # The help formatter needs a work around as well as our checks
+        # rely on things not guaranteed to be in commands.Context
+        # at the time of check.
+        if ctx._internal_view_use is None:
+            # discord.py doesn't give checks access to the actual command
+            # a user attempted to invoke, This sidesteps that somewhat.
+            view = ctx.view
+            previous = view.index
+            ctx._internal_view_use = ctx.command
+
+            while True:
+                view.skip_ws()
+                check_next = view.get_word()
+                try:
+                    possible_command = next(
+                        filter(
+                            lambda c: check_next in c.aliases or c.name == check_next,
+                            ctx._internal_view_use.all_commands.values(),
+                        )
+                    )
+                except (AttributeError, StopIteration):
+                    break
+                else:
+                    ctx._internal_view_use = possible_command
+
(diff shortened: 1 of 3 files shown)


mikeshardmind:
After more testing, this avenue for fixing this breaks the helpformatter behavior of hiding unavailable commands, indicating the proposed changes in #1899 will need to be finished and included for this to be a valid fix.

mikeshardmind:
This still works as intended, but after discussion with @Tobotimus this should likely be handled in a redesign of the underlying mechanisms while leaving the user-facing interfaces as similar as possible. Leaving open for such a decision to be made, but will not continue with this myself from here.

Tobotimus:
Indeed, unfortunately I think a redesign is needed. I'm going to leave this open in case that doesn't eventuate, however. This could act as a hotfix, if the help formatter was not affected.

mikeshardmind:
@Tobotimus Have you had a chance to come up with an alternative method, or should I look deeper into the current ways this breaks the help formatter, and how to avoid that?

mikeshardmind:
Solved alternatively with #2149

PR #1967, opened 2018-08-01
[v3] Sentry: use raven-aiohttp

calebj:
### Type
- [x] Enhancement

Switch to the nonblocking [raven-aiohttp](https://github.com/getsentry/raven-aiohttp/) transport for Sentry.

Added the logout function to the Red class to hook any async shutdown code, such as graceful shutdown of Raven's aiohttp session.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -295,6 +295,13 @@ class Red(RedBase, discord.AutoShardedClient):
     You're welcome Caleb.
     """
 
+    async def logout(self):
+        """Logs out of Discord and closes all connections."""
+        if self._sentry_mgr:
+            await self._sentry_mgr.close()
+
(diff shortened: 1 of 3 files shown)


mikeshardmind:
I'm curious if this is actually needed. Sentry shouldn't block currently, since the [default transport ](https://docs.sentry.io/clients/python/transports/#threaded-default) for raven is a threaded approach

calebj:
aiohttp is the superior transport because of connection pooling/reuse alone, but in general, event-driven async IO is much faster than multiprocessing. I think that if the entire rest of the bot code is using async, Sentry should too, even if (hopefully) it isn't called often.

Tobotimus (approved):
Happy with these changes. Thanks!

PR #1971, opened 2018-08-01
[V3] enforce commands as ours

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature

Prevents cogs not using our command module from being added (#1880)

### Why in cog load
Rather than do this at the command level, it makes more sense to do at the cog level to prevent the cog from even being partially added.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -300,7 +300,8 @@ class RedBase(BotBase, RPCMixin):
                     " is not using Red's command module, and cannot be added. "
                     "If this is your cog, please use from `redbot.core import commands`"
                     "in place of `from discord.ext import commands`. For more details on "
-                    "this requirement, see the documentation."
+                    "this requirement, see this page: "
+                    "http://red-discordbot.readthedocs.io/en/v3-develop/framework_commands.html"
                 )
         super().add_cog(cog)
 
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -296,7 +296,8 @@ class RedBase(BotBase, RPCMixin):
                 _attr, commands.Command
             ):
                 raise RuntimeError(
-                    f"{cog.__module__} is not using Red's command module, and cannot be added."
+                    f"{cog.__class__.__name} from {cog.__module__},"
+                    " is not using Red's command module, and cannot be added."
                 )
         super().add_cog(cog)
 
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -296,8 +296,11 @@ class RedBase(BotBase, RPCMixin):
                 _attr, commands.Command
             ):
                 raise RuntimeError(
-                    f"{cog.__class__.__name} from {cog.__module__},"
-                    " is not using Red's command module, and cannot be added."
+                    f"The {cog.__class__.__name__} cog in the {cog.__module__} package,"
+                    " is not using Red's command module, and cannot be added. "
+                    "If this is your cog, please use from `redbot.core import commands`"
+                    "in place of `from discord.ext import commands`. For more details on "
+                    "this requirement, see the documentation."
                 )
         super().add_cog(cog)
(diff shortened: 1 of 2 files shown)


Jamesp1989SL:
Forgive me but why is it important to use your system and not the built in ?

Tobotimus:
@useruser3 It's true that this change (especially being a breaking one) deserves some more explanation. I've opened #1972 to shed some light here and perhaps open up some discussion if necessary.

redbot/core/bot.py
```diff
@@ -289,6 +289,16 @@ def unload_extension(self, name):
             if pkg_name.startswith("redbot.cogs"):
                 del sys.modules["redbot.cogs"].__dict__[name]
 
+    def add_cog(self, cog):
+        for attr in dir(cog):
+            if isinstance(discord.ext.commands.Command, attr) and not isinstance(
+                commands.Command, attr
+            ):
```
Tobotimus: This is just iterating over names of attributes, not the attributes themselves

Tobotimus (dismissed):

Redjumpman:
I think this is a good change. If anything, it helps curb the muscle memory of always using discord.ext. My recommendation is to include a blurb in the docs the reasoning behind why we subclassed `commands.Command`. While on this topic, it might be beneficial to have a section of all Red specific utilities that should be used over standard d.py.

mikeshardmind:
More on the last point by @redjumpman we have a cog cookiecutter repo which will be documented soon to help alleviate any adjustments like this for new cog creators as well

Redjumpman (changes requested):
Running this currently results in a `TypeError`

`TypeError: isinstance() arg 2 must be a type or tuple of types.` 

I tried calling `type()` on attr and the return of `getattr` and was able to get it to move past *that* error. I think because when iterating over dir, not everything is returned as a type. 

However, this does get the runtime error to raise, but I think unexpectedly for the one time we need it.

`RuntimeError: redbot.core.core_commands is not using Red's command module, and cannot be added.`

mikeshardmind:
This should be working all the time now, rather than only in the small toy test case I was using for it before.

redbot/core/bot.py
```diff
@@ -289,6 +289,18 @@ def unload_extension(self, name):
...
                 del sys.modules["redbot.cogs"].__dict__[name]
 
+    def add_cog(self, cog):
+        for attr in dir(cog):
+            _attr = getattr(cog, attr)
+            if isinstance(_attr, discord.ext.commands.Command) and not isinstance(
+                _attr, commands.Command
+            ):
+                raise RuntimeError(
+                    f"{cog.__class__.__name} from {cog.__module__},"
```
Redjumpman: `cog.__class__.__name` should be `cog.__class__.__name__` otherwise it raises an attribute error.

Redjumpman (changes requested):
After the change I marked is fixed, the message reports as `RuntimeError: RussianRoulette from russianroulette.russianroulette, is not using Red's command module, and cannot be add
ed.` I think this could be written a little more clearly as "The RussianRoulette cog in the russianroulette.russianroulette package must use from redbot.core import commands, instead of from discord.ext import commands." Something along those lines.

mikeshardmind:
@Redjumpman Sorry, I feel like an idiot after that last minute change to give more detailed errors. That's been fixed, and I've added a documentation link to the error, along with a slight tweak to the documentation making it clear that we have additional methods there for red specific things.

Redjumpman:
Awesome. Everything works great. This is now ready for the next release.

Redjumpman (approved):

Tobotimus:
Thanks for reviewing @Redjumpman, but considering this is a breaking change we haven't announced for yet, and beta 19 being due this Sunday August 12, I'm going to block this until after beta 19.

PR #1974, opened 2018-08-03
[V3] Major dependency update

Tobotimus:
Alright, so this is the major dependency update we've been waiting for. These are the details of new dependency versions:
 - discord.py pinned to commit Rapptz/discord.py@8ccb98d395537b1c9acc187e1647dfdd07bb831b
 - Other dependencies and sub-dependencies have *all* been pinned [as shown in the updated setup.py](https://github.com/Cog-Creators/Red-DiscordBot/pull/1974/files?utf8=%E2%9C%93&diff=unified#diff-2eeaed663bd0d25b7e608891384b7298R16)
 - The same has been done with sub-dependencies in the extras (including dev dependencies). I know we've talked a bit about changing how we're treating extras but that's part of the tox vs pipenv discussion.

### Other changes (these are up for discussion)
 - I removed requirements.txt and put our requirements into setup.py. This seems to be the way python libraries and applications are doing it these days, requirements.txt is being retired and replaced by things like Pipfile.
 - discord.py>=1.0.0a0 is listed as a dependency in the requirements and is removed when on `READTHEDOCS` (since it seems to cause issues there, but nowhere else). Previously, it was being *added* if we weren't on `READTHEDOCS`. This makes it clearer that it is in fact a requirement for installation and deployment.
 - `distro` is installed conditionally using [PEP 508 environment markers](https://www.python.org/dev/peps/pep-0508/#environment-markers), which have actually been around for a while.
- `setup_requires` has been removed - it's no longer required for the package to be installed correctly.
- Added a `__name__ == '__main__'` guard to setup.py, although it's not entirely necessary it might be useful one day to import the module from another one.
 - `python_version` has been **updated to `>=3.6,<3.8` to allow the use of python 3.7**. The bot itself, and all of its dependencies, are no longer incompatible with python 3.7. :tada:

### Accounting for breaking changes
This PR shouldn't bring any breaking changes to this library. However, there were breaking changes in aiohttp-json-rpc and discord.py which needed accounting for. We might need to keep an eye on how these have been dealt with (specifically RPC, sorry @tekulvw) in case something's been done incorrectly.


setup.py
```diff
@@ -97,45 +91,82 @@ def glob_locale_files(path: Path):
...
+        author_email="",
+        description="A highly customizable Discord bot",
+        classifiers=[
+            "Development Status :: 4 - Beta",
+            "Framework :: AsyncIO",
+            "Framework :: Pytest",
+            "Intended Audience :: Developers",
+            "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
+            "Operating System :: OS Independent",
+            "Programming Language :: Python :: 3.6",
```
mikeshardmind, line 121: Since this will allow 3.7 usage, this should also include a 3.7 identifier for pypi

Tobotimus:
I've added the 3.7 pypi classifier, and also added a 3.7 build to tox and travis. Unfortunately with Travis, they don't have an option to use v3.7.0 of python, the latest 3.7 one they have is 3.7-dev which is actually 3.7.0a4. This'll have to do for now though.

The build failures are coming from tox trying to install a non-existent version of red-lavalink: this'll be fixed once that release has been made.

calebj:
`ClientSession.close()` is async-only now, so these need to be moved to async or changed to `detach()`:
- [redbot/pytest/core.py:181](https://github.com/Tobotimus/Red-DiscordBot/blob/7881630671df581efef4f184e35484a7f3f69c2f/redbot/pytest/core.py#L181)
- [redbot/cogs/image/image.py:27](https://github.com/Tobotimus/Red-DiscordBot/blob/7881630671df581efef4f184e35484a7f3f69c2f/redbot/cogs/image/image.py#L27)
- [redbot/cogs/audio/audio.py:1929](https://github.com/Tobotimus/Red-DiscordBot/blob/7881630671df581efef4f184e35484a7f3f69c2f/redbot/cogs/audio/audio.py#L1929)

Redjumpman:
The only thing I found at the moment is the issue with `ClientSession.close()` as @calebj pointed out. I'll continue testing as this PR evolves.

Tobotimus:
The `aiohttp.ClientSession.close()` calls in sync code have been changed to `detach()` calls as requested

tox.ini
```diff
@@ -6,15 +6,15 @@
...
     docs
     style
 
 [testenv]
 description = Run unit tests with pytest
 extras = voice, test, mongo
 deps =
-    https://github.com/Rapptz/discord.py/archive/7eb918b19e3e60b56eb9039eb267f8f3477c5e17.zip#egg=discord.py[voice]
-    -rrequirements.txt
+    https://github.com/Rapptz/discord.py/archive/f06563cb3520058c900ba6d4921ce38f2407b729.zip#egg=discord.py
```
mikeshardmind: This wont work as a dependency link anymore. Something changed on github's redirection and this breaks now.

```
fatal: unable to update url base from redirection:
  asked for: https://github.com/Rapptz/discord.py/tarball/f06563cb3520058c900ba6d4921ce38f2407b729/info/refs?service=git-upload-pack
   redirect: https://codeload.github.com/Rapptz/discord.py/legacy.tar.gz/f06563cb3520058c900ba6d4921ce38f2407b729
```
Link format should no longer point to an archive, but a commit as part of a repo.
```
git+https://github.com/Rapptz/discord.py@f06563cb3520058c900ba6d4921ce38f2407b729#egg=discord.py1.0.0a
```
Or we need to adjust the archive link and hope it doesn't change again. I don't like relying on that, and the redirection behavior to remain the same.

Tobotimus: I don't seem to be able to reproduce the issue.

I'm not really sure if `git+` links work as dependency links for a package in setup.py. But even if they did, they require all users to have git installed when installing the bot.

mikeshardmind: I went to get full reproduction steps and can't anymore either. I'll keep an eye out for if it comes up again in some way later on at which point it can be addressed, but until I can reproduce it even semi-consistently, this can be ignored for now. 

Worth noting, git is a listed dependency in all our documentation and installation instructions due to downloader's use of it.

Redjumpman:
Any news on when we might see the websockets version bump via #36 on Red-Lavalink merged?

Tobotimus:
@Redjumpman Whenever we hold a gun to someone's head and tell them to press the big green button :grin:

Tobotimus:
This is no longer blocked!

I've gone and done some extra stuff now. Here are the details:
 - Travis is now building on the Xenial distribution so we can test with python 3.7.0
 - Travis no longer does a `pipenv install --dev` at the start, since the only dependency it needs is tox. Now it just installs and runs tox.
 - I've included a `dependency_links.txt` file and added it to the `MANIFEST.in` because we're starting to copy the discord.py commit hash all over the place. This serves as a single place to change the commit hash whenever we update our discord.py version, and is used by tox, setup.py and RTD.
 - I've removed docs/requirements.txt as it's more boilerplate which is no longer needed  - we can have RTD install the `docs` extra instead.
 - The `[dev-packages]` in Pipfile now just reference the extras `[test,docs,style]` since this reduces more boilerplate, and it seems to work fine.

dependency_links.txt
```diff
@@ -0,0 +1 @@
+https://github.com/Rapptz/discord.py/tarball/8ccb98d395537b1c9acc187e1647dfdd07bb831b#egg=discord.py
```
mikeshardmind: This needs to be listed as
```
https://github.com/Rapptz/discord.py/tarball/8ccb98d395537b1c9acc187e1647dfdd07bb831b#egg=discord.py-1.0.0a
```
or else this: https://gist.github.com/mikeshardmind/2c84190bc3d972e18b5566928a4627f6
happens

I'm also going to continue the prior recommendation on this being changed to the below, but the above is sufficient.
```
git+https://github.com/Rapptz/discord.py@8ccb98d395537b1c9acc187e1647dfdd07bb831b#egg=discord.py-1.0.0a
```

mikeshardmind (changes requested):
Beyond needing to specify the versioning in discord.py to get it to install without using a cached version, I haven't found any breakage on python3.6 or python3.7 in any of the core cogs. This has a :+1: from me pending that change.

dependency_links.txt
Tobotimus: Good catch @mikeshardmind, just fixed that up. Gonna stick to a tarball link for now though.

mikeshardmind (approved):
This looks ready to ship now.

calebj (approved):
I also agree that this is ready.

Redjumpman (approved):

Redjumpman:
+1 from me now.

Tobotimus:
Alright, this is being merged :tada: Thanks @Redjumpman @calebj and @mikeshardmind for reviewing :)

PR #1976, opened 2018-08-04
[V3 Bank] Raise TypeError when passing a non-int transaction amount

### LINKED ISSUE #1975: [V3 Bank] bank.withdraw/deposit_credits breaks mongodb with floats
# Other bugs

If you give bank.deposit_credits or bank.withdraw_credit a float instead of an int while using MongoDB, the users balance is stuck as a double. Because of this many cogs break because they expect an int, but get a float.

I was trying to withdraw credits from a users account with

`await bank.withdraw_credits(ctx.author, X * 0.30)`

although due to a bug I accidentaly ran

```
await bank.withdraw_credits(ctx.author, X * 0.30)
await bank.deposit_credits(ctx.author, X * 0.30)
````

This is most likely unrelated to the bug, but I thought I should mention it anyway.

I expected red to either throw an error or convert the float to an int 

#### What actually happened?

Red carried on, as if the value was an int. It changed the value type of the balance in mongodb from an int32 to a double. This breaks many other cogs because they expect an int but get a double.

1. run the following command

```
await bank.withdraw_credits(ctx.author, X * 0.30)
await bank.deposit_credits(ctx.author, X * 0.30)
```

2. done

### PR DESCRIPTION
**Pull request by Tobotimus**
Resolves #1975

Non-int transaction amounts don't really make any logical sense, so instead of naïvely putting them into Config, we should make sure the user is supplying the correct type.

### DIFF
--- a/redbot/core/bank.py
+++ b/redbot/core/bank.py
@@ -214,8 +214,12 @@ async def withdraw_credits(member: discord.Member, amount: int) -> int:
     ValueError
         If the withdrawal amount is invalid or if the account has insufficient
         funds.
+    TypeError
+        If the withdrawal amount is not an `int`.
 
     """
+    if not isinstance(amount, int):
+        raise TypeError("Withdrawal amount must be of type int, not {}".format(type(amount)))
     if _invalid_amount(amount):
         raise ValueError("Invalid withdrawal amount {} < 0".format(amount))
 
@@ -245,8 +249,12 @@ async def deposit_credits(member: discord.Member, amount: int) -> int:
     ------
     ValueError
         If the deposit amount is invalid.
+    TypeError
+        If the deposit amount is not an `int`.
 
     """
+    if not isinstance(amount, int):
+        raise TypeError("Deposit amount must be of type int, not {}".format(type(amount)))
     if _invalid_amount(amount):
         raise ValueError("Invalid deposit amount {} <= 0".format(amount))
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: redbot/core/bank.py [hunk 1]
```diff
@@ -214,8 +214,12 @@ def _invalid_amount(amount: int) -> bool:
     ValueError
         If the withdrawal amount is invalid or if the account has insufficient
         funds.
+    TypeError
+        If the withdrawal amount is not an `int`.
 
     """
+    if not isinstance(amount, int):
+        raise TypeError("Withdrawal amount must be of type int, not {}".format(type(amount)))
```
Redjumpman: Missing a period at the end of this statement.

redbot/core/bank.py
```diff
@@ -245,8 +249,12 @@ def _invalid_amount(amount: int) -> bool:
     ------
     ValueError
         If the deposit amount is invalid.
+    TypeError
+        If the deposit amount is not an `int`.
 
     """
+    if not isinstance(amount, int):
+        raise TypeError("Deposit amount must be of type int, not {}".format(type(amount)))
```
Redjumpman: Missing a period at the end of this statement.

redbot/core/bank.py
```diff
@@ -269,14 +277,18 @@ def _invalid_amount(amount: int) -> bool:
...
     Raises
     ------
     ValueError
         If the amount is invalid or if ``from_`` has insufficient funds.
+    TypeError
+        If the amount is not an `int`.
 
     """
+    if not isinstance(amount, int):
+        raise TypeError("Transfer amount must be of type int, not {}".format(type(amount)))
```
Redjumpman: Missing a period at the end of this statement.

Redjumpman:
I marked a few lines for the type error being raised, where it was missing a period. Overall this looks pretty good. I think having it raise a TypeError versus changing floats to an int internally is a better way to go. An internal conversion could round the number to a value the user was not expecting.

To remark on the previous issue that was started, floats do not really make sense for economy, and it really causes some weird issues with the leader board when you get some crazy long float.

Other than the punctuation, this PR is ready.

**Note:** Only adding my supporting rationale for this PR in case this is revisited in the future.

Tobotimus:
Thanks @Redjumpman, the requested changes have been made.

Redjumpman (approved):
This is ready to go for B19

PR #1979, opened 2018-08-04
[V3 Modlog] Make the case number optional

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Like on V2's modlog, the PR makes the case number optional and, if not provided, the latest case will be chosen.

### Examples

I have a modlog with 15 cases

```
[p]reason 13 This is my new magic reason
```
> This will edit the case number 13.

```
[p]reason I'm bored.
```
> This will edit the case number 15.

### DIFF
--- a/redbot/cogs/modlog/modlog.py
+++ b/redbot/cogs/modlog/modlog.py
@@ -99,12 +99,22 @@ class ModLog:
 
     @commands.command()
     @commands.guild_only()
-    async def reason(self, ctx: commands.Context, case: int, *, reason: str = ""):
+    async def reason(self, ctx: commands.Context, reason: str = ""):
         """Lets you specify a reason for mod-log's cases
+        
         Please note that you can only edit cases you are
-        the owner of unless you are a mod/admin or the server owner"""
+        the owner of unless you are a mod/admin or the server owner.
+        
+        If no number is specified, the latest case will be used."""
         author = ctx.author
         guild = ctx.guild
+        potential_case = reason.split()[0]
+        if potential_case.isnumeric():
+            case = int(potential_case)
+            reason.replace(potential_case, '')
+        else:
+            case = await modlog.get_next_case_number(guild) - 1
+            # latest case
         if not reason:
             await ctx.send_help()
             return
--- a/redbot/cogs/modlog/modlog.py
+++ b/redbot/cogs/modlog/modlog.py
@@ -97,7 +97,7 @@ class ModLog:
         else:
             await ctx.send(embed=await case.get_case_msg_content())
 
-    @commands.command()
+    @commands.command(usage="[case] <reason>")
     @commands.guild_only()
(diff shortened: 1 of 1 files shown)


redbot/cogs/modlog/modlog.py
```diff
@@ -99,15 +99,22 @@ def __init__(self, bot: Red):
...
     @commands.guild_only()
-    async def reason(self, ctx: commands.Context, case: int, *, reason: str = ""):
+    async def reason(self, ctx: commands.Context, *, reason: str):
         """Lets you specify a reason for mod-log's cases
+        
         Please note that you can only edit cases you are
-        the owner of unless you are a mod/admin or the server owner"""
+        the owner of unless you are a mod/admin or the server owner.
+        
+        If no number is specified, the latest case will be used."""
```
Tobotimus, line 108: Since we've modified the command signature, the command syntax will show in help as `[p]reason <reason>`.

You can actually specify a custom signature with the [usage](https://discordpy.readthedocs.io/en/rewrite/ext/commands/api.html#discord.ext.commands.Command.usage) argument to `@commands.command()`. I think this is the best way to make it clear how the command can be used :)

Tobotimus (changes requested):

Redjumpman (commented):
I'm ready to further test this PR, after the changes requested by @Tobotimus have been made.

redbot/cogs/modlog/modlog.py
laggron42: So the decorator would look like this:
```py
@commands.command(usage="reason [case] <reason>")
```

Tobotimus: For the arguments, yes, that looks good. You can omit the command name. Play around with it :)

laggron42: Didn't know that was a thing, thanks for the tip!

redbot/cogs/modlog/modlog.py
```diff
@@ -97,17 +97,24 @@ def __init__(self, bot: Red):
...
         author = ctx.author
         guild = ctx.guild
-        if not reason:
-            await ctx.send_help()
-            return
+        potential_case = reason.split()[0]
+        if potential_case.isnumeric():
```
Redjumpman: This should be using `isdigit()` not `isnumeric()`. The later of these two is used when you need to make extra considerations for unicode numeric-type values like super/sub script, ½, etc.

Redjumpman (dismissed):
Other than the requested change, this PR is working. I'll approve after this has been completed.

Redjumpman:
Just need one more change for this. Good job @retke

redbot/cogs/modlog/modlog.py
laggron42: Done 👌

Tobotimus (approved):

PR #1980, opened 2018-08-05
[V3 Cleanup] Refactor get_messages_for_deletion

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This begins a more long-term fix. Part of that is changing a function signature, and as such, steps are being taken to do this with a proper warning for 3rd party devs using this utility function.

This is targeting beta 20, but promises changes in beta 21, and beta 22 
PRs will be opened and connected to this.

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -77,7 +77,8 @@ class Cleanup:
             # Don't know how many, or even if 3rd party cogs are using this
             warnings.warn(
                 "Function signature for `get_messages_for_deletion`"
-                "set to become kwarg only in 3.0.0b21, see #1980 for details"
+                "set to lose parameters and become kwarg only in 3.0.0b21, "
+                "see #1980, #2006 for details"
             )
 
         two_weeks_ago = datetime.utcnow() - timedelta(days=14)
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -1,4 +1,7 @@
 import re
+import warnings
+from datetime import datetime, timedelta
+from typing import Union, List, Callable
 
 import discord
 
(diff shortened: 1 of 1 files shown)


aikaterna:
I've tried this change on a couple bots with no change in the bot's response, cleanup after still returns 0 messages deleted.

mikeshardmind:
This has been further changed to not break early, and filter the resultant list:

From the discord.py documentation:

```py
"""
...
    reverse: bool
        If set to true, return messages in oldest->newest order. Recommended
        when using with "after" queries with limit over 100, otherwise messages
        will be out of order.
...
"""
```
This also (potentially) explains the previously inconsistent behavior. We couldn't use `reverse=True` with the previous loop strategy and an early break, and `reverse=False` breaking on a message date isn't safe due to this either.

Tobotimus (changes requested):
This fixes `[p]cleanup after` but breaks all other cleanup commands.

You've also refactored `get_messages_for_deletion` which changes a lot of the behaviour unexpectedly. For example, the idea is to delete `number` of messages, whereas `limit` is the amount of messages to be collected at a time (confusing but we're just fixing one command here, we're not rewriting a cog)

Tobotimus:
Why not just not use `get_messages_for_deletion` in this command? Just get all messages after the given message (as long as the message is less than 14 days ago), remove pinned ones (if requested), and `delete_messages`. Otherwise this helper method is just getting too messy and complicated.

Given that this PR is fixing a currently obselete command, I will merge it into beta 19 post-cutoff (we already are post-cutoff) as long as it doesn't affect the behaviour of the rest of the cog.

mikeshardmind:
This shouldn't break any other cleanup command and the fact that it did was a fault of me not handling a specific issue with it. This should be fixed in this method, as it was made a static method specifically so that other things can use it too, we don't need to make a special casing for this, just need to handle it correctly.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -68,35 +72,55 @@ def author_check(message):
...
-                    and check(message)
-                    and (ctx.message.created_at - message.created_at).days < 14
-                    and (delete_pinned or not message.pinned)
-                ):
-                    to_delete.append(message)
-                elif (ctx.message.created_at - message.created_at).days >= 14:
-                    too_old = True
-                    break
-                elif number and len(to_delete) >= number:
+        two_weeks_ago = datetime.utcnow() - timedelta(days=14)
```
Tobotimus: I'd rather use `ctx.message.created_at` and do slightly less than 14 days (like 13 days, 23 hours and 55 minutes), since the system time could be out of sync with discord's, `created_at` should be correct since it uses the snowflake to calculate the time.

mikeshardmind:
Because it's a timedelta, being out of sync on the base time isnt a
concern, and I'd prefer removing ctx as we don't need it, so it opens up
for event based usage without mocking an object.

I agree making it slightly under 14 days makes sense.

mikeshardmind:
Closing favoring #2013

PR #1981, opened 2018-08-05
[V3 ModLog] Fix error in command [p]case

### LINKED ISSUE #1978: [V3 ModLog] Error in command "case"
# Command bugs

#### Command name

[p]case

ModLog

To show me information about a case.

#### What actually happened?
I get this error:
image

1. Have a modlog with cases.
2. Do [p]case

Tobotimus:
Resolves #1978, which was caused by trying to call a nonexistent method `Case.get_case_msg_content()` -> The correct method is `Case.message_content()`

Also sends as plaintext (not embed) with respect to the optional embed setting.

### DIFF
--- a/redbot/cogs/modlog/modlog.py
+++ b/redbot/cogs/modlog/modlog.py
@@ -95,7 +95,10 @@ class ModLog:
             await ctx.send(_("That case does not exist for that server"))
             return
         else:
-            await ctx.send(embed=await case.get_case_msg_content())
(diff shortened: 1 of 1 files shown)


aikaterna:
Works as expected with both display settings.

aikaterna (approved):
Works as expected with both display settings.

Redjumpman (approved):

PR #1983, opened 2018-08-05
[V3 CogManager] Correctly separate core, install and other cog paths

### LINKED ISSUE #1982: [V3 CogManager] Paths being added implicitly and incorrectly
# Other bugs

Paths (specifically the core cogs path and the cog install path) are being added to CogManager's config implicitly and incorrectly whenever `CogManager.add_path()` or `CogManager.remove_path()` is called. Basically, the methods get the existing list of paths with `CogManger.paths()`, which returns all valid path directories including the cog install path and the core cogs path. The new/old path is then added/removed from this list and the list is set again.

This is undesirable since:
 1. The install path will stay in the list of valid cog paths even after it's been changed
 2. The core cogs path should only be valid within the scope of a single *install*, not the scope of an *instance* (multiple installs can use one instance)

1. Use `[p]addpath` or `[p]removepath` once in any valid way
2. Use `[p]paths` and take note of the install path
3. Use `[p]installpath` to change the install path
4. Use `[p]paths` again and notice that the install path is still in the list of valid paths
5. Also notice that the core cogs path (located at red's install location) is listed as a valid path, even if you change the install location.

Tobotimus:
### Type

 - Bugfix

### Description
Resolves #1982.

This also modifies `[p]paths` a bit so it displays the core cogs path separately from the other paths.

I will admit I cleaned up a few other things in this PR, I will go through and comment the rationale on the individual changes.

Minor breaking change:
 - `CogManager.core_paths()` has been changed to a static class attribute `CogManager.CORE_PATH`. This shouldn't really affect anyone since it's for internal use.

### DIFF
--- a/redbot/core/cog_manager.py
+++ b/redbot/core/cog_manager.py
@@ -35,12 +35,13 @@ class CogManager:
     bot directory.
     """
 
+    CORE_PATH = Path(redbot.cogs.__path__[0])
+
     def __init__(self, paths: Tuple[str] = ()):
         self.conf = Config.get_conf(self, 2938473984732, True)
         tmp_cog_install_path = cog_data_path(self) / "cogs"
         tmp_cog_install_path.mkdir(parents=True, exist_ok=True)
         self.conf.register_global(paths=(), install_path=str(tmp_cog_install_path))
-        self.core_path = Path(redbot.cogs.__path__[0])
         self._paths = [Path(p) for p in paths]
 
     async def paths(self) -> Tuple[Path, ...]:
@@ -55,7 +56,7 @@ class CogManager:
         conf_paths = [Path(p) for p in await self.conf.paths()]
         other_paths = self._paths
 
-        all_paths = _deduplicate(list(conf_paths) + list(other_paths) + [self.core_path])
+        all_paths = _deduplicate(list(conf_paths) + list(other_paths) + [self.CORE_PATH])
 
         if self.install_path not in all_paths:
             all_paths.insert(0, await self.install_path())
@@ -150,7 +151,7 @@ class CogManager:
 
         if path == await self.install_path():
             raise ValueError("Cannot add the install path as an additional path.")
-        if path == self.core_path:
+        if path == self.CORE_PATH:
             raise ValueError("Cannot add the core path as an additional path.")
 
         async with self.conf.paths() as paths:
@@ -213,7 +214,7 @@ class CogManager:
         """
         resolved_paths = _deduplicate(await self.paths())
 
-        real_paths = [str(p) for p in resolved_paths if p != self.core_path]
+        real_paths = [str(p) for p in resolved_paths if p != self.CORE_PATH]
 
         for finder, module_name, _ in pkgutil.iter_modules(real_paths):
             if name == module_name:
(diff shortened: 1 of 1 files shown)


redbot/core/cog_manager.py
```diff
@@ -54,18 +55,13 @@ def __init__(self, paths: Tuple[str] = ()):
...
-        all_paths = _deduplicate(list(conf_paths) + list(other_paths) + core_paths)
+        all_paths = _deduplicate(list(conf_paths) + list(other_paths) + [self.CORE_PATH])
 
         if self.install_path not in all_paths:
             all_paths.insert(0, await self.install_path())
         return tuple(p.resolve() for p in all_paths if p.is_dir())
 
-    async def core_paths(self) -> List[Path]:
-        core_paths = [Path(p) for p in redbot.cogs.__path__]
-        return core_paths
```
Tobotimus, line 67: There is only ever one core path, and it never changes during runtime - this has been moved to a static class attribute.

redbot/core/cog_manager.py
```diff
@@ -228,7 +227,8 @@ def _ensure_path_obj(path: Union[Path, str]) -> Path:
             " in any available path.".format(name)
         )
 
-    async def _find_core_cog(self, name: str) -> ModuleSpec:
+    @staticmethod
+    async def _find_core_cog(name: str) -> ModuleSpec:
```
Tobotimus, line 231: This method doesn't use `self`, thus I've made it static

redbot/core/cog_manager.py
```diff
@@ -310,7 +310,8 @@ def invalidate_caches():
 class CogManagerUI:
     """Commands to interface with Red's cog manager."""
 
-    async def visible_paths(self, ctx):
+    @staticmethod
+    async def visible_paths(ctx):
```
Tobotimus, line 314: Same with this one

redbot/core/cog_manager.py
```diff
@@ -322,11 +323,15 @@ class CogManagerUI:
...
-        msg = _("Install Path: {}\n\n").format(install_path)
+        cog_mgr = ctx.bot.cog_mgr
+        install_path = await cog_mgr.install_path()
+        core_path = cog_mgr.CORE_PATH
+        cog_paths = await cog_mgr.paths()
+        cog_paths = [p for p in cog_paths if p not in (install_path, core_path)]
+
+        msg = _("Install Path: {install_path}\nCore Path: {core_path}\n\n").format(
+            install_path=install_path, core_path=core_path
+        )
```
Tobotimus, line 334: I've made the core path appear separately to the rest of the paths, to show it's not set in Config.

redbot/core/cog_manager.py
```diff
@@ -428,16 +433,16 @@ class CogManagerUI:
         """
         loaded = set(ctx.bot.extensions.keys())
 
-        all = set(await ctx.bot.cog_mgr.available_modules())
+        all_cogs = set(await ctx.bot.cog_mgr.available_modules())
 
-        unloaded = all - loaded
+        unloaded = all_cogs - loaded
```
Tobotimus, line 438: This change just because it was shadowing the `all` builtin method

mikeshardmind (approved):
After the conversation with Will verifying that the core path did not need to remain a coroutine for future purposes, this looks solid. I was a little wary of changing that in particular, but since it was for internal use only, and it didn't need to remain that way, changes here are solid.

Redjumpman (approved):

PR #1984, opened 2018-08-05
[v3 core] fix unload_extension for module-less cogs

### LINKED ISSUE #1943: [V3 Unload/Reload] edge cases in bot.unload_extension
### The issue

Right now, `bot.unload_extension` will break unload and reload if any cogs exist on the bot where `Cog.__module__` is `None` 

While this is definitely an edge case, I think it is at least worth considering just filtering out any case where there isn't a module, and trust that in that case, the user of created that situation can handle any required cleanup.

### The use case

Adding a cog through eval to test a specific behavior temporarily, without creating an entire module for adding through standard means.

(yes, this is an extreme edge case, but it's how I've run into it, so even a "we won't support this" is a fine answer if that's the decision)

calebj:
### Type

- [x] Bugfix

Fixes #1943

This just skips cogs when  `__module__` is None when unloading by module name. Also:
- backported rapptz/discord.py#621
- moved RPC handler unregister to remove_cog()
- raise an exception when a load would overwrite an existing extension
- uppercase cog name when unloading every time, not just when checking
- edit rpc.py to correct comment saying it's lowercase

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -24,6 +24,10 @@ from .help_formatter import Help, help as help_
 from .sentry import SentryManager
 
 
+def _is_submodule(parent, child):
+    return parent == child or child.startswith(parent + ".")
+
+
 class RedBase(BotBase, RPCMixin):
     """Mixin for the main bot class.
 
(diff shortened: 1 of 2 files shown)


mikeshardmind (approved):
Looks completely solid to me, and works as intended.

Tobotimus (approved):
Thanks for this patch :)

PR #1988, opened 2018-08-06
Extending Mod functionalty: unban command, configurable default ban cleanup

Aryss:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

- **New command "Unban": accepts id, removes ban. Very handy for reversing accidental bans.**
- **New option for ban and hackban commands: Default clear days.** 
  Set to 1 by default.  This way simple ban will clear X number of days without the need to specifiy days explictly. Can be set via "bandelete" command.
- **Mention spam autoban case now lists the message that caused the ban as part of the reason in modlog**

### DIFF
--- a/cogs/mod.py
+++ b/cogs/mod.py
@@ -6,6 +6,7 @@ from __main__ import send_cmd_help, settings
 from datetime import datetime
 from collections import deque, defaultdict, OrderedDict
 from cogs.utils.chat_formatting import escape_mass_mentions, box, pagify
+import time
 import os
 import re
 import logging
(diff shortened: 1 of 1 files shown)


Twentysix26:
Hi, v2 is feature locked. If you're interested you can contribute to v3, which is currently [in development](https://github.com/Cog-Creators/Red-DiscordBot/tree/V3/develop).

Aryss:
Eh... the thing is we're using v2 on Fortnite discord and it'll be a while before v3 becomes stable enough that we can switch to it
¯\_(ツ)_/¯

Twentysix26:
A solution is to copy mod.py into a separate cog, unload the original and use the new one with your edits so that future updates won't clash with the original. This of course means that you might also want to port any future bugfix in.

Aryss:
This is precisely what we do, but not having to keep an eye on bugfixes is what made me create this PR

PR #1989, opened 2018-08-07
[v3 cog manager, utils] handle missing cogs correctly, add some helpful algorithms

calebj:
### Type

- [x] Bugfix
- [x] Enhancement
- [x] New feature

For cog loading, only show "cog not found" if the module in question was the one that failed to import. `ImportErrors` within cogs will now show an error as they should.

Also:
- Add a chain deduplicator, benchmarked to be the fastest method
- Add a semaphore-bounded variant of `asyncio.gather` and an async `as_completed`
- And of course, tests and docs for all additions

### DIFF
--- a/redbot/core/cog_manager.py
+++ b/redbot/core/cog_manager.py
@@ -3,9 +3,10 @@ import pkgutil
 from importlib import import_module, invalidate_caches
 from importlib.machinery import ModuleSpec
 from pathlib import Path
-from typing import Tuple, Union, List
+from typing import Tuple, Union, List, Optional
 
 import redbot.cogs
+from redbot.core.utils import deduplicate_iterables
 import discord
 
 from . import checks, commands
@@ -18,12 +19,13 @@ from .utils.chat_formatting import box, pagify
 __all__ = ["CogManager"]
 
 
-def _deduplicate(xs):
-    ret = []
-    for x in xs:
-        if x not in ret:
-            ret.append(x)
-    return ret
+class NoSuchCog(ImportError):
+    """Thrown when a cog is missing.
+
+    Different from ImportError because some ImportErrors can happen inside cogs.
+    """
+
+    pass
 
 
 class CogManager:
@@ -56,7 +58,7 @@ class CogManager:
         conf_paths = [Path(p) for p in await self.conf.paths()]
         other_paths = self._paths
 
-        all_paths = _deduplicate(list(conf_paths) + list(other_paths) + [self.CORE_PATH])
+        all_paths = deduplicate_iterables(conf_paths, other_paths, [self.CORE_PATH])
 
         if self.install_path not in all_paths:
             all_paths.insert(0, await self.install_path())
@@ -209,10 +211,10 @@ class CogManager:
 
         Raises
         ------
-        RuntimeError
-            When no matching spec can be found.
+        NoSuchCog
+            When no cog with the requested name was found.
         """
-        resolved_paths = _deduplicate(await self.paths())
+        resolved_paths = deduplicate_iterables(await self.paths())
 
         real_paths = [str(p) for p in resolved_paths if p != self.CORE_PATH]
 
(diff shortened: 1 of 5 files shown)


calebj:
@Tobotimus rebased.

redbot/core/utils/__init__.py
```diff
@@ -70,10 +95,105 @@ def safe_delete(pth: Path):
...
+    Raises
+    ------
+    TypeError
+        When invalid parameters are passed
+    """
+    if loop is None:
+        loop = asyncio.get_event_loop()
+
+    if semaphore is None:
+        if type(limit) != int or limit <= 0:
```
Tobotimus: Please use `not isinstance` instead of `type() !=`

redbot/core/utils/__init__.py
```diff
@@ -70,10 +95,105 @@ def safe_delete(pth: Path):
...
+            return await task
+
+    pending = [sem_wrapper(semaphore, task) for task in coros_or_futures]
+
+    while pending:
+        async with semaphore:
+            try:
+                done, pending = await asyncio.wait(pending, loop=loop, return_when=FIRST_COMPLETED)
+            except asyncio.CancelledError:
+                []
```
Tobotimus: Not sure what's going on with this `[]` statement, would prefer the use of `pass` or `contextlib.supress`

redbot/core/utils/__init__.py
```diff
@@ -70,10 +95,105 @@ def safe_delete(pth: Path):
...
+
+    if semaphore is None:
+        if type(limit) != int or limit <= 0:
+            raise TypeError("limit must be an int > 0")
+
+        semaphore = Semaphore(limit, loop=loop)
+
+    async def sem_wrapper(sem, task):
+        async with sem:
+            return await task
```
Tobotimus: Since this `sem_wrapper` is common to the one in `bounded_gather_iter`, it can be moved to the module-level and made private

redbot/core/utils/__init__.py
```diff
@@ -70,10 +95,105 @@ def safe_delete(pth: Path):
...
+    Raises
+    ------
+    TypeError
+        When invalid parameters are passed
+    """
+    if loop is None:
+        loop = asyncio.get_event_loop()
+
+    if semaphore is None:
+        if type(limit) != int or limit <= 0:
```
Tobotimus: `isinstance` here too

redbot/core/core_commands.py
```diff
@@ -23,6 +23,7 @@
 from redbot.core import checks
 from redbot.core import i18n
 from redbot.core import commands
+from redbot.core.cog_manager import NoSuchCog
```
Tobotimus: Unused import

redbot/core/cog_manager.py
```diff
@@ -209,10 +211,10 @@ def _ensure_path_obj(path: Union[Path, str]) -> Path:
 
         Raises
         ------
-        RuntimeError
-            When no matching spec can be found.
+        NoSuchCog
+            When no cog with the requested name was found.
         """
-        resolved_paths = _deduplicate(await self.paths())
+        resolved_paths = deduplicate_iterables(await self.paths())
```
Tobotimus: `paths()` already deduplicates the paths before it returns, so you can remove the call to deduplicate here

Tobotimus (changes requested):
@calebj This PR adds a lot of really nice things, thank you. Just a few little things that can be fixed up and then I think it's good to go.

redbot/core/utils/__init__.py
calebj: I think I had logic here for cancelling pending tasks, but I found something better (see the push)

mikeshardmind (approved):
I'm not seeing any issues with this after the changes made. I'm giving it a :+1:

Tobotimus (approved):
Awesome. Thanks @calebj and thanks @mikeshardmind for reviewing

PR #1995, opened 2018-08-09
[V3 Admin] Fix [p]addrole error when a hierarchy issue occurs

### LINKED ISSUE #1952: [V3] Addrole command errors on hierarchy issue
Cleaner output for hierarchy issue with [p] addrole instead of giving error.
image

Tobotimus:
### Type

 - Bugfix

### Description
Resolves #1952.

### DIFF
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -136,7 +136,7 @@ class Admin:
             # noinspection PyTypeChecker
             await self._addrole(ctx, user, rolename)
         else:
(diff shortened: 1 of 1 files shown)


mikeshardmind (approved):
This works. The minor double take I had at my initial glance at the diff may bring up a good point about using the `usage` kwarg in commands instead of relying on misleading variable names to accomplish the same in the core code, but this probably isn't the PR to start that.

Redjumpman (approved):

PR #1996, opened 2018-08-09
[V3 Audio] Embeds colour respecting its set value

Brramble:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

I have gone through the cog and removed all set colours from what it used to be and replaced with the embed request so it should now run uniform to the rest of the bot.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -131,10 +131,7 @@ class Audio:
                 embed = discord.Embed(
                     title="Queue ended."
                 )
-                if await ctx.embed_requested():
-                    await notify_channel.send(embed=embed)
-                else:
-                    await notify_channel.send(embed=embed)
+                await notify_channel.send(embed=embed)
 
         if event_type == lavalink.LavalinkEvents.QUEUE_END and status:
             if playing_servers == 0:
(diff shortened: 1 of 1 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -130,9 +129,12 @@ def __init__(self, bot):
...
                 notify_channel = self.bot.get_channel(notify_channel)
                 embed = discord.Embed(
-                    colour=notify_channel.guild.me.top_role.colour, title="Queue ended."
+                    title="Queue ended."
                 )
-                await notify_channel.send(embed=embed)
+                if await ctx.embed_requested():
+                    await notify_channel.send(embed=embed)
+                else:
+                    await notify_channel.send(embed=embed)
```
Tobotimus: This isn't doing anything. The condition statement is meaningless - the bot will send the exact same thing no matter the condition.

Tobotimus (changes requested):
I appreciate what you're trying to do here. I think the function you're looking for is `await ctx.embed_colour()`. Checking `if ctx.embed_requested()` is just checking whether embeds are enabled, for which you're sending an embed if they're disabled anyway. Best remove all of those statements.

Brramble:
Thank you for the help I will rework it.

Brramble:
Ive also added a [p]thumbnail command which will display the songs thumbnail in an embed. Furthermore with thumbnails in the [p]queue and [p]now commands. 

Me being stupid I pushed it all at once so allow me to go back to remove them for this PR as this is just for embeds

PR #1998, opened 2018-08-10
[V3 Audio] Add thumbnail display with toggle

aikaterna:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Adds a toggle to display a thumbnail for YouTube tracks on notify messages, [p]now and [p]queue for the current song. Thanks to Bramble for the suggestion/request.


redbot/cogs/audio/audio.py
```diff
@@ -1909,6 +1930,14 @@ def _track_creator(player, position=None, other_track=None):
             track_obj[key] = value
         return track_obj
 
+    @staticmethod
+    def _thumbnail(track):
+        if "youtube" in track.uri:
+            values = track._info.values()
+            id = list(values)[1]
```
calebj: Will this always be the second item in `values()`? Would it have a key instead? Also, is it possible to patch Red-Lavalink for this?

aikaterna: Changed it up a bit. And the thumbnail fetching maybe should be moved to Red-Lavalink.

Crossedfall:
Tested in Windows (10/Server 2016) and Linux (Debian/Ubuntu/Centos 7) in python 3.6.5 and 3.6.6 with no issues. Also tested this using the "built-in" version of lavalink and using an externally hosted node, no differences there even when the external node had experimental changes applied.

**Testing Procedure**:

1. Begin playing a single song
2. Run the `[p]now` and `[p]queue` commands
3. Verify that there are no thumbnails
4. Enable thumbnails with `[p]audioset thumbnail`
5. Run the `[p]now` and `[p]queue` commands again
6. Verify that there are thumbnails
7. Add a large playlist to the queue (100+ songs) and begin skipping through the songs
8. Re-verify that thumbnails are displayed between every third skip for the `[p]now` and `[p]queue` commands respectively 5 times
9. Disable thumbnails and repeat step 8, instead verifying that thumbnails are not being displayed

redbot/cogs/audio/audio.py
```diff
@@ -1909,6 +1930,14 @@ def _track_creator(player, position=None, other_track=None):
             track_obj[key] = value
         return track_obj
 
+    @staticmethod
+    def _thumbnail(track):
+        if "youtube" in track.uri:
+            for v, k in track._info.items():
```
calebj: Why not
```py
if "youtube" in track.uri and "identifier" in track._info:
    return "https://img.youtube.com/vi/{}/mqdefault.jpg".format(track._info["identifier"])
```
or even
```py
    return "https://img.youtube.com/vi/{identifier}/mqdefault.jpg".format(**track._info)
```
The `return None` is redundant behavior.

These are acceptable... but I still prefer changing the library instead.

calebj (changes requested):

redbot/cogs/audio/audio.py
aikaterna: After https://github.com/Cog-Creators/Red-Lavalink/pull/38 is merged and released, I'll make another PR to change audio to use it.

calebj (dismissed):
Looking good. I also recommend adding a TODO comment referencing Cog-Creators/Red-Lavalink#38.

Tobotimus (approved):
Thanks @aikaterna!

Issue #1999, opened 2018-08-10
[V3 Streams] Youtube alerts added by ID have NoneType name


JJW8071:
# Command bugs

Red v3.0.0.b18

#### Command name

[p]streamalert youtube UC79_r-MK8-bmJCYmyzAkvPw

#### What cog is this command from?

Streams 

#### What were you expecting to happen?

Add non-partner youtube channel ("Cade") to streamalert settings.
Add <id: channel-name>"Cade"</channel-name> name to streams settings in lieu of a Partner-Name.

#### What actually happened?
**Streams cog is completely disabled**

If the Youtube Channel ID is used <UC....xzy> instead of a Partner channel name AND if the youtube channel is *not* a Youtube Partner, then Streams sets the channel name as 'null'.

```
Exception during loading of cog
Traceback (most recent call last):
  File "C:\Users\jason\AppData\Local\Programs\Python\Python36\lib\site-packages\redbot\core\core_commands.py", line 87, in _load
    await bot.load_extension(spec)
  File "C:\Users\jason\AppData\Local\Programs\Python\Python36\lib\site-packages\redbot\core\bot.py", line 222, in load_extension
    await lib.setup(self)
  File "C:\Users\jason\AppData\Local\Programs\Python\Python36\lib\site-packages\redbot\cogs\streams\__init__.py", line 6, in setup
    await cog.initialize()
  File "C:\Users\jason\AppData\Local\Programs\Python\Python36\lib\site-packages\redbot\cogs\streams\streams.py", line 71, in initialize
    self.streams = await self.load_streams()
  File "C:\Users\jason\AppData\Local\Programs\Python\Python36\lib\site-packages\redbot\cogs\streams\streams.py", line 618, in load_streams
    return [x for x in streams if not (x.name.lower() in seen or seen_add(x.name.lower()))]
  File "C:\Users\jason\AppData\Local\Programs\Python\Python36\lib\site-packages\redbot\cogs\streams\streams.py", line 618, in <listcomp>
    return [x for x in streams if not (x.name.lower() in seen or seen_add(x.name.lower()))]
AttributeError: 'NoneType' object has no attribute 'lower'
```

#### How can we reproduce this issue?

[p]streamalert youtube UC79_r-MK8-bmJCYmyzAkvPw
[p]reload streams
[error]

mikeshardmind:
Can repro on both b18 and V3/dev This makes the cog completely inoperable.

[...]

PR #2002, opened 2018-08-11
[V3] temp fix to prevent broken streams cog before next beta

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This prevents even attempting to call the youtube API without having all required header parameters. Discovery of what causes that, and a more comprehensive fix should still be applied, but are beyond the scope of a fix during the cutoff window. Details of the issue in #1999


mikeshardmind:
Temp fix unneeded, other changes made should already cover this case sufficiently enough for temporary use.

PR #2004, opened 2018-08-12
[V3 Cleanup] Hotfix for [p]cleanup after

Tobotimus:
### Type

- Hotfix

Since `[p]cleanup after` is obselete right now, I'd like to push a hotfix out there before beta 19. This only edits code within `[p]cleanup after` as to avoid any collateral damage.

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -68,6 +68,12 @@ class Cleanup:
         - The message is less than 14 days old
         - The message is not pinned
         """
+        if after is not None:
+            log.error(
+                "The `after` parameter for the `Cleanup.get_messages_for_deletion` method is "
+                "currently broken, see PRs #1980 and #2004 for details."
+            )
+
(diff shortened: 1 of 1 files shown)


mikeshardmind:
My preferences in order on this:

1. Mark cleanup after as disabled on b18
2. Go with the refactor in #1980 if others have tested
3. Go with this, (pending testing) but add something raising an error if anything is passed to the utility function for after which points to this commit and related issues in case 3rd party cogs are using this.

I'll test in a moment here.

mikeshardmind (changes requested):
Works as intended, I've made my preferences here known already, The only change I'm going to request is a temprary logging.error if after is passed as a keyword to the utility func with a link to this discussion to make it easier for 3rd party devs to figure out what the hell is breaking.

Tobotimus:
@mikeshardmind Log warning has been added

mikeshardmind (approved):
Good to go.

Redjumpman (approved):

PR #2008, opened 2018-08-12
[V3] remove self-bot support

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature

It's against the terms of service, we tell people not to use them, and we even have special warnings in the code for it. I know a lot of this was prior to discord's stance being firm on this one, but we need to drop the support for this, we shouldn't support API misuse of any kind.


Tobotimus (approved):
Thanks for the patch!

PR #2007, opened 2018-08-12
[3.0.0b22] cleanup finish deprecation cycle

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature

### Description of the changes

completes deprecation started in #1980 and continued in #2006

PR #2006, opened 2018-08-12
[3.0.0b21] b21 cleanup deprecation continuation

mikeshardmind:
### Description of the changes
This continues changes made in #1980

PR #2013, opened 2018-08-13
[V3 Cleanup] refactor internals 

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Per discussion with @Tobotimus in discord, an alt to #1980 , #2006 , #2007 since deprecating the current usage shouldn't be necessary. 

This reverts the hotfix #2004 , and refactors `get_messages_for_deletion`. This also adds a note in the doc string about the unsafe usage case no core code uses, and how to handle that case should we care about it in the future, rather than apply any special casing here.


Tobotimus:
On #1980 you stated this:

> Because it's a timedelta, being out of sync on the base time isnt a concern, and I'd prefer removing ctx as we don't need it, so it opens up for event based usage without mocking an object.

I agree we should remove ctx and I like that change. However discord's time being out of sync with the system time can still cause problems, see this example, simply achieved by changing my PC's system time:
image
```
discord.errors.HTTPException: BAD REQUEST (status code: 400): You can only bulk delete messages that are under 14 days old.
```


To be honest I'm not *too* concerned about it, but it could be resolved by adding an optional `now` parameter to `get_messages_for_deletion` for callers to pass in `ctx.message.created_at`, but falls back to `datetime.utcnow()` if not provided.

mikeshardmind:
You're right, the timedelta still gets compared to a snowflake, which isn't isolated, I think a better solution is if the first message the bot receives seems more than 2 minutes out of sync (`discord.utils.snowflake_time`, compared to `datetime.utcnow()`) to raise a warning. It shouldn't be a burden for each time sensitive piece of code to handle, but we can still take steps to alert users to that issue.

mikeshardmind:
@Tobotimus I still don't believe that handling bad system clock times should be handled here, and have opened up #2036 to handle this instead.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -244,35 +244,23 @@ def check(m):
             await ctx.send(_("This command can only be used on bots with bot accounts."))
             return
 
-        try:
-            message = await channel.get_message(message_id)
-        except discord.NotFound:
-            await ctx.send(_("Message not found."))
-            return
+        after = await channel.get_message(message_id)
```
Tobotimus: This still needs to be in a try/except

Tobotimus (changes requested):
Will give the tick of approval after this change

Tobotimus (approved):

PR #2018, opened 2018-08-14
[Config] Group.__call__() has same behaviour as Group.all()

Tobotimus:
### Type

- Bugfix

This modifies `config.Group.__call__` to now have the same behaviour of `config.Group.all()`, and in fact `all()` is now simply an alias for `__call__`. Both of these functions are able to be used as a context manager as well.

Furthermore, to match the new behaviour of `__call__`, `get_raw` will mix in defaults when a sub-group is being retrieved by it.


Tobotimus:
### Notes about breaking changes
The situations where these changes could be breaking are where you are retrieving data whose default type is a dict, by either calling a `config.Group` object, or using `get_raw()` to retrieve a dict. Previously, these functions would not return defaults mixed into the returned dict - the new behaviour is that they do.

As an example:
```py
# old behaviour

config.register_global(foo={'bar': True})
await config.foo()  # -> {}

# new behaviour

config.register_global(foo={'bar': True})
await config.foo()  # -> {'bar': True}

```

And an example with `get_raw()`:
```py
# old behaviour

config.register_global(foo={'bar': True, 'baz': False})
await config.foo.baz.set(True)
await config.get_raw('foo')  # -> {'baz': True}

# new behaviour

config.register_global(foo={'bar': True, 'baz': False})
await config.foo.baz.set(True)
await config.get_raw('foo')  # -> {'bar': True, 'baz': True}

```

### Concerns
If you have any concerns about how you will update to account for these changes, let us know (in #coding perhaps) and we will do our best to assist you.

mikeshardmind:
A slight clarification, this is only a breaking change in those cases if you were using that behavior as a method for checking that no value was set. In most cases, registering a default should indicate that you have a default value and do not care if one is not already set, however, if you do care, there are a few ways to handle this.

- Sentinel Values which will never exist in normal operation as defaults. (`None` for a value which can't normally be None as an example)
- Checking the all_members, all_roles, all_guilds (etc) method for their existence.

redbot/core/bank.py
```diff
@@ -400,19 +400,18 @@ def _invalid_amount(amount: int) -> bool:
 
     """
     if await is_global():
-        acc_data = (await _conf.user(member)()).copy()
-        default = _DEFAULT_USER.copy()
+        all_accounts = await _conf.all_users()
```
mikeshardmind, line 403: Would it make more sense to use a context manager with a sentinel value default for account creation which would be impossible? It's not a required change, but it should simplify the new user case significantly.

mikeshardmind (approved):
This looks solid and appears to be behaving with the new intended behavior. I've left a comment about a particular refactor option in the bank, but it's not a necessity to change the way this was handled, just an option for discussion.

redbot/core/bank.py
Tobotimus: That's true. For now this works so I'm going to leave it as it is, however sentinel values are possible as defaults... You just need to be careful whether you specify a sentinel value as an empty dict, or `None`.

When using a `NoneType` sentinel value, that value will not be treated as a `config.Group`, even if a dict later replaces it. Using an empty dict instead, the value will be treated as a `config.Group`, and programmers can utilise the extra methods of this object.

Kowlin (approved):

Issue #1968, opened 2018-08-01
[v3 Downloader] RFC: installing cogs/shared libs from PyPI + cog discovery


calebj:
# Feature request

#### Select the type of feature you are requesting:
- [x] Core (Downloader)

#### Describe your requested feature
Since redbot can be installed directly using `pip`, and will be usable without `git` once discord.py v1.0+ hits PyPI, I think that Red should support cogs or cog collections installed as python packages.

Fine-grained management via Downloader (beyond just calling `[p]pipinstall`) would be nice to have, but using [setuptools plugin discovery](https://packaging.python.org/guides/creating-and-discovering-plugins/#using-package-metadata) ([docs](https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins)) is a good first step.

#### Benefits:
- Dependency management and updates are already handled
- Supports variants/dependencies for specific OSes or via [setuptools "extras"](https://setuptools.readthedocs.io/en/latest/setuptools.html#id16)
- Versioning, stability level and other metadata is well-defined
- Cog installation can be done in one step using `pip install`, without adding any repos
- Cogs can support pre-compiled native code packaged as a wheel
- Release pipelines are easier to set up

#### Current issues:
1. Package discovery for normal packages is independent of repo contents
   - packages not installed will not appear during discovery
   - sources of cog list/metadata could be a "base" package, or cogs.red.
   - base package would have to be updated when cogs are added/removed, description changes, etc.
   - cogs.red would have to perform discovery (or just rescan when a webhook fires)
2. PyPI has no tag/classifier for Red ([yet](https://github.com/pypa/warehouse/issues?q=is%3Aissue+is%3Aopen+label%3A%22classifier+request%22))
3. Hybrid repositories
   - setup.py reading from info.json?
   - repo manager should tolerate namespace packages in git repos
   - installation may need to parse requirements.txt instead of JSON requirements array

mikeshardmind:
Definitely interested in this being supported. Would make distributing my cogs with proper versioning much simpler. Would this also include better interaction between the downloader and python venvs? (I Should probably collect the list of odd behavior's I've noticed with venv usage mixed with current downloader using `--target` usage if not to ensure they are intended)

calebj:
> Would this also include better interaction between the downloader and python venvs?

[...]

PR #2019, opened 2018-08-14
Update submodules

Proximyst:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Instead of cloning empty submodules, it should now automatically pull all of the modules as it is from remote. This allows for repositories to be their own cogs and rather use submodules in an "indexing" repository, as we do with https://github.com/r-Norge/tannhjul

### DIFF
--- a/cogs/downloader.py
+++ b/cogs/downloader.py
@@ -609,9 +609,9 @@ class Downloader:
                 if "@" in url: # Specific branch
                     url, branch = url.rsplit("@", maxsplit=1)
                 if branch is None:
-                    p = run(["git", "clone", url, folder])
+                    p = run(["git", "clone", "--recursive", url, folder])
                 else:
-                    p = run(["git", "clone", "-b", branch, url, folder])
+                    p = run(["git", "clone", "--recursive", "-b", branch, url, folder])
                 if p.returncode != 0:
                     raise CloningError()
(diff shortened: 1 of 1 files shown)


Tobotimus:
Hi, thanks for your PR.

Unfortunately I'm gonna freeze this as V2 is at a very low priority (technically it's bugfix-only). It will probably be closed if it doesn't get any attention over the next few months.

mikeshardmind:
I'll just toss on a note, that if you want this behavior in V3's downloader, you may want to add a note on #1968

Proximyst:
@Tobotimus The PR is mostly to help out people using such, as us over at r/Norge wish to use it for the bot we call Askeladden (A Norwegian folk tale icon), and we wish to have the repositories in an easy way to handle with proper permissions. Seeing as the server is relatively small at little over 3k users, there is without a doubt others who most likely also would like to see this be added for their own organisational usecases and simply for the fact that it's simple to handle.

@mikeshardmind I will be sure to do so in a moment, thank you!

mikeshardmind:
Right now, V3's downloader updates submodules. As this is a behavior change, and not a critical fix or a non-code change, We're not going to include this for V2's EOL support, but if there's anything else with this you need for V3, feel free to let us know.

PR #2023, opened 2018-08-15
[i18n] Use redgettext over pygettext

Tobotimus:
### Type

- Enhancement

I mentioned a while ago that I started to write a modified version of pygettext specifically for Red. The rationale was to add extra functionality to the tool which would enhance our i18n workflow. The main feature here is extracting command/cog help, without having to extract every docstring in the file (which is the only option when using pygettext). However I've also tweaked the tool to generally modernise it, removed redundant features and have also tweaked the default settings and file discovery for our ease of use.

Here is a list of the changes which have been made in this PR:
* Clear out and ignore autogenerated `messages.pot` files 
  * Rationale: we don't need to track these as they just get overwritten every time Travis deploys to Crowdin anyway - we only need to track `.po` files
* Remove redundant `regen_messages.py` files
  * Rationale: using the new `--recursive` file discovery feature of redgettext means that all we need to do to regenerate all `messages.pot` files is to run `redgettext --recursive redbot`. This removes a bunch of boilerplate and we no longer need to remember to add paths to new modules in existing `regen_messages.py` files
* Replace `generate_strings.py` with `make gettext` in the Makefile, since the commands for crowdin deployment are now vastly simpler we don't need to use a python script
* Install redgettext in Travis Crowdin job
  * This is the only place I've specified redgettext to be installed. I don't think we need it in an extra or anything like that.
* Clean up some problematic usages of gettext function
  * Usages of the `_()` function should *only* contain string literals. In some places this was not the case. This fixed that to remove some warning messages.


Makefile
```diff
@@ -2,3 +2,6 @@ reformat:
 	black -l 99 `git ls-files "*.py"`
 stylecheck:
 	black --check -l 99 `git ls-files "*.py"`
+gettext:
+	redgettext --command-docstrings --verbose --recursive redbot
```
mikeshardmind: Swap with
`````
redgettext --command-docstrings --verbose `git ls-files "redbot/**/*.py" --exclude "redbot/pytest/**"`
`````
to avoid including pytest

mikeshardmind (changes requested):
Pytest should be excluded, I've included a way to do this in a comment on the makefile

Makefile
mikeshardmind: ED: This actually still includes it, one moment

mikeshardmind:
`````
redgettext --command-docstrings --verbose `git ls-files "redbot/**/*.py" | grep -v "redbot/pytest"`
`````
Works, but is ugly, the alternative is manual exclusions (from file ideally) of everything we want to exclude, with no matching pattern including wildcards, which is uglier) or a directoy exclusion option in redgettext if you are agreeable to implementing that as well.

Tobotimus: FTR: Unfortunately this solution doesn't work, I don't have a great understanding of why, so I'm gonna add an option on redgettext for a file exclusion pattern

mikeshardmind (approved):
This appears to properly exclude `redbot/pytest` now, thanks for adding the exclude option to redgettext.

Kowlin (approved):

PR #2025, opened 2018-08-16
[Launcher] Always force reinstall when updating

Tobotimus:
In lieu of recent support issues involving users updating dependencies, it shows that dependency links are wasted when upgrading unless we force reinstall and ignore the cache. Only then, will pip install the up-to-date discord.py specified in the dependency link.

### DIFF
--- a/redbot/launcher.py
+++ b/redbot/launcher.py
@@ -71,7 +71,7 @@ def parse_cli_args():
     return parser.parse_known_args()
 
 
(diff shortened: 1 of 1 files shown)


aikaterna (approved):

Tobotimus:
Superseded by #2027. Another PR can be opened to refactor the reinstall options in the launcher menu.

Issue #1249, opened 2018-01-13
[V3 Docs] Virtual Environment


mikeshardmind:
### Type:

- [X] Suggestion
- [ ] Bug

Docs should give steps on setting up a virtual environment using the builtin `python3 -m venv` , especially as we are now moving towards pip installable cogs to reduce dependency conflicts

Tobotimus:
Resolved by #2029.

PR #2027, opened 2018-08-17
[V3 Launcher] append --user if the bot isn't in a venv

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This adds the `--user` flag to `redbot-launcher` which allow people to install and update their packages without admin permissions. This will basically add the `--user` flag to all `pip install` commands.

### DIFF
--- a/redbot/launcher.py
+++ b/redbot/launcher.py
@@ -27,7 +27,7 @@ if sys.platform == "linux":
 
 PYTHON_OK = sys.version_info >= (3, 6)
 INTERACTIVE_MODE = not len(sys.argv) > 1  # CLI flags = non-interactive
-if not INTERACTIVE_MODE and "--user" in sys.argv:
+if not len(sys.argv) > 2 and "--user" in sys.argv:
     # only --user CLI flag
     INTERACTIVE_MODE = True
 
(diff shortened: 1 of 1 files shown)


Tobotimus:
We would much prefer if users installed the bot in a virtual environment (see #1249), as this would greatly simplify people's issues with permissions and version conflicts. Also see #2025 which conflicts with this PR.

Having said that, I can understand that when users, for whatever reason, are not using a virtual environment, the libraries should be installed to user site packages, always. With respect, I think it's ugly to expose a flag on our application which simply passes it on to another, especially in a way where it is used as a startup command which impacts the behaviour of options in an interactive menu. If the user is to have control over the flag, they can just invoke `pip` directly.

I suggest that we *always* include `--user` when upgrading *as long as* we are not in a virtual environment, as it is not allowed in that circumstance. You can check if the bot is installed in a virtual environment during runtime with `hasattr(sys, "real_prefix")`.

laggron42:
This was updated following @Tobotimus's review. Now the bot will always append the `--user` flag if the bot is not in a virtualenv or in a venv.

aikaterna (approved):
Tested on Win 7, Win 2012 R2, and Ubuntu 17.10. More favorable than #2025 as it doesn't run into permissions issues on some Windows installs when reinstalling.

Tobotimus (approved):

PR #2029, opened 2018-08-19
[Docs] Add virtualenv guide and compress install guides

Tobotimus:
Finally! I'm proposing to add this section to our docs, however this PR does more than that.

I realised whilst modifying our install guide that I had to copy and paste a little section linking to the venv guide from *8 different guides*, which all were basically the same, except for the commands used for installing pre-requirements.

This is why I've decided to compress the guides into two different ones:
 - Windows
 - Non-windows (a.k.a Linux or Mac)

In the Non-Windows guide, I've simply included a section for each OS under the pre-requirements heading, and the rest is common across all of those distributions.

I've kept Windows separate because as we know that guide needs to be as straightforward and linear as possible, and also because of the slightly annoying differences in `python`/`python3` commands.

[Here is a RTD site for this PR's proposed changes.](https://red-discordbot-test.readthedocs.io/en/v3-venv_docs/)


docs/install_linux_mac.rst
```diff
@@ -0,0 +1,194 @@
...
+    brew cask install java8
+
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+Ubuntu 18.04 Bionic Beaver
+~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+.. code-block:: none
+
+    sudo apt install python3.6-dev python3-pip build-essential libssl-dev libffi-dev git unzip default-jre -y
+
```
mikeshardmind, line 106: This is at the same visual seperation from other distros, but assumes the user will continue with the 2nd half of the ubuntu instructions below. I know duplicate code and all, but in this case, it needs to be

mikeshardmind (changes requested):
I think this is also an appropriate time to add a small change to the systemd instructions to use [below changes] since we're updating setup guides, but I can open a seperate PR for that if you'd prefer.

`sudo -e /etc/systemd/system/red@.service`

in place of `sudo nano /etc/systemd/system/red@.service`

(`sudo -e` preferable to `sudo nano` for environment reasons)

docs/install_linux_mac.rst
Tobotimus: Hmm? I don't think that's the case. Python 3.6, pip, git and java are installed in this step. Xenial has separate steps to get those.

aikaterna (approved):

Kowlin (approved):

PR #2035, opened 2018-08-20
[Audio] Do less strict matching for java version

### LINKED ISSUE #2033: (V3) Audio cog doesnt load
# Other bugs

I was normally, like everyone does, trying to load audio cog.

I was expecting it to load normally and to be usable.

#### What actually happened?

It failed to load, throwing in the terminal this:

```py
[20/08/2018 02:50] ERROR core_commands _load 89: Package loading failed
  File "/home/ruben/.local/lib/python3.6/site-packages/redbot/core/core_commands.py", line 87, in _load
  File "/home/ruben/.local/lib/python3.6/site-packages/redbot/core/bot.py", line 226, in load_extension
    await lib.setup(self)
  File "/home/ruben/Downloads/Red-DiscordBot-3-develop/redbot/cogs/audio/__init__.py", line 53, in setup
    await start_lavalink_server(bot.loop)
  File "/home/ruben/Downloads/Red-DiscordBot-3-develop/redbot/cogs/audio/manager.py", line 68, in start_lavalink_server
    java_available, java_version = await has_java(loop)
  File "/home/ruben/Downloads/Red-DiscordBot-3-develop/redbot/cogs/audio/manager.py", line 47, in has_java
    version = await get_java_version(loop)
  File "/home/ruben/Downloads/Red-DiscordBot-3-develop/redbot/cogs/audio/manager.py", line 63, in get_java_version
    major, minor = version_string.split(".")[:2]
ValueError: not enough values to unpack (expected 2, got 1)
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x7f8b24284ef0>
```

Im not sure, as i only installed like in the guide on arch linux, and it failed to load the cog right after the 1st startup, and i had java installed 
Version below v
[ruben@arch redbot]$ java -showversion
Picked up _JAVA_OPTIONS: 
java version "1.8.0_181"
Java(TM) SE Runtime Environment (build 1.8.0_181-b13)

Edit: I was using python 3.6.6 from the aur, as the python that was in the repos was already python 3.7

Tobotimus:
### Type

- Bugfix

The java version checking was a bit strict before - it was looking through the output of `java -version` with a very clear idea of what it should look like. #2033 shows us that it can look different sometimes.

Resolves #2033.


mikeshardmind (changes requested):
This entire version check is meaningless.

There are multiple ways in which the version can be expressed, one of these is with compatible oracle java:

```
$ java -version
openjdk version "10.0.1" 2018-04-17
```

Note that we're expecting that version to be expressed as 1.8.x in comparison elsewhere. (which it can be...)
```
# java -version
openjdk version "1.8.0_181"
```

These are both compatible open-jdk installs on different systems. 

We either need to constrict ranges and match either, check the JVM for specific capability with an import-less test class, or just attempt to launch and catch the failure cases.

Edit: and despite oracle java being compatible, we aren't compatible with oracle java 7, which is still out there, and would pass this version check too.

mikeshardmind (approved):

Kowlin (approved):

PR #2036, opened 2018-08-20
[V3] handle time innacuracy by warning owner

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

We shouldn't need to build in code checks everywhere for inaccurate system time. It is the responsibility of the host to ensure this. This adds a warning on a detected system clock inaccuracy that any time-sensitive code in the bot may fail if detecting a significant difference in discord message times to the system clock.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -106,6 +106,7 @@ class RedBase(BotBase, RPCMixin):
 
         self.counter = Counter()
         self.uptime = None
+        self.checked_time_accuracy = None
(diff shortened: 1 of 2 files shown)


calebj (dismissed):
Looking good. I wish there was a heartbeat event to hook into, but the message one is probably the best candidate without one.

Tobotimus (approved):
This looks good to me. Thanks for the patch

PR #2041, opened 2018-08-22
[V3] Global data path

### LINKED ISSUE #2039: [V3] Be able to have global settings and data (not instance-specific)
# Feature request

- [ ] Cog
- [ ] Command
- [x] API functionality

Currently, the 3rd party cogs are always specific and depending to one single instance. The data is stored in the instance's files and cannot "interact" with other instances. An option to get a global data path would be great and reduce disk usage/memory usage/CPU usage for specific functions of some cogs.

Currently, `redbot.core.data_manager.cog_data_path()` returns something like this: `InstanceName/cogs/YourCog/`. Adding another function like `global_data_path` which would lead to `/usr/local/share/Red-DiscordBot/YourCog/` (on Linux) would be great for hosting something that use a lot of disk usage for example.

As an example, with that feature, the Audio cog could download Lavalink.jar in that global folder (I host 4 bots, only one Lavalink server but still 4 identic files) and make shared Lavalink server easier.

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This will add two new functions to `redbot.core.dataconverter`: `global_cog_data_path` and `global_core_data_path`. These functions are similar to `cog_data_path` and `core_data_path`, except they will redirect to a folder common to all instances. This resolves #2039.

**Important:** I tested it quickly, the base functions are working, but this might need more deep testing since this modifies the main code from `__main__` (starting a bot and grabbing data).

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -9,6 +9,7 @@ from redbot.core.cog_manager import CogManagerUI
 from redbot.core.data_manager import (
     create_temp_config,
     load_basic_configuration,
+    global_core_data_path,
     config_file,
     _convert_base_config,
 )
@@ -131,6 +132,7 @@ def main():
         )
         cli_flags.instance_name = "temporary_red"
         create_temp_config()
+    _convert_base_config()
     load_basic_configuration(cli_flags.instance_name)
     log, sentry_log = init_loggers(cli_flags)
     red = Red(cli_flags=cli_flags, description=description, pm_help=None)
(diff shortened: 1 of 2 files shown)


Tobotimus:
Thank you for your PR @retke, but due to the reasons I mentioned on #2039 I'm not going to merge this, instances should not be able to modify any global data, that is the job of `redbot-setup`. We can look into moving the Lavalink.jar into site-packages or something similar using setuptools.

PR #2049, opened 2018-08-25
[Warnings] Help users understand custom reasons

### LINKED ISSUE #2014: [V3 Warnings] AllowCustomReasons -> This is not a registered reason
# Command bugs

#### Command name

warn

warnings

A user to be warned with whatever reason I just specified

#### What actually happened?

Error message: "That is not a registered reason!"

1. Enable custom reasons `[p]warningset allowcustomreasons 1`/true/yes
2. Warn a user with a custom reason: `[p]warn <user> <reason>`
3. "That is not a registered reason!"

------

I've asked about this on multiple beta versions and still didn't get a clear reply if this is supposed to work how I think it is. At least I can't figure out why else there would be a subcommand called "allowcustomreasons". If this is supposed to do something else, can you clarify that somehow in the command help? Because I interpret this that when allowcustomreasons is disabled, you can only use added reasons on the warn, and when it's enabled you can specify any reason.

Tobotimus:
Resolves #2014.

This helps users out with some common misunderstandings about how custom reasons work in Warnings.

### DIFF
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -221,8 +221,8 @@ class Warnings:
         if user == ctx.author:
             await ctx.send(_("You cannot warn yourself."))
             return
+        custom_allowed = await self.config.guild(ctx.guild).allow_custom_reasons()
         if reason.lower() == "custom":
-            custom_allowed = await self.config.guild(ctx.guild).allow_custom_reasons()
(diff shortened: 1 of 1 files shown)


Redjumpman:
FWIW I did some testing on this one. The new info message is more clear.

Redjumpman (approved):

PR #2051, opened 2018-08-25
[V3 CustomCom] Custom Command Parameters

Zephyrkul:
### Type

- [x] New feature

Allows specifying more parameters for CC's via {0}, {1}, etc. that will be filled by the user invoking the CC. Python-style type hinting and attribute access is also allowed for Discord and builtin types.

> [p]cc add simple greet Hi, {0.mention:Member}!
> ...
> [p]greet zephyrkul
> Hi, @zephyrkul!

The bot will reply with the standard help messages if the cc is incorrectly executed.

> [p]greet me
> Member "me" not found


Zephyrkul:
Oh, I forgot that I removed some variable definitions on my copy that weren't used since my linter kept bugging me, sorry!

on_message also had some variable name changes so I could better keep track of what variable was doing what.

Zephyrkul:
I'm unsure what the travis-ci failure is - I looked at it but I don't know what it's trying to say.

Zephyrkul:
@aikaterna told me that it was because I didn't style it according to your style guidelines, so I've formatted it accordingly. Thanks, aikaterna!

Zephyrkul:
As a small note - a side effect of how I managed the parsing and help replies for this PR is the fact that CC's now have command event dispatching, like `on_command` and `on_command_error`. Should CC permissions be included in the future, this may make permissions easier to handle via modifying the fake command's checks attribute before invokation, but as that is outside the scope of the goal of this PR I haven't looked into it.

Kowlin:
Could I request some documentation of some sort (in command help, or its own doc page) for this? This might confuse newer users or users who aren't code oriented.

Zephyrkul:
I added [cog_customcom.rst](https://github.com/zephyrkul/Red-DiscordBot/blob/patch-1/docs/cog_customcom.rst) for documentation. Feel free to give it a read to make sure the documentation makes sense.

Zephyrkul:
I've been messing around with my branch and found that with a few minor changes I can make custom commands work with the Permissions cog.

I've stated before that this is outside the scope of this PR. Combined with that the approach I made, while simple to implement, has a couple downsides. I've created another PR with further details on my own fork:
https://github.com/zephyrkul/Red-DiscordBot/pull/1

I would be willing to merge that PR into this one if the downsides are acceptable. Otherwise I can leave this PR as-is.

Zephyrkul:
Any opinions on the above comment / PR?

mikeshardmind:
I don't see anything necessarily wrong with this implementation, but I would much prefer a string templating based design to conditional formatting. While I can't find a way to break this, that isn't to say it can't be done, or that something upstream couldn't change allowing this to be broken later. With templating, we provide what can be changed, and nothing more.

mikeshardmind:
As for the permissions part, I'd hold off on that. The permissions cog is likely going to be reworked slightly on the backend. However, if you run the global checks on the context object created and it correctly fills in the command (custom-command in this case) name, then permissions will automatically work with it, no extra effort required. The only issue with this, is I'm not certain get_context and how you are parsing does this without registering CCs as command objects on the bot (I have not tested this)

If it does, then it will work out of the box, if it doesn't, I would not worry about this until after the current issues with the permissions cog are resolved, as your fake command objects may need internal modifications to support it not being real and managed by the bot.

Redjumpman:
First I just want to say this is an excellent pull request. I apologize for not getting to this sooner. I really like that you went the extra mile and added in some docs. I've been playing around with this a lot and like @mikeshardmind , I couldn't find a way to break it. String templating is more or less the go to solution for these types of obstacles, but I'm not 100% convinced that it is absolutely necessary in this case. 

I only had three small things requested for change. Just look over my review for more details.

Once these bits are handled I'll give it the green light.

Redjumpman (dismissed):
I encountered some behavior that might be troublesome for new users. If you put the numbering out of order or start with `1` instead of `0`, it will give a help message that doesn't really address what is happening. Here are some screenshots of what I am talking about:

**Using 10 first and 0 last:**
https://gyazo.com/a77a2df8a9b48b57daeba5f1ae2e1500

**Starting at 1 instead of 0:**
https://gyazo.com/26b66081737a5bee2f3486f152ccbed7

Not sure if there is a good solution to that, but at the very list maybe slip in an added note that you need to start at 0, and the numbers must be sequential. As programmers we are used to things starting at 0, but it might not be natural to gen-pop. 

Lastly, I hate to pimp out your PR for this, especially since I know it isn't your fault, but would you care to freshen up the doc string for `cc add` ? 

https://gyazo.com/fce4b7d3f15ff4f06cfc720d168e5709

Zephyrkul:
All points addressed.

### Non-sequential arguments:
https://gyazo.com/6e61853a692fb78d4bb09d587b6b20a8

### Arguments starting at non-zero:
https://gyazo.com/58da58f9f49693fb0cbd340f6bc5c99a
While this one was already documented in cogs_customcom.rst, it was only a little work to "forgive" the user for this transgression so I had it simply correct it and continue parsing.

### Helpstring
https://gyazo.com/34aa26b39556cab9048799d88304e41c

### Miscellaneous
I added a couple other fixes. When creating / editing a custom command, it will verify that the arguments will parse. (Incorrectly formatting the brace-syntax will still do as author, etc. do.) This fixes a bug that I had known about for a while where random-response custom commands would randomly fail if the arguments did not match.

Tobotimus (approved):
Wow @zephyrkul this is great! Thank you so much for this PR. Also thanks @Redjumpman for reviewing and spotting those issues. I've done some testing and reading and I'm more than happy with it.

PR #2064, opened 2018-08-29
[V3 Filter] implement exempting other users/roles from the filter

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

Adds user/role exemptions for the filter. Will also be extending with channel-specific filters


redbot/cogs/filter/filter.py
```diff
@@ -135,65 +315,53 @@ def __unload(self):
...
 
         return removed
 
     async def check_filter(self, message: discord.Message):
         server = message.guild
         author = message.author
-        word_list = await self.settings.guild(server).filter()
+        word_list = set(
+            await self.settings.guild(server).filter()
+            + self.settings.channel(message.channel).filter()
```
Redjumpman: `self.settings.channel(message.channel).filter()` needs to be awaited.

Redjumpman (dismissed):
Aside from the co-routine needing to be awaited on line 363, I only have a few things.

When exempting a user or a role, you can simply type in the role name, but a user must be mentioned. It won't throw an error message in discord, but it will raise a `BadUnionArgument` in the console.

Finally, there isn't a way to remove the exempted users or roles. 

Not sure if this would exceed the scope of the PR, but it would be nice if there was a way to view what roles/users have been exempted as well.

mikeshardmind:
Sorry to tack something else onto the above from RJM...

After the discussion in the discord about adding a more global setting (instead of each cog including 3rd party needing these settings where applicable), #2129 once merged will add such a setting, It also includes adding logic to filter to handle interaction based on the current filter code, but this does somewhat obsolete (some of) this. I did add in filter immunity based on them in that PR, but am not sure how this interacts with the other planned changes your description alludes to.

palmtree5:
I've actually gone ahead and dropped the exemptions stuff in favor of the automated mod action immunity settings introduced in #2129

Redjumpman (approved):
This is good to go now. We can always extend this little by little in the future. No need to have on huge PR to add all the different things.

Issue #1904, opened 2018-06-30
[V3 Filter] Unable to filter sentences


Crossedfall:
# Command bugs

#### Command name

`filter add "<sentence>"`

#### What cog is this command from?

Filter [core]

#### What were you expecting to happen?

Whenever the full sentence is posted within the filtered Discord server, it should be removed by the bot.

#### What actually happened?

Nothing. Sentences will be deleted if you remove the spaces from the trigger phrase, however.

#### How can we reproduce this issue?

1. Create a new filter with spaces
2. Attempt to trigger the filter

![filter gif](https://user-images.githubusercontent.com/26130695/42119867-c0200498-7bd6-11e8-9c8e-689eb104a67a.gif)

_Note: it would seem that some sentences are read whereas most are not. Multiple tests are advised_

#### Red Version
![image](https://user-images.githubusercontent.com/26130695/42119883-e7749ff4-7bd6-11e8-9b0e-75283ea3d080.png)

mikeshardmind:
Unfortunately, that gif isn't a demonstration of anything not working.

The filter will never impact other bots, the bot owner, mods or admins:

https://github.com/Cog-Creators/Red-DiscordBot/blob/3bcf375204c22271ad3ed1fc059b598b751aa03f/redbot/cogs/filter/filter.py#L248-L251

https://github.com/Cog-Creators/Red-DiscordBot/blob/3bcf375204c22271ad3ed1fc059b598b751aa03f/redbot/core/utils/mod.py#L143-L150

if you need to test filter's behavior, if you launch the bot with `--dev` you can use `[p]mockmsg` or you can get someone else to help you test it.

Crossedfall:
@mikeshardmind if you watch the gif to the end you'll note that it filters the message if I remove all of the spaces. That'd imply that it does actually filter messages from a bot owner and guild admin.

If it really isn't supposed to do that, then there might be another issue at play, possibly unrelated to this report. 

Edit: I have tested this with other users and in other servers. I can confirm that even without this oddity it still happens.

mikeshardmind:
Does the audit log confirm it was your bot, and are you running more than one bot on that token?

If the answer to those is "yes" and "no" respectively, something is seriously wrong here.

Crossedfall:
As far as I can tell, filtering words is not tracked in the log; even with the `--debug` flag. No other bots are running with this token.

[...]

PR #2065, opened 2018-08-29
[V3 Filter] fix issue with filtering sentences

palmtree5:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes the issue with filtering sentences outlined in #1904 by actually adding spaces back in when constructing what needs to be added/removed from the filter


mikeshardmind (approved):
Works without issue.

Tobotimus (approved):

PR #2067, opened 2018-08-29
[V3 Audio] Add [p]sing

Zephyrkul:
### Type

- [X] Enhancement

I need my Transistor songs fix.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -5,6 +5,7 @@ import discord
 import heapq
 import lavalink
 import math
+import random
 import re
 import time
 import redbot.core
@@ -689,6 +690,21 @@ class Audio:
             if not player.current:
                 await player.play()
         await ctx.send(embed=embed)
+
+    @commands.command()
+    @commands.guild_only()
+    async def sing(self, ctx):
+        """Makes Red sing one of her songs"""
+        ids = (
+            "zGTkAVsrfg8",
+            "cGMWL8cOeAU",
+            "vFrjMq4aL-g",
+            "WROI5WYBU_A",
+            "41tIUr_ex3g",
+            "f9O2Rjn1azc",
+        )
+        url = f"https://www.youtube.com/watch?v={random.choice(ids)}"
+        await ctx.invoke(self.play, query=url)
 
     @commands.group()
     @commands.guild_only()
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -691,21 +691,6 @@ class Audio:
                 await player.play()
         await ctx.send(embed=embed)
 
-    @commands.command()
-    @commands.guild_only()
-    async def sing(self, ctx):
-        """Makes Red sing one of her songs"""
-        ids = (
-            "zGTkAVsrfg8",
-            "cGMWL8cOeAU",
-            "vFrjMq4aL-g",
-            "WROI5WYBU_A",
-            "41tIUr_ex3g",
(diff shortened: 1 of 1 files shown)


Crossedfall:
What benefit does this actually add to audio? Instead of hard-coding random videos and needlessly bulking up Audio, you could easily accomplish this with a custom playlist. You could go even further by creating an alias called `sing` that plays it.

At best, in my opinion, we should hold off on looking to add meme functions until after we do our final release. At worst, memes should be reserved for cogs.

Zephyrkul:
The functionality of `[p]sing` can't be replicated by aliases and playlists since it chooses one of the voiced Transistor songs at random and plays only that one. This functionality currently can only be gained via a coded command.

I am unsure how this "bulks up audio" considering it makes one import and one command with two local variables. The only reason it's more than eight lines is due to the requirement of formatting via `black`. Hell, its sixteen lines are less than one percent of Audio's nearly two thousand lines, not even counting files that aren't `audio.py`.

As for the benefit, it's a fun little command that gives a nod to Transistor, the game whose main character this bot was named after. While you could argue this is a "meme" function based on the scientific definition, I think this nod to Red's origins helps give this bot part of its unique feel. The fact that it can lead people to a great game (I personally bought and played Transistor because of this funciton in V2) is also a nice benefit.

Also, this functionality was in Red V2, so I doubt the unfounded concern of "bulking up Audio" is much of an issue, either there or here. I could understand delaying this until 3.1 since this PR is so late, but I do believe this should be a command in Audio.

Crossedfall:
You can still use an alias, the only difference is that you need to ensure shuffle is enabled. Different means to the same end, it will randomly pick a song from the playlist and play it from there.

I don't see the merit in saying that Audio is already bulky so there shouldn't be any harm in adding a few extra lines for a meme function. If anything, this should be an added case against adding such non-essential functions in.

I suppose I could rephrase my question a bit better, what benefit is there to adding it in as core? Even acknowledging your given sentiments, it still seems like a function that would be better serviced as a 3rd party cog.

I close by referring back to my prior statement about Audio's current bulk. Adding in non-essential features seems needless and sets a poor standard for future audio development, in my humble opinion. I respect your differing stance on this, but I felt it worthwhile to voice myself here.

Zephyrkul:
It's a good thing my sole argument wasn't regarding Audio's bulk, then.

Your method still causes every song on the list to play, so your method does not reach the same end.

I already told you the benefit to adding it to core. Please feel free to re-read my argument. You wouldn't get the same results from leaving it to 3rd party since anyone can put any identity into their third party - I could, for instance, make a cog filled with references to the song "Paint It Black". Making it core would reinforce Red's core identity despite any third party making their own identity with their commands.

I disagree that this is non-essential. While it's not required to make the bot *function*, I believe it is essential to the bot's *identity*. You wouldn't see me PRing a `[p]paint` command that makes the bot play "Paint It Black" because that goes against the bot's origins and currently established identity.

Twentysix26 (approved):

skeith:
How about this? 

`sing` by default will play a hardcoded playlist for transistor song in shuffle and repeat. It accepts playlist name as optional argument which will play that playlist in shuffle and repeat

Additional update:
I believe that in V3, Red discord bot evolve, thus, I also belive `sing` command should also evolve

Tobotimus:
@Twentysix26 has spoken. This is being added :smiling_imp: 

I think it's good for Red to keep some of its novelty and character. This command adds exactly that. The unique behaviour where it plays a single random song is fitting for this command.

Tobotimus (approved):

PR #2068, opened 2018-08-30
[V3 Core] Support missing .po files

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Sometimes, Red comes without the .po files. If this happens, the bot will be typing forever and never send anything. I just added a small message to explain the problem instead of not responding.

*Note:* This new string isn't translated because... if no translations are found, it will just stay in english.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1066,7 +1066,7 @@ class Core(CoreLogic):
             red_dist = pkg_resources.get_distribution("red-discordbot")
             red_path = Path(red_dist.location) / "redbot"
             locale_list = sorted(set([loc.stem for loc in list(red_path.glob("**/*.po"))]))
-            if locale_list == []:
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -1067,8 +1067,11 @@ def check(m):
             red_path = Path(red_dist.location) / "redbot"
             locale_list = sorted(set([loc.stem for loc in list(red_path.glob("**/*.po"))]))
             pages = pagify("\n".join(locale_list))
-
-        await ctx.send_interactive(pages, box_lang="Available Locales:")
+        
+        if pages == []:
```
Zephyrkul: Since `pagify` returns a generator this will always be `False`.

Tobotimus (approved):
Thanks @retke :)

PR #2072, opened 2018-08-31
[V3 Docs] Update sample cog

aikaterna:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

The v3 example cog on the docs isn't functional and refers to the old discord commands import.

### DIFF
--- a/docs/guide_cog_creation.rst
+++ b/docs/guide_cog_creation.rst
@@ -44,10 +44,13 @@ In that file, place the following code:
 
 .. code-block:: python
 
-    from discord.ext import commands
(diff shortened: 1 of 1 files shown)


Redjumpman:
This is good to go.

Tobotimus:
Good catch, but could you remove the bit creating `self.bot`? It's often obsolete in V3 since the bot object isn't often needed and all commands now take `ctx`, so we can get the bot from `ctx.bot`

Tobotimus (approved):

PR #2081, opened 2018-09-02
[V3] More mention filtering

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Escapes out bad mention content from usernames in the `[p]names` command. Notably, a user with any fake mention in their name or display name won't be rendered as such in output.

### DIFF
--- a/redbot/core/utils/common_filters.py
+++ b/redbot/core/utils/common_filters.py
@@ -17,8 +17,7 @@ INVITE_URL_RE = re.compile(r"(discord.gg|discordapp.com/invite|discord.me)(\S+)"
 
 MASS_MENTION_RE = re.compile(r"(@)(?=everyone|here)")  # This only matches the @ for sanitizing
 
-# negative lookbehind used here to avoid canceling our own filter strategy.
-OTHER_MENTION_RE = re.compile(r"(?<!\\)(<(@?[!&]|#)\d+>)")
+OTHER_MENTION_RE = re.compile(r"(<)(@[!&]?|#)(\d+>)")
 
 # convenience wrappers
 def filter_urls(to_filter: str) -> str:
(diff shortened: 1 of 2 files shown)


redbot/core/utils/common_filters.py
```diff
@@ -16,6 +17,8 @@
 
 MASS_MENTION_RE = re.compile(r"(@)(?=everyone|here)")  # This only matches the @ for sanitizing
 
+# negative lookbehind used here to avoid canceling our own filter strategy.
+OTHER_MENTION_RE = re.compile(r"(?<!\\)(<(@?[!&]|#)\d+>)")
```
calebj: This should be
```py
r"(?<!\\)(<(@[!&]?|#)\d+>)"
```

calebj: Unfortunately, users and roles are still pinged when there's a backslash in front of the full mention. It only prevents it from being formatted as a mention. Also, the character group for `[!&]` should be optional, not the `@`. Something like this:
```py
r"<((@[!&]?|#)\d+)>"
```
called by:
```py
OTHER_MENTION_RE.sub(r"<\\\1>", to_filter)
```
should do the trick.

Regex101 [here](https://regex101.com/r/ofMqwv/1).

mikeshardmind: Yeah, misplaced the optional, fixed. Wasn't aware the ping still happened, so I've modified the escape strategy to avoid that.

calebj (approved):
Looks good to go. The pattern with three groups might be *very* slightly slower than the example with just one, though. It just depends on how much traffic this code sees.

Tobotimus (approved):

PR #2096, opened 2018-09-04
Ignore fuzzywuzzy slow sequence matcher warning

Tobotimus:
Many users who see this warning seem to panic and think it is worth reporting. I'd rather this just be hidden.

### DIFF
--- a/redbot/__init__.py
+++ b/redbot/__init__.py
@@ -1,6 +1,7 @@
 import sys
+import warnings
 import discord
-from colorama import init, Back
+from colorama import init
 
 init()
 # Let's do all the dumb version checking in one place.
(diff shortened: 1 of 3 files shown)


mikeshardmind:
hmmm. I don't really like hiding this one. It gives users who actually understand the message enough info to get a performance boost if they handle it. If we can intercept the warning and make it more clear that this is not a breakage, I'd find such a solution to be better.

Tobotimus:
I'd rather just have it as a logging.info call then. warnings outputs a code snippet and it looks ugly and causes users to panic

mikeshardmind:
We can detect that we have the slow version from inside red right? If so, keeping the warning suppression, and adding the logging.info call for it would be sufficient

Tobotimus:
Yeah that's exactly what I'm doing :smile:

mikeshardmind (approved):

PR #2077, opened 2018-09-01
[V3 Audio] Add queue search

aikaterna:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

### Description of the changes
This adds [p]queue search to Audio so users can search the queue for tracks, returning the track's place in the queue for use with commands like [p]remove and [p]bump.

PR #2097, opened 2018-09-04
[V3 Audio] Local playback

aikaterna:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This builds on top of #2077 and could be used in favor of it.

Filetype support: .mp3, .flac, .ogg

Local tracks can be queued a few ways:

- `[p]play localtracks/foldername/song_to_play.mp3`
- `[p]local play` (interactive)
- `[p]local folder` (interactive, queues an entire localtracks folder)
- `[p]local search` (interactive, queues the user-picked track via menu)

Informational display will favor ID3 tag information, but if it's not available, it'll use the file name.

If you're looking to test this as an existing user:

- You must be using an internal lavalink server, or your jar should live in the audio data folder if it's external.

- Tracks should go in a localtracks folder under the audio data directory. Place tracks inside an additional subfolder. You can have as many subfolders under localtracks as you want, however audio only looks one folder deep for tracks.

For example:

data_directory/cogs/Audio/localtracks/Album Folder Name/Artist - Track.mp3

data_directory/cogs/Audio/localtracks/Another Great Album/Another Artist - Another Track.mp3

are both valid paths for tracks.

- Replace the lavalink.jar in your audio data directory with build no. 584 (https://ci.fredboat.com/repository/download/Lavalink_Build/4825:id/Lavalink.jar?guest=1)
- Change `local: false` in the application.yml file in audio's data path to `true`.
- If your files return "Nothing found.", check your error.log or the lavalink.log in the data directory to see if it was an encoding error.

General error reporting in Discord is disabled for local tracks. Lavalink is not best-suited for local track playback and can complain at the end of every mp3 file but play the entire thing, for example.

Something to be considered before merging: There should be a way to determine if someone has an existing lavalink.jar and application.yml and be able to overwrite them both, or set them to download & overwrite by default on the release this is included in.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -2,21 +2,25 @@ import aiohttp
 import asyncio
 import datetime
 import discord
+from fuzzywuzzy import process
 import heapq
 import lavalink
 import math
+import os
 import random
 import re
 import time
 import redbot.core
 from redbot.core import Config, commands, checks, bank
+from redbot.core.data_manager import cog_data_path
 from redbot.core.utils.menus import menu, DEFAULT_CONTROLS, prev_page, next_page, close_menu
 from redbot.core.i18n import Translator, cog_i18n
+from urllib.parse import urlparse
 from .manager import shutdown_lavalink_server
 
 _ = Translator("Audio", __file__)
 
-__version__ = "0.0.6d"
+__version__ = "0.0.7"
 __author__ = ["aikaterna", "billy/bollo/ati"]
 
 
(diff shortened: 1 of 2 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -1111,22 +1358,34 @@ def check(m):
...
             )
             await ctx.send(embed=embed)
 
     @commands.command(aliases=["q"])
     @commands.guild_only()
-    async def queue(self, ctx, page: int = 1):
-        """Lists the queue."""
+    async def queue(self, ctx, *, page="1"):
+        """Lists the queue.
+        Use [p]queue search <search terms> to search the queue."""
```
Tobotimus: Literally my only qualm with this PR right now, could you add a linebreak in this help message? :D

mikeshardmind (changes requested):
Seems to work fine for the features added. Ran into a small issue at first testing since this was missing bugfixes added since work started, but just merged those locally and was fine.

I did notice something that appears to be by design(especially with comments in the PR description), but I think changing it is preferable.

While I understand not traversing directories infinitely in an interactive menu, It would be nice to be able to specify a specific path contained in local tracks to play from, or a specific file and have it work correctly in any amount of nesting. I have audio files in my music directory as 
```
Music Folder/Artist Name/Album Name/track number - Trackname.extension
```
I can't just symlink this folder to where localtracks is expected due to this limitation on directory traversal.

aikaterna:
That's planned/intended (being able to have a different directory than the internal lavalink dir). Just not in this PR yet.

Also, I won't be able to support infinite folders. It'll need to be one directory deep from the localtracks dir.

mikeshardmind (dismissed):
Modifying prior review to approving since the changes requested are planned improvement in future PR instead, and the changes requested do not necessarily need to happen for the rest of this PR to be valid.

mikeshardmind:
I'm not saying it should support infinite folder depth, but I do think it should support more than just 1 if passed the path directly, and not using the interactive menu.

Tobotimus (approved):

PR #2098, opened 2018-09-04
[V3] removal of invoke_without_command where used

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature

This keyword arg doesn't provide any benefit which we actually leverage, and has some pitfalls as discussed in the discord. This shouldn't impact regular usage.


mikeshardmind:
.... We actually do need one of those behaviors, so the pitfalls need to be addressed differently

PR #2116, opened 2018-09-09
[V3 Core] Support already loaded packages

laggron42:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

When loading an already loaded package, this error shows:
```py
Exception during loading of cog
  File "/usr/local/lib/python3.6/dist-packages/redbot/core/core_commands.py", line 96, in _load
  File "/usr/local/lib/python3.6/dist-packages/redbot/core/bot.py", line 221, in load_extension
    raise discord.ClientException(f"there is already a package named {name} loaded")
discord.errors.ClientException: there is already a package named audio loaded
```

This PRs catch the exception and explain what's wrong.
screen shot 2018-09-09 at 21 33 58


Tobotimus:
Thanks for opening this Pull Request @retke! I've been wanting this for a while :smile:

calebj (changes requested):
While you're working on this, I think you should make the behavior of Red's Bot.load_extension() match upstream. Since the return value of that function is always None, it doesn't hurt to enhance it in Red. I suggest having it return the initialized cog object upon success, None if it's already loaded, or ClientException if there is no setup() function like d.py does.

The reason for the suggested change is: currently, Red's load_extension throws the same exception for both when the module is missing a setup() function AND when the cog is already loaded. Matching the upstream behavior removes this ambiguity and allows your code to distinguish between the failure cases.

Tobotimus (approved):
Thanks @retke :smiley:

PR #2117, opened 2018-09-10
Set 3.6.6 as minimum python version on Windows

Tobotimus:
Due to a bug in ProactorEventLoop prior to 3.6.6, this is necessary, one user reported this last night straight after beta 21 was released.

Relevant bug: http://bugs.python.org/issue26819

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -53,7 +53,7 @@ Red's repository is configured to follow a particular development workflow, usin
 
 ### 4.1 Setting up your development environment
 The following requirements must be installed prior to setting up:
- - Python 3.6.2 or greater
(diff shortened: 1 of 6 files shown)


mikeshardmind (changes requested):
I do not like the special casing here post install, for something that will never work as intended on that version. setup.py should be modified to refuse to install in this case. Doc changes are fine.

setting `python_requires` to the result of the below function should be sufficient to achieve this, without special casing this in multiple locations otherwise
```py
def get_python_versions():
    # # Due to issues with ProactorEventLoop prior to 3.6.6 (bpo-26819)
    if sys.platform == "win32":
        return ">=3.6.6,<3.8"
    return ">=3.6.2,<3.8"
```

mikeshardmind (approved):

PR #2118, opened 2018-09-10
[Audio] More robust startup and unload

Tobotimus:
Currently, Audio is loading before the Lavalink connection is properly established. This can cause problems if a user runs a command during the setup time.

Also, if Audio ever fails to load, the tasks and aiohttp sessions the cog created are destroyed whilst still pending/open. This causes an ugly traceback possibly other side-effects as well.

This change ensures the lavalink connection is established prior to load, and tasks are cancelled properly when audio is either deleted or unloaded.

This does the same for streams, which can have a similar cleanup problem to audio.

### DIFF
--- a/redbot/cogs/audio/__init__.py
+++ b/redbot/cogs/audio/__init__.py
@@ -52,6 +52,6 @@ async def setup(bot: commands.Bot):
         await maybe_download_lavalink(bot.loop, cog)
         await start_lavalink_server(bot.loop)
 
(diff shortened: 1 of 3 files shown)


aikaterna (approved):
From a user standpoint this is a nice improvement for the lack of tracebacks on exit when closed unexpectedly and the inability to use audio commands before the jar is loaded. Thanks Toby.

PR #2119, opened 2018-09-10
[Downloader] Parse tree URLs when cloning repos

Tobotimus:
This addition handles a seemingly common misuse of `[p]repo add`, where users include the trailing `/tree/branch` part of a repo URL. The intended behaviour of how these are parsed should be explained by the supplied test case.

### DIFF
--- a/redbot/cogs/downloader/repo_manager.py
+++ b/redbot/cogs/downloader/repo_manager.py
@@ -491,6 +491,7 @@ class Repo(RepoJSONMixin):
 
 class RepoManager:
 
+    GITHUB_OR_GITLAB_RE = re.compile("https?://git(?:hub)|(?:lab)\.com/")
     TREE_URL_RE = re.compile(r"(?P<tree>/tree)/(?P<branch>\S+)$")
 
     def __init__(self):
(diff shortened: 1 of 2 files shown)


mikeshardmind:
I'll test this in more depth later, it does appear to work, but I think since this is so easily detected and parsed, a more satisfactory solution would cover this behavior for each major online VCS, not just githubs url syntax.

Edit note, my phrasing left that unclear, it's been updated to clarify.

mikeshardmind (changes requested):
Works properly for GitHub, and GitLab. Does not work for bitbucket. Have not looked at others. While I'm not sure it matters enough to special case each other online VCS, I'd at least like a sanity check that we're not trying to do anything fancy with a URL scheme which isn't expected, so I'd like this to not match if not using a supported site, and require the old format in that case (a clone URL + branch name)

Tobotimus:
@mikeshardmind See updated, thanks

mikeshardmind (approved):
Appreciate the quick changes. Have tested, and this now behaves as expected in the github/gitlab case, without modifying other cases without the same defined behavior.

Issue #2031, opened 2018-08-19
Fuzzy search not working


Tobotimus:
# Other bugs

It appears we have an issue with fuzzy search. When enabled in DMs, sending a misspelled command yields this:
```
[19/08/2018 19:47] ERROR events on_error 181: Exception in on_command_error
Traceback (most recent call last):
  File "/home/tobyh/PycharmProjects/Red-DiscordBot/.venv/src/discord.py/discord/client.py", line 220, in _run_event
    await coro(*args, **kwargs)
  File "/home/tobyh/PycharmProjects/Red-DiscordBot/redbot/core/events.py", line 234, in on_command_error
    fuzzy_result = await fuzzy_command_search(ctx, ctx.invoked_with)
  File "/home/tobyh/PycharmProjects/Red-DiscordBot/redbot/core/utils/__init__.py", line 52, in fuzzy_command_search
    is_alias, alias = await alias_cog.is_alias(ctx.guild, term)
  File "/home/tobyh/PycharmProjects/Red-DiscordBot/redbot/cogs/alias/alias.py", line 60, in is_alias
    server_aliases = await self.unloaded_aliases(guild)
  File "/home/tobyh/PycharmProjects/Red-DiscordBot/redbot/cogs/alias/alias.py", line 41, in unloaded_aliases
    return (AliasEntry.from_json(d) for d in (await self._aliases.guild(guild).entries()))
  File "/home/tobyh/PycharmProjects/Red-DiscordBot/redbot/core/config.py", line 761, in guild
    return self._get_base_group(self.GUILD, guild.id)
AttributeError: 'NoneType' object has no attribute 'id'
```

~~And server fuzzy doesn't seem to work at all~~ (Edit: It was just a high ratio, I'm going to lower it). I'll investigate.

Tobotimus:
Having resolved the above traceback on my local branch, I ran into this one which seems to be caused by typing a command similar to a guild-only one whilst in DM:
```
Traceback (most recent call last):
  File "/home/tobyh/PycharmProjects/Red-DiscordBot/.venv/src/discord.py/discord/client.py", line 220, in _run_event
    await coro(*args, **kwargs)
  File "/home/tobyh/PycharmProjects/Red-DiscordBot/redbot/core/events.py", line 231, in on_command_error
    fuzzy_result = await fuzzy_command_search(ctx, ctx.invoked_with)
[...]

PR #2122, opened 2018-09-11
Refactor fuzzy help and clean up help command

Tobotimus:
I'm really sorry this is a big diff. There was lots to refactor though.

What's changed:
- Fixed issues mentioned on #2031
- Fuzzy help displays more like help manual
- Fuzzy help is easier and more flexible to use
- Fuzzy help string-matching ratio lowered to 80
- Help formatter is more extendable
- Help command has been optimized, cleaned up and better incorporates fuzzy help
- Added async_filter and async_enumerate utility functions because I was using them for this PR, then no longer needed them, but decided they might be useful anyway.
- Moved `clean_prefix` from the help formatter to `commands.Context`
- Added `Command.can_see(ctx: Context) -> bool` method, which tries to mimic help formatter behaviour over whether a command can be seen in the current context.

### DIFF
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -4,6 +4,7 @@ This module contains extended classes and functions which are intended to
 replace those from the `discord.ext.commands` module.
 """
 import inspect
+import itertools
 import weakref
 from typing import Awaitable, Callable, TYPE_CHECKING
 
@@ -105,6 +106,37 @@ class Command(commands.Command):
 
             # We should expose anything which might be a bug in the converter
             raise exc
+
+    async def can_see(self, ctx: "Context"):
+        """Check if this command is visible in the given context.
+
+        In short, this tries to emulate the behaviour of the help formatter
+        to see if this command would be displayed when the help command is
+        invoked in the given context.
+
+        Parameters
+        ----------
+        ctx : `~redbot.core.commands.Context`
+            The invocation context to check with.
+
+        Returns
+        -------
+        bool
+            ``True`` if this command is visible in the given context.
+
+        """
+        for cmd in itertools.chain((self,), self.parents):
+            if cmd.hidden:
+                return False
+            try:
+                can_run = await self.can_run(ctx)
+            except commands.CheckFailure:
+                return False
+            else:
+                if can_run is False:
+                    return False
+
+        return True
 
     def disable_in(self, guild: discord.Guild) -> bool:
         """Disable this command in the given guild.
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -116,7 +116,7 @@ class Command(commands.Command):
 
         Parameters
         ----------
-        ctx : `~redbot.core.commands.Context`
+        ctx : `Context`
             The invocation context to check with.
 
         Returns
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -4,7 +4,6 @@ This module contains extended classes and functions which are intended to
 replace those from the `discord.ext.commands` module.
 """
 import inspect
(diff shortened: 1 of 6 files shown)


redbot/core/commands/commands.py
```diff
@@ -106,6 +107,37 @@ def parents(self):
...
+        ctx : `Context`
+            The invocation context to check with.
+
+        Returns
+        -------
+        bool
+            ``True`` if this command is visible in the given context.
+
+        """
+        for cmd in itertools.chain((self,), self.parents):
```
mikeshardmind: Is this really worth importing itertools for? This could easily just be

`for cmd in (self, *self.parents):`

redbot/core/commands/commands.py
```diff
@@ -106,6 +107,37 @@ def parents(self):
...
+                return False
+            try:
+                can_run = await self.can_run(ctx)
+            except commands.CheckFailure:
+                return False
+            else:
+                if can_run is False:
+                    return False
+
+        return True
```
mikeshardmind, line 137: This logic only holds up with the default help formatter behavior, which can be changed while the bot is running by cogs. Do we care about this, or should we query the behavior of how the help formatter handles check failures?

redbot/core/commands/context.py
```diff
@@ -237,3 +237,12 @@ def embed_color(self):
...
             return await self.send(message)
+
+    @property
+    def clean_prefix(self) -> str:
+        """str: The command prefix, but a mention prefix is displayed nicer."""
+        if self.guild is not None:
+            me = self.guild.me
+        else:
+            me = self.bot.user
+        return self.prefix.replace(me.mention, f"@{me.display_name}")
```
mikeshardmind: I think it might be worth adding `me` to `context` as
```py
@property
def me(self):
    return self.guild.me if self.guid else self.bot.me
```
And use that here inside clean_prefix. There are many cases where the user/member object may be desired, and such a shortcut is nice to have.

redbot/core/help_formatter.py
```diff
@@ -20,25 +20,24 @@
 e.g. format_help_for(ctx, ctx.command, "Missing required arguments")
 
 discord.py 1.0.0a
-Experimental: compatibility with 0.16.8
 
 Copyrights to logic of code belong to Rapptz (Danny)
 Everything else credit to SirThane#1780"""
```
mikeshardmind: These credits at the top don't seem accurate anymore. The majority of the logic below is not equivalent to the work Danny or SirThane did. I'd like to see this modified to more appropriately credit their work as the work this is now building off of, rather than all of this being their work.

mikeshardmind (changes requested):
Okay, first of all, this works wonderfully, and I love the improvements this makes from a UX perspective.

There are a couple mostly small things I've noted below, and while I would not call any of them "must change", I'd prefer discussing them if you think there's a reason not to.

redbot/core/commands/commands.py
Tobotimus: Good point, it's not like `parents` is going to be a particular memory-hungry sequence.

redbot/core/commands/commands.py
Tobotimus: I'm happy with `can_see` being a simple combination of check verification and `hidden`, however I am happy to make that more clear in the docstring. This method could yet see a little tweak in the upcoming permissions overhaul too.

mikeshardmind: I guess that's fair enough. Anything calling it should be aware that if the formatter has been told to display all visible commands, this loses meaning, and a doc string change is enough to signal that.

mikeshardmind (approved):
:+1: This one's ready to go.

PR #2126, opened 2018-09-13
[V3 commands] Introduction to Red's converters + TimeConverter

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This PR adds a `converters.py` file, following d.py's style. This comes with the `TimeConverter` converter, which converts a string to a `datetime.timedelta` unit. It can be called manually or be used as a function annotation, like other converters.

You can provide many types of strings and this will be converted to a valid class, thanks @mikeshardmind for the regex thing.

This PR will be needed for the future timed mute, that should be made by me if everything's alright.

### Example

- `5m` > `timedelta(seconds=300)`
- `10 minutes` > `timedelta(seconds=600)`
- `1 second` > `timedelta(seconds=1)`
- `5d` > `timedelta(days=5)`

### DIFF
--- a/redbot/core/commands/converters.py
+++ b/redbot/core/commands/converters.py
@@ -1,0 +1,64 @@
+import re
+from datetime import timedelta
+from . import Converter, Context, BadArgument
+
+__all__ = ["Time"]
+
+
+class Time(Converter):
+    """
+    Convert to a :class:`datetime.timedelta` class.
+
+    The converter supports seconds, minutes, hours and days.
+
+    Time can be attached to the unit (only use the first letter) or not
+    (then use the full word).
+
+    .. admonition:: Example
+
+    *   `50s` = `50 seconds` -> :class:`datetime.timedelta(seconds=50)`
+    *   `12m` = `12 minutes` -> :class:`datetime.timedelta(seconds=720)`
+    *   `4h` = `4 hours` -> :class:`datetime.timedelta(seconds=14400)`
+    *   `1d` = `1 day` -> :class:`datetime.timedelta(days=1)`
+
+    Using the converter with a command:
+    .. code-block:: python3
+
+        @commands.command()
+        async def timer(self, ctx, time: commands.Time):
+            await self.start_timer(time)
+
+    Using the converter manually:
+    .. code-block:: python3
+
+        async def convert_time(ctx: RedContext, text: str) -> datetime.timedelta:
+            time = await commands.Time.convert(ctx, text)
+            return time
+
+    Arguments
+    ---------
+    ctx: Context
+        The context of the command.
+    argument: str
+        The string you want to convert.
+
+    Returns
+    -------
+    datetime.timedelta
+        The :class:`datetime.timedelta` object.
+
+    Raises
+    ------
+    """
+
+    async def convert(ctx: Context, argument: str) -> timedelta:
+        TIME_RE = re.compile(
+            r"((?P<days>\d+?)\s?(d(ays?)?))?\s?((?P<hours>\d+?)\s?(hours?|hrs|hr?))?\s?((?P<minutes>\d+?)\s?(minutes?|mins?|m))?\s?((?P<seconds>\d+?)\s?(seconds?|secs?|s))?\s?",
+            re.I,
+        )
+        matches = TIME_RE.match(argument)
+        params = {k: int(v) for k, v in matches.groupdict().items() if None not in (k, v)}
+        time = timedelta(**params)
+        if str(time) == "0:00:00":
+            raise BadArgument("No time could be found.")
+        return time
--- a/redbot/core/commands/converters.py
+++ b/redbot/core/commands/converters.py
@@ -2,10 +2,10 @@ import re
 from datetime import timedelta
 from . import Converter, Context, BadArgument
 
-__all__ = ["Time"]
+__all__ = ["TimeConverter"]
 
 
-class Time(Converter):
+class TimeConverter(Converter):
     """
     Convert to a :class:`datetime.timedelta` class.
 
--- a/redbot/core/commands/converters.py
+++ b/redbot/core/commands/converters.py
@@ -39,7 +39,7 @@ def timedelta_converter(argument: str) -> timedelta:
         .. code-block:: python3
 
             async def convert_time(ctx: Context, text: str) -> datetime.timedelta:
-                time = await commands.TimedeltaConverter().convert(ctx, text)
+                time = commands.TimedeltaConverter().convert(ctx, text)
                 return time
 
     Arguments
--- a/redbot/core/commands/converters.py
+++ b/redbot/core/commands/converters.py
@@ -13,7 +13,7 @@ TIME_RE = re.compile(
 )
 
 
-class TimedeltaConverter(Converter):
+def timedelta_converter(argument: str) -> timedelta:
     """
     Convert to a :class:`datetime.timedelta` class.
 
@@ -45,35 +45,26 @@ class TimedeltaConverter(Converter):
             async def convert_time(ctx: Context, text: str) -> datetime.timedelta:
                 time = await commands.TimedeltaConverter().convert(ctx, text)
                 return time
+
+    Arguments
+    ---------
+    argument: str
+        The string you want to convert.
+
+    Returns
+    -------
+    datetime.timedelta
+        The :class:`datetime.timedelta` object.
+
+    Raises
+    ------
+    ~discord.ext.commands.BadArgument
+        No time was found from the given string.
     """
 
-    async def convert(self, ctx: "Context", argument: str) -> timedelta:
-        """
-        Manually convert a string to a :class:`datetime.timedelta` class.
-
-        .. warning:: This should not be called as a command function annotation.
-            This is for manual calls.
-
-        Arguments
-        ---------
-        ctx: Context
-            The context of the command.
-        argument: str
-            The string you want to convert.
-
-        Returns
-        -------
-        datetime.timedelta
-            The :class:`datetime.timedelta` object.
-
-        Raises
-        ------
-        ~discord.ext.commands.BadArgument
-            No time was found from the given string.
-        """
-        matches = TIME_RE.match(argument)
-        params = {k: int(v) for k, v in matches.groupdict().items() if None not in (k, v)}
-        if not params:
-            raise BadArgument("No time could be found.")
-        time = timedelta(**params)
-        return time
+    matches = TIME_RE.match(argument)
+    params = {k: int(v) for k, v in matches.groupdict().items() if None not in (k, v)}
+    if not params:
+        raise BadArgument("No time could be found.")
+    time = timedelta(**params)
+    return time
--- a/redbot/core/commands/converters.py
+++ b/redbot/core/commands/converters.py
@@ -48,7 +48,7 @@ class TimedeltaConverter(Converter):
 
     async def convert(self, ctx: "Context", argument: str) -> timedelta:
         """
-        Convert manually a string to a :class:`datetime.timedelta` class.
+        Manually convert a string to a :class:`datetime.timedelta` class.
 
         .. warning:: This should not be called as a command function annotation.
             This is for manual calls.
--- a/redbot/core/commands/converters.py
+++ b/redbot/core/commands/converters.py
@@ -7,8 +7,6 @@ from redbot.core.commands import Converter, BadArgument
 if TYPE_CHECKING:
     from .commands import Context
 
-__all__ = ["TimedeltaConverter"]
-
 TIME_RE = re.compile(
     r"((?P<days>\d+?)\s?(d(ays?)?))?\s?((?P<hours>\d+?)\s?(hours?|hrs|hr?))?\s?((?P<minutes>\d+?)\s?(minutes?|mins?|m))?\s?((?P<seconds>\d+?)\s?(seconds?|secs?|s))?\s?",
     re.I,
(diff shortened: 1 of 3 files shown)


redbot/core/commands/converters.py
```diff
@@ -0,0 +1,74 @@
+import re
+from datetime import timedelta
+from . import Converter, Context, BadArgument
+
+__all__ = ["TimeConverter"]
+
+
+class TimeConverter(Converter):
```
mikeshardmind: This should probably be `TimedeltaConverter`

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,74 @@
...
+        -------
+        datetime.timedelta
+            The :class:`datetime.timedelta` object.
+
+        Raises
+        ------
+        ~discord.ext.commands.BadArgument
+            No time was found from the given string.
+        """
+        TIME_RE = re.compile(
```
mikeshardmind: This is a compiled constant, it belongs outside the class at the top of the file

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,74 @@
...
+            r"((?P<days>\d+?)\s?(d(ays?)?))?\s?((?P<hours>\d+?)\s?(hours?|hrs|hr?))?\s?((?P<minutes>\d+?)\s?(minutes?|mins?|m))?\s?((?P<seconds>\d+?)\s?(seconds?|secs?|s))?\s?",
+            re.I,
+        )
+        matches = TIME_RE.match(argument)
+        params = {k: int(v) for k, v in matches.groupdict().items() if None not in (k, v)}
+        time = timedelta(**params)
+        if str(time) == "0:00:00":
```
mikeshardmind: This disallows a zero second timedelta, which could be allowed by checking against `if not params` earlier in the code instead.

mikeshardmind (changes requested):
I haven't tested the implementation yet, but I already noticed a few things with it. And next time, careful with those GitHub pings.

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,74 @@
+import re
+from datetime import timedelta
+from . import Converter, Context, BadArgument
```
Tobotimus: I know this package is a bit confusing due to the way it imports all of discord.py's stuff to the top level, then writes our stuff afterwards.

When importing something we've overwritten from within the actual `commands` package, import it from the sibling-module, not from the top level. So in this example, it would be `Context`, you want to do `from .context import Context`.

Also, whenever you're doing a relative import which is only for a type-hint, please use [`typing.TYPE_CHECKING`](https://docs.python.org/3.7/library/typing.html#typing.TYPE_CHECKING) and forward references to avoid any circular imports.

redbot/core/commands/converters.py
laggron42: What do you mean? Put arguments outside of the class and keep what's related to time inside the class?

redbot/core/commands/converters.py
laggron42: But if the user sets, for example `2 weeks`, Regex will return 0:00:00, while the user is actually expecting something else. For example, with timed mute, if the user input a wrong unit, what will happen is that the user will get muted and immediately unmuted.

I'm trying to follow `discord.py`'s style for converters, and when it can't convert, it raises `BadArgument` (instead of returning None or something like that).

redbot/core/commands/converters.py
laggron42: okay for `Context`, but what about `Converter` and `BadArgument`? Should I import `discord.ext.commands` instead?

redbot/core/commands/converters.py
mikeshardmind: Your handling of this places `0 seconds` as invalid, when the intent is to not handle `2 weeks`. Meanwhile, if `params` is an empty dict here, you can raise the bad argument there (having matched nothing expected)

redbot/core/commands/converters.py
mikeshardmind: You shouldn't be calling `re.compile` each time a conversion happens. Make it a module level constant, but don't place it in `__all__`'s definition.

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,80 @@
...
+                await self.start_timer(time)
+
+        Using the converter manually:
+
+        .. code-block:: python3
+
+            async def convert_time(ctx: Context, text: str) -> datetime.timedelta:
+                time = await commands.Timedelta.convert(ctx, text)
+                return time
+    """
```
Zephyrkul: Couple issues with these code blocks.
1. lines 39 and 47: it's TimedeltaConverter, not Timedelta.
2. line 48: You're using the class to call an instance's method. You need to use an instance to call an instance method.

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,80 @@
...
+        Using the converter manually:
+
+        .. code-block:: python3
+
+            async def convert_time(ctx: Context, text: str) -> datetime.timedelta:
+                time = await commands.TimedeltaConverter().convert(ctx, text)
+                return time
+    """
+
+    async def convert(self, ctx: "Context", argument: str) -> timedelta:
```
Redjumpman: I'm not sure what ctx is doing here. It's not used for the rest of the code in this function, and the "type" checking is just "Context". Did you mean to check for `commands.Context` ?

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,80 @@
...
+        """
+        Convert manually a string to a :class:`datetime.timedelta` class.
```
Redjumpman: Can we change this to` Manually convert a string ... ` instead of `Convert manually a string...`?

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,80 @@
...
+            The string you want to convert.
+
+        Returns
+        -------
+        datetime.timedelta
+            The :class:`datetime.timedelta` object.
+
+        Raises
+        ------
+        ~discord.ext.commands.BadArgument
```
Redjumpman: '~' character not needed.

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,80 @@
+import re
+from datetime import timedelta
+from typing import TYPE_CHECKING
+
+from redbot.core.commands import Converter, BadArgument
+
+if TYPE_CHECKING:
+    from .commands import Context
+
+__all__ = ["TimedeltaConverter"]
```
Redjumpman: Not sure why dunder all is needed here. TimedeltaConverter is literally the only thing here. Typically you use all to "hide" some of the helper functions that build up the stuff you want developers to use.

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,80 @@
...
+        *   ``4h`` = ``4 hours`` -> :class:`datetime.timedelta(seconds=14400)`
+        *   ``1d`` = ``1 day`` -> :class:`datetime.timedelta(days=1)`
+
+        Using the converter with a command:
+
+        .. code-block:: python3
+
+            @commands.command()
+            async def timer(self, ctx, time: commands.TimedeltaConverter):
+                await self.start_timer(time)
```
Redjumpman: This is not a great example of how these deltas could be used. Perhaps change this very ambiguous classmethod to something like `await asyncio.sleep(time.seconds)`

Redjumpman:
I added some comments to this PR as well. I think this could be a helpful addition, but it needs some cleaning up.

redbot/core/commands/converters.py
Zephyrkul: ctx is required for converters that inherit from d.py's Converter. That said, if there's no need for ctx, perhaps just a normal module function would be better. d.py takes any callable as a converter, after all.

redbot/core/commands/converters.py
Zephyrkul: `time.total_seconds()` might be better so it picks up on days and up, which this converter converts.

redbot/core/commands/converters.py
laggron42: I tried to follow what the `discord.py` docs recommand. Please read this, and if you still want to remove that argument, tell me.
https://discordpy.readthedocs.io/en/rewrite/ext/commands/commands.html#advanced-converters

redbot/core/commands/converters.py
laggron42: Isn't that character supposed to only show `BadArgument` in the docs instead of showing `discord.ext.commands.BadArgument`? 🤔

redbot/core/commands/converters.py
laggron42: I followed `converters.py` as an example for this PR, and tried to match with their style, but if it's not needed I'll remove it 👍

redbot/core/commands/converters.py
Redjumpman: True. `total_seconds` would be better.

redbot/core/commands/converters.py
Redjumpman: If that's the case, I'll resolve this then.

redbot/core/commands/converters.py
Zephyrkul: the discord.py docs recommend using basic converters unless you require ctx. Since you don't require ctx, then a basic converter will likely be a better approach.
https://discordpy.readthedocs.io/en/rewrite/ext/commands/commands.html#basic-converters

laggron42: I find it easier for the users to use the advanced converters, because the core discord.py's converters are advanced. The users are used to call a function like that:
```py
await commands.MemberConverter.convert(ctx, text)
```
Using the basic converters would give this:
```py
commands.timedelta_converter(text)
```
Using a different style is quite a mess, especially since **`redbot.core.commands` inherits from `discord.ext.commands`** and include their converters. We should just keep the same style as the other converters. Just imagine:
- For a member, I call `commands.MemberConverter.convert()`
- For a channel, I call `commands.TextChannelConverter.convert()`
- For a role, I call `commands.RoleConverter.convert()`
- For time, I call `commands.timedelta_converter()`

That doesn't match at all with the other converters and will confuse the users. That's why I'm using advanced converters, so it works the same as the other ones, which are often used.

Basic converters may be useful for a personnal cog, but not for an API that already uses an other type of converters under the same module!

laggron42:
Travis is failing because of a Python 3.6.6 problem.

redbot/core/commands/converters.py
Zephyrkul: discord.py includes basic converters as well, ex `clean_content`. More than that, basic converters are all over the bot, just look at any `int` or `str` typehints. Those are basic converters in d.py's eyes as well. 

The advanced converters are advanced *because* they require `ctx`. The advanced converters require `ctx` because they are getting data about the context to return Discord objects. Your converter does not require `ctx` because it's neither returning any Discord objects nor getting any data about Discord's context.

Creating a convoluted method to convert to match Discord converters which *require* that convoluted method to function is quite simply absurd. You should match the functionality of what you require, keep it as simple as possible, and above all actually match how d.py functions with its typehint conversions rather than make nonsense excuses for what you coded.

Besides which, what happens if I want to convert an argument to a timedelta in a Discord event? Your method *requires* me to have a Context object when it's not needed, so I'd have to somehow find or create one or be SOL. I could pass `None`, but what looks better and is easier to remember in your eyes?

`commands.timedelta_converter(my_string)`
`await commands.TimedeltaConverter().convert(None, my_string)`

mikeshardmind: Actually, on that note @zephyrkul , I'd prefer if the actual conversion happened in a static method which was called from the convert to allow usages without the overhead of a creating a coroutine when being called manually. The only reason this needs to be a coroutine is it's use as a converter, so by offering manual usage as a feature, we should better support it as well.

Zephyrkul: I'd prefer it if it wasn't an advanced `Converter` in the first place, which would remove the coro requirement (as well as the `ctx` requirement).
Did you mean to ping Laggron with that? Your phrasing seems to sound like you're addressing the person coding this PR.

palmtree5:
uhh, Travis isn't failing on this PR atm @retke

redbot/core/commands/converters.py
mikeshardmind: Actually, you're right. There's nothing used here which would need a coro or ctx, so This can easily just be a basic converter.

Tobotimus:
I restarted the job that failed, was just Travis having its periodic connection issues when downloading python

laggron42:
The 10 reviews are closed, is there something else to add?

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,66 @@
...
+        *   ``12m`` = ``12 minutes`` -> :class:`datetime.timedelta(seconds=720)`
+        *   ``4h`` = ``4 hours`` -> :class:`datetime.timedelta(seconds=14400)`
+        *   ``1d`` = ``1 day`` -> :class:`datetime.timedelta(days=1)`
+
+        Using the converter with a command:
+
+        .. code-block:: python3
+
+            @commands.command()
+            async def timer(self, ctx, time: commands.TimedeltaConverter):
```
Zephyrkul: This and line 42 need updating with the new syntax.

redbot/core/commands/converters.py
```diff
@@ -0,0 +1,66 @@
...
+            async def timer(self, ctx, time: commands.timedelta_converter):
+                await asyncio.sleep(time.total_seconds())
+                await ctx.send("Time's up!")
+
+        Using the converter manually:
+
+        .. code-block:: python3
+
+            async def convert_time(ctx: Context, text: str) -> datetime.timedelta:
+                time = await commands.timedelta_converter(text)
```
Zephyrkul: `timedelta_converter` is not awaitable.

Tobotimus:
This was superseded by #2140

PR #2128, opened 2018-09-14
[V3] Hackban fixes

### LINKED ISSUE #2127: [V3] Ban peers with hackban
# Command bugs

#### Command name

[p]hackban

Mod

When called by a mod on another mod, display error message: "I cannot let you do that. You are not higher than the user in the role hierarchy." and do nothing.

#### What actually happened?

Mod got banned, message displayed: "Done. The user will not be able to join this server."

1) Set a role with ban permissions and access to the Mod cog and hackban command.
2) Give the role to two members, it should be the highest role for both of them.
3) Have one of them execute [p]hackban <other's id>

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

If the member is in the guild, delegates to existing ban logic. (Fixes #2127)

Additionally, check that we have ban perms prior to attempting to fetch the existing ban list.

### DIFF
--- a/redbot/cogs/mod/mod.py
+++ b/redbot/cogs/mod/mod.py
@@ -452,9 +452,8 @@ class Mod:
         user = guild.get_member(user_id)
         if user is not None:
             # Instead of replicating all that handling... gets attr from decorator
-            return await self.ban.callback(
-                self, ctx, user, None, reason=reason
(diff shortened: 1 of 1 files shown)


Tobotimus (approved):
Thanks for this PR!

PR #2132, opened 2018-09-15
[V3] json_io atomic write improvements.

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes a deprecation issue with lock usage.

```
Deprecated since version 3.7: Lock acquiring using await lock or yield from lock and with statement (with await lock, with (yield from lock)) are deprecated.
```

Forces a file fsync prior and a directory fsync (where available) after rename to prevent issues with left behind temp files.

### DIFF
--- a/redbot/core/json_io.py
+++ b/redbot/core/json_io.py
@@ -5,8 +5,8 @@ import asyncio
 import logging
 from uuid import uuid4
 
-# This is basically our old DataIO, except that it's now threadsafe
-# and just a base for much more elaborate classes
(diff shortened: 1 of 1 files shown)


mikeshardmind:
Also appears to fix a temp files issue, so maybe there's an unspoken reason that method was being deprecated?

Kowlin:
Can confirm that this does resolve issues with temp files (and possible write speeds) on the Json driver. I would like it if someone else can confirm this too before merging.

Tobotimus (approved):

PR #2134, opened 2018-09-15
[V3] Fix role check in local blacklist/whitelist

### LINKED ISSUE #2133: [V3] local blacklist ignores roles
# Command bugs

#### Command name

localblacklist

core

Adding a role to the local blacklist would/should disallow the person to use commands.

#### What actually happened?

The user with the blacklisted role could sue the command without issues.

1. Blacklist a role
2. Assign the role to someone and let them use any command

#### Code

Support for blacklisting roles was added here:
https://github.com/Cog-Creators/Red-DiscordBot/blob/2784730f7f0d2eff5d59bb9947663210c281c769/redbot/core/core_commands.py#L1492-L1498

But it looks like it was missed during the actual check:
https://github.com/Cog-Creators/Red-DiscordBot/blob/2784730f7f0d2eff5d59bb9947663210c281c769/redbot/core/global_checks.py#L29-L34

mikeshardmind:
fixes #2133 , 

`discord.Role.is_default` is a method, not a property.

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/core/global_checks.py
+++ b/redbot/core/global_checks.py
@@ -26,7 +26,7 @@ def init_global_checks(bot):
         local_blacklist = await guild_settings.blacklist()
         local_whitelist = await guild_settings.whitelist()
 
-        _ids = [r.id for r in ctx.author.roles if not r.is_default]
+        _ids = [r.id for r in ctx.author.roles if not r.is_default()]
         _ids.append(ctx.author.id)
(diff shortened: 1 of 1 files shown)


Tobotimus (approved):
Thanks :smiley:

PR #2137, opened 2018-09-18
Data backend migration

Tobotimus:
This adds a fairly small script for migrating data from JSON to Mongo, and can be extended in the future for other backends if required. The script has a console entry point, `redbot-migrate`.

It operates by loading all of the data from an instance given its name, and dumping it to the destination storage backend. Users can then create a new instance which uses the new backend, and reference specify wherever they dumped the migrated data.

I've also cleaned up some other usage of the `redbot.core.drivers` package throughout the bot - no more lazy loading, driver classes are accessible at the top of the `redbot.core.drivers` package (it makes no sense to have to access a single class through a submodule as well) and some more documentation has been added.

This could also be used (although slightly cumbersome) to restore backed up JSON data to a mongo instance.

### DIFF
--- a/docs/framework_config.rst
+++ b/docs/framework_config.rst
@@ -371,18 +371,3 @@ Driver Reference
 
 .. automodule:: redbot.core.drivers
     :members:
-
-Base Driver
-^^^^^^^^^^^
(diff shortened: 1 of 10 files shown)


Tobotimus:
Needs re-purposing

PR #2140, opened 2018-09-20
[V3] timed/temporary mutes [Don't review current, see comments]

### LINKED ISSUE #2103: [V3 Mod] Timed mute
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality

An included timed mute for the mod cog is requested a lot currently, and would be a nice feature. I'm thinking about something like Dyno for the arguments, and would give something like this:

```
[p]mute guild @someone 30m my random reason
```

`30m` = 30 minutes
`2h` = 2 hours
`4d` = 4 days
...

Please give your feedback about this to know what would be the best, I'm planning to code this for the next beta (if there is one before RC)

mikeshardmind:
### Type

- [ ] Bugfix
- [x] Enhancement
- [X] New feature

Adds support for temporary mutes.

Usage examples:

```
[p]mute channel @Sinbad#0001 1 day being rude
```

```
[p]mute server @Blaze 4hours 20 minutes Won't stop talking about weed in every channel. 
```

```
[p]mute channel @AngryGamer 10 minutes muted while a mod mediates
```

This resolves #2103


mikeshardmind:
Going to modify the unmute task somewhat to handle time resolution a little more closely.

Tobotimus:
I understand the appeal of argparse-style commands, but unfortunately I don't agree that they have a place in core commands; not yet at least. My reasons being:
1. There is no reason why mute should be the only argparse-style command
2. A considerable proportion of our userbase has never used a unix shell and are unlikely to understand how to use this; users are used to positional command arguments as this is the style of all other bot commands
3. We now have `Greedy` and `Optional` converters which would work perfectly for the purposes of temp mute without the break from convention

In the future I would happily accept a utility for more general argparse-style converters, but if it's to be used in the core bot, it should not force any command signatures to change.

mikeshardmind:
Actually, the greedy converter and optional one wont work for this. Greedy behaves more like a typehint for sequence, than a greedy match. And optional does not play nice with arbitrary strings which can exist in a reason. @tobotimus

mikeshardmind:
An example: 

```
mute channel LoudUser wont wait 1 minute for a mod to mediate 5 minutes
```

```
mute channel user could not shut up for 1 minute
```



Parse those unambiguously using d.py builtins ( and any other possible one) and I'll agree. Otherwise I'm not budging on implementation.

mikeshardmind:
Since this has been dropped from RC1, close this if that's the final stance on argparse, and I'll make my own. I'm available on discord, as usual, to discuss up until then. 

ED: discussed, and see below

mikeshardmind:
I've updated this to not use argparse and flags. While both the time and reason remain optional, now the time must be provided first if it is going to be provided.

mikeshardmind:
I've also decided to leave the original timedelta converter without reason support in the PR as I do still think this is a useful utility for other creators, but this can be removed if this is no longer wanted there.

mikeshardmind:
Stuff here needs some more cleaning up, and likely re-handling after or with #2232 . Stale for now

laggron42:
Can you move `redbot.core.converter.timedelta_converter` in a different pull request, or reopen #2126 because I need the converter in my cogs and knowing that this PR is frozen is not really reassuring for me...

mikeshardmind:
As 3.0.0 is in RC phase and not receiving new features, you should use a cog local converter if you need it before 3.1 when this would be included.

mikeshardmind:
Closing this as I will not have the time to have a polished implementation of this before 3.1

The approach used here should likely not be reused after conversations in discord anyway.

Issue #1921, opened 2018-07-07
[V3] Better handling of delete delay / CCs


mikeshardmind:
# Feature request

- [X] API functionality

### Details

So right now, things like custom commands get handled by delete delay by us actually parsing to see if it was a CC,
What if we had a red specific event which cogs could trigger when they are finished with a message which is handled through on_message as if it were a command for things like it? This would allow vastly simplifying how delete delay is handled and also allow handling of it for any 3rd party cogs with similar behavior

something like `on_custom_command(message, cog)` and `on_custom_command_error(message, cog, error)` for signatures.

Tobotimus:
I'm going to freeze this for the moment as I don't see it as a high priority, however I'll give my 2 cents.

I wanted to implement a system to add aliases and CC's to the bot's registered commands, which would also simplify this issue, whilst opening up opportunities to solve others (e.g. conflicting command names). I simply haven't had the time to do it (yet). I'm not treating it as a high priority at this point, and I don't mind if it's done after V3's release.

Those events you've suggested would be fine though. I'm just apprehensive to implement them if we plan on going down the route I've explained above.

mikeshardmind:
Closing this old issue, alias and cc both actually get handled as commands now, and we'll have other means for allowing other "not quite commands" to work this way

Issue #1920, opened 2018-07-05
[V3] Cleanup bot does not clean aliases


pabloleban:
# Command bugs

#### Command name

`[p]cleanup bot`

#### What cog is this command from?

Cleanup

#### What were you expecting to happen?

Cleans command messages (including aliases) and messages from the bot.

#### What actually happened?

It does not delete the alias commands.

#### How can we reproduce this issue?

Create an alias:
`[p]alias add rickroll play https://www.youtube.com/watch?v=dQw4w9WgXcQ`

Call alias:
`[p]rickroll`

Call cleanup:
`[p]cleanup bot 10`

`[p]rickroll` command never gets deleted.

Kowlin:
Can confirm that this is a bug.

pabloleban:
Note: If the `[p]cleanup bot 10` is called with an alias like `[p]cleanbot`, this alias command is deleted.

rguan72:
I saw that this issue had no assignees. Just to double check, do you guys want someone to work on this?

Tobotimus:
@rguan72 Sure, no one is working on it yet as far as I know :)

Tobotimus:
Resolved by #2213.

PR #2144, opened 2018-09-22
[WIP] [V3] Cog-specific command mapping

Zephyrkul:
# DO NOT MERGE - WORK IN PROGRESS

<sub><sup>and also because my code is terrible</sup></sub>

This PR only has the framework for this to function so far. I am currently working on refactoring CustomCommands and Alias to take advantage of the framework currently in this PR.

I've decided to open it early so I could get feedback and discuss the changes as they're done, and to mark this change as being actively worked on. As the framework is already committed to my fork, and as it is a little hacky, I would appreciate feedback on it before I get too far into refactoring CCs and Alias.

Alternate implementation ideas are, as always, welcome.

### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

Allows cogs to create their own command mapping, like `bot.all_commands`, that can resolve command names into commands. These mappings are unique to each cog and cannot overwrite each other. If the bot fails to find a command by that name, it will look through all cog mappings instead before failing. This allows cogs to create specialized commands that can't be created through `@commands.command`.

The mapping is automatically removed from the bot when the cog is unloaded; no `__unload` necessary. A cog can get its mapping via `bot.get_cog_map(self)`.

This change allows specialized and/or dynamic commands, like those from CustomCommands and Alias, to count as full commands as far as the bot is concerned, while still allowing for dynamic changes simply via modifying the cog's own mapping without fear of damaging command data from other cogs or the bot itself.

Once complete, this PR will address #1920 and #1921, since special commands can be found via `bot.all_commands` / `bot.get_command`, simplifying commands like `[p]cleanup bot` that look for command messages. Special commands will also dispatch command events as normal. (CCs already has this feature due to #2051, but this will extend the feature to other special cog commands.)

### Known issues
- The cog mappings are searched through in the order they are loaded, and stop searching when they hit a match. I don't believe this to be much of an issue, especially since the standard commands are always first in line, but I figured that it was worth noting.
- This changes `bot.all_commands` and `bot.get_command` to search through all mappings when read. A `core_map` property is available to read from only the bot's standard commands, ignoring cog mappings. I plan to implement an `others_map` method to return a map that only looks through mappings not belonging to the cog. Writes and deletions to `bot.all_commands` still only affect standard commands, so this won't change that side of the bot's functionality.


Zephyrkul:
Pushed some CC changes. Feel free to take a look while I work on alias next.

Zephyrkul:
Closing this in favor of a planned rewrite later on down the line.

PR #2149, opened 2018-09-25
Permissions redesign

### LINKED ISSUE #1931: [V3] Permissions nested group issues.

Tobotimus:
This PR proposes a complete redesign of how the permissions cog works, how our default check decorators work, amongst some other things.

One major difference is how default permissions are described. An instance of the new `Requires` class in the [`redbot.core.commands.requires`](https://github.com/Tobotimus/Red-DiscordBot/blob/V3/redesign_permissions_model/redbot/core/commands/requires.py) will be attached to every `commands.Command` object and every `commands.Cog` object at the `requires` attribute. This object describes the bot's and user's required permissions for the command, the privilege level required, the overrides for the command/cog, and any checks which should be checked *after* overrides. In my opinion this improves the current design of how there are managed since it is a clear representation of everything related to the command's required permissions, and provides a simple  and efficient API for modifying these requirements during runtime.

How permissions rules are applied to `Requires` objects is another question. One of the main motivations of this redesign was to resolve #1931 in a way which is sane, reliable, and doesn't impact performance. The main focus was trying to delegate complex tasks to the application and/or removal of rules, rather than having them be resolved every time a command is called. This is why now, when a rule is explicitly allowed or denied, rules for supercommands will have rules applied as well. Overrides are now resolved much like a state machine; see the `requires` module for more info on how this works.

This does mean rules stored in the config of the Permissions cog need to be loaded into these objects either when the permissions cog is loaded, or the receiving command/cog is loaded (whichever comes first). Events, setup and teardown tasks have been added to facilitate this accordingly.

API changes:
- Cogs must now inherit from `commands.Cog` (see #2151 for discussion and more details)
- All functions which are not decorators in the `redbot.core.checks` module are now deprecated in favour of their counterparts in `redbot.core.utils.mod`. This is to make this module more consistent and end the confusing naming convention.
- `redbot.core.checks.check_overrides` function is now gone, overrideable checks can now be created with the `@commands.permissions_check` decorator
- Command, Group, Cog and Context have some new attributes and methods, but they are for internal use so shouldn't concern cog creators (unless they're making a permissions cog!).
- `__permissions_check_before` and `__permissions_check_after` have been replaced:  A cog method named `__permissions_hook` will be evaluated as permissions hooks in the same way `__permissions_check_before` previously was. Permissions hooks can also be added/removed/verified through the new `*_permissions_hook()` methods on the bot object, and they will be verified even when permissions is unloaded.
- New utility method `redbot.core.utils.chat_formatting.humanize_list`
- New dependency [`schema`](https://github.com/keleshev/schema)

User-facing changes:
- When a `@bot_has_permissions` check fails, the bot will respond saying what permissions were actually missing.
- All YAML-related `[p]permissions` subcommands now reside under the `[p]permissions acl` sub-group (tbh I still think the whole cog has too many top-level commands)
- The YAML schema for these commands has been changed
- A rule cannot be set as allow and deny at the same time (previously this would just default to allow)

Documentation:
- New documentation for `redbot.core.commands.requires` and `redbot.core.checks` modules
- Renewed documentation for the permissions cog
- `sphinx.ext.doctest` is now enabled

Note: standard discord.py checks will still behave exactly the same way, in fact they are checked before `Requires` is looked at, so they are not overrideable.


redbot/core/utils/chat_formatting.py
```diff
@@ -317,3 +320,24 @@ def escape(text: str, *, mass_mentions: bool = False, formatting: bool = False)
...
+
+
+def humanize_list(words: List[str]):
+    """Get comma-separted list, with the last element joined with *and*.
+
+    Examples
+    --------
+    .. doctest::
+
+        >>>humanized_list(["One", "Two", "Three"])
```
mikeshardmind: This doctest looks like it isn't right....

`humanize` != `humanized`

Tobotimus:
This PR should now be ready for the first round of reviews! I've updated the header with the details. Sorry if I've missed some stuff.

mikeshardmind:
I'll note on the removal of `__permissions_check_before` and `__permissions_check_after`, that I don't think getting rid of them in entirety is a good thing. Getting rid of the after should be mostly fine, but the before did serve an important purpose. 

Neither the previous iteration or permissions, nor this one can properly support a negative condition. (i.e. deny to anyone _not_ is a specific group, without modifying the rule for that specific group to do it ). An example of where this can be problematic actually shows up in how the docs are being modified to hedge the actual intent behind locking audio, and how using this method will actually prevent a discord server owner (not bot owner) from further restricting those commands. These extra hooks allowed for quick extra rules to be added for people with such cases, without needing to design an entire permission model to make that possible.

redbot/core/commands/requires.py
```diff
@@ -0,0 +1,667 @@
...
+    "guildowner",
+    "guildowner_or_permissions",
+    "admin",
+    "admin_or_permissions",
+    "mod",
+    "mod_or_permissions",
+]
+
+_T = TypeVar("_T")
+GlobalPermissionModel = Union[
```
mikeshardmind, line 57: This is being used as a union converter in a couple places, which isn't safe. There's no such converter for guild, breaking every attempt to add a rule not via acl, and also the error from this is not user facing (traceback in console, but silent bot response).

I'd suggest adding a guild converter in our subclassed commands to accommodate this.

mikeshardmind (changes requested):
Setting via ACL appears to be working as intended. I've not found any major issues. I have found an inability to set rules via non-acl methods, (cause: converter, see comment).

This is a much more sustainable solution than before and seems to be performing well. With a fix for the converters that are problematic, this should be ready to go.

mikeshardmind (approved):
All changes needed have been made.

PR #2155, opened 2018-09-29
[V3 Audio] Playlist list and notify msg changes

aikaterna:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

Notify messages now have currently playing song duration and the requester in the footer.
[p]playlist list has had its output revamped to be more descriptive and include number of tracks and playlist author.
Volume now cannot be set to less than zero. Thanks to @retke for reporting it.
Removed an unused import.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -1,5 +1,3 @@
-import contextlib
-
 import aiohttp
 import asyncio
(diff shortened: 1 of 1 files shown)


Tobotimus (approved):
Awesome, thanks. Works from my testing :smile:

PR #2158, opened 2018-09-29
SQLite config driver

Twentysix26:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

I based this work on the JSON flat files driver since, except for the different source of data, they should be quite similar.
This should prevent the same issues fixed by #1458 from cropping up in this driver too.

Open points:
1) Currently there's no config whatsover if SQLite is chosen as data backend.  
The database filename is hardcoded as "database.db" and is saved in the root of the user chosen data folder, just like flat files.
Should there be a config process to open up more customization options? If so what would you like to see?
Interesting ones might me allowing the db to be [memory-only](https://www.sqlite.org/inmemorydb.html) (for test purposes or similar) and [WAL mode](https://www.sqlite.org/wal.html) (assuming it works for our use case)
2) At the moment I'm doing db operations by running them in a executor, bounded by a driver-wide lock. SQLite doesn't like different threads using the same connection, *however*, from what I've read it's safe to do this when you're using a lock. It must be tested thoroughly to see if this method causes any issues under heavy writes.


Twentysix26:
I made a ghetto cog to test the performances of this driver.  
As it turns out, SQLite with _Journal Mode_ set to **DELETE** (which is the default) is significantly slower than the flat files' driver. However, when set to **WAL** it is faster than flat files.

Each entry of the following results are the average of 10 rounds of config's `.set()` ran for 10 consecutive times, on Windows

```
Flat files:
0.05227529800000017
0.036639565700000175
0.040477251600000486
0.040705172199999494
0.04258879390000061
0.038067932200000244
0.0418592610999994
0.03912166909999968
0.04689988030000016
0.04548916160000047

SQLite (Journal mode: DELETE):
0.07245000770000445
0.0721971821999972
0.07643642170000078
0.0758659764000015
0.07444021469999881
0.07434559920000083
0.0752516328999974
0.08126998639999954
0.07271632479999539
0.08469479320000062

SQLite (Journal mode: WAL):
0.012791467299996384
0.013603470100005666
0.01428761209999152
0.012186546999998882
0.014279037300002528
0.013228753399997118
0.011774692500000582
0.013407830099993134
0.011949846999999635
0.012053739600000312
```

mikeshardmind:
Since the downsides of `WAL` are circumvented by the fact that we are holding current values in memory, given this benchmarking, It's likely worth enforcing this with 
```py
conn.execute('pragma journal_mode=wal')
```

Given how the json driver is designed, this may still see a lower performance overall, as the json driver can handle multiple simultaneous writes (from separate config instances), whereas this cannot. Any time where the amount changed in the table is close to the size of the table as a whole is a performance loss with this solution, so the mod cog on large servers is an obvious winner here, but other cogs may have worse performance with this

A possible way around this would be for each cog to have its own DB, much like each cog currently has it's own json file with the json driver. I'm not entirely certain this would be worth it, and would want to see the actual performance differences in a real environment before deciding, however, doing this from the start would be easier than changing it later.

Twentysix26:
A different db for each cog doesn't really feel clean to me but I can't say I have a better solution for this particular problem. Perhaps it's worth making a test case with multiple cogs doing heavy writes at the same moment and see how much of an impact they have on the overall performances, so we can decide how to proceed with this driver

Tobotimus:
In my opinion the writes need to be much smaller, and we could also rethink our caching method. One way to do this would be to store less information in the JSON objects and more in schema names, table names and rows. This would mean much smaller writes and would also result in faster lookups, possibly moving towards not using our current caching method.

Here would be my design:
- Each Config has a schema named `CogName2626` where 2626 is the unique cog identifier.
- Each schema contains tables, one for each registered scope, i.e. a table for guild data, user data, member data etc.
- Each table would have the primary key be the object's ID (guild ID, user ID etc.) and the JSON data would be stored in the data column. The only exception to this would be the table for member data, which would have a coupled primary key across two rows - (guild ID, user ID).

I think it'd also be beneficial to take advantage of [JSON1 extension](https://www.sqlite.org/json1.html), in particular the `json_set` method, and perhaps the `json_extract` method if we rethink the caching design.

I also don't really think a ThreadPoolExecutor is the way to go for SQLite. Considering SQLite is just file handles, and our writes will tend to be small if using the aforementioned JSON functions, just using the driver synchronously could be best. Otherwise a design like [aiosqlite](https://github.com/jreese/aiosqlite) could be good, which utilises the `queue` module and runs within a single shared thread per connection.

Twentysix26:
@Tobotimus I think it's certainly a valid proposal. I'll look a bit more into the schema though and see if it's worth having each category in a separate table (it would be 5 tables for each cog+identifier)

mikeshardmind:
I was under the impression the json1 extension for SQLite was not guaranteed to be available on the platforms we support. If we were to require PostgreSQL availability, we could make use of json columns to mirror current storage of dictionaries (for nested values). Otherwise, we'd need to either take steps to ensure availability of the extension (likely during install of red) or handle load/unload of these objects in the driver (serialization to text fields?)

Twentysix26:
Seems pretty solid from my tests, I'd be curious to know the results from a server with *much* higher traffic though.
I'm not confident I have completely tested every functionality of Config with this so be wary

Drapersniper:
After a conversation with @Twentysix26  in discord. Closing this PR in favour of #3922, or any other more up to date implementation.

PR #2159, opened 2018-09-29
[V3 Config] Support MongoDB URIs

yamikaitou:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Allows the user to specify using mongodb or mongodb+srv. Specifically adds support for MongoDB Clusters through use of DNS SRV records. This does add an additional dependency of dnspython

### DIFF
--- a/redbot/core/drivers/red_mongo.py
+++ b/redbot/core/drivers/red_mongo.py
@@ -9,7 +9,7 @@ _conn = None
 
 
 def _initialize(**kwargs):
-    uri = kwargs["URI"]
+    kwargs.get("URI", "mongodb")
     host = kwargs["HOST"]
     port = kwargs["PORT"]
     admin_user = kwargs["USERNAME"]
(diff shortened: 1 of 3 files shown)


redbot/core/drivers/red_mongo.py
```diff
@@ -9,18 +9,24 @@
 
 
 def _initialize(**kwargs):
+    uri = kwargs["URI"]
```
Tobotimus: This breaks existing installations using Mongo since URI won't be a key in its storage details. I'd be happy for you to simply change this to `kwargs.get("URI", "mongodb")` as a solution.

Tobotimus (changes requested):
Thanks for this PR. Just one thing needs to be fixed, otherwise I think it's good :smiley:

redbot/core/drivers/red_mongo.py
yamikaitou: Change made, thanks

Tobotimus (approved):
Thank you for this PR :smiley: All good to go for me.

PR #2161, opened 2018-09-30
[V3 Mod] Remove tempban from data after unbanning

### LINKED ISSUE #2160: [V3 Mod] Tempban doesn't remove the user from settings on unban
# Command bugs

#### Command name

tempban

Mod

When the user is unbanned and then permanently banned, they remain permanently banned.

#### What actually happened?

The user is unbanned by the bot on the next loop due to their previous tempban having expired.

1. Tempban a user.
2. User is unbanned and rejoins.
3. User is permanently banned. Another tempban overrides the settings, so an extended tempban sidesteps the issue.
4. User is unbanned by the bot when the unban task loop runs again.

Zephyrkul:
### Type

- [X] Bugfix

Resolves #2160 
Also resolves another issue I found where the bot will error out when getting the guild's invites when it doesn't have the Manage Server permission.

### DIFF
--- a/redbot/cogs/mod/mod.py
+++ b/redbot/cogs/mod/mod.py
@@ -714,10 +714,13 @@ class Mod:
         to send the newly unbanned user
         :returns: :class:`Invite`"""
         guild = ctx.guild
-        if "VANITY_URL" in guild.features and guild.me.permissions.manage_guild:
-            # guild has a vanity url so use it as the one to send
-            return await guild.vanity_invite()
-        invites = await guild.invites()
+        if guild.me.permissions.manage_guild:
+            if "VANITY_URL" in guild.features:
+                # guild has a vanity url so use it as the one to send
+                return await guild.vanity_invite()
+            invites = await guild.invites()
+        else:
+            invites = []
         for inv in invites:  # Loop through the invites for the guild
             if not (inv.max_uses or inv.max_age or inv.temporary):
                 # Invite is for the guild's default channel,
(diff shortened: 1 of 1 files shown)


redbot/cogs/mod/mod.py
```diff
@@ -1392,23 +1395,24 @@ def __unload(self):
...
+                            await self.settings.member(member(uid, guild)).banned_until()
+                        )
+                        now = datetime.utcnow()
+                        if now > unban_time:  # Time to unban the user
+                            user = await self.bot.get_user_info(uid)
+                            queue_entry = (guild.id, user.id)
+                            self.unban_queue.append(queue_entry)
+                            try:
+                                await guild.unban(user, reason="Tempban finished")
+                                guild_tempbans.pop(i)
```
Tobotimus: If two tempbans are finished in this same `for` loop, the 2nd one will pop the wrong one off the list. After the first one is removed, elements in `guild_tempbans` are shifted meaning `i, uid` becomes misaligned with it. Best to use `.remove(uid)`, no need to enumerate (tempting I know due to extra performance of `.pop(i)`)

Tobotimus (changes requested):
Thanks for this PR. Right idea, just one thing needs to be fixed.

Zephyrkul:
Changes addressed, thanks for the sharp eye.

Tobotimus (approved):
Thanks for this PR @zephyrkul :heart:

PR #2162, opened 2018-09-30
[V3 Trivia] Handle smart quotes

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

Adds a new filter function for substituting out smart-quotes.

Makes trivia use it.


Twentysix26:
My only nitpick here is that the name of the function looks pretty weird with `un` being separate and `smartquote` as a single word. As far as I know the `un` prefix is never used as a word on its own in English.  
I would have went with something like `normalize_smart_quotes` (`replace` could work too), which also respects the other function names being plural instead of singular.
In this case the `to_filter` parameter name should also be renamed since you're not filtering, you are transforming text.

Tobotimus (approved):
Thanks for this PR :+1:

PR #2037, opened 2018-08-21
[V3] Docker Compose

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

### Description of the changes

This is a WIP for a proper docker-compose solution with persistence.
This hasn't been tested at all yet, even a simple run. It's a quick skeleton to get the direction this is going out there and open to discussion.

PR #2163, opened 2018-09-30
[V3] Docker Images for x86/armhf

SuperSandro2000:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

I did some docker images for red-discordbot. I based this contribution of @mikeshardmind pull request #2037 but redid they entire iamges from the ground up with two main goals: small images and armhf support. I would have loved to base those images on Alpine but as it only supports openjre8 it was't possible. Docker requires at least openjre9 or even better 10, which Debain Buster has, to achieve at least acceptable audio quality even on the Raspberry Pi 3b.

The 4 docker files support x86 and armhf architecture and can be build from source or pip. This is represented by the file extension of the Dockerfiles.

|arch|    source    |       pip        |
|----|--------------|------------------|
|x86 |  Dockerfile  |  Dockerfile.pip  |
|arm |Dockerfile.arm|Dockerfile.pip.arm|

The Docker Image names should look like:

|arch|  source  |    pip   |
|----|----------|----------|
|x86 |  source  |  latest  |
|arm |source-arm|latest-arm|

pip is comparable to release.

Example images are being build here: https://hub.docker.com/r/supersandro2000/red-discordbot/.

I would suggest using Docker Hub to build the images. If any help is required I am more than willing to help. If another CI service would be more preferable like Travis I am more than happily to adapt the travis.yml.

This PR almost fully deprecates #2037 except that it doesn't use MongoDB. If it is wished I could add 4 more Images with MongoDB support in them.


Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM debian:buster
+
+MAINTAINER Sandro Jäckel <sandro.jaeckel@gmail.com>
+
+WORKDIR /app
+
+COPY ["docker/config.json", "/root/.config/Red-DiscordBot/"]
```
mikeshardmind, line 7: You either didn't properly setup the file needed, or (more likely) this got excluded because you also did not update `.gitignore` to allow this file to be added automatically

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+
+RUN apt-get update \
```
mikeshardmind, line 9: From a stylistic point, line continuations should be indented.

Dockerfile.pip.arm
```diff
@@ -0,0 +1,35 @@
...
+
+RUN [ "cross-build-start" ]
+
+WORKDIR /app
+
+COPY ["docker/config.json", "/root/.config/Red-DiscordBot/"]
+COPY ["docker/libconnector.so", "/files/"]
+COPY ["docker/run.sh", "/files/"]
+
+RUN apt-get update \
```
mikeshardmind, line 13: same as above with the indenting line continuations

docker/run.sh
```diff
@@ -0,0 +1,4 @@
+#! /bin/bash
+mkdir -p /data/cogs/Audio
+cp -u /files/Lavalink.jar /data/cogs/Audio/
```
mikeshardmind: This shouldn't be needed, Red will fetch Lavalink on first audio load if not there.

mikeshardmind (changes requested):
I appreciate the changes made to support armf. There are a couple small things I'd like to see changed.

 I have no issue with defaulting to the json driver instead of mongo, this can be changed after the fact by anyone preferring mongo, and the stability issues with the json driver have been fixed.

docker/run.sh
SuperSandro2000: The problem is that the fetched jar does not support armhf out of the box and needs manual patching.
It is missing natives/linux-arm/libconnector.so. I am unsure if this can be patched in a better way or we can add the .so file to the build process as I am never touched maven in my life. If you can think of a better way to fix that on arm I am very open to do it this way.

mikeshardmind: We should be including a jar with armhf support in releases (To be handled.) We don't switch lavalink versions without reason, so ensuring this is patched first should be reasonably accommodated. That said, this is something which has only been discussed, not actually acted on so far.

SuperSandro2000: I would suggest merging this until a arm lavalink.jar is provided as I take the one from the releases tab.

SuperSandro2000:
Finally all builds are building on my docker hub.
https://hub.docker.com/r/supersandro2000/red-discordbot/tags/

If it would be possible I would like that more people could test this feature wise as I currently almost exclusively tested the audio streaming on my RPI3b.

causefx:
I can confirm that https://hub.docker.com/r/supersandro2000/red-discordbot/ works....

image

SuperSandro2000:
I reworded it a bit and thanks for the report. I wanted to say that it would be great if someone with maybe more experience than me could test if some feature/cog broke because a missing package or some other querk. I tried to keep the image as bare bones as possible which is quiet hard with a headless jdk and python.

causefx:
So i tired installing some plugin from Jumper and i'm not sure if the issue lies with the plugin or with my install...
https://github.com/Redjumpman/Jumper-Plugins/issues/153

SuperSandro2000:
The last line states that the cog needs an update and I want to add that my intention isn't to support 3rd party cogs that may require more installed packages to keep the image as slim as it can be.
FYI: Alpine got some openjdk 12 images yesterday. I am not sure how nice they would work with this. I wouldn't promise to much gain from them either. They are already 200 MB (I assume uncompressed) and adding python on that will probably yield the exact same or marginally smaller size. Additionally Alpine is always a bit behind in openjdk releases so I probably don't switch bases.

causefx:
The issue wasn't your image, it was the plugin.  It is fixed on their end and we are solid now.

mikeshardmind:
I was able to trim minimal amounts of space off the base image size using alpine. I think the choice to use debian buster is fine given your reasoning and we can always change this later should the size difference become more noticeable in the future. I haven't tested this as extensively as I'd like to have yet, but this seems to be working well.

SuperSandro2000:
I noticed that on some armhf machines audio streaming does not work. I tried it with my friends rpi 3b and it just didn't work with no error messages despite it working on mine.
And switching to Alpine is no option as it probably don't not have any jdk at all in the future.

SuperSandro2000:
I tested this image at a friend's place and together we found two little issues.

calebj:
I've been working on my own docker deployment using compose, and suggest the following for your PR:
1. Use the official [fredboat/lavalink](https://hub.docker.com/r/fredboat/lavalink/) image, since it removes the java dependency in the red image.
2. Switch to python:3.7-alpine for a smaller image size. The [discord.py rewrite](https://hub.docker.com/r/gorialis/discord.py/) image is a good place to start, and I haven't had any problems with it so far.
3. Build the container using the source tree instead of pulling the tarball from github.
4. Include a configuration variant to run a MongoDB container in the composition, or just to specify the Mongo URI in the environment.

I'll probably PR what I have when I'm happy with it, since making Lavalink external requires a change to the audio cog's configuration handling.

If you consider these changes, you'll also need to include instructions for building the Lavalink image on ARM. The openjdk image that Lavalink builds upon is available for that and several other architectures, and the Lavalink image is simple enough to be its own standalone `build:` entry.  I'll experiment with it on my armbian board sometime soon.

**UPDATE:**
Getting lavaplayer to build is a mess, and is apparently required for Lavalink v2. Until Red can use Lavalink v3, it looks like the libconnector.so hackery is necessary on any non-x86 architectures. I'm opposed to including binaries in the source tree, especially if they're architecture specific.

SuperSandro2000:
1. I don't know why we should use a 3rd party image for that. This creates more problems and unecessary dependencies.
2. No matter of it is python Alpine or not all official python images are very big and bulky. Nevertheless this is not an option because Alpine jdk is only officially jdk8 and has some jdk12/13 builds that are very vague and not guaranteed to stay. So no.
3. What difference does that make? I am not sure and shouldn't matter.
4. I could make 4 more variants with Mongodb but it was discussed that this is not necessary.

If you are not happy with the hacky labeling solution then build one do a PR that builds it for arm and if you test it change the volume of the bot because then it always crashes.

EelcoLos:
@SuperSandro2000 I'm running your docker image in a container atm, ty for building it :)
2 questions: 
1: can you update the docker image (build) to rc2
2: can you volume the /data folder, this way the data can be retained over container builds

SuperSandro2000:
1. Is building. I can't say for sure if it builds because I did not look at the upstream changes. Would be appreciated if someone could test it , too. 
2. isn't it already volumed? anyway the compose file is only a suggestion and you are free to modify it.

I may or may not incoperate this image in my new docker build system with daily updates. It would have the latest source and release builds together with a multi-arch image. Could be done over the weekend. No promises.

EelcoLos:
> 
> 
>     1. Is building. I can't say for sure if it builds because I did not look at the upstream changes. Would be appreciated if someone could test it , too.
> 
>     2. isn't it already volumed? anyway the compose file is only a suggestion and you are free to modify it.
> 
> 
> I may or may not incoperate this image in my new docker build system with daily updates. It would have the latest source and release builds together with a multi-arch image. Could be done over the weekend. No promises.

1. your build has succeded, my watchtower container updated the image immediatly.
2. no, it isn't volumed in the dockerfile. I'll suggest it ;) 👍

SuperSandro2000:
> 2. no, it isn't volumed in the dockerfile. I'll suggest it ;) 👍

then I misunderstood you. I am not a fan of that. We have an example docker-compose.yml that includes that.

SuperSandro2000:
If someone is interested: I am doing new images with daily rebuilds here https://hub.docker.com/r/supersandro2000/reddiscord/

mikeshardmind:
I'm not in favor of using this as is. 

I like a lot of what's gone into this, but I think promoting a docker compose solution which actually utilizes the advantages of using compose makes more sense here, i.e. using the official lavalink image as one of the services started as well as a mongo image for handling storage.

This would enable significantly smaller image sizes, as well as prevent re-downloading the lavalink jar (among other things, including not needed to bundle natives ourselves)

As such, I'm recommending this be put on hold until cog-creators/red-lavalink#41 is merged, and the approach used here be modified to leverage this.

SuperSandro2000:
TL, DR:
- I want a lavalink.jar with libconnector.so in it
- I don't see any advantage using MongoDB over json
- Reasons not to use the official Lavalink image:
  - is just the jar in the image.
  - it has no armhf support
  - we have no control over it


> using the official lavalink image as one of the services started

To run Lavalink in a separate container would be very docker-like but the official Lavalink image has several problems:
- It does not support armhf
- to support armhf we would have to build it ourselves
- it is literally just a copy of the binary so we could just download a jar into the docker image and have much more control over it

> as well as a mongo image for handling storage.

I just used the most basic setup for Red. I don't see the need to run mongodb as it does not provide to much except another running docker container. Maybe for a giant discord with multiple thousand of users or if you run the bot on 10 servers it would be beneficial but at this point you most likely use a custom solution for your needs anyway.
I see the docker images as a quick way to try the bot out without throwing tons of stuff at your system you need to cleanup afterwards if you decide not to use it anymore.

> This would enable significantly smaller image sizes

No not really. We could only reduce it if we download lavalink on run and not on compile them and this could lead to dangerous bugs.
I could even imagine that using the lavalink image could significantly increase the image size as we would increase the layers by 2 to 3. The end user would need to download all layers with potentially unused diffs. And again the lavalink image just copys the jar in and has no armhf support.

> as well as prevent re-downloading the lavalink jar (among other things, including not needed to bundle natives ourselves)

As long as fred does not offer a armhf jar we need to repack it anyway. If such a jar would be provided on the releases page then it would reduce the complexity of the arm images drastically. Or make the startup script aware of the system architecture and led it download the correct version of the jar. It does not do this right now so I need to manually download it, repack it and then force the images always to use the repacked jar or the image breaks (hence the copy on startup). I could have moved the jar to another place in the image but I am not familiar with that in Red.

> As such, I'm recommending this be put on hold until Cog-Creators/Red-Lavalink#41 is merged, and the approach used here be modified to leverage this.

Sure we can wait until this is merged but I would suggest to build a proper lavalink.jar for armhf I don't need to monkey patch with libconnector.so. I am not familiar with java or even cross compiling so I would need someone else to do this.

mikeshardmind:
I'm not inclined to go out of our way to support armhf with docker, and I'd actually consider this an edge case on supported hardware (to be clear, the desire to use docker on armhf, not armhf itself.)

With that said, see sedmelluq/lavaplayer#137 for more details on larger architecture support. It's being worked on, but should not be something we monkey patch in on our end. 

If this is a problem for you, as you've already said, anyone can create their own docker images. I'm just not inclined to support an official docker setup which does things in this manner.

SuperSandro2000:
Then I would suggest we wait until this is merged and then continue this.

Let's say I do a mongodb image. Which would be the right approach?
- Another 4 images.
- add an interactive setup to chose the backend
- Or should I just drop the source images to reduce the amount of images and at 2 mongodb  images that need another container running with the db?

mikeshardmind:
I personally believe the best option here is to provide a working compose setup for mongo as the backend with lavalink and mongo both as separate running images in compose. Providing this should also provide enough that anyone knowledgeable can tweak this to suit their own needs, including, but not limited to, using json instead of mongo.

This would then come down to being 2 images: latest(stable) and develop.

SuperSandro2000:
My plan for now: 
- wait until the mentioned above PR is merged
- do a multi-arch lavalink image if it is needed
- do a multi-arch Red image
- do the compose file

Multi-arch images are just a fancy abstraction to remove the need to chose between latest and arm-latest but we need a CI for that cause the build service of docker hub does not support it.

mikeshardmind:
I think it's worth waiting unless someone else wants to create this in the short term with the proper warnings about supported architectures included in the docs. 

I'd want to see some size benchmarking on multi-arch images. If it doesn't add significantly to the image size, I don't have an objection to doing it that way, but I'd prefer separate arm images otherwise (and would be willing to help there if needed)

SuperSandro2000:
I planned using docker manifests. The official multi arch images should use this, too. If you want to read something about them here is a short article https://link.medium.com/VegaeCurlS .

In a nutshell: we still build the amd64 and armhf images separately but after pushing them create a manifest that points to the correct image for your architecture.
eg: we have the tag latest that is a manifest and points to amd64-latest for amd64  machines and to arm-latest for armhf  machines.
It could look something like this https://hub.docker.com/r/supersandro2000/reddiscord/tags/

If you are interested how I created those tags you can take a look here https://github.com/SuperSandro2000/docker-images/blob/master/reddiscord/Makefile#L48

Drapersniper:
Do we have any udates on the status of this

SuperSandro2000:
I am no longer using RedDiscord because music streaming did not work very reliable and kept having issues.

Drapersniper:
That issue is unrelated to Red. But okay  so is it safe to assume you are no longer working on this issue?

SuperSandro2000:
Nope, I don't have planned to work in this.

PR #2164, opened 2018-10-01
[Mod] Enhanced [p]hackban

Twentysix26:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This PR makes use of the recently introduced Greedy and Optional to enhance the `[p]hackban` command, allowing multiple user IDs to be passed. It also adds the optional "message deletion days" parameter.

Work in progress and not completely tested yet (especially the modlog's case creation)
Feedback welcome


redbot/cogs/mod/mod.py
```diff
@@ -39,7 +68,7 @@ class Mod:
 
     def __init__(self, bot: Red):
         self.bot = bot
-        self.settings = Config.get_conf(self, 4961522000, force_registration=True)
+        self.settings = Config.get_conf(self, 4_961_522_000, force_registration=True)
```
Kowlin: Black went a bit nuts here.

redbot/cogs/mod/mod.py
```diff
@@ -1286,8 +1349,8 @@ def __unload(self):
             user = author
 
         #  A special case for a special someone :^)
-        special_date = datetime(2016, 1, 10, 6, 8, 4, 443000)
-        is_special = user.id == 96130341705637888 and guild.id == 133049272517001216
+        special_date = datetime(2016, 1, 10, 6, 8, 4, 443_000)
+        is_special = user.id == 96_130_341_705_637_888 and guild.id == 133_049_272_517_001_216
```
Kowlin: And here too.

Kowlin (dismissed):
Black went on its own way and said that our int's aren't good enough.

redbot/cogs/mod/mod.py
Twentysix26: I think it does that for readability but it's not really useful for us in this case. I guess we have to tweak it a bit.

redbot/cogs/mod/mod.py
```diff
@@ -429,62 +448,111 @@ def __unload(self):
...
-        if user is not None:
-            # Instead of replicating all that handling... gets attr from decorator
-            return await ctx.invoke(self.ban, user, None, reason=reason)
-        user = discord.Object(id=user_id)  # User not in the guild, but
+        for user_id in user_ids:
+            user = guild.get_member(user_id)
+            if user is not None:
+                # Instead of replicating all that handling... gets attr from decorator
+                try:
+                    await ctx.invoke(self.ban, user, days, reason=reason)
```
Twentysix26: Edge case still to be fixed here: at the moment if the invoked ban fails, the banning is still successful as far as the hackban command is concerned.

Twentysix26:
I think I covered all the edge cases after the last commit. Should be ready to be tested fully.

Kowlin:
Gonna give it a shot.

aikaterna:
If you hackban someone and the mod-log channel is not set to a channel, you get this error:
```
[05/10/2018 15:28] INFO mod hackban 525: aikaterna(15449707214864****) hackbanned 49561788060847****
[05/10/2018 15:28] ERROR events on_error 191: Exception in modlog_case_create
  File "/usr/local/lib/python3.6/site-packages/discord/client.py", line 225, in _run_event
    await coro(*args, **kwargs)
  File "/usr/local/lib/python3.6/site-packages/redbot/cogs/mod/mod.py", line 1620, in on_modlog_case_create
    mod_channel = await modlog.get_modlog_channel(case.guild)
  File "/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/redbot/core/modlog.py", line 694, in get_modlog_channel
    raise RuntimeError("Failed to get the mod log channel!")
RuntimeError: Failed to get the mod log channel!
```
I'm not sure if this is the case on brand new installs, this was on a bot that had a mod-log channel set and then removed and when that removal occurs the mod-log channel saved in the settings.json is set as "null" for that guild.

Twentysix26:
@aikaterna Is this related to what this PR introduces? I'm doing the same call (and error handling) as the other commands so if this is a separate problem it might be worth opening a dedicated issue

Twentysix26:
I put the whole `[p]ban` logic into its own function, so that it can be used by both commands. This was necessary to handle the `[p]ban` invocation from `[p]hackban` cleanly: before, if any error occurred, it would send a separate message for it. So, for example, if 9 bans out of 10 failed, you would get the `[p]hackban`'s "results" message plus 9 separate messages coming from `[p]ban`.

mikeshardmind:
I'd like to take a look at this later, but I'm in favor of anything that de-duplicates shared logic, and moves as much out of command objects as is sane to do (increases ability for things written to be used by other things sanely).

PR #2169, opened 2018-10-02
Finish and Refactor Predicate Utility

Tobotimus:
* Uses classmethods to create predicates
* Classmethods allow using a combination of different parameters to describe context
* Some predicates assign a captured `result` to the predicate object on success
* Added `ReactionPredicate` equivalent to `MessagePredicate`
* Added `utils.menus.start_adding_reactions`, a non-blocking method for adding reactions asynchronously
* Added documentation
* Uses these new utils throughout the core bot
Happened to also find some bugs in places, and places where we were waiting for events without catching `asyncio.TimeoutError`

### DIFF
--- a/docs/framework_utils.rst
+++ b/docs/framework_utils.rst
@@ -22,10 +22,16 @@ Embed Helpers
 .. automodule:: redbot.core.utils.embed
     :members:
 
-Menu Helpers
-============
+Reaction Menus
+==============
 
 .. automodule:: redbot.core.utils.menus
+    :members:
+
+Event Predicates
+================
+
+.. automodule:: redbot.core.utils.predicates
     :members:
 
 Mod Helpers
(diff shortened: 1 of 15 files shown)


Redjumpman:
Holy crap dude. This is amazing. I can't believe you took my tiny enhancement and turned it into this robust utility. I'm currently going through all of the methods testing them separately, and then checking the implementation applied to core. Also the documentation is 💯 .

Redjumpman (changes requested):
This tested out fine with me. The only thing I saw that could be changed is in downloader. We are using `MessagePredicate.equal_to` where a `MessagePredicate.lower_equal_to` might be more appropriate with the target string becoming "i agree".

Tobotimus:
@Redjumpman Thanks for reviewing :smiley:

PR #2172, opened 2018-10-02
[V3 Trivia] Append only takes one argument. This was silently failing

bobloy:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This was incorrect syntax of `list.append` and was silently failing. Adjusted to use `list.extend`


Tobotimus (approved):
Thanks for catching this

Issue #2168, opened 2018-10-01
[V3 Trivia] Trivia answers should ignore some special characters


pabloleban:
# Feature request

#### Select the type of feature you are requesting:

- [x] Cog
- [ ] Command
- [ ] API functionality

#### Describe your requested feature

Right now I can see this situation with the special characters in some of the trivia's answers:

From **games.yaml**
`Mega Man's traditional nemesis in the Mega Man series is whom?:
- Dr. Wily
- Dr wily

Who is Mega Man's creator?:
- Dr. Light
- Dr Light`

From **leagueults.yaml**
`Which champion's ultimate is named: Void Rush?:
- Rek'sai
- reksai
- rek sai

From **gamesofthrones.yaml**
`What are the official words of House Martell?:
- Unbowed, Unbent, Unbroken
- Unbowed Unbent Unbroken`

I would suggest to ignore at least this special characters: `. : , '` from the answers on the trivia list and from the user's answers to avoid having so many options, so you can be more flexible with your answers.

So following the first example you can have:

`Mega Man's traditional nemesis in the Mega Man series is whom?:
- Dr. Wily`

Then `dr. wily` or `dr wily` are correct answers, just by having one answer in the trivia. There are a lot of cases like this in all the trivia lists. I've just named a few of them.

pabloleban:
I would also consider interpreting & as "and"

SuperSandro2000:
I would add apostrophes to the list. Also what about number formating if an answer is a number? Currency? Date?
And as I am German I need to add Umlaute (ä=ae, ü=ue, ö=oe) and ß (equal to ss) to the list.

Bakersbakebread:
Fuzzy string matching for typos, too?

Tobotimus:
To be honest a lot of this is already possible (and done in existing trivia lists) by adding more alternate answers to the trivia lists. For example, if an answer contains &, Umlaute or other characters which have alternate ways of specifying, simply have multiple answers:
```yaml
Question 1:
  - One & two
  - One and two
Question 2:
  - Straße
  - Strasse
```

I would prefer this than trying to process these alternate spellings within the trivia cog, as I'm sure there are a plethora of these characters.

pabloleban:
[...]

PR #2173, opened 2018-10-02
[V3 Trivia] Option to ignore special characters in guesses and answers

bobloy:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Option to satisfy #2168
Removes specific special characters from guesses and answers when configured.

I predict this would be better configured **per trivia list** using the config override rather than the **per guild** setting


bobloy:
Switch from `re.sub` to `string.translate` to avoid additional import and increase speed.

Regex:
image

String.translate:
image

Jackenmen:
Closing this since we already closed the original issue:
https://github.com/Cog-Creators/Red-DiscordBot/issues/2168#issuecomment-570758219

PR #2177, opened 2018-10-03
[V3 Trivia] On / Off also resolve to bool with YAML

bobloy:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

`On` and `Off` also resolve to boolean values in YAML and should be equally addressed.

For example, the answer to 
`Name this video game: https://www.youtube.com/watch?v=xajMfQuVnp4`
is 
`Off`


Tobotimus (approved):
Thanks. In the future I might think about modifying the YAML loader so it doesn't parse bools at all

Issue #1841, opened 2018-06-10
[V3 Utils] Make markdown helpers escape special chars


Tobotimus:
# Feature request

#### Select the type of feature you are requesting:

- API functionality

#### Describe your requested feature
Currently we have a bunch of helpers in `redbot.core.utils.chat_formatting`, including those which simply return some string formatted in some way with markdown, i.e. `bold`, `italics`, `inline` etc. All they do now is wrap the input string in the formatting delimiters and return - people are more likely to do this by just including the delims in their raw string, since it takes up less characters and doesn't require an import.

A way to actually make these functions useful would be to have them escape any relevant markdown formatting syntax *within* the input string.

##### Inline
Current behaviour:
```py
>>> inline("a string` containing` backticks")
"`a string` containing` backticks`"
```
Formatted: `a string` containing` backticks`

After changes:
```py
>>> inline("a string` containing` backticks")
"``a string` containing` backticks``"
```
Formatted: ``a string` containing` backticks``

##### Italics
Current behaviour:
```py
>>> italics("A string* containing *asterisks")
"*A string* containing *asterisks*"
```
Formatted: *A string* containing *asterisks*

After changes:
```py
>>> italics("A string* containing *asterisks")
"*A string\* containing \*asterisks*"
```
Formatted: *A string\* containing \*asterisks*

Tobotimus:
Partially resolved by #2182, the unresolved ones don't seem to be solve-able right now.

PR #2182, opened 2018-10-04
[V3 Utils] Markdown helpers escape special chars

Twentysix26:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Implements #1841
From my tests escaping italics (`*`) that get passed to `italics()` doesn't work, the string still won't show up properly. If you can think of a way let me know.


Tobotimus (approved):
Thanks, I can't think of a way to escape italics right now, and unfortunately we know code-blocks can't be escaped properly, but this will do for now :smiley:

PR #2183, opened 2018-10-05
[V3 Menus] ctx.me so this will work in DM's

bobloy:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

`ctx.me` handles using `ctx.guild.me` if `ctx.guild is not None`
`ctx.guild.me` directly error in DMs


Tobotimus (approved):
Thanks :smiley:

PR #2185, opened 2018-10-06
[V3/Config] New config methods

Redjumpman:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

This adds some quality of life methods to config. The two new operations are `inc` and `toggle`. These work like `set`, but for specific value types. They also return the new value. This aims to solve repetition in calling.

**Examples for Incrementing**
```py
# Add 3 to a number
new = await config.foo.inc(3)
await ctx.send(f"Incremented foo from {new - 3} to {new}.")

# Subtract 5
new = await config.foo.inc(-5)
await ctx.send(f"Incremented foo from {new - 5} to {new}.")
```
**Example for Toggle**
```py
new = await config.foo.toggle()
await ctx.send(f"Changed foo from {not new} to {new}.")
```

**Example of doing this without these methods**
```py
current = await config.foo()
await config.foo.set(value=(not current))
await ctx.send(f"Changed foo from {current} to {not current}.")
```
I included some tests, and this is working with the both the json and mongo drivers. This shouldn't break any current usage of config. 

**DISCLAIMER**
I'm setting this up to be implemented in 3.1 and not prior to release.

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -151,6 +151,64 @@ class Value:
         """
         await self.driver.set(*self.identifiers, value=value)
 
+    async def inc(self, value: int) -> int:
+        """Increment and return the value of the data element pointed to by `identifiers`.
+
+        Example
+        -------
+        ::
+
+            # Increments and returns a global value "foo" by 3.
+            new = await conf.foo.inc(3)
+
+            # You can also decrement a value using by negative numbers
+            new = await conf.foo.inc(-3)
+
+            # Floats are also supported
+            new = await conf.foo.inc(3.834)
+
+        Parameters
+        ----------
+        value : int
+            Integer that will be used to increment a value.
+
+        Returns
+        -------
+        `int` or `float`
+            The stored value plus the value given as a integer or float.
+
+        """
+        if isinstance(self.default, bool) or not isinstance(self.default, (int, float)):
+            raise TypeError("Config inc method only supports integer or float values.")
+        try:
+            new_value = await self.driver.inc(*self.identifiers, value=value, default=self.default)
+        except TypeError as e:
+            raise TypeError("Config inc method requires an integer or float for value.") from e
+        return new_value
+
+    async def toggle(self) -> bool:
+        """Toggles a Boolean value between True and False as pointed to by `identifiers`.
+
+        Example
+        -------
+        ::
+
+            # Assume the global value "foo" is currently True.
+            # The following will change it to False, and return the result.
+            new = await conf.foo.toggle()
+
+        Returns
+        -------
+        `bool`
+            The opposite of the originally stored value.
+
+        """
+        if not isinstance(self.default, bool):
+            raise TypeError("Config toggle method can only toggle Boolean values.")
+
+        new_value = await self.driver.toggle(*self.identifiers, default=self.default)
+        return new_value
+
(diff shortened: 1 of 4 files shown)


tests/core/test_config.py
```diff
@@ -385,6 +385,39 @@ def test_nested_group_value_badreg(config):
...
+        await config.foo.toggle()
+
+
+@pytest.mark.asyncio
+async def test_inc_bad_values(config):
+    defaults = {"foo": "bar", "baz": True}
+    config.register_global(foo=defaults)
+    with pytest.raises(TypeError):
+        await config.foo.inc(10)
+        await config.baz.inc(10)
```
Tobotimus, line 419: These two operations should be in separate `with` statements; `pytest.raises` will catch the first exception then exit the context manager

Tobotimus (commented):
Thanks for this PR, these new methods are awesome :+1: All I found from browsing the code was 1 thing, I'll do more local tests in the next few days :smiley:

Once I've approved this I'll merge it into a feature branch as you suggested to me earlier.

tests/core/test_config.py
Redjumpman: Done.

Redjumpman:
Anything left for me to do on this?

mikeshardmind:
The only additional thing I'd request is on the subtraction/increment operators, allow a conditional to be passed for a minimum start/end value, which would raise an error and not commit the change.

This would allow some inlining of single access to those types while still keeping validation for the result.

redbot/core/config.py
```diff
@@ -151,6 +151,64 @@ def __call__(self, default=...) -> _ValueCtxManager[Any]:
...
+            Integer that will be used to increment a value.
+
+        Returns
+        -------
+        `int` or `float`
+            The stored value plus the value given as a integer or float.
+
+        """
+        if isinstance(self.default, bool) or not isinstance(self.default, (int, float)):
+            raise TypeError("Config inc method only supports integer or float values.")
```
Tobotimus, line 182: Might also want to say that the method only supports *registered* integer or float values

redbot/core/config.py
```diff
@@ -151,6 +151,64 @@ def __call__(self, default=...) -> _ValueCtxManager[Any]:
...
+            new = await conf.foo.toggle()
+
+        Returns
+        -------
+        `bool`
+            The opposite of the originally stored value.
+
+        """
+        if not isinstance(self.default, bool):
+            raise TypeError("Config toggle method can only toggle Boolean values.")
```
Tobotimus, line 207: Same here

Tobotimus (changes requested):
The Mongo driver with these new methods is making 3 queries for each inc/toggle (except for one case). This is actually more than we would otherwise use for a simple get and set.

Because of the way we store data in mongo, the use of the `$inc` and `$bit` operators isn't worth it. It would make more sense if each member/user/channel/guild etc.'s data included all defaults on initial set, but unfortunately that simply isn't the case.

tekulvw (changes requested):
The `Base` driver class should also be updated with these new methods. Mongo driver will be updated as necessary in #2536

tekulvw:
Superseded by #2635 due to the underlying driver changes and addition of MongoDBV2

PR #2187, opened 2018-10-06
V3/fix/info fix

palmtree5:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This fixes the issue with the info command in RC1 that relates to the outdated bot check (for good measure, I made the same changes to the message that appears on startup as well). I'm dropping the use of `distutils.version.StrictVersion` in favor of reading in the upload time for both the currently running release and the most recent release straight from PyPI's API (using `datetime.datetime.strptime` because `datetime.datetime.fromisoformat` is new in Python 3.7 and we support 3.6)

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -13,7 +13,6 @@ from collections import namedtuple
 from pathlib import Path
 from random import SystemRandom
 from string import ascii_letters, digits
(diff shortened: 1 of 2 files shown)


Tobotimus:
Superseded by #2188.

PR #2194, opened 2018-10-06
[V3 Alias] Fix alias help

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Alias help would only return the first character of the invoked command previously. This change shows help for basic commands that are aliased (i.e. just `ping`) or aliased commands that have an argument included (i.e. `audioset role beep` with `beep` being a role name)


Redjumpman (approved):
This is good to go.

Tobotimus (approved):
Thanks @aikaterna and @Redjumpman :smiley:

PR #2200, opened 2018-10-07
[Permissions] Fix ValueError for "default" rule in config/ACL

Tobotimus:
Error was thrown when the "default" key existed and Permissions tried to iterate over the list mapping keys to `int`.


Kowlin (approved):
Bug fixes work as intended.

Issue #820, opened 2017-06-20
[V3 CustomCommands] Allow option to fetch Markdown version of the text


smlbiobot:
Unless I have mistaken, if the text I use in the custom commands contains markdown for formatting, there is currently no way to copy the full markdown from the app for easy edits. Right now I have to re-create the markdown every time because I can only get plain text if copied from the app.

This is not an issue if it was the same person who created the original text, but it became a hassle whenever I need to edit someone else’s text. In some servers that I admin, the customcoms are rather long so I would like an option for me to copy the markdown source, perhaps placed inside a code block.

Please let me know if that option is already there and I just didn’t know about it. And if not, it would be awesome if you can implement that feature.

Thanks!

skmendez:
I have a PR that would add this, but the problem is that there is no way in Discord to show 3 contiguous backticks inside of a codeblock. In normal Markdown, we can display backticks as such:

```````
We use 5 backticks:
`````
So we can use 3 backticks inside the codeblock:
```
Look at em!
```
And now we exit the code block
`````
Isn't that magical!
```````
I also used 7 backticks to escape that codeblock. Discord, unfortunately, does not support this.

Tobotimus:
As stated on #1352 I'd happily accept this as an addition to V3 :)

Tobotimus:
Resolved by #2204.

PR #2204, opened 2018-10-07
[V3 CustomCommands] Add [p]cc show

Twentysix26:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

Implements #820.  
Also fixes a minor graphical issue with [p]cc delete's help.  
I was hesitant to modify the current CC's API (which I found to be quite inconsistent) to prevent current 3rd party cogs that interact with it from breaking, so I just added an additional method.
I went with a code block instead of an embed but it can be changed if an embed is preferable. I think it looks clear:

picture

### DIFF
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -8,7 +8,7 @@ from typing import Mapping, Tuple, Dict
 import discord
 
 from redbot.core import Config, checks, commands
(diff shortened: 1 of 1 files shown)


Redjumpman (approved):
This is great! Working as intended.

Tobotimus (approved):
Thanks @Twentysix26 and @Redjumpman :smiley:

PR #2206, opened 2018-10-08
Refactor commands.Cog, create commands.CogMeta

Tobotimus:
This does quite a few things (without any breaking changes to public API):

#### New Features
- Cogs can now be disabled in particular guilds with `[p]command disable`, in the same way commands can (I purposefully avoided allowing a global disable since users should really unload the cog instead)
- Cog classes can now be decorated with checks, and such checks will apply to all top-level commands (and subsequently all subcommands)
#### Changes
- Moved `commands.Cog` into new module `redbot.core.commands.cog`.
- `commands.Cog` is an instance of the metaclass `commands.CogMeta`, which in turn inherits from what `commands.Cog` previously inherited from.
- `requires` attribute is now attached to cog classes rather than cog instances, as is more consistent with commands.
#### Bugs fixed
- Rules are now cleared from the cache of cogs/commands when they are unloaded (previously this could have resulted in issues for commands if the rules changed whilst the command was unloaded)
- Fixed permissions hooks being cleaned up incorrectly in `Red.remove_cog` (this is something which slipped through from #2149, I meant to come back and fix this before that PR was merged)


Tobotimus:
Blocking this until it has some more parity with discord.py's recent changes

Tobotimus:
Closing - this needs to be redone in light of other changes

PR #2207, opened 2018-10-08
[V3 Core] Specify an error message on package loading

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This will allow cog creators to explain clearly why a cog cannot load by raising  `redbot.core.errors.CogLoadError`. Instead of having to check in the console what's wrong, the message will directly be sent in the context channel.

### Example

`__init__.py`
```py
from .my_cog import MyCog
from redbot.core.errors import CogLoadError

try:
    import pillow
    has_lib = True
except ImportError:
    has_lib = False

def setup(bot):
    if not has_lib:
        raise CogLoadError("You need to type `[p]pipinstall pillow` to load this cog.")
    n = MyCog(bot)
    bot.add_cog(n)
```
Gives this:
screen shot 2018-10-08 at 22 06 45

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -583,7 +583,9 @@ class Core(commands.Cog, CoreLogic):
     async def reload(self, ctx: commands.Context, *cogs: str):
         """Reloads packages"""
         async with ctx.typing():
-            loaded, failed, not_found, already_loaded = await self._reload(cogs)
+            loaded, failed, not_found, already_loaded, failed_with_reason = await self._reload(
+                cogs
+            )
 
         if loaded:
             fmt = "Package{plural} {packs} {other} reloaded."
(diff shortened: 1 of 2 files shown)


Tobotimus:
@retke Thanks for this PR! This is actually a really cool idea.

I know you'd like this to be in 3.0 but this'll be headed towards 3.1 since it's a new feature :P Bugfixes only for the 3.0 branch. I've just rebased it onto V3/develop.

laggron42:
This is actually because the `errors.py` file was not in the `V3/develop` branch, so I pulled from the 3.0 branch.

mikeshardmind (approved):
:+1: My calculator cog thanks you for this.

Tobotimus (approved):
Thanks @retke :smiley:

PR #2212, opened 2018-10-09
[MongoDB] Escape special characters in keys

### LINKED ISSUE #2038: [V3 Config] Config does not reject bad keys
Including a `.` or starting a keyname with `$` needs to be invalid in config because of mongoDB backend support. 

Config will work just fine for this on the json backend, but not with the mongodb one because these characters are not allowed in field names for mongo. Instead, it works with json, and throws uncaught exceptions on mongo.

I'd be in favor of limiting keynames to alpha-numeric and space, underscore, and hyphens. to prevent future issues like this with other future possible backends without possibly breaking cogs twice for this.

Tobotimus:
Essentially resolves #2038, although this is escaping and not rejecting keys as that issue implies.

### DIFF
--- a/redbot/core/drivers/red_mongo.py
+++ b/redbot/core/drivers/red_mongo.py
@@ -1,6 +1,11 @@
-import motor.motor_asyncio
-from .red_base import BaseDriver
+import re
+from typing import Match, Pattern
 from urllib.parse import quote_plus
+
+import motor.core
+import motor.motor_asyncio
+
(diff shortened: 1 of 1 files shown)


mikeshardmind (approved):
There could potentially be some really extreme edge case where someone wants to use 
`\\U&00000024`
as a dict key, but this is so far out there, that this is a satisfactory solution, and that person can use a sentinel value of their own.

PR #2215, opened 2018-10-09
[CustomCom] Better display for [p]cc list

### LINKED ISSUE #2104: [V3] Escape markdown properly on Custom Command list
# Command bugs

Yes I found bug

#### Command name

`cc list`

customcom

It list the cc beautifully 

#### What actually happened?

image

Just have several custom commands with code block / inline code block markdown

Tobotimus:
Uses a menu, optionally embedded with respect to the embed settings, for scrolling through the custom command list, each cc with a ~50 character preview. Format is purposefully similar to the help menu.

Resolves #2104.

### DIFF
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -8,9 +8,9 @@ from typing import Mapping, Tuple, Dict
 import discord
 
 from redbot.core import Config, checks, commands
-from redbot.core.utils.chat_formatting import box, pagify
 from redbot.core.i18n import Translator, cog_i18n
 from redbot.core.utils import menus
+from redbot.core.utils.chat_formatting import box, pagify, escape
 from redbot.core.utils.predicates import MessagePredicate
 
 _ = Translator("CustomCommands", __file__)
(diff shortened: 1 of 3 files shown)


redbot/cogs/customcom/customcom.py
```diff
@@ -338,24 +339,42 @@ def __init__(self, bot):
...
-
-        if len(_commands) < 1500:
-            await ctx.send(box(_commands))
+            # Don't put a line-break in preview
+            newline_pos = result.find("\n")
+            if newline_pos != -1:
+                result = result[:newline_pos] + "..."
+            # Cut preview to 52 characters max
+            if len(result) > 52:
+                result = result[:49] + "..."
```
mikeshardmind, line 357: I don't like this pagification method with embed. I think this is a good case for field based pagination to allow new lines in previews. While this method is certainly cleaner looking than the previous one, It's also less functional for many purposes.

As an example, I have many CCs in a codiing discord utilizing code blocks. The only thing I get is the language specifier and escaped backticks for those.

mikeshardmind (changes requested):
I don't think these changes are of the quality we should push for. It may make the list cleaner, but significantly less functional. If we need an RC3 to accomplish this, so be it in my mind.

Twentysix26:
I like it, certainly leagues better than the previous form.  
Maybe the issue with the code blocks could be resolved by replacing the new lines characters in the previews with a space.
Minor nitpick and maybe up to personal preference: I think it may look better displayed on two lines for each command
```
cmd1
preview
cmd2
preview
(...)
```

mikeshardmind (approved):
After discussion in discord, I'm modifying my review to approval.

If nobody else does, I'll eventually be looking to add an additional way to browse CCs myself later with more detail, but this shouldn't block unbreaking the current behavior for.

Tobotimus:
@Twentysix26 I actually hadn't thought of replacing newlines with spaces, I will actually give that a go. For displaying the preview on the next line, I think I personally prefer the current method for similarity with the help menu, but I will do it and post some screenshots FTR.

Tobotimus:
### Screenshots
<h4>
<details>
  <summary>Original PR style embed (click me)</summary>
  
</details>
</h4>

<h4>
<details>
  <summary>With newlines replaced with spaces</summary>
  
</details>
</h4>

<h4>
<details>
  <summary>With preview on following line</summary>
  
</details>
</h4>

I prefer the 2nd one, so I'm gonna go with it :smiley:

PR #2218, opened 2018-10-09
[Warnings] Fix actions not being taken

### LINKED ISSUE #2106: [V3] Warnings cog issue
# Command bugs

#### Command name

.warn

Warnings

When a user goes above the points threshold, the assigned action should occur.

#### What actually happened?

User gets warned and points get added however no action is issued.

Create a warnreason, assign it a point value and set that point value to ban a user.
 Then warn the user.

Here's an example of what I have setup and the bot has done nothing to the user

http://evac.gg/i/pthl6.png

When setting up the action, i used:

ban {user}

Tobotimus:
When multiple warning actions were registered, and the user didn't exceed the points for the highest action on the list, no action was taken.

Resolves #2106.

Also commented out the casetype registration for warnings, since it's not actually using modlog yet.

### DIFF
--- a/redbot/cogs/warnings/helpers.py
+++ b/redbot/cogs/warnings/helpers.py
@@ -19,9 +19,11 @@ async def warning_points_add_check(
     act = {}
     async with guild_settings.actions() as registered_actions:
         for a in registered_actions:
+            # Actions are sorted in decreasing order of points.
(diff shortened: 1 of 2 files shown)


mikeshardmind (approved):
:+1:

PR #2222, opened 2018-10-11
[i18n] Remove @cog_i18n decorator

Tobotimus:
Continues on from #2206.

These are just general design improvements and some cleanups for i18n, possible now through the `commands.Cog` class and by extension, the `commands.CogMeta` metaclass.
- `Translator` no longer accepts a name as its first argument, as it's unused and unnecessary
- Removes @cog_i18n decorator in favour of passing the translator as an argument to `commands.Cog.__init_subclass__`
- Assigns `__translator` object to cog classes, and adds new method `commands.CogMeta.translate()` which is used by both commands and cogs to translate help strings
- Removes `commands.Command.translator` attribute


redbot/core/commands/cog.py
```diff
@@ -0,0 +1,74 @@
...
+
+    def translate(cls, untranslated: str) -> str:
+        return cls.__translator(untranslated)
+
+
+class Cog(metaclass=CogMeta):
+    """Base class for a cog."""
+
+    def __init_subclass__(cls, **kwargs) -> None:
+        cls.__translator = kwargs.pop("translator", lambda s, t: t)
```
mikeshardmind: This should default to `None` and compare to `None` where needed. None type comparison is significantly cheaper than a lambda when considering every user-facing string.

redbot/core/commands/cog.py
```diff
@@ -0,0 +1,75 @@
...
+        if cog_instance is None:
+            raise discord.ext.commands.BadArgument(_('Cog "argument" not found.'))
+        return type(cog_instance)
+
+    def translate(cls, untranslated: str) -> str:
+        if cls.__translator is not None:
+            return cls.__translator(untranslated)
+
+
+class Cog(metaclass=CogMeta):
```
mikeshardmind, line 79: Edit: the below changes are un-needed. It's been a rarity I've needed multiple metaclasses at once, so I had some alarm bells going off from a slightly different, but importantly syntactically different, case.
____
<details><summary>outdated</summary><p>
I'd prefer if the stuff in this were one more level of abstracted to better allow metaclass compatability.

This would entail chainging this to:

```py
class CogBase:
    """Base class for a cog."""
    def __init_subclass__(cls, **kwargs) -> None:
        cls.__translator = kwargs.pop("translator", None)
        super().__init_subclass__(**kwargs)

    @classmethod
    async def convert(cls, ctx: "Context", argument: str) -> "Cog":
        cog_instance = ctx.bot.get_cog(argument)
        if cog_instance is None:
            raise discord.ext.commands.BadArgument(_('Cog "argument" not found.'))
        return cog_instance

class Cog(CogBase, metaclass=CogMeta):
    pass
```

Doing this could allow for cooperative metaclassing in the following manner:

```py
# .abcs

class _MyMeta(type):
    ...

class MyMeta(_MyMeta, commands.CogMeta):
    pass

class CogBase(commands.CogBase, metaclass=MyMeta):
    pass
```
While I understand this is an edge case, it's reasonably accomodated, and avoids things like nearly completely breaking cogs using metaclasses to define APIs for static typing tools like mypy (see python/mypy#1996 )
</p></details>

it should be enough for anyone needing this to do
```py
# .abcs

class MyMeta(commands.CogMeta):
    pass

class MyCogBase(commands.Cog, metaclass=MyMeta):
    pass
```

Tobotimus:
Currently rewriting this, will open a new PR when it's ready

PR #2223, opened 2018-10-11
[V3 Streams] remove communities support

palmtree5:
Per https://blog.twitch.tv/introducing-tags-and-new-categories-33744ef7b04f, communities on Twitch have been discontinued. Therefore I am opening this PR to remove them


palmtree5:
This should be ready at this point. Note that I left `streamalert twitch channel` in place because of work I am doing on implementing support for teams

Redjumpman (dismissed):
This probably needs a bump, because it's a stale PR, but this works.

Tobotimus (approved):

PR #2233, opened 2018-10-12
Only trigger reload of updated cogs which are loaded

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -372,7 +372,7 @@ class Downloader(commands.Cog):
                 await self._reinstall_libraries(installed_and_updated)
                 message = _("Cog update completed successfully.")
 
-                cognames = [c.name for c in installed_and_updated]
+                cognames = {c.name for c in installed_and_updated}
                 message += _("\nUpdated: ") + humanize_list(tuple(map(inline, cognames)))
             else:
                 await ctx.send(_("All installed cogs are already up to date."))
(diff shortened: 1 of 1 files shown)


redbot/cogs/downloader/downloader.py
```diff
@@ -402,7 +402,7 @@ def __init__(self, bot: Red):
             if can_react:
                 with contextlib.suppress(discord.Forbidden):
                     await query.clear_reactions()
-
+            cognames &= set(ctx.bot.cogs.keys())  # only reload loaded cogs
```
Tobotimus: This needs to be `await ctx.bot.db.packages()` rather than `ctx.bot.cogs.keys()`, and also doing this here means `cognames` as it is passed to `[p]reload` can be empty. Perhaps it is best to create this set before prompting for a reload and exit early if it's empty.

Tobotimus (changes requested):

redbot/cogs/downloader/downloader.py
mikeshardmind: Using `ctx.bot.extensions.keys()` instead. I initially went with that approach but we can skip a co-routine here, and this can be incorrect if the bot is loaded with the `--no-cogs` arg (as you noted in discord later)

Tobotimus (approved):
Thanks :+1:

PR #2234, opened 2018-10-13
Fix permissions hook removal

Tobotimus:
Some in-progress work slipped through #2149, and I figure it should be fixed before RC2.

I've also just decided to allow discovery of permissions hooks from superclasses as well. We should try to be more aware of the possibility of cog superclasses moving forward.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1,4 +1,5 @@
 import asyncio
+import inspect
 import os
 import logging
 from collections import Counter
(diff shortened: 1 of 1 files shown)


mikeshardmind (approved):
Appears to work as intended using this [toy implementation](https://github.com/mikeshardmind/SinbadCogs/blob/v3/audiohook/audiohook.py) and the logic appears solid as well. :+1:

PR #2235, opened 2018-10-13
Config reserved keywords

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This PR adds a small warning in the Config docs, explaining which values should **not** be used to prevent class breaking. I myself encountered a bug by using `default` as a value, and it wasn't specified that it shouldn't be used, so I made that quickly with the help of Sinbad.

I also modified the Python interpreter in `redbot/docs/Makefile` so it uses `python3` instead of `python`, was quite annoying for me since it was calling Python 2.7

### Preview

screen shot 2018-10-13 at 17 52 10

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -3,7 +3,7 @@
 
 # You can set these variables from the command line.
 SPHINXOPTS    =
-SPHINXBUILD   = python -msphinx
+SPHINXBUILD   = python3 -msphinx
 SPHINXPROJ    = Red-DiscordBot
 SOURCEDIR     = .
 BUILDDIR      = _build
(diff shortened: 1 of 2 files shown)


Tobotimus:
It's perfectly OK to use these as keys through the `get_attr`/`get_raw`/`set_raw`/`clear_raw` methods. In cases where variable-size dicts are used, such occurrences would be almost unavoidable (this is the reason `Group.get_attr` exists, because it bypasses python's default attribute access exhibited by the built-in function `getattr`).

However, you're right that these names should be avoided when users are planning on accessing these keys through dot notation, and hence this warning is relevant there. Please try to make this warning refer only to the use of dot notation (and registered defaults, because by config's design, defaults should be accessed through dot notation).

Perhaps we ought to obscure the names of a few attributes/methods for the `Group` and `Config` classes to make these collisions less likely to occur. Doing so would be a breaking change, but may be worthwhile.

PR #2242, opened 2018-10-15
Flatpak (100% Working)

lionirdeadman:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Added a .yml file to allow a flatpak to be built on any distribution

### Notes

Breaking points for Flathub inclusion : 
- Pip is used directly
- It's not fully reproducible because it installs whatever is newest on pypi
- Doesn't check hash
This was unavoidable because of the links dependency

Non-working : 
~~- Audio cog because OpenJDK10 has a broken certificate (https://github.com/flathub/org.freedesktop.Sdk.Extension.openjdk10/issues/4)~~

Possible improvements : 
- Add .desktop to easily launch from app menu
- Add .AppData.xml to see general information, screenshots and updates (Mainly useful for Flathub)
- Icon folder for looks

Feel free to test and tell me what you think of it. We can publish the bundles with the releases until the breaking points for Flathub inclusion are fixed.


lionirdeadman:
Audio is now working so everything works now :tada:

lionirdeadman:
Can anyone review or try this?

I think I might just close it and keep it in another repo for whoever wants it.

Tobotimus:
Please keep it open for now, I'm sorry it hasn't been reviewed yet, but we've mostly been focused on 3.0.0 stuff. I will take a closer look at it soon :smiley:

mikeshardmind:
I can try and get a review of this together soon (No absolute guarantees on timeframe). There may be some changes required (minor) on the base install command (depending on merge order) due to pip19's removal of `--process-dependency-links` (see #2387 for how we're handling this)

PR is appreciated, but the focus has been elsewhere recently.

lionirdeadman:
Yeah, that's fine, I simply thought that no one was interested in this hence why I was proposing to close it.

The removal of `--process-depency-links` would actually probably a good thing towards Flathub inclusion since It'll make a fully reproducible build which doesn't update possible (even though that's not what we're doing at the moment)

mikeshardmind:
@lionirdeadman I'm still interested in flatpak support. Our dependency issues should finally be cleaned up, so it may be a good time to re-discuss anything which is necessary for that. 

While I don't think this will be ready (and reviewed) prior to 3.1, there's not necessarily a reason it could not be part of 3.1.1 or 3.1.2

lionirdeadman:
I'm uncertain how we should deal about cog requirements if we make a proper flatpak since they're immutable. Is it possible to move those in a certain place?

mikeshardmind:
cog reqs are already downloaded to a shared_lib path in the bot's configured path (not to system shared lib path)

Not sure how this does/doesn't work here, would require testing to see.

lionirdeadman:
Oh then that should work just fine, I need to look how I can use flatpak-pip-generator's output (JSON) in my YAML manifest, I'm asking a friend right now. 

If you want, you can make an com.github.Red.DiscordBot.appdata.xml and desktop file which I could use in the flatpak, [for reference on how those might look like](https://github.com/flathub/org.xonotic.Xonotic)

Oh and we need to look at how we can prevent the launcher from trying to update since it won't be able to inside the flatpak (to do so, they'd install a new bundle or if we had a repository, they'd use `flatpak update` from outside it. 

Probably need some docs on doing PM2 and systemd configurations with this new install method.

Drapersniper:
Mongo is being depricated, this should be update to reflect it

Drapersniper:
@lionirdeadman  is this something you are still actively working on or have you dropped this PR. Unsure whether you are currently working on this locally and just haven't pushed any commits.

lionirdeadman:
No, it isn't something I'm working on privately, I think a containerized setup makes more sense and I used to use this simply to refuse filesystem access. The container setup I use now has private home setups so I don't see the point in making this anymore for my personal use.

This can give a package that can be used on all distros but I don't think Red wants to go with this so there's no point having yet another package to maintain, I think. If someone is interested, I can give guidance or make it work with current Red but I won't be maintaining it unless there is significant interest.

Kowlin:
I'm curious in this method my self, and am probably willing to maintain the system if I've read into it, but as of right now I have literally 0 experience in Flatpacks. So I'll be reading into this setup.

lionirdeadman:
This flatpak doesn't use good practice. I'll modify it for best practice tomorrow. We can talk about it on the discord if you want explanations.

We need requirements.txt for `flatpak-pip-generator` hm

Flame442:
Due to the age of this PR and the fact that Flatpak [isn't able to be used on a desktopless server](https://flatpak.org/faq/#Can_Flatpak_be_used_on_servers_too_), it doesn't seem like something we are interested in at this time.

PR #2243, opened 2018-10-15
[Downloader] Unload extensions on uninstall

### LINKED ISSUE #2216: [Downloader] Cogs don't get unloaded after uninstalling.
# Other bugs

When uninstalling a cog from downloader the cog does not get unloaded.

The cog being deleted **&** unloaded.

#### What actually happened?

The cog only got deleted, The cog stayed active and in memory.

``cog uninstall <3rd party cog>``
Check that the cog is still loaded in memory
``cogs``

Tobotimus:
Resolves #2216.


mikeshardmind (approved):
:+1:

PR #2245, opened 2018-10-16
[V3 Audio] Restrict toggle for commercial sites

### LINKED ISSUE #2244: [V3 Audio] Restrict remote audio stream locations by default.
# Feature request

- [ ] Cog
- [ ] Command
- [X] API functionality

For multiple reasons, we shouldn't be allowing the play command to grab arbitrary remote things.

aikaterna:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Added [p]audioset restrict. This command will be True by default, and will restrict audio usage to main commercial sites like YouTube, Soundcloud, Vimeo, Mixer, Bandcamp and Twitch. It's an owner-only command, so if the bot owner wishes to allow http links, http streams, or other valid links to play that are not commercially-based, they will need to switch this toggle to False. Resolves #2244.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -2611,7 +2611,8 @@ class Audio(commands.Cog):
             track_obj[key] = value
         return track_obj
 
-    def _url_check(self, url):
+    @staticmethod
+    def _url_check(url):
         valid_tld = [
             "youtube.com",
             "youtu.be",
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -2620,7 +2620,7 @@ class Audio(commands.Cog):
             "bandcamp.com",
             "vimeo.com",
             "mixer.com",
-            "twitch.com",
+            "twitch.tv",
             "localtracks",
         ]
         query_url = urlparse(url)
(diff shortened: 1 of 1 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -2586,6 +2611,24 @@ def _track_creator(player, position=None, other_track=None):
...
+        valid_tld = [
+            "youtube.com",
+            "youtu.be",
+            "soundcloud.com",
+            "bandcamp.com",
+            "vimeo.com",
+            "mixer.com",
+            "twitch.com",
+        ]
+        site_check = []
```
mikeshardmind: This solution is not robust. A file hosted at `twitch.compromising-link.xyz/phishy-file.mp3` would be allowed.

redbot/cogs/audio/audio.py
```diff
@@ -2586,6 +2611,24 @@ def _track_creator(player, position=None, other_track=None):
...
+        for site in valid_tld:
+            if site in url.lower():
+                site_check.append(True)
```
mikeshardmind: And on matching a solution that was, this can just be `return True` here, and everything else below it in the function would be omitable

mikeshardmind (changes requested):
Outside of how the urls are being matched here, this is working and fine. Some changes are needed (direction was already briefly discussed in discord)

mikeshardmind (approved):
Seems to work properly, giving this a :+1:

Kowlin (approved):

PR #2246, opened 2018-10-16
 [i18n] Context-based Locale

Tobotimus:
Continues on from #2222 (which continues on from #2206). ~~Also blocked by #2394.~~

This is a WIP proof-of-concept for context-based locale. It is only available in Python 3.7 due to contextvars support only being available since then; 3.6's contextvars backport does not have asyncio support and thus is not very useful to us.

I will poke around and see what improvements I can make, all I can say is that this is on the table for 3.1. I'll probably make a feature branch whenever we agree on a design.

For a nicer diff exclusive to this PR, see [here](https://github.com/Tobotimus/Red-DiscordBot/compare/V3/remove_cog_i18n..V3/personalise_locale).


Tobotimus:
Currently rewriting this, will open a new PR when it's ready

Issue #1309, opened 2018-02-15
[V3] Scheduler


The-Tasty-Jaffa:
- [x] Suggestion
- [ ] Bug

Add a time/looping feature for more effective and less repetitive code.

Why?
1. Listening for events has many advantages over repeatedly checking for a particular event/condition
2. Removes the need of repetitive functions in cogs which all perform similar tasks

Sitryk:
Do you mean like, more events that cogs can listen for?
Any examples of some you might use

The-Tasty-Jaffa:
> Do you mean like, more events that cogs can listen for?

Yes.



As an example, Instead of having something like this 
```
def check():
    DELAY = 60 #Delay in seconds
    while self == self.bot.get_cog("cog"):
        #body
```
It could be done like this
```
@utils.loop(time=60) #also delay in seconds
def check():
        #body
```

The-Tasty-Jaffa:
Or however it ends up getting implemented. So it could work something more like listeners currently do in discord.py

Sitryk:
Developers could start using the `bot.dispatch` which would be really nice for cogs like audio and economy, which could dispatch events for stuff like song changing and cogs could act on those events

Bakersbakebread:
There's also the listen decorator. Would be an easy integration

https://discordpy.readthedocs.io/en/rewrite/ext/commands/api.html#discord.ext.commands.Bot.listen

The-Tasty-Jaffa:
However, how would calling it at different times work (for bot.listener)? (it doesn't allow for other options. For example, datetime objects cannot be passed or number of seconds.) While there could be a set time of 60 seconds I feel that doing this it would remove the usefulness of any such additions.

mikeshardmind:
`bot.dispatch` would be the *correct* way to have core red send events to listeners in cogs if this were to be added. We'd need documentation on any events we'd send, as well as ensuring zero overlap with internal discord.py dispatch events.

This would be as simple for cog creators as having a 
```py
async def on_event_name(self, event_arg1, ...):
```
line. it would be no harder to use from the perspective of a cog creator than listening for an on message event.

Also, because of how bot.dispatch is designed, if talking about an internal scheduler for tasks, when scheduling them, cogs could just send their own event name to the scheduler (this may not be a good idea if we don't want cogs to be able to schedule for other cogs though)

Sitryk:
[...]

PR #2247, opened 2018-10-16
V3/scheduler

tekulvw:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Adds some new decorators to declare cog methods as called once in the future, called at shutdown, or looped forever.

This is a WIP and documentation is incoming. Based off #2206.


The-Tasty-Jaffa:
Does this solve #1309 ?

tekulvw:
Needs to be rewritten.

PR #2255, opened 2018-10-17
[V3 Help] Menus in help

bobloy:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This modifies the help_formatter to use the `utils.menus` system to display help on servers.

It does not use it for pm, as you cannot manage reactions in pm.

It uses the built in configuration for maximum page size to split cog commands into multiple pages.
It uses the built in `helpset pagecharlimit` to set a guideline on how many cogs are displayed per page.

This obsoletes `max_pages_in_guild` and so it is removed.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -52,7 +52,6 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):
             color=15158332,
             fuzzy=False,
             help__page_char_limit=1000,
-            help__max_pages_in_guild=2,
             help__tagline="",
             disabled_commands=[],
             disabled_command_msg="That command is disabled.",
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1014,25 +1014,6 @@ class Core(commands.Cog, CoreLogic):
         await ctx.bot.db.help.page_char_limit.set(limit)
         await ctx.send(_("Done. The character limit per page has been set to {}.").format(limit))
 
-    @helpset.command(name="maxpages")
-    async def helpset_maxpages(self, ctx: commands.Context, pages: int):
-        """Set the maximum number of help pages sent in a server channel.
-
-        This setting only applies to embedded help.
-
-        If a help message contains more pages than this value, the help message will
-        be sent to the command author via DM. This is to help reduce spam in server
-        text channels.
-
-        The default value is 2 pages.
-        """
-        if pages < 0:
-            await ctx.send(_("You must give a value of zero or greater!"))
-            return
-
-        await ctx.bot.db.help.max_pages_in_guild.set(pages)
-        await ctx.send(_("Done. The page limit has been set to {}.").format(pages))
-
     @helpset.command(name="tagline")
     async def helpset_tagline(self, ctx: commands.Context, *, tagline: str = None):
         """
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1074,7 +1074,6 @@ class Core(commands.Cog, CoreLogic):
         """Toggle whether help will be sent in a DM or not.
          If this is enabled, the help message will be sent to the
         command author via DM.
-        I
          The default value is None
         """
         current_setting = ctx.bot.pm_help
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1069,6 +1069,22 @@ class Core(commands.Cog, CoreLogic):
         await ctx.bot.db.help.tagline.set(tagline)
         await ctx.send(_("The tagline has been set to {}.").format(tagline[:1900]))
 
+    @helpset.command(name="dm")
+    async def helpset_toggle_dm(self, ctx: commands.Context):
+        """Toggle whether help will be sent in a DM or not.
+         If this is enabled, the help message will be sent to the
+        command author via DM.
+        I
+         The default value is None
+        """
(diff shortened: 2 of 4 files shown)


bobloy:
This may need some adjustment to skip menus when there is only one page.

Additionally, `pm_help` is not the same thing has using `help` in a pm and may need tweaking as well.

bobloy:
This is ready for review. 
https://github.com/Cog-Creators/Red-DiscordBot/pull/2255/commits/1a3750f2b0fea31bd1414f78b33c4ce917e34948 moved `pm_check` to be a function that could be used in the help command.
https://github.com/Cog-Creators/Red-DiscordBot/pull/2255/commits/40b7ea85a842e9a9e75657353e790381a6053d50 fixed using a menu with only one page. Additionally, it will check the `use_embeds` setting based on the destination context instead of the channel context. This resolves the bot sending code_blocks when embeds are requested but the bot lacks permission to post in the channel.

aikaterna:
This is a really neat feature that a lot of people have requested over time. One thing I noticed while testing is that even with a small character page size set (like 300-400 characters per help page), I would get 2 smaller cogs and 1 large cog on the first page (Admin, Alias, Audio) which led to a really large page vs. what I expected to see because Audio has so many commands comparatively. I felt like maybe a good suggestion for pages would be to break it up per cog, but then you would have at least one case where a help page would be rather small, like one or two commands, because that's all the cog has - and ganging up more cogs on that page seems like the better option. Is there an easy way to implement per-cog paging without inconveniencing people that wouldn't like the super small help pages?

Ianardo-DiCaprio:
Issue, By removing the max_pages_in_guild, the bot will always try to post the help into the server and there is no current way to make it be forced to DM's like there is now with max_pages, so an option to force to DM's will need to be added.

bobloy:
> Issue, By removing the max_pages_in_guild, the bot will always try to post the help into the server and there is no current way to make it be forced to DM's like there is now with max_pages, so an option to force to DM's will need to be added.

After doing some digging, I found that `pm_help` is a parameter that is passed when creating the bot instance. This parameter currently handles forcing all help dialogue to PM's. 
My guess is this is a relic of help coming from discord.py, and could safely be replaced with a config value instead.

bobloy:
>One thing I noticed while testing is that even with a small character page size set (like 300-400 characters per help page), I would get 2 smaller cogs and 1 large cog on the first page (Admin, Alias, Audio) which led to a really large page vs. what I expected to see because Audio has so many commands comparatively.

#2281 Is my proposed fix to this problem.

> Is there an easy way to implement per-cog paging without inconveniencing people that wouldn't like the super small help pages?

I've done some digging on this. The current setup groups by cog unless the subcommands are >1000 characters, then it splits it into multiple pages using built-in `pagify`. Afterwards, it groups the cogs until it is **greater** than the set `page_char_limit` in helpset. This leads to inconsistent page sizes.

As for a solution, I think #2281 will reign in the weirdness by adjusting the secondary grouping to aim for **less than** `page_char_limit` grouping. Per cog grouping could go a number of ways though.

Uniform pages could be better accomplished by dropping the alphabetical precedent of ordering help. 
Per cog grouping with no regard for page size could be another option.

I'm currently toying around with it and will update here when I have something I like.

bobloy:
@aikaterna I've got a fix now for the grouping issues you brought up.

- Pagifying has been moved to the `group_fields` function
- Max 1000 characters is enforced
- `(Continued)` lines are respected for their actual length
- This replaces #2281 and can be closed if this is accepted.

bobloy:
One more commit to address the comment by @Iangit1 
Created a command to adjust the default `bot.pm_help` using `[p]helpset dm`

BurritoWrapped:
@bobloy Any progress on this?  Can I help in any way?

bobloy:
> @bobloy Any progress on this? Can I help in any way?

This PR is ready for review. You can help by testing it and reporting whether you see any issues or not.

Kowlin:
As a reminder, all pull requests that are **not** fixes are ignored for the time being while we're in this RC period

bobloy:
Adjusted the timeout for reactions being removed to 60 seconds from the default of 30 seconds.

The purpose is to avoid timing out while reading long help pages. I could see it going up to 120 seconds even, but let me know what you think.

aikaterna:
A couple things: helpset pagecharlimit may need to be reduced to a range. Anything under approx 8 characters seems to crash the bot, or at least lock it up so that it doesn't respond. Seems to start looking appropriate at around 100 characters minimum, but even then only one command seems to be shown at once. Also, there is no upper limit and while I wasn't able to get the bot to error out or produce an overly large embed, that should be limited to slightly under the embed character limit for that field.
Also, while many people are excited about menu help and I appreciate the help in DM's option that was added recently, you might want to add an option to this to have the old default long-page style help in guild. There were a couple people I spoke to about this PR that seemed to be disappointed it was only a menu option in-guild without any of the old help (multiple messages).

Ianardo-DiCaprio:
As a request for the helpset toggle. Could you make it so it's server specific. So I could have it send the help in some servers to send this new help menu, and some servers have it to send to DM's.

mikeshardmind:
I'd like this to be an option. I'd rather not remove the old formatter behavior (it's nice to just have a full help menu in places), but provide this as an alternative output for it.

mikeshardmind:
Closing due to large upstream changes in the help formatter.

I'd love to see this working, but it needs large retooling to be compatible with discord.py changes being added in #2587

PR #2258, opened 2018-10-17
[V3 Permissions] Find things uniquely for models

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This is a safety measure to prevent accidentally passing a user when you wanted a role with the same name. See discord discussion, and @Kowlin 

If it can't find it uniquely, it should tell them to use an id.

### DIFF
--- a/redbot/cogs/permissions/converters.py
+++ b/redbot/cogs/permissions/converters.py
@@ -1,71 +1,132 @@
+import itertools
 import re
-from typing import NamedTuple, Union, Optional, cast, Type
+from typing import NamedTuple, Union, Optional
+
+import discord
 
 from redbot.core import commands
 from redbot.core.i18n import Translator
 
 _ = Translator("PermissionsConverters", __file__)
 
-MENETION_RE = re.compile(r"^<(@[!&]?)|#([0-9]{15,21})>$")
+MENTION_RE = re.compile(r"^<(?:(?:@[!&]?)|#)(\d{15,21})>$")
+
+
+def _match_id(arg: str) -> Optional[int]:
+    if arg.isdigit() and 15 <= len(arg) <= 21:
+        return int(arg)
+    else:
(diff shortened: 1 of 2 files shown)


Tobotimus:
Needs a rebase before merging to be included in 3.0.0

mikeshardmind:
:+1: Could have sworn I branched off of V3/release/3.0.0, but was branched off of FETCH_HEAD instead.

redbot/cogs/permissions/converters.py
```diff
@@ -6,6 +6,58 @@
 _ = Translator("PermissionsConverters", __file__)
 
 
+class GlobalUniqueObjectFinder(commands.Converter):
+    @staticmethod
```
Tobotimus: Unfortunately static methods don't work as convert methods because they don't pass the `inspect.ismethod` predicate. They need to be class methods instead

redbot/cogs/permissions/converters.py
```diff
@@ -6,6 +6,58 @@
...
+            lambda obj: str(obj) == arg,
+            lambda obj: obj.name == arg,
+        ):
+            maybe_matches = list(filter(func, objects))
+            if len(maybe_matches) == 1:
+                return maybe_matches[0]
+            if len(maybe_matches) > 0:
+                raise commands.BadArgument(
+                    "Could not uniquely match that, please use an id or mention for this", arg
+                )
```
Tobotimus: Message args to `commands.BadArgument` are user-facing strings, and thus need to be translated

Tobotimus (dismissed):
This is a good change but could be done more efficiently. If the user passes an ID or mention it should find the object in constant time, or O(guild count) time for global.
This also won't find members from their nickname, if supplied.

Tobotimus (approved):

PR #2262, opened 2018-10-22
[V3 Filter] Filter based on words for non-phrases

### LINKED ISSUE #2261: Filter Regex v3
# Feature request

- [ ] Cog
- [x ] Command
- [ ] API functionality

The ability to use regex in [p]filter so if "hell" is added for example, it won't delete Hello.
<!--
Feel free to describe in as much detail as you wish.
To make Filter not delete everything that has the filtered word in.

If you are requesting a cog to be included in core: 
    - Describe the functionality in as much detail as possible
    - Include the command structure, if possible
    - Please note that unless it's something that should be core functionality,
      we reserve the right to reject your suggestion and point you to our cog 
      board to request it for a third-party cog
      
If you are requesting a command:
    - Include what cog it should be in and a name for the command
Filter
    - Describe the intended functionality for the command
To not censor all words that have a filtered word inside it.
    - Note any restrictions on who can use the command or where it can be used
Mod, admin, server owner and bot owner.

### LINKED ISSUE #2292: [Filter] [V3] Removes words not added to the filter.
# Other bugs

we added certain words to the filter to be filtered on the discord server.
and we want the words that consist of the bad word to remain on the channel.

for the word consisting of the bad. so not being the bad word.
For example: ass is on the filter but if someone then types assist or assistant it should leave those words in the channel, because they are not bad words.

#### What actually happened?

Someone mentioned that he needs assistance with a specific problem and the bot removed the word assistance because it has the word ass in it that is indeed on the filter

the word has been added with: !filter add ass

1. add a word to the the filter by typing for example: !filter add ass
2. type a word that has the filtered word in it like: assistant
3. The bot will remove the word assistant, event though the word itself isn't filtered but the word ass is.

I hope i did this reporting right, i've made an account with github just to report this issue.
Thanks for the great bot!

calebj:
### Type

- [x] Enhancement

Filters based strictly on words (ignoring punctuation) if filter entry isn't a phrase.

Resolves #2292.

### DIFF
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -1,13 +1,13 @@
 import discord
 import re
-from typing import Union
+from typing import Union, Set
 
 from redbot.core import checks, Config, modlog, commands
 from redbot.core.bot import Red
 from redbot.core.i18n import Translator, cog_i18n
 from redbot.core.utils.chat_formatting import pagify
 
-RE_WORD_SPLIT = re.compile(r'[^\w]')
+RE_WORD_SPLIT = re.compile(r"[^\w]")
 _ = Translator("Filter", __file__)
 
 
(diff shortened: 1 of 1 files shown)


Redjumpman (approved):
The solution here does in fact fix the problem that was addressed in 2292 for me, and I haven't found any issues with it to date.

Resolves #2292

PR #2264, opened 2018-10-23
Fix a bug

mikeo:
When a picture is sent, it doesn't contain any string. This means it checks if "" is in a string, which always returns True

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature


mikeo:
might be with an old version actually

Twentysix26:
Please give meaningful titles to your PRs in the future. Also, explain what your PR fixes / adds in terms of behaviour rather than describing what the code does, as the latter is not always self explanatory

PR #2269, opened 2018-10-25
Fix text formatting issue in session.py

FixedThink:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Removed two asterisks from a string: the "Question number {}" message was being emboldened twice from the left side, which made the bot send messages with two redundant asterisks.


Tobotimus (approved):
Thanks for catching this

PR #2270, opened 2018-10-28
[V3 Audio] Match openJDK 11 on Ubuntu

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Using the OpenJDK 11 from java.net on Ubuntu 18 reports "11" as the version, failing the Java version check on loading audio on a new instance. This change will return "11 0" as the version, passing the check, instead of just "11".


Tobotimus (approved):
Tested and works :+1:

PR #2273, opened 2018-10-29
[v3] [wip] docker + docker-compose support

calebj:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Allows a user to spin up a Red instance by configuring `.env` and pressing go. The Red image is built in-tree, and the Lavalink container uses the official image. There is also a Dockerfile included for custom builds of the Lavalink image, but the jar must be provided.

Since the Audio cog must be loaded and connected to Lavalink to configure an external server (see #2335), an environment variable check had to be added. It's the simplest way to pass a default hostname through to the LL server connection settings.

### TODO:
- [ ] Include docs for non-x86 w/ alternate lib paths (sedmelluq/lavaplayer#137)
- [ ] Use a regular volume OR store data in ~/.config like normal instances


SuperSandro2000:
Similar approach #2163

calebj:
Not so blocked anymore; this is open for testing and QA while I get documentation together.

mikeshardmind:
This is working on Linux Hosts (tested x86 and x64), and on Windows 10 hosts (tested x64)

I can't test on any other hosts currently.

palmtree5:
I can confirm this works on a Windows 10 host, though I want to note that `[p]shutdown` just leads to the bot restarting automatically which could be unexpected

docker/red-json/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM python:3.7-alpine
+
+WORKDIR /app
+ENV LD_LIBRARY_PATH /usr/local/lib:/usr/lib
+
+RUN echo "http://dl-8.alpinelinux.org/alpine/edge/community" >> /etc/apk/repositories
+
+RUN \
+    apk --no-cache add -q git openssl openssl-dev openssh alpine-sdk build-base gnupg linux-headers xz && \
```
SuperSandro2000: Why are the linux headers and alpine sdk kept in the final image?

calebj:
> [...] though I want to note that `[p]shutdown` just leads to the bot restarting automatically which could be unexpected

The last commit is a tweak that lets you use `--abort-on-container-exit` in docker-compose to get the desired behavior. This'll be noted in the docs.

SuperSandro2000:
@palmtree5 you can't stop anything I know from isnide a docker container.

I would set the restart policy of everything to:
``restart: unless-stopped`` or you easily get sticky containers that can be hard to remove and if one of them errors shit can't just stop it and examine the issue or attach a console.

Kowlin:
Out of curiosity, what's the status of this PR?

SuperSandro2000:
I still have a problem with:
* The restart policy
A container could potentially loop with a start up crash and I think this is not a good suggestions.
* The base image
Debian offers no advantage over Alpine except increasing the image size. IIRC it was the only image that had jre >9 a couple of months ago but this got resolved. If we stick with Debian we just force the user to download about 60 MB more.
Most popular pip packages have a previous alpine for all arches anyway.
* The dist-upgrade command that is really really bad practice.
Basically the user needs to download the core packages twice for no reason and it significantly increases the image size especially on the end of an image is life cycle.

@Kowlin you can basically use the images but you will probably need to download more data than necessary and need more storage space than necessary.
And look at the todo.
Edit: If you desperately need an image you can take a look at my old multi-arch one https://hub.docker.com/r/supersandro2000/reddiscord/

docker/red-json/Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+RUN apt-get update && \
+    apt-get install -y --no-install-recommends git ca-certificates xz-utils libssl-dev build-essential && \
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    pip install -U pip Cython pytest -q --retries 30 && \
+    rm -rf /root/.cache/pip/*
+
+COPY . /red_src/
+
+RUN \
```
SuperSandro2000: Put the copy before the first run and combine them or your removal process hurts the image more than it helps.

docker/red-json/docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+services:
+  lavalink:
+    # Comment image and uncomment build to use a custom .jar
+    image: fredboat/lavalink:master-v3
+    # build: ../lavalink
+    restart: always
+    volumes:
+      - ../lavalink/application.yml:/opt/Lavalink/application.yml:ro
+    expose:
+     - "2332"
```
SuperSandro2000: If you connect to the lavalink container from another docker container via ``lavalink`` you don't need to expose anything on the host.
We still need to consider that if those ports are used the container won't start.

docker/red-mongodb/Dockerfile
```diff
@@ -0,0 +1,29 @@
+FROM python:3.7-slim-stretch
+
+WORKDIR /app
+ENV LD_LIBRARY_PATH /usr/local/lib:/usr/lib
+
+RUN apt-get update && \
+    apt-get dist-upgrade -y && \
```
SuperSandro2000: You don't run a dist-upgrade in a docker container!!! Either upgrade a package you need to update or build a base image with the updated packages. You potentially create a >400mb diff here and ship those megabaytes twice to the user.

docker/red-mongodb/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+    apt-get install -y --no-install-recommends git ca-certificates xz-utils libssl-dev build-essential && \
```
SuperSandro2000: After the build we can remove build essentials if they are not needed on run time and they get removed in the same RUN command.

docker/red-mongodb/Dockerfile
```diff
@@ -0,0 +1,29 @@
...
+    apt-get clean && \
+    rm -rf /var/lib/apt/lists/* && \
+    pip install -U pip Cython pytest -q --retries 30 && \
```
SuperSandro2000: I think that 3 retries are enough or the container build could be blocked for 1 or 2 minutes before it fails.

mikeshardmind:
`restart: unless-stopped` breaks the ability to shut down the bot from discord. 
Currently possible using `--abort-on-container-exit` in compose

`restart: on-failure` makes sense for this use case.

SuperSandro2000:
@mikeshardmind makes sense. We could argue a whole day about should you be able to but we don't need to.
But we could change the other containers. I just don't like container that start even if I shut them off manually.

SuperSandro2000:
@calebj I did a quick branch with all my changes that I think are necessary before this can be merged over [here](https://github.com/SuperSandro2000/Red-DiscordBot/commits/docker). Feel free to merge or reuse them.

abelfodil:
https://github.com/abelfodil/Red-DiscordBot-Docker

Drapersniper:
> https://github.com/abelfodil/Red-DiscordBot-Docker

This docker image has some things we wouldn’t want by default, the Prefix is set to “-“ and the mentionable flag is also set, one thing worth mentioning too is that’s JSON

Drapersniper:
Another thing worth flagging up is that Mongo is being depricated, and as such this should be updated to remove anything thats mongo specific (and possibly replace it with postgres)

Drapersniper:
@calebj is this something you are still actively working on or have you dropped this PR. asking to figure out wether to close and open an issue for Docker support or keep this open as you are working on it but haven't just pushed any changes.

Drapersniper:
Closing due to innactivity.

Note: Docker support is something we want, however for it to be implemented we require both educating SS as making is avaliable and documented would result in us actively supporting it. There is also other factors to consider here, but if anyone wishes to work on a full docker implementation by all means do so after talking to us in the main server about it (<https://discord.gg/red>)

SuperSandro2000:
> educating SS

Not sure what this means but I hope not what google offered me.

yamikaitou:
SS as in Support Squad

Kowlin:
No it isn't, it refers to support squad, it's an bit of an unfortunate mixing of 2 acronyms

PR #2281, opened 2018-10-31
[V3 Help] Group Fields respect page_char_limit

bobloy:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

It would groups the cogs until it is **greater** than the set `page_char_limit` in helpset. This leads to inconsistent page sizes when a large cog was appended to something barely under the limit.

I think this commit will reign in the weirdness by adjusting the secondary grouping to aim for **less than** `page_char_limit` grouping.


Tobotimus (approved):
Very well spotted. Thanks!

PR #2282, opened 2018-10-31
[V3 Mod] Support global users with userinfo

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

The `[p]userinfo` command can now fetch informations about any Discord user, even if he doesn't share a server with the bot, using the `bot.get_user_info(ID)` endpoint.

### Example

image

### DIFF
--- a/redbot/cogs/mod/mod.py
+++ b/redbot/cogs/mod/mod.py
@@ -2,7 +2,7 @@ import asyncio
 import contextlib
 from datetime import datetime, timedelta
 from collections import deque, defaultdict, namedtuple
-from typing import cast
+from typing import cast, Union
 
 import discord
 
@@ -42,7 +42,7 @@ class Mod(commands.Cog):
     def __init__(self, bot: Red):
         super().__init__()
         self.bot = bot
-        self.settings = Config.get_conf(self, 4961522000, force_registration=True)
+        self.settings = Config.get_conf(self, 4_961_522_000, force_registration=True)
 
         self.settings.register_guild(**self.default_guild_settings)
         self.settings.register_channel(**self.default_channel_settings)
@@ -1315,8 +1315,8 @@ class Mod(commands.Cog):
             is_member = True
 
         #  A special case for a special someone :^)
-        special_date = datetime(2016, 1, 10, 6, 8, 4, 443000)
-        is_special = user.id == 96130341705637888 and guild.id == 133049272517001216
+        special_date = datetime(2016, 1, 10, 6, 8, 4, 443_000)
+        is_special = user.id == 96_130_341_705_637_888 and guild.id == 133_049_272_517_001_216
 
         since_created = (ctx.message.created_at - user.created_at).days
         user_created = user.created_at.strftime("%d %b %Y %H:%M")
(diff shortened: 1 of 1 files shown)


redbot/cogs/mod/mod.py
```diff
@@ -1304,36 +1304,50 @@ def __unload(self):
 
         if not user:
             user = author
+        if isinstance(user, int):
+            user_id = user
+            user = await self.bot.get_user_info(user_id)
```
Zephyrkul: `get_user_info` isn't like `get_member` and the like - if the user ID is not found, it raises a 404 Not Found error (specifically `discord.NotFound`) instead of returning `None`.

It's also worth noting that there's a `get_user` method of `bot`, which will fetch the user from the bot's global cache of connected users. It may be worth checking that before using `get_user_info`.

redbot/cogs/mod/mod.py
```diff
@@ -1304,36 +1304,53 @@ def __unload(self):
 
         if not user:
             user = author
+        if isinstance(user, int):
+            user_id = user
+            user = self.bot.get_user(user_id)
+            if not user:
+                try:
+                    user = await self.bot.get_user_info(user_id)
+                except discord.errors.Forbidden:
```
Zephyrkul: `Forbidden` is a 403 error. You want to catch `NotFound`, a 404 error, as per my last comment.

mikeshardmind:
Considering the ratelimits that endpoint is subject to, I think this belongs in a separate command which isn't available to all users.

Zephyrkul:
How about allowing cache access (`get_user`) for anyone but only allowing `get_user_info` if it's the bot owner?

Kowlin:
I'd feel more comfortable with only allowing to look up the member from the cache if the the person the author is looking up has a mutual server with him. This avoids the issue of trying to brute force information on public Red bots if we ever decide to add more information to userinfo

mikeshardmind:
I've given this a little more thought as well. Rather than expand `userinfo` in this manner, I think it is appropriate to add a `finduser` command.

The following internal filtering would need to occur in such a command:

1. If the user is in bot.users:
  If any of the guilds are mutually shared between the bot, queried user, and author, allow the info.
  If the author is the bot owner, allow the info
  Otherwise, don't share the info

2. If the user is not in bot.users
   Allow only to the owner.

Additionally, this lookup should use `redbot.core.utils.antispam`  (Not a straight cooldown) to limit this commands repeated usability to non-bot owners.

laggron42:
@mikeshardmind good idea, but should this stay inside the mod cog?

mikeshardmind:
@retke I think so, this has very little practical purpose outside of moderation. I'm open to the new command being placed elsewhere with a logical reason for doing so though.

mikeshardmind:
I've pushed a sample implementation of this command into a `staging` cog on my cog repo. 
I'll PR this to exist within core mod if the implementation here looks reasonable enough for general use.

[Source](https://github.com/mikeshardmind/SinbadCogs/blob/v3/staging/staging.py)

laggron42:
Will be rewritten into a new command.

PR #2286, opened 2018-11-03
[V3 Core] Graceful shutdown when SIGTERM is received

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

When the bot receives SIGTERM, it will close the connexion (with `bot.shutdown`) correctly. Without this handler, the process is killed and the gateway is stuck, with the bot still connected for a while. Using `sudo systemctl stop red.service <instance>` is now safe to correctly shutdown the bot.

What it looks like in the logs:
```
laggron@discordbots:~/RedDB-fork$ sudo systemctl stop redfork-discordbot.service
laggron@discordbots:~/RedDB-fork$ systemctl status redfork-discordbot.service
● redfork-discordbot.service - Red-DiscordBotV3:Red
   Loaded: loaded (/etc/systemd/system/redfork-discordbot.service; disabled; vendor preset: enabled)
   Active: inactive (dead)

[...]
Nov 03 11:50:28 discordbots pipenv[12401]: [03/11/2018 11:50] INFO __main__ sigterm_handler 115: SIGTERM received. Quitting...
Nov 03 11:50:28 discordbots pipenv[12401]: [03/11/2018 11:50] INFO manager shutdown_lavalink_server 115: Shutting down lavalink server.
```

### DIFF
--- a/docs/autostart_systemd.rst
+++ b/docs/autostart_systemd.rst
@@ -49,6 +49,14 @@ To set the bot to start on boot, you must enable the service, again adding the i
 
 :code:`sudo systemctl enable red@instancename`
 
+If you need to shutdown the bot, you can use the ``[p]shutdown`` command or
+you can stop the service, still by adding the instance name after the **a**:
+
+:code:`sudo systemctl stop red@instancename`
+
+.. warning:: If the service doesn't stop in the next 10 seconds, the process is killed.
+    Check your logs to know the cause of the error that prevents the shutdown.
+
 To view Red’s log, you can acccess through journalctl:
 
 :code:`sudo journalctl -u red@instancename`
--- a/docs/autostart_systemd.rst
+++ b/docs/autostart_systemd.rst
@@ -28,6 +28,7 @@ Paste the following and replace all instances of :code:`username` with the usern
     Restart=always
     RestartSec=15
     RestartPreventExitStatus=0
+    TimeoutStopSec=10
 
     [Install]
     WantedBy=multi-user.target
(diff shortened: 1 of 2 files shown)


laggron42:
Note: this was only tested on Ubutnu 18.10 with systemdaemon, it should work the same with upstart, but I don't know how Windows handle this, so this might need testing on this OS (I don't have Red installed on Win)

docs/autostart_systemd.rst
```diff
@@ -48,6 +49,14 @@ To set the bot to start on boot, you must enable the service, again adding the i
 
 :code:`sudo systemctl enable red@instancename`
 
+If you need to shutdown the bot, you can use the ``[p]shutdown`` command or
+you can stop the service, still by adding the instance name after the **@**:
```
Tobotimus: Grammar is slightly off here, maybe just say

> ...or you can stop the service with the following command:

redbot/__main__.py
```diff
@@ -139,6 +145,7 @@ def main():
     if cli_flags.dev:
         red.add_cog(Dev())
     loop = asyncio.get_event_loop()
+    loop.add_signal_handler(SIGTERM, lambda: asyncio.ensure_future(sigterm_handler(red, log)))
```
Tobotimus: This method is not implemented on non-unix systems, so this needs to be guarded under an `if os.name == "posix"` condition.

Tobotimus (changes requested):
Thanks for adding this. Just need a few changes and it should be good to go.

Tobotimus (approved):

PR #2290, opened 2018-11-06
[V3 Audio] Disallow seek during active vote

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Users are able to use seek to skip songs while voteskip is active. This change prevents users from doing so.


Tobotimus (approved):
:+1:

PR #2304, opened 2018-11-19
[V3 Downloader] Findcog fix with no repo installed

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

The findcog command errors out if a repo is removed. To reproduce, install a repo, install and load a cog, remove the repo, and then use the findcog command on a command in the installed cog. This change informs the user that the repo is not installed instead of erroring out.


Tobotimus (approved):
Looks good to me :+1:

PR #2311, opened 2018-11-24
[Mod] Fixed not handling of nicknames that are too long.

Kowlin:
### Type

- [x] Bugfix

Fixed Mod not handling nicknames that exceed 32 characters.

### DIFF
--- a/redbot/cogs/mod/mod.py
+++ b/redbot/cogs/mod/mod.py
@@ -866,27 +866,42 @@ class Mod(commands.Cog):
     @commands.guild_only()
     @commands.bot_has_permissions(manage_nicknames=True)
     @checks.admin_or_permissions(manage_nicknames=True)
-    async def rename(self, ctx: commands.Context, user: discord.Member, *, nickname=""):
+    async def rename(self, ctx: commands.Context, user: discord.Member, *, nickname: str = ""):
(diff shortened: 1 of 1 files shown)


mikeshardmind (changes requested):
This still isn't completely sufficient for preventig a traceback with this. see <https://discordapp.com/developers/docs/resources/user#usernames-and-nicknames>, should wrap in a try/except additionally.

redbot/cogs/mod/mod.py
```diff
@@ -877,7 +877,15 @@ def __unload(self):
         if len(nickname) > 32:
             await ctx.send("The nickname you have provided is too long.")
         else:
-            await user.edit(reason=get_audit_reason(ctx.author, None), nick=nickname)
+            try:
+                await user.edit(reason=get_audit_reason(ctx.author, None), nick=nickname)
+            except discord.Forbidden:
```
mikeshardmind, line 895: I'd prefer doing a hierarchy and perm check before the API call here.

The other error still needs to be caught for cases where the nick limitation is a result of discord not sharing all invalid nick parameters.

mikeshardmind (changes requested):

Tobotimus (approved):

PR #2326, opened 2018-12-10
Tools for marking things unsafe for general use

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Tangentially related to #2324.

This adds some quick tools for marking things as unsafe for general use. It does not block their usage but will raise some runtime warnings if they are used. These warning can be suppressed in known safe contexts with `warnings.catch_warnings`


Tobotimus:
This looks good at first glance, but I think the `safety` module belongs in `utils`. It doesn't have to be documented for 3rd party developers, but I see it more as a utility than part of core functionality.

mikeshardmind:
I've gone ahead and moved it to `redbot.core.utils.safety` I've intentionally not added anything to RTD about this.

Kowlin (dismissed):

tekulvw (dismissed):

tekulvw (approved):

PR #2331, opened 2018-12-13
[Mod] Fixed loud RuntimeError on modlog cases

Kowlin:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This change stops the loud modlog error from occurring if no modlog channel is found.


Tobotimus (approved):
Thanks for this, I also corrected the docstring of `get_modlog_channel`, as it never returns None. I also removed the comment you added, since I think it assumed the previous docstring was correct

PR #2334, opened 2018-12-16
[V3 General] Make urban require a nsfw channel

mikeshardmind:
Due to the nature of the likely results of most uses of this command, this really needs it to comply with discord's guidelines.

### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Note: Because this is making a change to comply with discord guidelines, I'm calling it a behavior fix.

### DIFF
--- a/redbot/cogs/general/general.py
+++ b/redbot/cogs/general/general.py
@@ -247,6 +247,7 @@ class General(commands.Cog):
         except discord.Forbidden:
             await ctx.send(_("I need the `Embed links` permission to send this."))
 
+    @commands.is_nsfw()
     @commands.command()
     async def urban(self, ctx, *, word):
         """Search the Urban Dictionary.


Kowlin:
I'd like to propose that bot mods and higher can do it anywhere in a guild. Seeing as moderators are *usually* trusted people within a community

mikeshardmind:
This is quickly leaving the scope of something that could be included in 3.0 as a minor behavior fix.
Between this, and the potential in the image cog, I think we should override the default `nsfw` check to allow overrides in permissions as well as add `nsfw_or_mod` and `nsfw_or_admin` if we're going this route.

Tobotimus:
I think overriding the nsfw check to use `@commands.permissions_check` instead would be the best way to go about this now, I know we have to adhere to Discord's guidelines but I am a little wary this will annoy users.

Tobotimus:
Actually, taking a closer look at discord's guidelines, I don't think this is necessary: 

>Not properly labeling NSFW channels [is not tolerated]. Some content just isn't appropriate for kids under 18, so let's do the right thing in our community and label any channel containing adult content as NSFW. This will add an age gate to the channel, protecting those who wish not to see.

The term "adult content" here is kind of open-ended but I don't think the urban command is really covered by it. Obviously there are lots of different interpretations of what is "appropriate" for kids between ages 13 and 18, but the occasional dirty joke isn't going to cause a riot.

Moreover the guidelines seem to suggest the admin labelling channels should be held liable, and they're still able to disable the command in channels which are not NSFW if they wish.

PR #2272, opened 2018-10-29
[Audio] Lavalink 3.0/3.1 compatibility updates

calebj:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

### Description of the changes
- Update to red-lavalink v0.2.0 (blocked by ~~Cog-Creators/Red-Lavalink#41~~ ~~red-lavalink release~~)
- Force lavalink to use TLSv1.2 on java 11+ ~~(blocked by #2270)~~

I would add equalizer support now, but there's no way to know the full Lavalink version (3.1.1+) and thus whether it's supported ahead of time. It's being worked on, though.

PR #2335, opened 2018-12-17
[v3 audio] connect to lavalink in the background

### LINKED ISSUE #2306: [V3 Audio] Unable to set external lavalink
Right now, audio can't be loaded if lavalink can't be started. If the reason it can't be started is the port is already in use by another instance or by something completely unrelated, then the cog can't be configured to use an external lavalink server, and thus gets locked in an inability to load it.

I think the best method to fix this would be to load the cog even without the lavalink server, but place a check on commands that need lavalink that blocks them if there is no lavalink server connection. This would allow remedying this without having to interfere with other running processes, shut down other instances, or manually modify configuration files.

calebj:
### Type

- [x] Bugfix
- [x] Enhancement

See title, as well as:
- restart and reconnect if connection settings change
  - shutdown and restart if not configured to use external
- show a message in [p]play et al. when the connection hasn't been made
- move the JAR download to manager so audio.py can access it
- only start if no process exists

Resolves #2306


Tobotimus:
I'm considering having this included in 3.0.1 as a bugfix

calebj:
That's fine. It's the final step for working split containers in Docker/compose without an environmental variable hack. Now that the LL v3 compatibility patch (#2272) is in, #2273 is a go after this issue is fixed. At least, for x86 users.

Tobotimus (dismissed):
I've tested this and it's a good step forward, thanks :smiley:

Tobotimus (approved):

PR #2338, opened 2018-12-20
[V3 core] on_message_without_command dispatch

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

Implements a new dispatch for messages which were not treated as a command. 

Should reduce overhead compared to the current best method of each event generating a context for the message to check validity if they want to process commands differently to normal messages, while still handling both.


Kowlin (approved):
Works like a charm.

PR #2339, opened 2018-12-20
[Permissions] Help menu respects rules

### LINKED ISSUE #2249: [V3 Help] Help filtering bug.
# Other bugs

Using permissions, default deny a Cog, reallow a command from it to some scope.

Help should show that cog in scopes where there is at least 1 usable command from it.

#### What actually happened?

Help can be invoked for the command, command works, but the cog help is not shown.

 - Using permissions, default deny a Cog, reallow a command from it to some scope.
 - Invoke help for the cog

### Note:
Reporting this for @Kowlin as he seems to have gotten distracted, behavior was first observed in Fortnite server.

Tobotimus:
Resolves #2249.


mikeshardmind (approved):
I wouldn't mind an extra set of eyes on this, but it looks good to go to me.

PR #2342, opened 2018-12-21
[DataManager] Don't copy bundled data to cog data folder

### LINKED ISSUE #2280: [Core] DataManager/Bundled_Data_Path cannot be used in __init__
# Other bugs

have bundled_data_path in a variable initialized in `__init__`

Successful initialization of the variable with the data path

#### What actually happened?

Init occurred faster than the bundled data folder creation, causing the cog to fail with a "dir could not be found"

Put the following in any cog that has data and try loading the cog:

```
def __init__(self):
self.path = bundled_data_path(self)
```

### LINKED ISSUE #2329: [V3] Bundled data is not properly nuked, old data files stick around

Remove files from the bundled data

The files removed from the repo would be removed from bundled data/cog data folders on update

#### What actually happened?

The new files are in place, but so are the old files.

Set up a data folder, put 1.jpg 2.jpg and 3.jpg in there. Load the cog via github, push a commit that deletes 2.jpg, then update the cog. You'll have all three files still available to your data path.

```
high 冬(winter) energyToday at 11:18 PM
<<weeed set backgroundImage list

RedbotNGBOTToday at 11:18 PM
backgrounds: ['CrlxLkzUEAEzEan.jpg', 'CrlxLk4UkAAww6Z.jpg', 'forest3.jpg', 'beach-paradise-beach-desktop.jpg', 'backround.jpg.cf8567d9efb8925809436a2c1f01e8b0.jpg', 'CrlxLk1VYAAQdzi.jpg', 'backround2.jpg.6f994a20cc67e6a0058a6bed9cb7c123.jpg', 'forest1.jpg', 'glitchscene3.jpg', 'forest2.jpg', 'glitchscene.jpg', 'glitchscene2.jpg', 'backround3.jpg.5fa80a45458028f986e20c096082e888.jpg']
```

unknown

Tobotimus:
- `load_bundled_data` is now deprecated and obselete
- Bundled data is now no longer copied to the cog's data folder, greatly simplifying its operations under the hood. Instead, `bundled_data_path` will now return the path to the folder from which the files would have previously been copied.
- Since we're already explicitly asking cog creators not to modify anything inside the bundled data folder, this shouldn't be a problem for anyone.

Resolves #2329.

Resolves #2280.

### DIFF
--- a/redbot/core/data_manager.py
+++ b/redbot/core/data_manager.py
@@ -154,7 +154,7 @@ def core_data_path() -> Path:
 
 
 # noinspection PyUnusedLocal
-@deprecated("This function is now obselete.")
+@deprecated("bundled_data_path() without calling this function")
 def load_bundled_data(cog_instance, init_location: str):
     pass
 
@@ -187,8 +187,7 @@ def bundled_data_path(cog_instance: commands.Cog) -> Path:
         If no bundled data folder exists.
 
     """
-
-    bundled_path = Path(inspect.getfile(cog_instance)).parent / "data"
+    bundled_path = Path(inspect.getfile(cog_instance.__class__)).parent / "data"
 
     if not bundled_path.is_dir():
         raise FileNotFoundError("No such directory {}".format(bundled_path))
--- a/redbot/core/data_manager.py
+++ b/redbot/core/data_manager.py
@@ -1,15 +1,15 @@
-import sys
-import os
-from pathlib import Path
-from typing import List
-from copy import deepcopy
-import hashlib
-import shutil
+import inspect
 import logging
+import os
+import sys
+import tempfile
+from copy import deepcopy
+from pathlib import Path
 
 import appdirs
-import tempfile
+from discord.utils import deprecated
 
+from . import commands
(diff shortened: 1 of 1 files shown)


redbot/core/data_manager.py
```diff
@@ -153,124 +153,28 @@ def core_data_path() -> Path:
...
-        if exists_checksum is not None:
-            orig_checksum = hash_bytestr_iter(file_as_blockiter(orig.open("rb")), hashlib.sha256())
-
-        if exists_checksum != orig_checksum:
-            shutil.copy(str(orig), str(poss_existing))
-            log.debug("Copying {} to {}".format(orig, poss_existing))
-
-
+# noinspection PyUnusedLocal
+@deprecated("bundled_data_path() without calling this function")
```
Tobotimus, line 157: Just because this looks a little weird without extra understanding of `@discord.utils.deprecated`: This will print a warning saying "load_bundled_data is deprecated, use bundled_data_path() without calling this function instead"

mikeshardmind (approved):
Good to go.

PR #2345, opened 2018-12-22
[CogManager] Removal of implicit paths and general cleanup

### LINKED ISSUE #2330: [V3 CogManager] Paths issues
# Command bugs
#### Command name

This affects [p]paths and [p]removepath, and loading v3 core cogs when there is a v2 install on the same machine/vps

CogManagerUI

For removepath to be able to remove a path, also for a brand new install to have only the Install Path and the Core Path listed

#### What actually happened?

2 users in support and a few of my RC2 bots show an additional path listed that were not added at instance creation or later. For example:
```
Install Path: /home/user/.local/share/Red-DiscordBot/cogs/CogManager/cogs
Core Path: /usr/lib/python3.7/site-packages/redbot/cogs

1. /home/user/Red-DiscordBot/cogs
```
While there is a v2 install located at `/home/user/Red-DiscordBot/`, this location was not added by the user and on a new instance creation, this path was still added to the paths list automatically. If the user tries to use [p]removepath with this location in no. 1, it will say the path is removed but it never is removed from the [p]paths list.

This error makes the bot try to load v2 cogs for default v3 cogs, resulting in a traceback.

A RC2 bot on Heroku:
```
Install Path: /app/cogs/CogManager/cogs
Core Path: /app/.heroku/python/lib/python3.6/site-packages/redbot/cogs

1. /app/cogs
```
An RC2 bot on Ubuntu 18.04:
```
Install Path: /root/.local/share/Red-DiscordBot/cogs/CogManager/cogs
Core Path: /usr/local/lib/python3.6/dist-packages/redbot/cogs

1. /root/cogs
```
Strangely another instance that shares that vps/Red v3 install with that previous bot, does not have an additional path listed in 1.

I am not quite sure what makes this occur. Might be related to us grabbing the cwd in <https://github.com/Cog-Creators/Red-DiscordBot/blob/3b50ed819212ac77a517a24c0138c801031b1a61/redbot/core/bot.py#L114> and having a v2 install on the cwd path.

Tobotimus:
- Removed memory-sided `CogManager(paths=...)` argument and `CogManager._paths` attribute, as they have no practical use and were causing problems (resolves #2330)
- `[p]removepath` now removes the actual path displayed with the index specified in `[p]paths` (phantom-paths could previously be "removed" using `[p]removepath`)
- New coroutine method (`CogManager.user_defined_paths`) for retreiving a deduplicated list of user-defined paths as `Path` objects
- General cleanup so we don't have to do so much head-scratching next time an issue arises here

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -111,7 +111,7 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):
 
         self.main_dir = bot_dir
 
-        self.cog_mgr = CogManager(paths=(str(self.main_dir / "cogs"),))
+        self.cog_mgr = CogManager()
 
         super().__init__(*args, formatter=Help(), **kwargs)
 
--- a/redbot/core/cog_manager.py
+++ b/redbot/core/cog_manager.py
@@ -3,7 +3,7 @@ import pkgutil
 from importlib import import_module, invalidate_caches
 from importlib.machinery import ModuleSpec
 from pathlib import Path
-from typing import Tuple, Union, List, Optional
+from typing import Union, List, Optional
 
 import redbot.cogs
 from redbot.core.utils import deduplicate_iterables
@@ -25,8 +25,6 @@ class NoSuchCog(ImportError):
     Different from ImportError because some ImportErrors can happen inside cogs.
     """
 
-    pass
-
 
 class CogManager:
     """Directory manager for Red's cogs.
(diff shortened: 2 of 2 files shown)


aikaterna (approved):
These changes resolved the "ghost path" issue from 2330 on the bot that I tested it on and paths were recognized in order appropriately for cog loading, and was able to add, remove, and reorder paths with no issues. The only thing I found which is rather inconsequential is that the reorderpath command allows for values out of the range of path numbers, but it just places the specified path at the end of the list with no error which is fine in my opinion.

PR #2349, opened 2018-12-27
Updates to Filter

Ianardo-DiCaprio:
### Type

- [ ] Bugfix
- [ x] Enhancement
- [ ] New feature

Updated help of Filter to show it auto immunes bot owners, server owners/admins and it now sends a messages saying it needs the manage message perm to work if it doesn't have the perm. (requested in #feature-ideas)


Twentysix26:
The help string is fine, other than the blank line you removed: that "breaks" the short description.
The check is making incorrect assumptions:  
filters are server wide. The user might want to configure them in a channel but need them to take effect somewhere else, thus making the check limiting for no reason.  
I wouldn't check for any permission, you don't know where the user want the filters to have effect in.  
A better option could be explaining that filters will only have effect in channels where the permissions `manage_messages` is granted; a good moment to do that is when the user sets a new filter

Twentysix26:
Thanks for the changes, I remade it in #2357 and co-authored you

PR #2351, opened 2018-12-28
[Mod] Context-based voice checks

### LINKED ISSUE #2296: [Mod] Voice checks
I'm talking about [these checks](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/mod/checks.py), there are currently 2 issues with these checks.

The first issue is that these checks **do not** respect Permissions and will ignore allows or denies, since they do not use the core ``mod_or_permissions`` check for example.

The second issue is a bit more ingrained in the checks it self. As soon as these checks come across the first channel in a server that they do not have the permission to mute or deafen members. (for example lets take a channel override deny mute members) The entire check will fail and will return False, leaving the bot unable to actually use the command in that guild at all.

Affected commands:
``mute voice``
``unmute voice``
``voiceban``
``voiceunban``

Tobotimus:
- Removed `redbot.cogs.mod.checks` module
- Moved logic for formatting a user-friendly list of permissions to `redbot.core.utils.chat_formatting`
- `[p]voice(un)ban` and `[p](un)mute voice` now check permissions in the user's voice channel

Resolves #2296.


mikeshardmind (changes requested):
On both the muting and unmuting, there is an attempt to send error handling to the voice channel, instead of to context. There's also an unnecessary return statement at the end of the voice mute command.

Other than that, this works and is ready to go.

Tobotimus:
Thanks for the review @mikeshardmind, addressed.

PR #2358, opened 2019-01-05
[V3 Audio] Sort local files

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

When using the `[p]local` command, the files were given in a weird order (directly returned from `os.listdir()`) and it was a problem, especially with the playlists whose files are indexed. This PR just add a `sorted()` where the bot must fetch the local tracks.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -40,7 +40,7 @@ class Audio(commands.Cog):
     def __init__(self, bot):
         super().__init__()
         self.bot = bot
-        self.config = Config.get_conf(self, 2711759130, force_registration=True)
+        self.config = Config.get_conf(self, 2_711_759_130, force_registration=True)
 
         default_global = {
             "host": "localhost",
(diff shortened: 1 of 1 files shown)


Nodja:
I modified my code similarly to this and was gonna make a pull request, but I saw this here.

The only change I would make would be to use `sorted([], key=lambda s: s.casefold())` so that the case doesn't affect file order.

redbot/cogs/audio/audio.py
```diff
@@ -596,11 +596,13 @@ def __init__(self, bot):
...
-            for f in os.listdir(current_folder)
-            if (f.lower().endswith(allowed_files)) and (os.path.isfile(current_folder + f))
-        ]
+        folder_list = sorted(
+            [
+                f
+                for f in os.listdir(current_folder)
+                if (f.lower().endswith(allowed_files)) and (os.path.isfile(current_folder + f))
+            ]
+        )
```
Tobotimus, line 606: You can use `sorted` directly on a generator expression, so you can remove the square brackets:
```diff
- folder_list = sorted(
-     [
-         f
-         for f in os.listdir(current_folder)
-         if (f.lower().endswith(allowed_files)) and (os.path.isfile(current_folder + f))
-     ]
- )
+ folder_list = sorted(
+     f
+     for f in os.listdir(current_folder)
+     if (f.lower().endswith(allowed_files)) and (os.path.isfile(current_folder + f))
+ )
```

Tobotimus (changes requested):
Just a tiny change suggested :smile:

Also I agree with @Nodja that `casefold` should be used.

redbot/cogs/audio/audio.py
laggron42: 👍

Tobotimus (approved):

PR #2363, opened 2019-01-08
Catch exceptions in [p]backup

### LINKED ISSUE #2354: [V3 Backup] Uncaught exceptions
Backup command does not properly handle filesize issues or closed DMs. Needs fixing for RC3.

Tobotimus:
Resolves #2354.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1138,10 +1138,14 @@ class Core(commands.Cog, CoreLogic):
             with tarfile.open(str(backup_file), "w:gz") as tar:
                 for f in to_backup:
                     tar.add(str(f), recursive=False)
+                file_size = tar.gettarinfo().size
             print(str(backup_file))
             await ctx.send(
                 _("A backup has been made of this instance. It is at {}.").format(backup_file)
             )
+            if file_size > 8_000_000:
+                await ctx.send(_("This backup is to large to send via DM."))
+                return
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -1152,10 +1155,18 @@ def get_bot_uptime(self, *, brief: bool = False):
...
+                            await ctx.author.send(
+                                _("Here's a copy of the backup"),
+                                file=discord.File(str(backup_file)),
+                            )
+                    except discord.Forbidden:
+                        await ctx.send(
+                            _("I don't seem to be able to DM you. Do you have closed DMs?")
                         )
+                    except discord.HTTPException:
+                        await ctx.send(_("Sending the backup file failed. It may be too large."))
```
mikeshardmind:
```suggestion
                        await ctx.send(_("I could not send the file. It is still available at {location}").format(backup_file))
```
This should serve as a better catch-all since we no longer expect this to be a size issue, but there are other things such as API stability which could cause this.

mikeshardmind (approved):
This is good to go. There's a relatively minor UX change I've suggested, but I don't think it is necessary for approving this, so I leave that to your discretion.

redbot/core/core_commands.py
Tobotimus: Yeah you're right, I'm just gonna leave it as an open-ended "I could not send the file" rather than try to guess the issue that occurred

PR #2364, opened 2019-01-09
[V3 Docs] Add easier to use windows install instructions using chocolately + PS

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

### DIFF
--- a/docs/install_windows.rst
+++ b/docs/install_windows.rst
@@ -8,6 +8,25 @@ Installing Red on Windows
 Needed Software
 ---------------
 
+The following software dependencies can all be installed quickly and easily through powershell, 
+using a trusted package manager for windows called `Chocolatey <https://chocolatey.org>`_
+You can manually install all of it instead if you want.
+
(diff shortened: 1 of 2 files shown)


docs/install_windows.rst
```diff
@@ -8,6 +8,25 @@ Installing Red on Windows
 Needed Software
 ---------------
 
+The following software dependencies can all be installed quickly and easily through powershell,
```
palmtree5, line 11: Would suggest making the powershell/chocolatey install and the manual install separate subsections under "Needed Software

palmtree5 (dismissed):

docs/install_windows.rst
mikeshardmind: I've gone ahead and made those separate sections while leaving a brief description of the options for install above those subsections.

Tobotimus (approved):
Great, thanks for this guide, hopefully, this does make it easier for our special Windows users :)

PR #2365, opened 2019-01-09
[V3 Docs] Added third party developer documentation.

Redjumpman:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Added crucial information about the third party cog creator program so that users can easily access this information in an official way. This documentation is brought over from [this gist](https://gist.github.com/Redjumpman/3c07803f1202c3221d6e83e339b39995). Future changes should now be easier to add and maintain.


docs/tpd_requirements.rst
```diff
@@ -0,0 +1,164 @@
...
+- No cog contains malicious code.
+- No stolen or copied code found without proper credits.
+- Disclose if a cog has heavy memory or I/O usage.
+- Disclose if your cog collects any data.
+- Disclose if your cog contains any NSFW material.
+- Disclose if your cog creates or comes bundled with data outside of using Config.
+- Your code is async compatible (not blocking).
+- Does not break the Discord TOS.
+- Does not conflict with core cogs. (E.G., it does not cause a core cog to fail to load).
+- Contains at least three cogs. (See `ctx.send_filtered <tpd_requirements.html#quality-vs-quantity>`__)
```
mikeshardmind: The link is relevant, the hyperlink text doesn't seem to belong here.
```suggestion
- Contains at least three cogs. (We reserve the right to make exceptions for  `quality over quantity <tpd_requirements.html#quality-vs-quantity>`__)
```

docs/tpd_requirements.rst
```diff
@@ -0,0 +1,164 @@
...
+- Sanitize outputs when sending a user specified text. We recommend using `ctx.send_filtered <framework_bot.html#redbot.core.bot.RedBase.send_filtered>`__
+- Respect the role hierarchy. Don't let a lower role have a way to grant a higher role.
+- Don't write your own path to your included data. Use bundled data_path when possible.
+- Your main class should inherit from commands.Cog.
+- Use instance variables over globals whenever possible.
+- f-strings are incompatible with translation, so choose one or the other.
+- For most use-cases you should be using config for your I/O.
+- Use converters in command arguments instead of strings when applicable. 
+
+We have created a lot of tools to help eliminate boilerplate code. We may, at times, ask that you use some of these tools instead of writing your own, when appropriate. For example, ctx.send_filtered vs your own sanitized output.
```
mikeshardmind, line 128: maybe extend this line somewhat to indicate the purpose of this is to ensure if new problematic strings are added. Like, what if discord adds a new thing like `@mobile`, and `@desktop` (now that the API exposes and tracks this, I wouldn't put this past them) there's not double work being done to fix this, and bot hosts get a consistent behavior for normal output from the bot.

Not really a requirement, but to point out our reasons for asking devs to use the prebuilt stuff where possible may help people accept it.

docs/tpd_requirements.rst
```diff
@@ -0,0 +1,164 @@
...
+    installation message.
+
+Cog Package Folder
+^^^^^^^^^^^^^^^^^^
+The cog folder has the following requirements:
+
+- The folder name, must match the cog.py name. It should be in all lowercase letters
+- The folder should contain an info.json file. Read above for more details on the specific keys needed
+- An __init__.py for the cog
+- The .py file for the cog. (in all lowercase letters)
```
mikeshardmind, line 85: > The .py file for the cog. (in all lowercase letters)

This line can be removed. There's nothing in the downloader's design requiring this to be how a cog folder is structured, just that a setup file is discoverable in the folder, and that the folder is structured as a module.

docs/tpd_requirements.rst
```diff
@@ -0,0 +1,164 @@
...
+*******************
+
+- Class names must be CamelCase.
+- The main cog class and every command must have a doc-string.
+- Sanitize outputs when sending a user specified text. We recommend using `ctx.send_filtered <framework_bot.html#redbot.core.bot.RedBase.send_filtered>`__
+- Respect the role hierarchy. Don't let a lower role have a way to grant a higher role.
+- Don't write your own path to your included data. Use bundled data_path when possible.
+- Your main class should inherit from commands.Cog.
+- Use instance variables over globals whenever possible.
+- f-strings are incompatible with translation, so choose one or the other.
```
mikeshardmind, line 124: > - f-strings are incompatible with translation, so choose one or the other.
This isn't entirely accurate. A correction to this is that our translator can't translate an f-string.

An example of the distinction:
```py
T_ = Translator("SomeCog", __file__)
_ = lambda s: s
STATUS = _("Status:")
_ = T_

...
    await ctx.send(f"{_(STATUS)} {some_bool}")
```
This would work fine 

That said, due to language differences, for anything more complicated supporting i18n, developers should use keyword format to allow rearranging where variables would be inserted appropriately for the language.

It may be worth a footnote reference with a note on i18n best practices.

docs/tpd_requirements.rst
```diff
@@ -0,0 +1,164 @@
...
+- Credited other authors when applicable.
+
+*******************
+Coding Requirements
+*******************
+
+- Class names must be CamelCase.
+- The main cog class and every command must have a doc-string.
+- Sanitize outputs when sending a user specified text. We recommend using `ctx.send_filtered <framework_bot.html#redbot.core.bot.RedBase.send_filtered>`__
+- Respect the role hierarchy. Don't let a lower role have a way to grant a higher role.
```
mikeshardmind, line 120: This seems appropriate to include to an extent, but could get excessive if we were to list every common issue in this doc.

It may be worth an extra doc later on as we come across more commonly overlooked issues. For now, this should be fine here.

docs/tpd_requirements.rst
```diff
@@ -0,0 +1,164 @@
...
+- Don't write your own path to your included data. Use bundled data_path when possible.
+- Your main class should inherit from commands.Cog.
```
mikeshardmind, line 122: note on inheritance doesn't seem necessary here since one of our requirements is for it to not be broken, and this is enforced by the bot at cog load.

docs/tpd_requirements.rst
```diff
@@ -0,0 +1,164 @@
...
+- Disclose if your cog collects any data.
+- Disclose if your cog contains any NSFW material.
+- Disclose if your cog creates or comes bundled with data outside of using Config.
+- Your code is async compatible (not blocking).
+- Does not break the Discord TOS.
+- Does not conflict with core cogs. (E.G., it does not cause a core cog to fail to load).
+- Contains at least three cogs. (See `ctx.send_filtered <tpd_requirements.html#quality-vs-quantity>`__)
+- Does not over-saturate the current list of cogs.
+- Unusable commands are hidden.
+- Default locale must be English.
```
mikeshardmind, line 110: I think clarifying the why on this here may be worth it as well, but possibly as an RST referenced footnote given the dual nature of it and fitting into this list.

mikeshardmind (changes requested):
Tossed a few notes on this based on what I think of this. Really looks fine even if these changes aren't all made.

Redjumpman:
I'm having trouble rebasing, so I'll make a new PR with the changes requested here and internally from QA at a later date.

PR #2375, opened 2019-01-15
Reconcile permission hooks with ctx.permission_state

### LINKED ISSUE #2374: [V3 Permissions] Permission hooks don't set state
Permission hooks forcibly allowing a command don't set the permission state to allowed, meaning it's impossible to detect that normal behavior was overridden in this case. 

As already shown in voice checks for mod cog, having access to the intended state is important.

It might be worth adding to the enum for allowed/denied by hook (denied by hook too for anyone logging check failures)

mikeshardmind:
resolves #2374

See mod.py's voice mute and related issue details of why this may be necessary to some developers.

### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -500,7 +500,12 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):
             if result is not None:
                 hook_results.append(result)
         if hook_results:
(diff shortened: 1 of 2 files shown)


Tobotimus (approved):
Thanks, this is nice and elegant.

PR #2384, opened 2019-01-21
Install SHARED_LIBRARY requirements during library installation

### LINKED ISSUE #2381: [Downloader] Install requirements for shared libraries
# Feature request

- [ ] Cog
- [ ] Command
- [X] API functionality

As far as I can tell, installables which have a type of `SHARED_LIBRARY` [do not have their requirements also installed](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/downloader/downloader.py#L303-L315), since the actual installation of requirements happens during Downloader's `install` command, which only operates on things in `available_cogs`.

I create a lot of cogs that share similar utilities which reside in a few shared libraries (eg, communicating with various AWS services). It would be great if I could define the requirements of things in that library once in the library's `info.json`, rather than having to do it in every cog which consumes the library.

From a quick test, adding a call to `Repo.install_requirements` from inside `Repo.install_libraries` seems to do the trick, but that may not be the most performant solution and I'm not sure if there's any terrible downstream impacts of doing this.

Please forgive me in advance if I'm missing functionality which already exists. 🙁

sepdotgg:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This change modifies downloader/repo_manager to also install requirements for `SHARED_LIBRARY` InstallableType's. If multiple cogs in a repository utilize the shared library, this allows the requirements needed by specifically that library to be defined once in the library's `info.json` file, rather than having to define them in each cog which consumes the library.

Additionally, adds tests to:
- Validate correct parsing of `SHARED_LIBRARY` type `info.json` files
- Validate the call path of `repo.install_libraries` when the repo contains available libraries.

resolves #2381

### DIFF
--- a/docs/framework_downloader.rst
+++ b/docs/framework_downloader.rst
@@ -40,8 +40,7 @@ Keys specific to the cog info.json (case sensitive)
   Downloader will not deal with this functionality but it may be useful for other cogs.
 
 - ``requirements`` (list of strings) - list of required libraries that are
-  passed to pip on cog install. ``SHARED_LIBRARIES`` do NOT go in this
-  list.
+  passed to pip on cog install.
 
 - ``tags`` (list of strings) - A list of strings that are related to the
   functionality of the cog. Used to aid in searching.
--- a/docs/framework_downloader.rst
+++ b/docs/framework_downloader.rst
@@ -40,7 +40,8 @@ Keys specific to the cog info.json (case sensitive)
   Downloader will not deal with this functionality but it may be useful for other cogs.
 
 - ``requirements`` (list of strings) - list of required libraries that are
-  passed to pip on cog install.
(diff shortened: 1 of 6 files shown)


sepdotgg:
I just noticed that when Downloader was originally added, the docs [explicitly stated](https://github.com/Cog-Creators/Red-DiscordBot/commit/53810b2262ad96fade5a3aa31b533b1d34a9d8f1#diff-74abe773d093a9596278b21a7cdfe7c4R44) that libraries did not have their requirements installed. 

This PR has been updated to remove that statement from the docs, but now I'm interested if this will introduce some known flaw since it was an intentional design choice. 🤔

calebj (approved):

Tobotimus (approved):
Thanks for your contribution @Seputaes :D

PR #2389, opened 2019-01-27
[V3 Core] Allow central storage of API keys

TrustyJAID:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Create a central location to store external API tokens that can be used between cogs without requiring each cog to be loaded to share keys.

This adds `[p]set api` to the core commands and a special converter to return a dict for any type of API token. Not all API's use the exact same naming convention and not all cog creators will want to use this but it *should* provide a means for cog creators to share keys between cogs and store any type of key they want. An example setting multiple tokens such as the client_id and client_secret would be `[p]set api twitch client_id,12345 client_secret,123456` to ensure flexibility on multiple API's. 

This also updates the Streams cog to utilize the central database.

### DIFF
--- a/redbot/cogs/image/image.py
+++ b/redbot/cogs/image/image.py
@@ -32,6 +32,7 @@ class Image(commands.Cog):
         imgur_token = await self.settings.imgur_client_id()
         if imgur_token is not None and "imgur" not in await self.bot.db.api_tokens():
             await self.bot.db.api_tokens.set_raw("imgur", value={"client_id": imgur_token})
+            await self.config.settings.imgur_client_id.set(None)
 
     @commands.group(name="imgur")
     async def _imgur(self, ctx):
--- a/redbot/cogs/image/image.py
+++ b/redbot/cogs/image/image.py
@@ -32,7 +32,7 @@ class Image(commands.Cog):
         imgur_token = await self.settings.imgur_client_id()
         if imgur_token is not None and "imgur" not in await self.bot.db.api_tokens():
             await self.bot.db.api_tokens.set_raw("imgur", value={"client_id": imgur_token})
-            await self.config.settings.imgur_client_id.clear()
+            await self.settings.imgur_client_id.clear()
 
     @commands.group(name="imgur")
     async def _imgur(self, ctx):
--- a/redbot/cogs/image/image.py
+++ b/redbot/cogs/image/image.py
@@ -32,7 +32,7 @@ class Image(commands.Cog):
         imgur_token = await self.settings.imgur_client_id()
         if imgur_token is not None and "imgur" not in await self.bot.db.api_tokens():
             await self.bot.db.api_tokens.set_raw("imgur", value={"client_id": imgur_token})
-            await self.config.settings.imgur_client_id.set(None)
+            await self.config.settings.imgur_client_id.clear()
 
     @commands.group(name="imgur")
     async def _imgur(self, ctx):
--- a/redbot/cogs/image/image.py
+++ b/redbot/cogs/image/image.py
@@ -19,11 +19,19 @@ class Image(commands.Cog):
     def __init__(self, bot):
         super().__init__()
         self.bot = bot
+        self.settings = Config.get_conf(self, identifier=2652104208, force_registration=True)
+        self.settings.register_global(**self.default_global)
         self.session = aiohttp.ClientSession()
         self.imgur_base_url = "https://api.imgur.com/3/"
 
     def __unload(self):
         self.session.detach()
+
+    async def initialize(self) -> None:
+        """Move the API keys from cog stored config to core bot config if they exist."""
+        imgur_token = await self.settings.imgur_client_id()
+        if imgur_token is not None and "imgur" not in await self.bot.db.api_tokens():
+            await self.bot.db.api_tokens.set_raw("imgur", value={"client_id": imgur_token})
 
     @commands.group(name="imgur")
     async def _imgur(self, ctx):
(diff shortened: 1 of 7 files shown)


Tobotimus:
Hi! At first glance this looks great, there are some things we could do to make it more user-friendly though:
- Automatically migrate keys from previously expected storage locations in core cogs, to avoid making users reset the key manually.
- In the commands where keys were previously set, instead of sending a big help message, only tell the user the information they need, i.e. simply send "This command has been moved to `[p]set api ...`"

I haven't done a full review yet but I thought you might appreciate getting some early feedback. Thanks :)

Redjumpman:
This is a great addition and does a great job of extending our platform. I agree with the two points that Toby posted. My one additional suggestion, (and this can come later), is to add some documentation with examples. I think the doc strings are written well, but the more functionality that we add to the platform, the further we move away from vanilla discord.py. 

Requiring new cog creators to learn two frameworks, especially if we are in some situations forcing creators to use these tools (inheriting from commands.Cog for example), and in other areas strongly encouraging, we need to provide information on these tools that is easy for everyone to access and digest, rather than digging through the docs to find information on a random function.

I don't want to slow down the work of this PR by tacking on more work, just something to think about as a future PR.

TrustyJAID:
So before I start digging into these suggested changes, which I fully agree with, I was curious what you may be expecting to approach the problems. 

First moving old keys to the new format should be seamless but would you recommend on cog load or when we check for the keys for the first time? I'm thinking on load or in `async def initialize`.

Secondly, I've had to spend most of the night thinking about removal of the old key commands documentation. Those I feel are invaluable especially with core cogs since when a user loads the cog and wants to set the keys they need to know the process to acquire the keys they need and this is not something that can currently be handled by a single function handling all API keys. This is not really an issue for third part cogs since they can include it in the cog install message but this is missing from core cogs and therefore we need a way to describe the steps to get the key they need, preferably inside the cog itself. I was considering for the streams cog at least to make one command `[p]streamset tokens` that describes how to get each token for each platform. 

Let me know what you guys think.

Tobotimus:
`initialize()` is a good place to do migrations. Doing them on cog load is a good idea. See the permissions cog for an example.

With your second point, you're right, and perhaps you're best leaving it the way it is now. I forgot `[p]streamset api` can't really help users retrieve API keys from arbitrary services. However, perhaps instead of sending the instructions via the help menu, have commands like `[p]streamset youtubekey` send just send the instructions as a message, kind of like `[p]permissions explain`. Sometimes when users use a command and see a command help menu show up, they think they're using the command wrong.

redbot/cogs/image/image.py
```diff
@@ -27,6 +27,13 @@ def __init__(self, bot):
     def __unload(self):
         self.session.detach()
 
+    async def initialize(self) -> None:
+        """Move the API keys from cog stored config to core bot config if they exist."""
+        imgur_token = await self.settings.imgur_client_id()
+        if imgur_token is not None and "imgur" not in await self.bot.db.api_tokens():
+            await self.bot.db.api_tokens.set_raw("imgur", value={"client_id": imgur_token})
+            await self.config.settings.imgur_client_id.set(None)
```
Tobotimus: Rather than setting the key to `None`, to reset a key to its default, it's best to completely remove the key from storage with `Value.clear()`.
```suggestion
            await self.config.settings.imgur_client_id.clear()
```

redbot/cogs/streams/streamtypes.py
```diff
@@ -124,6 +134,7 @@ def __init__(self, **kwargs):
         # self.already_online = kwargs.pop("already_online", False)
         self._messages_cache = kwargs.pop("_messages_cache", [])
         self.type = self.__class__.__name__
+        self.token_name = None
```
Tobotimus: Here you can remove token_name as an instance variable, but keep it as a class variable annotation:
```py
from typing import ClassVar

...

class Stream:

    token_name: ClassVar[str]

    def __init__(self, **kwargs):
```

redbot/cogs/streams/streamtypes.py
```diff
@@ -36,10 +36,14 @@ def __init__(self, **kwargs):
         self.channels = kwargs.pop("channels", [])
         self._messages_cache = kwargs.pop("_messages_cache", [])
         self._token = kwargs.pop("token", None)
+        self.token_name = "twitch"
```
Tobotimus: Since token_name is a constant, it's best to declare it as a class variable:
```py
class TwitchCommunity:

    token_name = "twitch"

    def __init__(self, **kwargs):
```

redbot/cogs/streams/streamtypes.py
```diff
@@ -149,6 +160,7 @@ class YoutubeStream(Stream):
     def __init__(self, **kwargs):
         self.id = kwargs.pop("id", None)
         self._token = kwargs.pop("token", None)
+        self.token_name = "youtube"
```
Tobotimus: Also needs to be a class variable

redbot/cogs/streams/streamtypes.py
```diff
@@ -232,14 +244,18 @@ class TwitchStream(Stream):
     def __init__(self, **kwargs):
         self.id = kwargs.pop("id", None)
         self._token = kwargs.pop("token", None)
+        self.token_name = "twitch"
```
Tobotimus: Also needs to be a class variable

Tobotimus (dismissed):
Getting this error on startup with streams:
```py
[06/02/2019 19:02] ERROR events on_ready 86: Failed to load package streams
  File "redbot/core/events.py", line 84, in on_ready
  File "redbot/core/bot.py", line 240, in load_extension
    await lib.setup(self)
  File "redbot/cogs/streams/__init__.py", line 6, in setup
    await cog.initialize()
  File "redbot/cogs/streams/streams.py", line 75, in initialize
    self.streams = await self.load_streams()
  File "redbot/cogs/streams/streams.py", line 680, in load_streams
    token = await self.bot.db.api_tokens.get_raw(_class.token_name, default=None)
AttributeError: type object 'YoutubeStream' has no attribute 'token_name'
```

redbot/cogs/streams/streams.py
```diff
@@ -71,22 +71,38 @@ def check_name_or_id(self, data: str):
...
+        youtube = await self.bot.db.api_tokens.get_raw("youtube", default={})
+        twitch = await self.bot.db.api_tokens.get_raw("twitch", default={})
+        for token_type, token in tokens.items():
+            if token_type == "YoutubeStream":
+                if "api_key" not in youtube:
+                    await self.bot.db.api_tokens.set_raw("youtube", value={"api_key": token})
+            if token_type == "TwitchStream":
+                # Don't need to check Community since they're set the same
+                if "client_id" not in twitch:
+                    await self.bot.db.api_tokens.set_raw("twitch", value={"client_id": token})
```
Tobotimus: The nested `if` statements can be made into one using an `and`

redbot/cogs/streams/streams.py
```diff
@@ -71,22 +71,38 @@ def check_name_or_id(self, data: str):
...
 
+    async def move_api_keys(self):
+        """Move the API keys from cog stored config to core bot config if they exist."""
+        tokens = await self.db.tokens()
+        youtube = await self.bot.db.api_tokens.get_raw("youtube", default={})
+        twitch = await self.bot.db.api_tokens.get_raw("twitch", default={})
+        for token_type, token in tokens.items():
+            if token_type == "YoutubeStream":
+                if "api_key" not in youtube:
+                    await self.bot.db.api_tokens.set_raw("youtube", value={"api_key": token})
```
Tobotimus: The nested `if` statements can be made into one using an `and`

redbot/cogs/streams/streamtypes.py
```diff
@@ -118,6 +131,9 @@ def __repr__(self):
 
 
 class Stream:
+
+    token_name: ClassVar[str]
```
Tobotimus: Since this attribute can be `None` you'll want to type-hint it as a `ClassVar[Optional[str]]`. You can import `Optional` from the typing module too. Also should probably initialize it to `None` here as in the suggestion.
```suggestion
    token_name: ClassVar[Optional[str]] = None
```

Tobotimus (changes requested):
Just some very minor stylistic changes, otherwise this is good :smiley:

Tobotimus (approved):

PR #2391, opened 2019-01-28
Added support for comparing PermissionOverwrites

DevilXD:
When trying to check if a specific permission overwrite is already set on a channel, it can be done by fetching the already existing overwrite and comparing it with a one you want to set. Comparing them directly wasn't possible before, and this small change allows for that.
Could be expanded for other operations, not sure how it should work though.

### DIFF
--- a/discord/permissions.py
+++ b/discord/permissions.py
@@ -543,6 +543,10 @@ class PermissionOverwrite:
     +-----------+------------------------------------------+
     | Operation |               Description                |
     +===========+==========================================+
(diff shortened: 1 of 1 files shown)


mikeshardmind:
This should be made as an upstream PR to discord.py

DevilXD:
Soooo, not here? I thought it's linked to the discord.py repo ;-;

Still confused how github works sometimes... Should I close it and make a pull request on the original then?

DevilXD:
I'll make a pull request to the original then. Closing this one.

PR #2394, opened 2019-01-29
Bump minimum python version to 3.7.0

Tobotimus:
This will be due for Red 3.1.

This is in anticipation of #2246, although I've written that PR to not break on 3.6, the feature itself is not usable on 3.6. So I think the best way forward is to simply require python 3.7. This also allows devs and cog creators to utilise all of the new features in 3.7.


Kowlin (approved):
Looks good.

PR #2405, opened 2019-02-01
global preinvoke hook to resolve issue with discord.Member.joined_at

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes this, not only in core code, but every case where it could occur:

Example of the issue this fixes.
image

### DIFF
--- a/redbot/core/events.py
+++ b/redbot/core/events.py
@@ -318,6 +318,21 @@ def init_events(bot, cli_flags):
             if command_obj is not None:
                 command_obj.enable_in(guild)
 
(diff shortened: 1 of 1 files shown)


Kowlin (changes requested):
Tested the fix, and it caused major performance issues. Every command no matter if it involved an user caused an delay on the bot.

mikeshardmind:
Closing this. Testing shows this to be too much a performance drain as a global hook. Will need to guard things requiring comparisons here.

PR #2406, opened 2019-02-01
prevent traceback 

mikeshardmind:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Alternative handling for issue observed in #2405 without the large performance consideration.

related to Rapptz/discord.py#1638


Kowlin (approved):
This will work, but I'd recommend hiding the member counter from the embed when this error occurs.

Twentysix26:
Would this handle the hypothetical case where the *target* of `[p]userinfo` is a member with a None `joined_at`? (see lines before the fix)

mikeshardmind:
@Twentysix26 That shouldn't occur unless someone goes out of there way to find out the IDs of someone using lurker mode. That said, no it doesn't. I don't know if we care about that case, but additional things can be added to handle that.

PR #2407, opened 2019-02-02
[V3 Core] Utilize consume rest, Union

Zephyrkul:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Some core commands weren't using consume-rest behavior or typing.Union.


Redjumpman (approved):
This change is not presenting any issues for me. It also fixes the situation when things like a username contain's spaces. The user_or_role Union is also working based on my testing. I think this is good to go, and may warrant a back port to 3.0.0.

PR #2411, opened 2019-02-04
Utilise setup.cfg, move version info to redbot package

Tobotimus:
- Setup metadata and options moved to setup.cfg. Since we no longer require arbitrary code execution in order to install reliably (no more compiler detection or platform-specific python_requires), we can move the options to the standardised setup.cfg for cleanliness and transparency.
- `redbot.__init__` now is safe to import without installing dependencies. Version info has also been moved here, so the version can now be retreived with `redbot.__version__`
- Now deploying binary wheel distribution from travis, meaning faster and safer installs.


setup.cfg
```diff
@@ -1,3 +1,105 @@
...
+name = Red-DiscordBot
+version = attr: redbot.__version__
+description = A highly customisable Discord bot
 long_description = file: README.md
 long_description_content_type = text/markdown; charset=UTF-8; variant=GFM
+author = Cog-Creators
+author_email = cogcreators@gmail.com
+url = https://github.com/Cog-Creators/Red-DiscordBot
+classifiers =
+    Development Status :: 4 - Beta
```
Kowlin: This should probably be changed. Since we're out of beta.

Kowlin (commented):
There is a remaining tag in Pypi that indicates that we're in beta.

Tobotimus:
@Kowlin Thanks for reminding me, I just absentmindedly copied all of those over. I added some new ones as well.

PR #2421, opened 2019-02-05
[V3 Downloader] Use shlex for subprocesses

Zephyrkul:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This allows for spaces in the bot's data dir without breaking the cloning process.
~~`{noquote_arg}` can be used if splitting normally is still necessary (see the `PIP_INSTALL` value).~~ Edit: See below comment.

### DIFF
--- a/redbot/cogs/downloader/repo_manager.py
+++ b/redbot/cogs/downloader/repo_manager.py
@@ -8,8 +8,9 @@ import re
 from concurrent.futures import ThreadPoolExecutor
 from pathlib import Path
 from subprocess import run as sp_run, PIPE
+from string import Formatter
 from sys import executable
-from typing import List, Tuple, MutableMapping, Union, Optional
+from typing import List, Tuple, Iterable, MutableMapping, Union, Optional
 
 from redbot.core import data_manager, commands
 from redbot.core.utils import safe_delete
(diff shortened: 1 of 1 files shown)


Zephyrkul:
I've pushed a different approach to this. My original approach I disliked due to the fact that it:
1. required undocumented syntax (`"{noquote_arg}"`);
2. forbade any positional formatting args; and
3. just generally looked weird and turned a simple format call into half a dozen or more.

I instead decided to implement a [`ProcessFormatter`](https://github.com/zephyrkul/Red-DiscordBot/blob/V3/git-space/redbot/cogs/downloader/repo_manager.py#L27) class, which properly shlex-quotes all formatting args, including positional if that ends up being used, and returns a shlex-split list that can be used for subprocesses. Non-string iterables passed into the formatter will be considered seperate arguments for the command, so `{noquote_arg}` is no longer necessary (just pass a tuple or list, e.g. [L488](https://github.com/zephyrkul/Red-DiscordBot/blob/V3/git-space/redbot/cogs/downloader/repo_manager.py#L488)). I'd be willing to move this class somewhere into utils if it'd be useful elsewhere, but it's a very small class so it should be fine where it is.

Zephyrkul:
I hate to bother, but this is the second most common but with Downloader, yet it's been six days since the last commit and eight days since this PR opened. Am I missing something? Is this PR being put aside for a later, larger fix? Is this a duplicate? Is there something wrong with the code that should be addressed?

Twentysix26:
@zephyrkul I don't see anything particularly wrong with it, just nobody got around reviewing the PR yet. Time permitting, I'll try to review it tonight myself

Twentysix26 (approved):

Issue #2423, opened 2019-02-06
Better management of RPC servers


laggron42:
# Feature request

#### Select the type of feature you are requesting:

- [ ] Cog
- [ ] Command
- [x] API functionality

#### Describe your requested feature

I'm planning to use the websockets server provided with Red but I found a potential issue, similar to the Lavalink issues.
If a user has multiple instances on his server (which is encouraged with the new launcher), he gets issues with ports, since only one program can use a specific port.
The issue is frequent with Lavalink but can be easily fixed by using the `[p]llset external` setting of the Audio cog, then using one instance's Lavalink server or even running his own server. However, if a cog requires the RPC server, there is no option to prevent the issue of multiple servers on one port.

Possible fixes could be:
- allowing to set his own port for RPC with flags or commands
- running only one RPC server for the whole redbot program, all instances will then connect to that server

I need to be sure that RPC can run on multiple instances before even considering the usage of this in my cogs.

mikeshardmind:
fixed in #2429

PR #2429, opened 2019-02-07
[RPC] Set custom port with flags

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

As mentioned in #2423, having multiple instances with an RPC server will cause issues, due to multiple programs using the same port. This is a small addition that prevents the issue by adding a `--rpc-port` flag to `redbot`, allowing to set his own port. If not specified, it is still default to 6133.

The best fix to the RPC issue would be to have only one running server for all instances, but that would require a lot of changes to make the instances communicate between them and I don't feel like I can do such changes.


mikeshardmind (dismissed):
This needs a changelog entry. Beyond that, the only other potential place for concern is an unpriveleged user attempting to use a low port number, but that's an end user responsibility at that point with non-standard settings

mikeshardmind (approved):

PR #2431, opened 2019-02-07
[help_formatter] Remove unused characters

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Removes unused `\n` that are in the footer of help messages since we can't use the line break in embeds footers.


Kowlin (approved):
Even though this change doesn't really affect anything. This is a technically a fix.

PR #2436, opened 2019-02-09
Update dependencies and copyright year

Tobotimus:


Tobotimus:
This'll be ready once aiohttp-json-rpc v0.12 is pushed to PyPI.

PR #2443, opened 2019-02-11
Added Filterlog

entchen66:
I added a log system for filtered messages.
It can set up a log channel. As default it is disabled, but when a channel is set up it can be enabled.
All is set up per guild.

If it is enabled, it will post every filtered message with user and message content to the channel wich was given in the setup.

### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Two new commands for the filter cog.

* [p]filterset logenabled - will toggle the "FilterLog" settings
* [p]filterset logchannel [channel] - will set the channnel for logging

### DIFF
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -383,12 +383,12 @@ class Filter(commands.Cog):
                         logenabled = await self.settings.guild(server).filter_log()
                         if logenabled and isinstance(logchannel, discord.TextChannel):
                             embed = discord.Embed(color=discord.Color.red())
-                            embed.set_author(name=_('Message get filtered'))
-                            embed.add_field(name=_('Member'), value='{0.name}#{0.discriminator}\n({0.id})'.format(author))
-                            embed.add_field(name=_('Channel'), value=message.channel.mention)
+                            embed.set_author(name=_("Message get filtered"))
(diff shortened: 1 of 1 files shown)


Kowlin:
I'm honestly against this PR, I feel like its not right for us to start logging messages by default. I fear that merging this PR will lead to similarly structured PRs whereby more behavior of logging will be introduced. And I feel that its not the place for core Red to store this kind of information in the first place. I'll leave this open for now. But I'm **strongly** against this.

mikeshardmind:
A similar sentiment from me, but probably different reasoning.

I'd rather this not be part of core red. I have no objections to 3rd party cogs adding message logging capability, as those can be added in compliance with relevant data privacy laws, or not opted into.

By placing this in the core experience, we potentially make the bot unusable for some people.

I'm aware this is a useful thing to have for some moderation teams, but I don't feel like it is includable in core red as implemented here. At most, I could see adding an event for filter deleting a message which could then be handled appropriately to each individual use case.

entchen66:
An event, custom cogs could listen to, would also create the possibility to use the core filter and also log these messages.

BurritoWrapped (approved):

Tobotimus:
I largely agree with Kowlin and Mike here. Firing an event when a message is filtered would be fine. If you're willing to implement that, please do so in another PR :)

PR #2446, opened 2019-02-12
Refactor CogManager and extension loading

Tobotimus:
This kicked off with me realising our use of [`Loader.load_module()`](https://docs.python.org/3/library/importlib.html#importlib.machinery.SourceFileLoader.load_module) as a hacky way to import modules from random locations is deprecated. Looking closer, I realised there were a couple of smelly things here which could be cleaned up. Here's what's changed:
- Paths in CogManager are now inserted into `redbot.ext_cogs.__path__`. This means to import any external cog (from install path, cogmanager path, whatever), the import path would be `redbot.ext_cogs.cog_to_import`. It makes the rest of the logic really straightforward.
- Importing cog modules is now done with `importlib.import_module` - this is a high-level function which allows us to delegate all of the nitty gritty import logic to Python itself, and it should be much more reliable. We're also no longer dealing with `ModuleSpec` objects, we just go straight from cog name to module.
- We're now using `importlib.reload` to reload modules like a good boy. This is done from the bottom-up, twice, to make sure *most* internal imports in the package are updated. Some old imports could hang around if people do lazy relative importing at the module level.
- `Red.load_extension` can take either a module, or a straight name. This means loading a package is as simple as `Red.load_extension(package_name)`, no more needing to know what method of `CogManager` to use.
- Some internal documentation was added in a module docstring of `redbot.core.cog_manager`.


calebj:
I haven't noticed any issues so far in using this, so it seems good. However, I would like to see some more widespread ~~battlefield~~ testing with a variety of third party cogs and external modules before giving my :+1:.

Tobotimus:
This is ready for review again and will be targetting 3.2

mikeshardmind:
I still have a few concerns with regard to using `importlib.reload` in this manner, but I also haven't seen this break any cogs which I use in any way which wouldn't already be broken with the current reload strategy. I'd like to do another pass over this later before merging it, but it should make the reload behavior of cogs more consistent from an end user standpoint at least.

redbot/core/bot.py
```diff
@@ -590,28 +591,35 @@ def list_packages():
...
+    # Pattern to match parent package name of cog module (redbot.cogs. or redbot.ext_cogs.)
+    _COG_PACKAGE_RE = re.compile(r"^redbot\.(?:ext_)?cogs\.")
+
+    async def load_extension(self, module: Union[str, types.ModuleType]) -> None:
         # NB: this completely bypasses `discord.ext.commands.Bot._load_from_module_spec`
-        name = spec.name.split(".")[-1]
-        if name in self.extensions:
-            raise errors.PackageAlreadyLoaded(spec)
+        if isinstance(module, str):
+            module: types.ModuleType = self.cog_mgr.load_cog_module(module)
```
mikeshardmind, line 600:
```suggestion
            module: types.ModuleType = self._cog_mgr.load_cog_module(module)
```

redbot/core/cog_manager.py
```diff
@@ -392,12 +377,7 @@ class CogManagerUI(commands.Cog):
             await ctx.send(_("Invalid 'from' index."))
             return
 
-        try:
-            all_paths.insert(to, to_move)
-        except IndexError:
-            await ctx.send(_("Invalid 'to' index."))
-            return
-
+        await ctx.bot.cog_mgr.reorder_path(to_move, to)
```
mikeshardmind, line 380:
```suggestion
        await ctx.bot._cog_mgr.reorder_path(to_move, to)
```

redbot/core/core_commands.py
```diff
@@ -75,35 +77,35 @@ def __init__(self, bot: "Red"):
...
+        cog_modules: List[Tuple[str, types.ModuleType]] = []
 
         for name in cog_names:
             try:
-                spec = await bot._cog_mgr.find_cog(name)
-                if spec:
-                    cogspecs.append((spec, name))
-                else:
-                    notfound_packages.append(name)
+                module = bot.cog_mgr.load_cog_module(name)
```
mikeshardmind, line 92:
```suggestion
                module = bot._cog_mgr.load_cog_module(name)
```

redbot/core/core_commands.py
```diff
@@ -75,35 +77,35 @@ def __init__(self, bot: "Red"):
...
                 failed_packages.append(name)
+            else:
+                cog_modules.append((name, module))
 
-        for spec, name in cogspecs:
+        for name, module in cog_modules:
             try:
-                self._cleanup_and_refresh_modules(spec.name)
-                await bot.load_extension(spec)
+                module = bot.cog_mgr.reload(module)
```
mikeshardmind, line 107:
```suggestion
                module = bot._cog_mgr.reload(module)
```

redbot/core/core_commands.py
```diff
@@ -2198,14 +2173,9 @@ class Core(commands.Cog, CoreLogic):
...
-
-        spec = await self.bot._cog_mgr.find_cog(cog_name)
-        if spec is None:
-            raise LookupError("No such cog found.")
-
-        self._cleanup_and_refresh_modules(spec.name)
-
-        await self.bot.load_extension(spec)
+        module = self.bot.cog_mgr.load_cog_module(cog_name)
+        module = self.bot.cog_mgr.reload(module)
```
mikeshardmind, line 2177:
```suggestion
        module = self.bot._cog_mgr.reload(module)
```

redbot/core/core_commands.py
```diff
@@ -2198,14 +2173,9 @@ class Core(commands.Cog, CoreLogic):
...
         cog_name = request.params[0]
-
-        spec = await self.bot._cog_mgr.find_cog(cog_name)
-        if spec is None:
-            raise LookupError("No such cog found.")
-
-        self._cleanup_and_refresh_modules(spec.name)
-
-        await self.bot.load_extension(spec)
+        module = self.bot.cog_mgr.load_cog_module(cog_name)
```
mikeshardmind, line 2176:
```suggestion
        module = self.bot._cog_mgr.load_cog_module(cog_name)
```

mikeshardmind (changes requested):

Drapersniper:
@Tobotimus is this something you are still actively working on or have you dropped this PR. Unsure whether you are currently working on this locally and just haven't pushed any commits.

Drapersniper:
Closing due to innactivity, Will convert into an issue so that other users can pick this up.

PR #2448, opened 2019-02-13
[Permissions] Allow multiple IDs when setting rules

### LINKED ISSUE #2214: [Permissions] Allow for mutiple IDs in permissions
# Feature request

- [x] Cog
- [x] Command
- [ ] API functionality

This idea is quite a simple one but I think this will go a long way in the usability and the speed of setting up permissions on the bot. Lets say I want to give a set of members the kick command (assuming kick is default deny all) without giving an role access to this. then I would type in:
``p addguildrule allow kick UserID1``
``p addguildrule allow kick UserID2``
etc.

However to greatly improve the speed at which this can be done we could allow the user to specify multiple IDs at once. As example:
``p addguildrule allow kick UserID1 UserID2 UserID3``

This might also work with commands and cogs to allow multiple of those in one go, but I suspect that would be much harder to implement correctly.

DiscordLiz:
Use `commands.Greedy` to apply multiple rules at once.

closes #2214

### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -278,7 +278,7 @@ class Permissions(commands.Cog):
         ctx: commands.Context,
         allow_or_deny: RuleType,
         cog_or_command: CogOrCommand,
-        who_or_what: GlobalUniqueObjectFinder,
+        who_or_what: commands.Greedy[GlobalUniqueObjectFinder],
     ):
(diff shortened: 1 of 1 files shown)


mikeshardmind (approved):
:+1:

tekulvw (approved):

PR #2458, opened 2019-02-15
Cooperate with custom error handlers

### LINKED ISSUE #2276: V3 Handler not shutting up after custom error handler.
# Other bugs

Replace a redbot error message (in this case the message displayed after CommandOnCooldown was triggered) with a custom message via @my_function.error:

    @my_function.error
    async def myfunction_error(self, ctx, error):
        if isinstance(error, commands.CommandOnCooldown):
             await ctx.send("I feel a little tired now. Ask me again in {:.0f}s".format(error.retry_after))

To only see my custom on_cooldown message from the event above.

#### What actually happened?

The bot wrote:
I feel a little tired now. Ask me again in 117s
This command is on cooldown. Try again in 116.69s

Add cooldown to any command and extend the @functionname.error handler like I described for your command. Then cause the command cooldown and try to use the command again while on cooldown.

DiscordLiz:
Adds detection of custom error handlers to the global error handler.

Adds documentation of the updated behavior.

closes #2276

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -410,7 +410,7 @@ for commands with a dedicated error handler.
 Red's global error handler will ignore commands with a registered error handler.
 
 To have red handle specific errors with the default behavior, 
-call ctx.bot.on_command_error with `manually_recalled` set to True
+call ctx.bot.on_command_error with `unhandled_by_cog` set to True
 
 For example:
 
@@ -422,7 +422,7 @@ For example:
             if isisntance(error, MyErrrorType):
                 self.log_exception(error)
             else:
-                await ctx.bot.on_command_error(ctx, error, manually_recalled=False)
+                await ctx.bot.on_command_error(ctx, error, unhandled_by_cog=True)
 
 Parameters
 -----------
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -398,6 +398,29 @@ class Command(CogCommandMixin, commands.Command):
         return old_rule, new_rule
 
 
+Command.error.__doc__ = """
+A decorator that registers a coroutine as a local error handler.
+
+A local error handler is an :func:`.on_command_error` event limited to
+a single command.
+
+The on_command_error event is still dispatched 
+for commands with a dedicated error handler.
+
+Red's global error handler will ignore commands with a registered error handler.
+
+Parameters
+-----------
+coro : :ref:`coroutine <coroutine>`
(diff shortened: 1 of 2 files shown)


calebj (changes requested):
It looks like Red already ignores (or at least, doesn't post) errors if a cog-wide exception handler is present (see line 191). If that becomes unreachable code, better to just skip that check altogether.

DiscordLiz:
I handled the feedback brought up on Discord and on GitHub separately.

@calebj brought up in discord that cog authors may not want to handle every error red does themselves.

I added a keyword only argument in the latest commit to allow developers to pass their errors back to Red's handler and have it skip the checks for if there were a cog handler already.

Other solutions to this which were discussed in discord.

1. Split error handling up into another function and the event. (@mikeshardmind brought this up)

This would incure an additional function call per error which wasn't handled, but would leave the event signature untouched.

2. Move error handling into the cog base class. (@mikeshardmind brought this up)

This would require significant changes to error handling to select the first handler in a cog class's mro, but would allow passing of unhandled exceptions using super.

3. Create a new convention of a red specific error handler for each type of error. (@calebj brought this up)

This would put an additional learning curve to new developers transitioning from using discord.py without Red, and would still require significant changes to upstream code to prevent double error handling.

This would also not allow people to use a cog error handler to log exceptions broadly without duplicating code.

calebj (changes requested):
Is it `manually_recalled` or `manually_reinvoked`?

calebj (dismissed):

redbot/core/commands/commands.py
```diff
@@ -398,6 +398,44 @@ def clear_rule_for(
...
+Parameters
+-----------
+coro : :ref:`coroutine <coroutine>`
+    The coroutine to register as the local error handler.
+
+Raises
+-------
+discord.ClientException
+    The coroutine is not actually a coroutine.
+"""
```
Tobotimus: Could you please create the `error(self, coro)` method on our command class to just call `super().error(coro)`, and then move this docstring there? That way it'll show up in our docs, and is more obvious in the code.

Tobotimus (changes requested):
Just one change I'm requesting then it's good :)

redbot/core/commands/commands.py
DiscordLiz: I've created the method and moved the docstring.

Tobotimus (approved):

PR #2461, opened 2019-02-16
[V3 Audio] Change pause to a toggle

aikaterna:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

* Change pause to a toggle as it was confusing as an explicit alias for resume
* Remove resume alias in preparation for removing other audio aliases, so owners can choose to assign their own aliases for audio commands instead of us dictating what they are
* Added catch in stop command to stop and clear queue when a song is paused, as it would just silently do nothing previously if that was the player state


TrustyJAID (approved):
This works well after testing.

Tobotimus (approved):

PR #2465, opened 2019-02-17
[V3 Audio] Add track length restriction

aikaterna:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

* Added [p]audioset maxlength, which takes seconds or 00:00-style of formatted input for restricting the player to songs that have a length under that threshold.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -404,7 +404,9 @@ class Audio(commands.Cog):
             msg += _("Jukebox:          [{jukebox_name}]\n").format(jukebox_name=jukebox)
             msg += _("Command price:    [{jukebox_price}]\n").format(jukebox_price=jukebox_price)
         if maxlength > 0:
-            msg += _("Max track length: [{tracklength}]\n").format(tracklength=self._dynamic_time(maxlength))
+            msg += _("Max track length: [{tracklength}]\n").format(
+                tracklength=self._dynamic_time(maxlength)
+            )
         msg += _(
             "Repeat:           [{repeat}]\n"
             "Shuffle:          [{shuffle}]\n"
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -308,7 +308,7 @@ class Audio(commands.Cog):
     async def maxlength(self, ctx, seconds):
         """Max length of a track to queue in seconds. 0 to disable.
 
-        Accepts seconds or a value formatted like 00:00:00 (h:m:s) or 00:00 (m:s).
+        Accepts seconds or a value formatted like 00:00:00 (`hh:mm:ss`) or 00:00 (`mm:ss`).
         Invalid input will turn the max length setting off."""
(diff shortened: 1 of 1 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -302,6 +303,29 @@ def _restart_connect(self):
         await self.config.guild(ctx.guild).emptydc_timer.set(seconds)
         await self.config.guild(ctx.guild).emptydc_enabled.set(enabled)
 
+    @audioset.command()
+    @checks.mod_or_permissions(administrator=True)
+    async def maxlength(self, ctx, seconds):
+        """Max length of a track to queue in seconds. 0 to disable.
+
+        Accepts seconds or a value formatted like 00:00:00 (h:m:s) or 00:00 (m:s).
```
TrustyJAID: `(h:m:s)` should be `(hh:mm:ss)` or `(h\\:m\\:s)` otherwise discord renders the `:m:` as an emoji. Alternatively surround the time markers with backticks.

redbot/cogs/audio/audio.py
```diff
@@ -2671,6 +2746,33 @@ def _track_creator(player, position=None, other_track=None):
...
+            length = round(track / 1000)
+        if length > 259200:
+            return True
+        elif length >= maxlength:
+            return False
+        else:
+            return True
+
+    async def _time_convert(self, length):
+        match = re.compile(r"(?:(\d+):)?([0-5][0-9]):([0-5][0-9])").match(length)
```
TrustyJAID: Replacing this regex with `(?:(\d+):)?([0-5]?[0-9]):([0-5][0-9])` can help improve usability where users can supply `8:00` instead of `08:00` for setting the max length a minor usability change.

redbot/cogs/audio/audio.py
```diff
@@ -2671,6 +2746,33 @@ def _track_creator(player, position=None, other_track=None):
             track_obj[key] = value
         return track_obj
 
+    @staticmethod
+    def _track_limit(ctx, track, maxlength):
+        try:
+            length = round(track.length / 1000)
+        except AttributeError:
+            length = round(track / 1000)
+        if length > 259200:
```
TrustyJAID: I'm curious about this and if it might have potential problems. Is this in case we're playing a live stream we don't care about the maxlength? Will this cause issues for streams that don't return a value of 72 hours?

TrustyJAID (changes requested):

redbot/cogs/audio/audio.py
aikaterna: Decided to move this value more towards the upper limit instead and added a comment regarding livestream length.

redbot/cogs/audio/audio.py
aikaterna: Thank you for this.

TrustyJAID (approved):
Working good after testing changes.

Tobotimus (approved):

PR #2467, opened 2019-02-18
Remove permissions's `p` alias

mikeshardmind:
People can readd it with the alias cog, but core red should not monopolize short aliases.

What about those servers that want play even shorter?

### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature


Tobotimus:
Good point. I can't help but feel this will annoy and/or confuse users slightly but allowing customisation is more important.

Tobotimus (approved):

PR #2469, opened 2019-02-19
Delete cooldown messages when expired

PredaaA:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

It delete cooldown message when cooldown is expired to avoid having unnecessary messages in channels.

Note : It's a suggestion, personally I use it because it's a thing that was reported to me and finally I prefer like that.


mikeshardmind (approved):
:+1:

tekulvw (approved):

mikeshardmind:
For context, there was some discussion about adding this within the last set of changes to cooldown handling but needed discussion.

tekulvw:
The "queued" build actually passed and the GH/Travis connection died.

PR #2470, opened 2019-02-19
[V3 Audio] Seek command can now seek to position

aikaterna:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

* Seek can now seek to a specific position, formatted like 00:00:00 or 00:00. Using negative or positive ints still functions the same as previously and will seek ahead or behind by that value instead.
* This PR requires the _time_convert func added in #2465 and should be merged after that one.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -2129,7 +2129,9 @@ class Audio(commands.Cog):
     @commands.command()
     @commands.guild_only()
     async def seek(self, ctx, seconds):
-        """Seek ahead or behind on a track by seconds or a to a specific time."""
+        """Seek ahead or behind on a track by seconds or a to a specific time.
+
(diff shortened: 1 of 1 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -2128,8 +2128,8 @@ def _restart_connect(self):
 
     @commands.command()
     @commands.guild_only()
-    async def seek(self, ctx, seconds: int = 30):
-        """Seek ahead or behind on a track by seconds."""
+    async def seek(self, ctx, seconds):
+        """Seek ahead or behind on a track by seconds or a to a specific time."""
```
TrustyJAID: Might be good to include similar documentation here as #2465 for how to skip to a specific time in the track.

TrustyJAID (changes requested):

TrustyJAID (approved):

Tobotimus (approved):

PR #2471, opened 2019-02-19
[V3 Warnings] Utilize modlog cases

TrustyJAID:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This update utilizes previously commented out modlog cases for warnings and slightly improves usage of custom warnings if enabled. Custom warnings can be done by `[p]warn user reason` if enabled instead of `[p]warn user custom reason`

### DIFF
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -2,7 +2,6 @@ from collections import namedtuple
 from typing import Union, Optional
 
 import discord
-import asyncio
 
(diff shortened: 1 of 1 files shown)


jonttm:
Looks great to me! Exactly what I needed! Just one thing: I noticed that the reason in `[p]warn user reason` is completely ignored while the description is used as the reason in the mod log. Not sure if this is intentional?

tekulvw (approved):

PR #2472, opened 2019-02-19
[V3 Audio] Add option for dc at queue end

aikaterna:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This addition adds a toggle for having the bot instantly disconnect when the queue ends. It takes precedence over the audioset emptydisconnect setting as it disconnects immediately when the queue or single song is finished.


Zephyrkul:
Would it be possible to set this on a timer? Or would there not be much point to that

Edit: From a conversation on Discord: no

TrustyJAID (approved):
Works great after testing.

Tobotimus (approved):

PR #2473, opened 2019-02-20
[V3 Audio] Emptydisconnect and status refactor

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

* Refactored disconnect_timer, the function behind audioset emptydisconnect, to be more appropriately responsive (thanks to TrustyJAID)
* Refactored status clearing/status changing when emptydisconnect or other Lavalink player statuses would have a TRACK_END event but no QUEUE_END event. This should clear or modify the bot's status event appropriately when [p]audioset status is on and the bot disconnects due to emptydisconnect and ideally play a little nicer with other cogs that set statuses.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -2619,7 +2619,7 @@ class Audio(commands.Cog):
                         stop_times[sid] = None
                         try:
                             await lavalink.get_player(sid).disconnect()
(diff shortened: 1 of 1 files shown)


TrustyJAID (approved):
This is working as expected.

Tobotimus (approved):

PR #2483, opened 2019-02-24
[V3 Filter] Throw a filter Event on_filter_message_delete

entchen66:
reference to https://github.com/Cog-Creators/Red-DiscordBot/pull/2443

### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Like Sinbad and Kowlin talked in PR 2443, I now add an event to the filter cog. It is thrown when a message got deleted and provide the message to custom cogs, just listen to the on_filter_message_delete event.

### DIFF
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -363,8 +363,8 @@ class Filter(commands.Cog):
 
         if hits:
             try:
-                self.bot.dispatch("filter_message_delete", message)
                 await message.delete()
+                self.bot.dispatch("filter_message_delete", message)
             except discord.HTTPException:
                 pass
             else:
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -363,6 +363,7 @@ class Filter(commands.Cog):
 
         if hits:
             try:
+                self.bot.dispatch("filter_message_delete", message)
                 await message.delete()
             except discord.HTTPException:
                 pass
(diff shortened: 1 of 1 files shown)


redbot/cogs/filter/filter.py
```diff
@@ -363,6 +363,7 @@ def __unload(self):
 
         if hits:
             try:
+                self.bot.dispatch("filter_message_delete", message)
```
Tobotimus: The event should only be dispatched if the deletion succeeds. To do this, move this line down to 371, just after the `else:` line :smiley:

Tobotimus (changes requested):
Thanks for taking the time to open this PR. Just would like you to change it a bit :smile:

Tobotimus:
It might also be useful to provide `hits` as an argument to event listeners, so for example with your logging use case, the log message could display exactly which words/phrases were detected in the message.

redbot/cogs/filter/filter.py
entchen66: oh yeah, sure^^

redbot/cogs/filter/filter.py
```diff
@@ -364,6 +364,7 @@ def __unload(self):
         if hits:
             try:
                 await message.delete()
+                self.bot.dispatch("filter_message_delete", message, hits)
```
Tobotimus: Please move it to line 371, after the `else:` line :D

Tobotimus (changes requested):

Tobotimus (approved):
Thanks for your contribution @entchen66 :smiley:

PR #2486, opened 2019-02-25
Update Translations

Tobotimus:
Also included a Makefile recipe which makes use of the Crowdin CLI's `crowdin download` command. This requires whoever is using it to provide the project's API key in an environment variable, but we may automate this at some point.


Kowlin (approved):
Working as intended.

PR #2487, opened 2019-02-25
[V3] Access the instance name throught the Red class

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

I'm developping a cog and a feature needs the name of the instance (not the name of the bot itself), and I was surprised that we cannot access it, even through multiple attributes, so I added this quick change to the Red class.

The instance name is now accessible with `bot.instance_name`, this is the name used for saving data or starting the bot with `redbot`, **not** the name of the Discord bot account.


Tobotimus:
It's available at `redbot.core.data_manager.instance_name`

PR #2496, opened 2019-02-27
Spelling correction of method name in Tunnel

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fix for wrong spelling in method name `files_from_attatch`, I changed it to `files_from_attach` in all its appearances.

This is actually breaking change, because of change of method's name, lol.

### DIFF
--- a/redbot/cogs/reports/reports.py
+++ b/redbot/cogs/reports/reports.py
@@ -165,7 +165,7 @@ class Reports(commands.Cog):
         if channel is None:
             return None
 
-        files: List[discord.File] = await Tunnel.files_from_attatch(msg)
+        files: List[discord.File] = await Tunnel.files_from_attach(msg)
 
         ticket_number = await self.config.guild(guild).next_ticket()
         await self.config.guild(guild).next_ticket.set(ticket_number + 1)
--- a/redbot/core/utils/tunnel.py
+++ b/redbot/core/utils/tunnel.py
@@ -158,6 +158,8 @@ class Tunnel(metaclass=TunnelMeta):
                 files.append(discord.File(_fp, filename=a.filename))
         return files
 
+    files_from_attatch = files_from_attach
+
     async def communicate(
         self, *, message: discord.Message, topic: str = None, skip_message_content: bool = False
     ):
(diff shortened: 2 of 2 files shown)


mikeshardmind (commented):
I'd prefer if this was done in a non-breaking fashion for now given how much QA has recommended people use this method when they need files in a quick and easy way.

```py

    @property
    def files_from_attatch(self):
        return self.files_from_attach

```

Adding this to the Tunnel class with your changes would be sufficient for that, the property could then be removed later on.

Jackenmen:
Because it is static method this would still be breaking change, because now you would have to create instance before using the method. I think it should be enough to just do a simple assignment to old name after defining the method.
```py
files_from_attatch = files_from_attach
```

mikeshardmind:
Right, forgot that was made a static method for a moment.

Jackenmen:
I added it then.

Jackenmen:
https://github.com/Cog-Creators/Red-DiscordBot/blob/e3cb29fbda6cb5d5f7ca30bec209cec96edd9848/redbot/core/utils/tunnel.py#L73-L76
also, I wanted to add PR for this, because it looks like bad logic to me, this will always send the message to the side that started the tunnel, either original channel, or the sender that the bot reacts to on this channel - I think it should be:
```py
send_to = self.recipient if uid == self.sender.id else self.origin
```
I just thought that maybe I could do it in this PR (assuming this isn't some strange logic that I just don't understand)

mikeshardmind:
your logic is solid there and properly fixes the intent (of sending a message of the tunnel being closed to the other end)

Jackenmen:
Can I do it here or should I just do another PR?

Tobotimus:
@jack1142 Another PR would be great for that change.

As for this PR, could you just add this comment above line 161?
```py
# Backwards-compatible typo fix (GH-2496)
```

Jackenmen:
Ayy, I wanted to do that from the beginning (I wouldn't have thought of adding PR number though), I'll commit it in a second

Jackenmen:
@Tobotimus updated, I'll add the other PR later today.

Tobotimus (approved):
Thanks @jack1142!

Issue #2500, opened 2019-03-01
[3.1] Mod Sprint


Zephyrkul:
With 3.1 closing in with #2499, I've collated a list of various issues and pull requests previously suggested on this repo. If I'm missing anything, feel free to either [open an issue](https://github.com/Cog-Creators/Red-DiscordBot/issues/new/choose) about it or comment below.

- [x] **#2501: Class compositing**
  - [ ] #2501: Move some Mod functionality to Core.
- [x] #2512 | #2501: Immediate modlog
- [ ] #2490: Move modlog-based listeners to modlog
- [ ] #2441: Default days for [p]ban
- [ ] #2437: Customize deleterepeats
- [ ] #2282: Global users with [p]userinfo
- [ ] #2231: Mute improvements
  - [ ] #2103 | #2140: Temporary Mutes
  - [ ] #1852: Guild mute by role
  - [ ] #1607 | #2599: Mute by category
  - [ ] #2179 | #2599: Mute quality of life
    - [ ] Remove voice mute / unmute
    - [ ] Category mute / unmute
    - [ ] Mass mute / unmute info logging
    - [ ] Target other channels with mute / unmute
    - [ ] Mute / unmute default to channel / category
  - [ ] Mute dodging by leaving / rejoining guild

mikeshardmind:
Adding to this, we need to ensure all modlog creations are made prior to sending any feedback in context channel, see #2512

PR #2501, opened 2019-03-01
[V3 Mod] Use a composite class for mod

mikeshardmind:
This turns mod.py into a composite class

I've split things in this up based on purpose, including `movetocore`

`movetocore` is a set of things which likely belong in the core bot and not relying on mod being loaded.

This is part of #2500

Per discussion in discord, this should be the first thing in #2500 merged

This also has some # TODO statements added as I noticed issues while splitting this for later investigation

The goal of this is not to handle all issues in mod, just to get the initial compositing done.

This should be a drop-in replacement for mod, but I'd appreciate another set of eyes that I did not miss anything in doing this.


mikeshardmind:
I've had this in use long enough without anyone yelling at me for breaking bot features, so scream test successful.

Wasn't anticipating breakage from this, but still worth testing/looking over.

mikeshardmind:
This includes fixes for #2512 now as well. A backport for this fix will have to happen as a separate PR as the location of the issue will be different prior to 3.1

palmtree5 (approved):
Sorry for the delay. Tested on a test bot and am not seeing anything unusual here so I'm approving

PR #2502, opened 2019-03-02
Logging enhancements and cleanup

Tobotimus:
- Removed debug log messages every time `Config.get_conf` is used or a JSON file is read/saved. The basic configuration is now logged once with DEBUG when the bot starts up instead.
- Changed logging output format to reverse date order, include seconds, and use the logger's name instead of the module, function and line number.
- Log files are now kept in the `DATAPATH/core/logs` directory. ~~Each time Red is restarted, a new log is created, and the old ones renamed in a rotating fashion.~~ *(see discussion below)* There can be a maximum of 9 logs in total.
- Each log file now has a smaller max size of ~~500KB~~ 1MB before it will be split into multiple parts. There are also a maximum of 9 parts of each log.
- Discord.py logger now uses the same output formatter as red's loggers
- Moved logging setup code into `redbot.logging` module.

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -2,29 +2,31 @@
 
 # Discord Version check
 
+import asyncio
+import logging
+import os
 import sys
+
 import discord
+
+import redbot.logging
 from redbot.core.bot import Red, ExitCodes
 from redbot.core.cog_manager import CogManagerUI
-from redbot.core.data_manager import create_temp_config, load_basic_configuration, config_file
 from redbot.core.json_io import JsonIO
 from redbot.core.global_checks import init_global_checks
 from redbot.core.events import init_events
 from redbot.core.cli import interactive_config, confirm, parse_cli_flags
 from redbot.core.core_commands import Core
 from redbot.core.dev_commands import Dev
-from redbot.core import modlog, bank
+from redbot.core import modlog, bank, data_manager
 from signal import SIGTERM
-import asyncio
-import logging.handlers
-import logging
-import os
 
 # Let's not force this dependency, uvloop is much faster on cpython
 if sys.implementation.name == "cpython":
     try:
         import uvloop
     except ImportError:
+        uvloop = None
         pass
     else:
         asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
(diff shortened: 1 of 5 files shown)


mikeshardmind:
> Each time Red is restarted, a new log is created, and the old ones renamed in a rotating fashion. There can be a maximum of 9 logs in total.

I'm not sure this is useful. I'd prefer to just have startup/shutdown events in the logs in a searchable form if the logical seperation is needed and rotate based on file size as this may make it harder than necessary to track down issues which cause bot crashes. (Such as the intermittent heartbeat issue we've had in #support in discord recently)

Tobotimus:
@mikeshardmind How does it make it harder to track down issues? Logical separation based on runtime makes sense because we often want to examine a runtime from start to finish. I don't see how separate files for each runtime is a bad method of doing this logical separation. I would find it much easier to navigate through the runtimes if they were separate files than CTRL+F'ing through a big file.

mikeshardmind:
Because we're destroying logs every 9 starts, combined with recommending an auto restart setup with systemd. Ideally, the issue would present the same each time, but this isn't a guarantee and can mask issues.

I'd be fine with a most recent session or even 2 as separate files in addition to existing logic if you want to be able to examine a recent session from start to end more easily, but I don't think it's wise to discard logs based on restarts.

Tobotimus:
@mikeshardmind Right, I can see your point now.

I suppose it'd be pretty easy to have a `red.log` which exhibits the old behaviour and a `latest.log` and a `previous.log` which are individual runtimes. I think this'd be a good middle ground which will probably maintain lots of runtimes and won't clutter up the logs folder.

One other thing is that I reduced the maximum file size for log files because I think 10MB is too large for a log file. Somewhere around 1MB would be enough to hold about 15,000 lines of log, without busting up some old text editors and reducing time to transfer the files over a network.

Tobotimus:
Made some changes as per the discussion:
- We maintain a red.log over multiple runtimes, alongside a latest.log and previous.log for individual runtimes.
- Naming convention changed a bit. E.g. when latest.log is just one part, it will be named latest.log. When it becomes two parts, latest.log is renamed latest-part1.log and an empty new file is named latest-part2.log.
- Rotation direction is reversed. This means the files end up being named in chronological order.

tekulvw (approved):

PR #2504, opened 2019-03-03
Fix translations of multiline strings

### LINKED ISSUE #2408: Partial translation display
# Other bugs

I translated several files into Russian on crowdin. I downloaded them and replaced the files on my server. I did not make any more changes on the server. Then I changed the language to Russian (!set locale ru-RU).

I expected that all the cogs that I translated will be in Russian. Module files are fully translated, I checked everything myself.

#### What actually happened?

In fact, the translation is only partially displayed. For example, when I call for !help, I see that some of the points are in English, and some in Russian. It is observed in almost all cogs.

[Here you can see clearly.](https://imgur.com/a/OC3f9T7)

1. Install clear redbot on ubuntu.
2. Download, for example, core cog with my 100% translation from crowdin.
3. Replace ru-RU file from /core/locales/ru-RU.po with my translation
4. Change bot language to Russian
5. Type !help, press enter

#### My observations

I think that the problem is multi-line, as I noticed in this pattern. Although I had a small sample.

Tobotimus:
Resolves #2408.

Also did a few little optimisations here and there, we're no longer just using copied code from another project.

### DIFF
--- a/redbot/core/i18n.py
+++ b/redbot/core/i18n.py
@@ -1,11 +1,12 @@
+import contextlib
+import io
 import os
-import re
 from pathlib import Path
-from typing import Callable, Union
+from typing import Callable, Union, Dict
 
 __all__ = ["get_locale", "set_locale", "reload_locales", "cog_i18n", "Translator"]
 
-_current_locale = "en_us"
+_current_locale = "en-US"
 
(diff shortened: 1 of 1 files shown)


Kowlin (approved):
Works like a charm.

PR #2505, opened 2019-03-04
Use menus and add information to [p]servers

lionirdeadman:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Use menus and add information to the menu

### Possible improvements

- Allow blacklisting (as to not allow servers to have the bot again)

image

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -513,12 +513,12 @@ class Core(commands.Cog, CoreLogic):
         embeds = []
         for i, server in enumerate(guilds, 1):
             embed = discord.Embed()
-            embed.title = "Servers list"
+            embed.title = _("Servers list")
             embed.description = server.name
             embed.set_thumbnail(url=server.icon_url)
-            embed.add_field(name="Total users", value=len(server.members))
+            embed.add_field(name=_("Total users"), value=len(server.members))
             embed.add_field(
-                name="Amount of bots", value=len([user for user in server.members if user.bot])
+                name=_("Amount of bots"), value=len([user for user in server.members if user.bot])
             )
             embeds.append(embed)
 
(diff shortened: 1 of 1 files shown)


lionirdeadman:
Good for a review now.

mikeshardmind:
I don't think this should switch to using an embed (at least without respecting the embed setting, this isn't something which embeds make better), and this hides the relevant info across many button presses, making it a worse tool for use IMO.

lionirdeadman:
Ah, yes I could respect the embed setting. But in this case, it does show the server picture which unless you wanna upload tons of server images is gonna help although I guess I'll remove that for non-embeds.

I mostly used an embed because it allows me to show more information in a beautiful manner and I wanted more information because only a name was proving to not be very useful (especially when you consider that there's no differentiation between servers who have the same name)

As for the many button presses, well, I guess that's a fair complaint, do you have suggestions on how I should go about changing this? This just seemed like the most appropriate way to me but I'm open to ideas of course.

mikeshardmind:
I don't really see a way a server-by-server menu can ever be an efficient way to do this.

There are public redbots out there which are in over 2k guilds, this menu could actually hit the recursion limit before visiting all guilds on one of those.

Best I can recommend is adding the ability to view a specific server's info by ID with serverinfo (limited to servers you share with the bot unless you are considered an owner by the bot), and add showing the server IDs in the servers command

I can see an argument for switching the pagification to use a menu at least, but scolling through one at a time is unacceptably slow in some cases.

lionirdeadman:
I'll put 50 servers with their IDs in a menu and add that feature to `serverinfo` in another PR since they're relatively unrelated, does that sound good?

lionirdeadman:
Actually doing that way would need a select for each page and make it overly complicated imo.

I'll close this PR. Maybe I can figure something out at some point but I can't seem to right now.

PR #2506, opened 2019-03-04
`[p]cog install` can now install multiple cogs

NeuroAssassin:
### Type

- [ ] Bugfix
- [X ] Enhancement
- [ ] New feature

Allows you to install multiple cogs from a repo with `[p]cog install <repo> <cogs>`.


mikeshardmind (changes requested):
This should probably move the per-cog install logic to another function here which the command then calls per cog and collects exit status codes for user feedback.

I'm also not 100% certain if this should be merged prior to work being done in #2446

NeuroAssassin:
Closing in favor of #2571

PR #2509, opened 2019-03-05
[Filter] Performance increases.

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

The filter was already using re to split words, this just does the entire search in re instead.

### DIFF
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -346,7 +346,7 @@ class Filter(commands.Cog):
                 word_list |= set(await self.settings.guild(channel).filter())
 
             if not word_list:
-                return
+                return set()
 
             pattern = re.compile("|".join(rf"\b{re.escape(w)}\b" for w in word_list), flags=re.I)
 
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -344,7 +344,7 @@ class Filter(commands.Cog):
             word_list = set(await self.settings.guild(guild).filter())
             if channel:
                 word_list |= set(await self.settings.guild(channel).filter())
-            
+
             if not word_list:
                 return
 
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -285,7 +285,7 @@ class Filter(commands.Cog):
 
     def invalidate_cache(self, guild: discord.Guild, channel: discord.TextChannel = None):
         """ Invalidate a cached pattern"""
-        self.pattern_cache.pop((guild, None), None)
+        self.pattern_cache.pop((guild, channel), None)
 
     async def add_to_filter(
         self, server_or_channel: Union[discord.Guild, discord.TextChannel], words: list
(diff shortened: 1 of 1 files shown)


Tobotimus:
First issue with this is that there's a fatal change of behaviour - with an empty word list, it will return a set containing a single empty string, causing the message to be deleted. Obviously we can just exit early instead.

Secondly, I've timed these changes and they're not necessarily any faster than before, in fact in many cases it's significantly slower. They seem to be showing a performance increase when the word list is small and the message is large. However as the parameters go in the opposite direction (bigger word list / smaller message) it shows a bigger slowdown.

I tried to give it a fair benchmark by taking a random sample of messages and testing them against an auto-generated banned word list of increasing size. I got the random sample by using the last 10,000 messages in #support. Here's a graph of my results:
graph

Tobotimus:
It also seems to be performing worse when the banned words and phrases themselves have a longer average length. In the previous test I used word lists in the form `["word1", "word2", ...]`, but here are my results using word lists in the form `["longerword1", "longerword2", ...]`:
graph

mikeshardmind:
1. The previous version was returning an empty set on no hits as well, this is *not* changing the API.
2. I can't reprofile this right now, but this was running *faster* with the test cases I used, so when I get home, I'll see if I can see what happened with that. 
3. This should show up as even faster on repeated use if the further changes which aren't made here to cache patterns are made, so I'll add those to this PR as well.

mikeshardmind:
The majority of the time difference without the cache was the generator expression which created the pattern.

Adding a pattern cache makes the regex solution multiple orders of magnitude faster on any message other than the first received in a channel with a filter.

(Side by side comparison from discord conversation, running the same above tests on the updated version, but in a profiler instead of just a straight time count.)

image

Context: `old_filter` is the behavior of the existing, the green block with re.compile at the same level is the initial message cost, and `new_filter` (again at the same level) is the amount of time excluding generating the cached pattern (not visible at that scale, extra image below):

image

Tobotimus:
RE: 1, You seem to have fixed the issue now, the issue was not returning an empty set on no hits. I said the issue was it returned a non-empty set on everything when filter was loaded but there were no banned words.

mikeshardmind:
yeah, I realized I misunderstood your meaning on that, and added that, but forgot to comment :sweat:

redbot/cogs/filter/filter.py
```diff
@@ -279,6 +283,10 @@ def __unload(self):
         else:
             await ctx.send(_("Names and nicknames will now be filtered."))
 
+    def invalidate_cache(self, guild: discord.Guild, channel: discord.TextChannel = None):
+        """ Invalidate a cached pattern"""
+        self.pattern_cache.pop((guild, None), None)
```
Tobotimus: Needs support for invalidating the cache in a channel as well

mikeshardmind: Toby, the key is a tuple there. That's intended.

Tobotimus:
Very nice, you managed to take it from linear time to constant time :grin: 
graph

redbot/cogs/filter/filter.py
Tobotimus: Why bother having a channel parameter if it's unused?

mikeshardmind: Fixed....

tekulvw (approved):

PR #2513, opened 2019-03-06
[Audio] Add songs when search-queuing

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Searching for a song and pressing the reaction to queue a song would not add the song to the queue if audioset maxlength was off. This was an omission from #2465.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -2212,7 +2212,8 @@ class Audio(commands.Cog):
                 player.add(ctx.author, search_choice)
             else:
                 return await self._embed_msg(ctx, _("Track exceeds maximum length."))
-        player.add(ctx.author, search_choice)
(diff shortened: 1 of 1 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -2212,6 +2212,7 @@ def _restart_connect(self):
                 player.add(ctx.author, search_choice)
             else:
                 return await self._embed_msg(ctx, _("Track exceeds maximum length."))
+        player.add(ctx.author, search_choice)
```
Tobotimus: Shouldn't this be garded behind an `else:`? Otherwise when maxlength is set and the track is under the max length, you'll be adding the song twice

Tobotimus (approved):

PR #2517, opened 2019-03-12
Cog events orgin update

NeuroAssassin:
### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

Enhancement: Dispatch `red_cog_remove` when unloading a cog, in order for cogs that may depend on this to listen for it and act accordingly.
Bugfix: As said by Danny, it may be better to convert to `red_cog_blah` instead of just `cog_blah` to be sure it does not conflict with discord.py.


mikeshardmind:
This is a breaking change with the event rename and should be marked as such.

While I agree with Danny's sentiment on avoiding a naming conflict, Still need to ensure anyone using the event will have a notice it changes.

mikeshardmind:
I'm not sure renaming this event at the current time is on our best interests. Leaving this open for now.

NeuroAssassin:
I could always remove the extra `red_`, save that for a later PR or something, and change this PR to just be dispatching an event for when a cog is unloaded.

mikeshardmind:
Closing this one as we are unlikely to do anything around this.

PR #2519, opened 2019-03-14
[V3 Audio] Cancel emptydisconnect when no longer alone

Zephyrkul:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Used to be that if you left and re-joined the voice channel, the bot would still disconnect after the empty disconnect timer. Now it will cancel that if someone joins during that time.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -2767,21 +2767,17 @@ class Audio(commands.Cog):
             for p in lavalink.players:
                 server = p.channel.guild
 
-                if server.id not in stop_times:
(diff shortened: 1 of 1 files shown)


aikaterna:
I commented on this in the Discord server, but wanted to put it here so it isn't lost. This doesn't seem to work after it leaves the channel for the first time - if it leaves the channel due to a disconnect and is called back to a channel with a play command or similar the loop doesn't start up again.

aikaterna (approved):
Works as intended now.

tekulvw (approved):

PR #2521, opened 2019-03-16
[Audio] Match v2 behavior for channel change

aikaterna:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

If the bot is not playing and there are no users in the channel with the bot(s), issuing an audio command should make the bot move to the new invoking channel. If the bot is playing something and a command is used where the invoker is in a different channel, the usual audio hierarchy will be used (mods can queue songs from outside of the channel, but the bot will stay put in the channel it's currently playing in, it will tell a pleb that they need to join the current channel, etc). Should match v2 audio behavior for this event.


PredaaA (dismissed):
I totally approve this PR. I have used it a lot (and I saw it has been used at several times on some servers), and for me it's working well as intented. Thanks for that, it's a useful feature!

tekulvw (dismissed):

tekulvw (approved):

PR #2525, opened 2019-03-17
[V3 Audio] If bot has move members perm, it can join to user-limited channels

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Before this, when you wanted to play some music on full voice channel and bot had permission to move members (which allows you to join the user-limited channel) it wouldn't join and reply with "I don't have permission to connect to your channel." even though it could. This PR fixes this behavior.

I'm not sure if it was a right choice to not put this check in `_userlimit` method. I can change it, if you think it will be better this way.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -986,8 +986,10 @@ class Audio(commands.Cog):
                     return await self._embed_msg(ctx, _("That URL is not allowed."))
         if not self._player_check(ctx):
             try:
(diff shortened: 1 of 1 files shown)


aikaterna:
Looks good and works as expected. For people reading this in the future if they are having problems making it work: make sure your bot has Read Text Channels & See Voice Channels checked, or if assigning the perm on the channel itself, View Channel, besides just the Move Members perm.

aikaterna (approved):

tekulvw (approved):

PR #2530, opened 2019-03-24
Increase default leaderboard limit

TimTheToaster:
Seems there's no purpose in having a limit at all, really. 10 names are displayed no matter what. The limit only constrains the functionality of the "<-" and "->" reaction buttons.

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This makes it so that users can hit the `<-` and `->` reactions on the output of `[p]leaderboard` without specifying a parameter. I've had to inform my users to write `[p]leaderboard 1000` if they want to look past the first 10 names.


mikeshardmind (changes requested):
I'm not opposed to increasing the default limit, but making the default limit 10,000 is a bit much. Something like top 100 by default would make a lot more sense here.

TimTheToaster:
So the issue here is that the limit only constrains how far the pagination goes. It doesn't constrain how many names are shown on each page. So if you set it to `100`, then it would still show 10 names per page but you'd only be able to look at the top 100 (that is to say, there'd only be 10 pages).

TimTheToaster:
So the issue here is that the limit only constrains how far the pagination goes. It doesn't constrain how many names are shown on each page. So if you set it to `100`, then it would still show 10 names per page but you'd only be able to look at the top 100 (that is to say, there'd only be 10 pages).

mikeshardmind:
The purpose of a limit here is that when considering a leaderboard, most people don't actually care about the 10,000th user. Most discord servers don't even have that many members. The defaults exist to provide a reasonable starting point (i.e. top 10, top 100 users) you've pointed out a good reason to increase from 10 in that by default, the first page should always contain that anyway. 

That said, this doesn't make it a good reason to arbitrarily increase the default limit or remove it all together.

TimTheToaster:
I agree, it doesn't make much difference when few will care but why restrict it? What harm is there in removing the limit (or making it arbitrarily high)? As a simple example, what if a user wanted to look at those in last place? They'd get to the user in 100th place and then it would just randomly stop paging. The users on my server have complained about this "bug". No one suspects that the command even has a parameter they can use to get around this problem unless they run `.help`. I don't think there needs to be a reason to increase it - the limit probably shouldn't have been there in the first place.

mikeshardmind:
For multiple reasons.

1. There's no reason for a leaderboard to be indefinitely large.
2. Due to the current menu util, anything larger than 100 pages is likely to cause issues if anyone actually pages through all the pages. (There is some work being done to handle this, but right now, nothing for it)
3. Even when 2 is actually handled, this is API spam to actually go through all of this. It would be better to accomplish being able to see who is at the bottom by adding new commands to query the total order by other things.

If you'd like some alternative ways to query the economy, that could be discussed, but this command should not have a default that large.

TimTheToaster:
I modified the code to make the parameter represent `the target rank to view`. So if you want to see who is in 200th place, you write `[p]leaderboard 200`. I understand if this can't be merged in as a result of the 2nd item on your list (which I don't understand but I'm sure you do) but perhaps this could be merged when that's resolved. The ability to jump to the area of interest in the list prevents the third item on your list, the API spam, from being an issue.

mikeshardmind:
This is a significant change in behavior which still does not address the issues with the prior iteration of your changes.

If you'd like a command to view a specific range from the leaderboard, rather than the top x, the answer is to add an additional command for other ways of viewing it, not changing the expected behavior of the command.

TimTheToaster:
Alright. Let's just call it a day on this. I don't think this functionality warrants its own command - it'd be bloat. I figured my modifications would be an improvement over an oversight but it appears the functionality is deliberate. I'm going to go ahead and close this. Thanks for taking the time to work with me on this potential improvement.

PR #2535, opened 2019-03-28
[V3 Downloader] Prevent installing already installed backages

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

The downloader currently installs all required pip packages by a cog, even if they're already installed with the base pip packages (the feature was already planned, as you can see there's a TODO about it currently in the code).

When installing requirements, the bot first grabs the pip packages, and split them into two lists : 
- `installed_packages` which contains the packages with the version number, like `pillow==5.3.0`
- `installed_packages_without_version` which contains the package names, without the version
If the cog requires a specific version of the packages, the bot checks if it's in `installed_packages`, else it checks in `installed_packages_without_version`.

I also added more logging, here's an example when installing Trusty's `imagemaker` cog with Pillow 5.3.0 already installed:
```
[28/03/2019 17:57] INFO repo_manager install_raw_requirements 507: Skipping installation of pillow because it is already installed.
[28/03/2019 17:57] INFO repo_manager install_raw_requirements 516: Installing the following pip packages at /Users/laggron/Documents/RedDB-Blue/cogs/Downloader/lib
opencv-python, numpy
```

Last thing I added is `ctx.typing()` to `[p]repo add` and `[p]cog install`, so the user can know the bot is not dead and still processing the request, which can get very long for some repos/cogs.


mikeshardmind:
The underlying command in use here is (currently)
```
PIP_INSTALL = "{python} -m pip install -U -t {target_dir} {reqs}"
```
This should already have an awareness of same version, and will use the already available version from the pip cache.

I'm going to need a concrete example of how using the already cached version is in any way problematic on this.

PR #2536, opened 2019-03-31
[V3 Config] Update Mongo document organization to bypass doc size restriction

tekulvw:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This PR reorganizes how the Mongo driver stores data on the database. It will create many more documents than the current implementation but they will each be significantly smaller than the current documents. Because of this reorganization, any existing Mongo data will not be readable by this PR and will need to be migrated using a future tool.

### Warning: Breaking Changes
* The `__init__` signature for Group and Value has changed in this PR.
* The signature for `driver.get()`, `driver.set()`, and `driver.clear()` has also changed.

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -269,12 +269,7 @@ class Group(Value):
             dict access. These are casted to `str` for you.
         """
         path = tuple(str(p) for p in nested_path)
-        identifier_data = IdentifierData(
-            self.identifier_data.uuid,
-            self.identifier_data.category,
-            self.identifier_data.primary_key,
-            path,
-        )
+        identifier_data = self.identifier_data.add_identifier(*path)
         await self.driver.clear(identifier_data)
 
     def is_group(self, item: Any) -> bool:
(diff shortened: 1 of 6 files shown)


mikeshardmind:
This breaks on *at least* the all_x methods, as well as on group all methods

From a quick test with a new bot on an isolated server:
image

The only correct result here is the one for past_nicks.

image

I'll look to see what else isn't working as intended with this later.

Stonedestroyer:
As discussed on red discord server apparently we can't use _id due to the fact that it can't be an array. This is an alternative. https://docs.mongodb.com/master/core/index-unique/

It would help making mongo faster than json, i think next step would allowing cog developers to define unique indexes for their own cogs.

Tobotimus:
@Stonedestroyer You can also simply make `_id` an object containing a single item with an array as a value, like so:
image

tekulvw:
The latest commit changes the layout of the collection/document structure to follow the format in the image below. There will be a single database for the bot, a collection for each combination of "Cog.CATEGORY" and a document for each unique dataset in "Cog.CATEGORY".
image

mikeshardmind (approved):
This seems ready to go now.

This is a breaking change. This does remove the ability to convert backend right now (which needs to go onto the 3.1 todo list (updating the conversion for the new organization).

Kowlin (approved):

PR #2545, opened 2019-04-03
[V3 Config] Require custom group initialization before usage

tekulvw:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This is a breaking change when using Config's custom groups feature. In order to safely and effectively convert data between data backends we must know how to interpret custom groups. Prior to this PR that was impossible because we wouldn't know the length of the custom group's primary key. The primary key is the list of identifiers passed into `Config.custom()`.

This PR adds the `Config.init_custom()` method which allows you to register the length of each custom group's primary key. This method must be called before `register_custom()` or `custom()`. ~~The information stored in config is written to disk the first time a get/set/clear happens.~~

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -100,7 +100,9 @@ class Permissions(commands.Cog):
         # Note that GLOBAL rules are denoted by an ID of 0.
         self.config = config.Config.get_conf(self, identifier=78631113035100160)
         self.config.register_global(version="")
+        self.config.init_custom(COG, 1)
         self.config.register_custom(COG)
+        self.config.init_custom(COMMAND, 1)
         self.config.register_custom(COMMAND)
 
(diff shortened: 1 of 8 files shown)


tekulvw:
Information will not be stored to disk in this PR. I need to make another PR to change `bot.add_cog` into a coroutine first.

mikeshardmind (changes requested):
Config objects need to be singletons with this change. The design direction this is going will require it here for some sanity instead of just at the driver.

Modification to object creation preferred over just the factory methods from me on this.

This also breaks the entire dataconverter cog. I'm fine with the removal of that cog until it can be rewritten in a more appropriate manner for future use.

mikeshardmind (approved):
Works as intended. Is breaking. Needs an announcement in #v3-burndown .

Kowlin (approved):

PR #2548, opened 2019-04-04
[V3 Config] Make bot.add_cog() a coroutine

tekulvw:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

### Description of the changes
This PR makes `bot.add_cog()` a coroutine and requires all cog `setup` functions to be coroutines as well. `bot.add_cog()` is overridden and used to save custom group information initialized during the cog's `__init__`.

This PR depends on #2545.

PR #2550, opened 2019-04-04
[V3 Config] Record custom group information using cog_add event

tekulvw:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

No new breaking changes! This is an alternative to #2548. Blocked by #2545.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -76,7 +76,7 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):
 
         self.db.register_user(embeds=None)
 
-        # self.db.init_custom(CUSTOM_GROUPS, 2)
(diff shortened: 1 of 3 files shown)


mikeshardmind (changes requested):
This would be breaking a few existing approved cogs which make use of config objects with a different cog name for config than the actual cog name used for the help menu

If we had a secondary weakvalue dictionary for config objects which had already been accounted for, we can avoid this issue and just update any which haven't been already, adding entries to the dictionary storing the "last updated state"

mikeshardmind (approved):
:+1: thanks for the changes, glad this one could avoid a breakage considering the required change that could end up being for 3rd party cog devs. Ready to go.

Kowlin (approved):

PR #2551, opened 2019-04-04
[V3 ModLog] Change register_casetypes behavior

### LINKED ISSUE #2543: [ModLog] modlog.register_casetypes() fails silently
This one is a bit obscure, but I promise there is a good bug here. Lets say you have a cog which registers Case A, install that cog and have it running for an day. And you update the cog with a new case type, (Case B). The ``modlog.register_casetypes()`` will fail silently and refuse to add the new case type.

Here's some quick code to verify your self. (Make your own init), run this cog once, unload it, and add another case below the first case.

```py
import discord
from redbot.core import modlog, commands
import asyncio

BaseCog = getattr(commands, "Cog", object)

class TestModlog(BaseCog):

    def __init__(self, bot):
        self.bot = bot
        asyncio,ensure_future(self.register_cases())

    @staticmethod
    async def register_cases():
        cases = [
            {
                "name": "testcase1",
                "default_setting": True,
                "image": "\N{HAMMER}",
                "case_str": "Test case one.",
            }
        ]
```

tekulvw:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Since `modlog.register_casetypes()` is often used when setting up a cog, this PR ignores all `RuntimeError`s as they just signify that a particular case type has already been registered.

resolves #2543

### DIFF
--- a/redbot/core/modlog.py
+++ b/redbot/core/modlog.py
@@ -746,7 +746,6 @@ async def register_casetypes(new_types: List[dict]) -> List[CaseType]:
 
     Raises
     ------
-    RuntimeError
     KeyError
     ValueError
     AttributeError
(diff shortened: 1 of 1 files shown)


Kowlin (approved):
Working as intended.

PR #2563, opened 2019-04-08
Fix message when user hits max credits

kennnyshiwa:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes the message a user should get when they have max credits and attempt to use the payday command


mikeshardmind (dismissed):
Side note: We have some inconsistent breakage in #testing with this.

tekulvw (approved):

PR #2566, opened 2019-04-09
Change command name and add aliases to show for its for resume and pause

NeuroAssassin:
### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

Chance command name to `rp` for `resume/pause`, and add aliases for the command to let users know that it is used for resuming also, as when I was looking, I thought it was only for pausing and could not find the resume command.


Jackenmen:
I would leave the name as it was and just add "resume" alias, because I think that short names should be available for custom aliases, if possible.

PR #2568, opened 2019-04-10
Change send_help for group of group to show correct help menu

NeuroAssassin:
Before, if I ran `[p]groupone grouptwo unknown_command`, it would show help as if I'd done `[p]help groupone`.  Now, it shows help as if I'd done `[p]help groupone grouptwo`.

### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Before, if someone ran `[p]groupone grouptwo unknown_command`, help would be displayed as if they had run `[p]help groupone`:
image
After this PR, if someone ran `[p]groupone grouptwo unknown_command`, help would be displayed as if they had run `[p]help groupone grouptwo`, since it would seem like the person would not know that is not a subcommand of `grouptwo`, and so displaying help for `groupone` would mean nothing, but showing help for `grouptwo` shows that that subcommand does not exist, and gives a list of the correct ones.
image

### DIFF
--- a/redbot/core/commands/context.py
+++ b/redbot/core/commands/context.py
@@ -26,6 +26,34 @@ class Context(commands.Context):
         super().__init__(**attrs)
         self.permission_state: PermState = PermState.NORMAL
 
+    async def command_from_message(self, message: discord.Message):
+        """Get the command from a message without arguments.
+        
+        Parameters
+        ----------
+        message: discord.Message
+            The message to get the command from.
+        
+        Returns
+        ----------
+        Optional[commands.Command]
+            The command invoked or None if no command was invoked.
+        """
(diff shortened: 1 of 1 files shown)


NeuroAssassin:
After reviewing #2569, and noticing that it's best to not create new views and new contexts, I changed it over the a per-word logic, that worked for me.  I might need to do more testing first, but it worked for me and provided the correct nested group helps.

mikeshardmind:
I'm sorry that I did not have time when I made the alternate to explain what was going on there. I am closing this in favor of said alternate now, but I do appreciate that you pointed out the issue.

For the record, It's not an issue to use a stringview, but we generally prefer not to modify the state of context's view to avoid potential breakage with discord.py

PR #2569, opened 2019-04-10
Fix ctx.send_help for nested groups

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This is an alternative to #2568 

It fixes the same problems without modifying the context's view, and without generating an additional context.

This may be made obsolete by planned changes to update the discord.py version in use later on, but can be safely backported as well if we do another 3.0.x release.


mikeshardmind:
Closing, root cause will be fixed in #2628 .

PR #2570, opened 2019-04-10
Add some errors for conversions and only allow MongoV2 creation

tekulvw:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This allows us to differentiate between mongo v1 and v2 for the conversion process. It also adds not implemented errors for conversions between MongoV2 and anything else for the time being.

### DIFF
--- a/redbot/setup.py
+++ b/redbot/setup.py
@@ -153,10 +153,10 @@ def basic_setup():
 
     storage = get_storage_type()
 
-    storage_dict = {1: "JSON", 2: "MongoDBV2"}
+    storage_dict = {1: "JSON", 2: "MongoDB"}
     default_dirs["STORAGE_TYPE"] = storage_dict.get(storage, 1)
 
-    if storage_dict.get(storage, 1) == "MongoDBV2":
+    if storage_dict.get(storage, 1) == "MongoDB":
         from redbot.core.drivers.red_mongo import get_config_details
 
         default_dirs["STORAGE_DETAILS"] = get_config_details()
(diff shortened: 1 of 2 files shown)


redbot/setup.py
```diff
@@ -272,12 +272,14 @@ def basic_setup():
...
                 #    await json_to_mongo(current_data_dir, storage_details)
-        else:
+        elif storage_dict.get(storage, 1) == "JSON":
             storage_details = instance_data["STORAGE_DETAILS"]
             default_dirs["STORAGE_DETAILS"] = {}
             if instance_data["STORAGE_TYPE"] == "MongoDB":
                 if confirm("Would you like to import your data? (y/n) "):
                     await mongo_to_json(current_data_dir, storage_details)
+            elif instance_data["STORAGE_TYPE"] == "MongoDBV2":
+                raise NotImplementedError("We cannot convert from MongoDB2 to JSON at this time.")
```
mikeshardmind:
```suggestion
                raise NotImplementedError("We cannot convert from MongoDB to JSON at this time.")
```
I don't think putting a 2 next to this is clear to the user. It has meaning internally to us, but shouldn't be user-facing.

Beyond that, I think this is ready to go, so since this is just cosmetic in nature, take or leave the change, it's fine.

mikeshardmind (dismissed):
Approving, but see comment about a suggested UX change this prior to deciding on merging.

redbot/setup.py
tekulvw: The only reason I kept that in there is because Mongo and MongoV2 are separate options in the conversion menu and I think they need to be. I'll take another look at it and see if I can condense them.

Kowlin (approved):

PR #2573, opened 2019-04-11
Add `[p]cc search` command

NeuroAssassin:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

Allows for a user to search through custom commands if fuzzy string matching is enabled in the server.  I thought this may be helpful for cases like in servers like Red, where they have a bot that has a lot of custom commands (Saltbot) and it can be difficult to remember them, and to search through the paragraph long of custom commands.


redbot/cogs/customcom/customcom.py
```diff
@@ -202,6 +203,30 @@ def __init__(self, bot):
         """Custom commands management."""
         pass
 
+    @customcom.command(name="search")
+    @commands.guild_only()
+    async def cc_serach(self, ctx: commands.Context, *, search):
```
Jackenmen: spelling issue :D - `cc_serach`

NeuroAssassin: Yeah, thanks for pointing it out.  I'll be pushing a commit in a second to address that.

mikeshardmind:
This can be included in 3.2 if the changelog + conflicting file are handled. Beyond needing that, this is ready for use.

Kowlin (dismissed):
This is directly touching config which Red doesn't allow anymore. Postponing this until the issue is fixed.

Jackenmen (approved):
I did a few changes to this PR and tested it and now it should be good to go.
I would appreciate a second pair of eyes though since after all, I did some edits to this PR.

Jackenmen:
Kowlin looked at this and gave it okay, merging

PR #2574, opened 2019-04-11
[V3 Docs] Add Central API Key Documentation

TrustyJAID:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Add documentation for central API key storage. Feel free to comment and let me know how you would like this to be changed for better explanations, etc.

### DIFF
--- a/docs/framework_apikeys.rst
+++ b/docs/framework_apikeys.rst
@@ -1,0 +1,48 @@
+.. V3 Shared API Key Reference
+
+===============
+Shared API Keys
+===============
+
+Red has a central API key storage utilising the core bots config. This allows cog creators to add a single location to store API keys for their cogs which may be shared between other cogs.
+
+There needs to be some consistency between cog creators when using shared API keys between cogs. To help make this easier service should be all **lowercase** and the key names should match the naming convetion of the API being accessed. 
+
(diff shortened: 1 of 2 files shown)


Jackenmen:
I think there should be some mention of naming services which names don't have only one word, for example Rocket League API could be written as `rocketleague` (IMO, looks better in those 2 examples, but it needs to be decided if that should be our convention) or `rocket_league`. Another example would be: `googlemaps` and `google_maps`.

On the other hand, sometimes you can find services, which are written both as separate words and as one long word, I don't have any particular example, but let's say that OpenWeatherMap can be written as one long word - "OpenWeatherMap" or as three words - "Open Weather Map", which may cause confusion, if we would actually use underscores in names as for first one you would write - `openweathermap` and for second `open_weather_map`, even though someone would follow convention with underscores.

The second thing is probably very rare, I'm sure I saw such services, which have their name both with and  without spaces, but I can't come up with any example, so this might not be important.

Edit: so I haven't said what I think would work better, so if it's not that clear from my comment, I'll tell it now: I prefer option without any underscores for services names with spaces.

Kowlin (approved):

PR #2576, opened 2019-04-12
Pretties up the output when libraries fail to install

### LINKED ISSUE #2575: "Failed to install libraries" message outputs a raw list
# Command bugs

#### Command name

`[p]cog install`

Downloader

A nice pretty output from the "Failed to install libraries" message.

#### What actually happened?

A horrible no good very bad list was outputted. 

Install a cog that has a library that cannot be installed, look at the resulting message.

#### Additional Information

I am making a PR for this issue in a few seconds, this issue is just for logging this problem.

Flame442:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

Fixes #2575

Previously, when a cog failed to install libraries, the resulting message would print the raw list of requirements. This PR changes that message to comma separate those libraries in order to look nicer.

Example before:

Before

Example after:

After

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -305,10 +305,11 @@ class Downloader(commands.Cog):
             return
 
         if not await repo.install_requirements(cog, self.LIB_PATH):
+            libraries = '`, `'.join(cog.requirements)
             await ctx.send(
                 _(
                     "Failed to install the required libraries for `{cog_name}`: `{libraries}`"
-                ).format(cog_name=cog.name, libraries=cog.requirements)
+                ).format(cog_name=cog.name, libraries=libraries)
             )
             return
 
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -305,7 +305,7 @@ class Downloader(commands.Cog):
             return
 
         if not await repo.install_requirements(cog, self.LIB_PATH):
-            libraries = '`, `'.join(cog.requirements)
+            libraries = "`, `".join(cog.requirements)
             await ctx.send(
                 _(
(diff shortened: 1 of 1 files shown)


Jackenmen:
Your code's completely fine, but I just wanted to let you know, that we have [inline](https://red-discordbot.readthedocs.io/en/v3-develop/framework_utils.html#redbot.core.utils.chat_formatting.inline) and [humanize_list](https://red-discordbot.readthedocs.io/en/v3-develop/framework_utils.html#redbot.core.utils.chat_formatting.humanize_list) functions (you can see, that we use those in other places in downloader) in [redbot.core.utils.chat_formatting](https://red-discordbot.readthedocs.io/en/v3-develop/framework_utils.html#module-redbot.core.utils.chat_formatting), where [inline](https://red-discordbot.readthedocs.io/en/v3-develop/framework_utils.html#redbot.core.utils.chat_formatting.inline) adds inline formatting to string and also makes sure to use double ticks, when tick is already in the text (that doesn't matter here, but just wanted to point that out) and [humanize_list](https://red-discordbot.readthedocs.io/en/v3-develop/framework_utils.html#redbot.core.utils.chat_formatting.humanize_list) functions (you can see, that we use those in other places in downloader) gives you comma-separated list, with the last element joined with *and*.
Link to docs:

We use those in other places where we send cogs or libraries names, so it might be better to use it.

You can use them together like that (it's used like this in all other places in downloader):
```py
humanize_list(tuple(map(inline, cog.requirements)))
```
or like that (that doesn't require converting map to tuple, so might be better, but since map is used everywhere, I'm not so sure, if it's better):
```py
humanize_list([inline(req) for req in cog.requirements])
```

tekulvw (approved):

PR #2577, opened 2019-04-12
Add fix for help docstring markdown

NeuroAssassin:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Before this change, it would add the ending underscores to the original string, and then grab the first 252 characters.  If it is a long docstring, it will end up in malformed markdown, as show below:
image
After this, it makes a string with the first two beginning underscores plus the string, then later when adding the field, it gets the first 250 characters and *then* adds the underscores to make sure they get added for proper markdown:
image

### DIFF
--- a/redbot/core/help_formatter.py
+++ b/redbot/core/help_formatter.py
@@ -159,11 +159,11 @@ class Help(dpy_formatter.HelpFormatter):
             # <long doc> section
             if self.command.help:
                 splitted = self.command.help.split("\n\n")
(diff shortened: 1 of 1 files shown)


mikeshardmind:
The help formatter needs to be handled entirely differently due to changes with discord.py coming with #2587 , closing this, but if you'd like to help with ensuring the new version handles this properly, that would be appreciated.

PR #2583, opened 2019-04-13
Help respects permission

Stonedestroyer:
# Enhancement request

#### Describe the enhancement

This PR makes sure that help respects commands that the user can't run and doesn't show them.

If users do [p]help shutdown as an example they see the command but this will make help silently fail instead similar to how [p]help Cogcategory works now. 

- [ ] Yes
- [x] No


mikeshardmind:
This is a change in functionality which I can see the appeal for, but should not be handled quite in this way.

Specifically, the method in which you check if a command can be run from outside the command's context is not safe here. (`change_permission_state=False` is a must here)

Additionally, the current iteration of the help formatter is being removed outright in #2587 . I'm leaving this open as the only current record of this as a feature request currently, but it should be included in the rewritten help formatter currently planned for inclusion after this.

Stonedestroyer:
This patch can be backported to 3.0.x if needed. I will put a link to this PR on the issue for help-formatter to track it better there. https://github.com/Cog-Creators/Red-DiscordBot/pull/2628

Stonedestroyer:
This is added to helpformatter now so closing.

PR #2328, opened 2018-12-12
[V3 Audio] Add Spotify support

aikaterna:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

### Description of the changes
This adds the ability to use open.spotify.com links or spotify:track:code type of Spotify tracks with the play command. Tracks, albums, playlists, and user playlists can be added using the url or code, but all within the play command. This requires that the bot owner sign up for a Spotify API key along with a YouTube Data API v3 key, and input the Spotify ID, Spotify client secret, and the YouTube API key in [p]audioset commands before those URLs/codes can be used. Thanks to TrustyJAID for his help in moving the YouTube API access to something quicker and more efficient than what I originally had in mind.

PR #2591, opened 2019-04-19
[Audio] Track search and append fixes

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

* Tracks had ytsearch: appended to it twice in some searching situations, accidentally introduced in the Spotify PR (#2328)
* Playlist append was joining strings with spaces in some situations before searching

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -1183,7 +1183,7 @@ class Audio(commands.Cog):
         player = lavalink.get_player(ctx.guild.id)
         guild_data = await self.config.guild(ctx.guild).all()
         if type(query) is not list:
-            if not (query.startswith("http") or query.startswith("localtracks")):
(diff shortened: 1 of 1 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -1183,7 +1183,11 @@ def _restart_connect(self):
         player = lavalink.get_player(ctx.guild.id)
         guild_data = await self.config.guild(ctx.guild).all()
         if type(query) is not list:
-            if not (query.startswith("http") or query.startswith("localtracks")):
+            if not (
```
Nodja, line 1186: You can pass a tuple to startswith:
```python
if not query.startswith(("http", "localtracks", "ytsearch"))
```

tekulvw (approved):

PR #2179, opened 2018-10-04
[V3 Mod] Mute / Unmute enhancements

Zephyrkul:
### Type

- [ ] Bugfix
- [X] Enhancement
- [x] New feature

### Description of the changes

- Resolves #2076 
- Remove the voice mute / unmute commands - due to a Discord limitation updating overwrites does not modify the member's permissions if the member is already in a voice channel. Mass mutes / unmutes still affect voice channels.
- Add category mute / unmute to mass mute / unmute all channels in a category. If used in a channel with no category, it will modify all channels without a category (excluding categories themselves).
- Helper methods mute_user and unmute_user now take GuildChannel instead of solely TextChannel.
- The bot will not attempt to mute a member with the Administrator permission, as that permission bypasses channel overwrites anyway. The bot will still unmute members with the Administrator permission (see #2076).
- Mass muting / unmuting now reports how many actions were successful, and how many actions were attempted.
- Issues when muting / unmuting are now translated. These issues are also info-logged when mass muting / unmuting.
- Channels and categories other than the current one can now be specified - this takes advantage of the new typing.Optional converter.
- Audit reasons are now specified for mass mutes / unmutes.
- mute and unmute now default to mute channel and unmute channel, and match the checks of their subcommands. If a category channel is passed to either, they will instead default to mute category and unmute category.

Issue #1852, opened 2018-06-11
[V3 Mod] Make mute server use a role


laggron42:
# Feature request

#### Select the type of feature you are requesting:

- [ ] Cog
- [x] Command
- [ ] API functionality

#### Describe your requested feature

The `[p]mute server` edit every channel permissions. This requires a lot of API calls and is really slow to operate (it takes about 6 seconds for 10 channels, so imagine on huge servers) and we could reduce that time by muting using a role instead of editing the channel permissions. 

This should be optional so users can decide what they want between a fast server mute with a role or a slow mute with channel perms.

Tobotimus:
This is a well presented suggestion, it has a few issues though. I agree that the current `[p]mute server` command is too API spammy. However applying a role instead is a change in behaviour, so this could be a separate command (*cough punish cough*). One use of `[p]mute server` is to quickly restrict a user to a small number of channels by doing `[p]mute server` followed by `[p]unmute channel`, because doing `[p]mute channel` in every other channel would take a long time. Applying a muted role to a person changes the behaviour in that `[p]unmute channel` wouldn't do anything to someone who is server-muted.

An alternate approach might be to allow roles to be muted the same way users can, then mods can just apply a muted role. This avoids changing the behaviour of `[p]mute` but doesn't solve the API spammy nature of the command itself. I'm open to more discussion :)

laggron42:
Maybe we could still use a role for guild mute and, when using channel unmute, force the `send_messages` permissions in the channel (ignore role perms)? But another issue would be to remove that forced permission, but when... 

I already started working on that feature, you can check the `mute` branch of my fork (the work currently done ignores what you suggested, but I will look after it). 

Another addition not specified in the issue is that, when using `mute` without subcommands, it will be default to a guild mute (my moderators were lost because they were just trying to `[p]mute @someone`)

calebj:
I had an idea for solving this problem in v2 punish that I'll try to implement and test today. Will comment with how it goes.

mikeshardmind:
This *cannot* be reliably done with a role without a server's permissions being designed for this usage. 

If this makes it into core, it needs to error in cases it would be ineffective with a clear reason why presented to the user.

Discord's developer docs on permissions explain this, but not really in good user-facing terms. If a user has any role allowing something (other than the default role), any role-based denies won't matter since non-default role allows are batch applied *after* batch applying non-default role denies.

[...]

PR #2599, opened 2019-04-23
[V3 Mod] Mute/Unmute QOL

### LINKED ISSUE #1607: [V3 Mod] Mute for channel categories
# Feature request

- [ ] Cog
- [x] Command

Instead of muting per channel, have an additional permission weather or not to mute catergorically, means less work for the bot and doesn't fill logs with per channel mutes/unmutes. All of this would be optional of course.

Edit; And then any channels that aren't muted via category it mutes per any channels left.

Zephyrkul:
### Type

- [X] Enhancement

Rewritten version of #2179 after class compositing. Part of #2500. There's a bit of weirdness with mute v. unmute in terms of permissions, strings, and attributes which I'd like to take another pass at tomorrow, before this is merged.

* ~~Guild mute / unmute will act on categories first, then on channels, to attempt to save on API calls. Resolves #1607.~~ lol nah fam
* Remove the voice mute / unmute commands - due to a Discord limitation updating overwrites does not modify the member's permissions if the member is already in a voice channel. Mass mutes / unmutes still affect voice channels.
* Add category mute / unmute to mass mute / unmute all channels in a category. If used in a channel with no category, it will modify all channels without a category (excluding categories themselves).
* Mass muting / unmuting now reports how many actions were successful, and how many actions were attempted.
* Channels and categories other than the current one can now be specified - this takes advantage of d.py's typing.Optional converter.
* mute and unmute now default to mute channel and unmute channel, and match the checks of their subcommands. If a category channel is passed to either, they will instead default to mute category and unmute category.
* Cleaned up various user-facing strings.

### DIFF
--- a/redbot/cogs/mod/casetypes.py
+++ b/redbot/cogs/mod/casetypes.py
@@ -56,17 +56,17 @@ CASETYPES = [
         "audit_type": "member_update",
     },
     {
-        "name": "vmute",
-        "default_setting": False,
-        "image": "\N{SPEAKER WITH CANCELLATION STROKE}",
-        "case_str": "Voice Mute",
-        "audit_type": "overwrite_update",
-    },
-    {
         "name": "cmute",
         "default_setting": False,
         "image": "\N{SPEAKER WITH CANCELLATION STROKE}",
         "case_str": "Channel Mute",
+        "audit_type": "overwrite_update",
+    },
+    {
+        "name": "catmute",
+        "default_setting": False,
+        "image": "\N{SPEAKER WITH CANCELLATION STROKE}",
+        "case_str": "Category Mute",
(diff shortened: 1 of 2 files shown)


mikeshardmind:
I have not looked into the implementation details yet, but I don't think this sufficiently closes the related issue of a category mute, and would prefer to add categories as a valid target for mutes to close that.

This also seems a bit fragile for server mutes given that an unmute can occur accidentally with channels being changed to no longer sync with parent channel, a better reduction of API calls would be for a configurable role to use for guild mutes (falling back on per channel behavior) see #1852

Ed: this adds category mutes, not just uses them for guild mutes. using them for guild mutes is the problematic portion, but this can close the prior mentioned related issue.

PR #2605, opened 2019-04-23
[V3 Downloader] Allow to specify minimum and maximum bot version in info.json

Jackenmen:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

Adds actually working way of specifying minimum and maximum bot version and removes not working `bot_version` key in info.json file
Downloader won't allow to install a cog if it doesn't support current version of Red. 

However it's not checked at all during cog update (and neither is minimum python version in that matter) I would like to do this in separate PR, preferably after #2571 as it has whole cog update process improved a lot and it might be easier to write with use of those changes. 

Also, shared libraries aren't checked for both python and bot version anywhere either, but again, it will be easier to do this with use of that rewritten code from #2571. 

BREAKING CHANGE:
- removal of `bot_version` attribute in `Installable`
- cog developers have to use `min_bot_version` instead of `bot_version` (`bot_version` didn't actually work, but still they need to know that)


Jackenmen:
Added changelog entries to this PR.

mikeshardmind (approved):

PR #2608, opened 2019-04-23
Add mods-only cc allowed to ping here and everyone

### LINKED ISSUE #2518: [V3 CC] CCs not pinging @here and @everyone
# Command bugs / Feature requests

Not sure if this is working as intended or a bug.

CC

If I type in `@everyone` or `@here` to a CC, I expect the respective annoyances to be pinged when executing the CC.

#### What actually happened?

CC inputs them as plaintext.

------

A user was asking in #support why his CCs don't ping @here and @everyone. After some snooping around people found that `Context.send` escapes mass mentions by default, and is the reason for them not working on CCs.

In my humble opinion, there should at least be an option to use them in CCs and actually make them ping. I was thinking if you don't want them to be enabled by default, maybe an argument like {everyone} would be possible? I can see the need for admins to create CCs that use these mentions, however annoying people might feel them to be.

Malarne:

Addition of the `[p]cc create mod`  command
it allows people to create a command, that is locked to mods, admins, and bot owner, that has the ability to ping everyone and here.
</img> is a good example of this command
This will just solve the problem of issue #2518 without having everyone in a server able to use those pings

### DIFF
--- a/redbot/cogs/bank/bank.py
+++ b/redbot/cogs/bank/bank.py
@@ -124,13 +124,4 @@ class Bank(commands.Cog):
         await bank.set_currency_name(name, ctx.guild)
         await ctx.send(_("Currency name has been set to: {name}").format(name=name))
 
-    async def on_command_error(self, ctx, exception):
-        author = ctx.author
-        command = ctx.command.qualified_name
-        data = await bank._conf.member(author).last_command_cost()
-        if data["command"] != command:
-            return
-        else:
-            await bank.deposit_credits(author, data["cost"])
-
     # ENDSECTION
--- a/redbot/cogs/bank/bank.py
+++ b/redbot/cogs/bank/bank.py
@@ -124,4 +124,13 @@ class Bank(commands.Cog):
         await bank.set_currency_name(name, ctx.guild)
         await ctx.send(_("Currency name has been set to: {name}").format(name=name))
 
+    async def on_command_error(self, ctx, exception):
+        author = ctx.author
+        command = ctx.command.qualified_name
+        data = await bank._conf.member(author).last_command_cost()
+        if data["command"] != command:
+            return
+        else:
+            await bank.deposit_credits(author, data["cost"])
+
     # ENDSECTION
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -88,7 +88,7 @@ class CommandObj:
         if not ccinfo:
             raise NotFound()
         else:
-            return ccinfo["response"], ccinfo.get("cooldowns", {})
+            return ccinfo["response"], ccinfo.get("cooldowns", {}), ccinfo["mod"]
 
     async def get_full(self, message: discord.Message, command: str) -> Dict:
         ccinfo = await self.db(message.guild).commands.get_raw(command, default=None)
@@ -97,7 +97,7 @@ class CommandObj:
         else:
             raise NotFound()
 
-    async def create(self, ctx: commands.Context, command: str, *, response):
+    async def create(self, ctx: commands.Context, command: str, *, response, mod : bool = False):
         """Create a custom command"""
         # Check if this command is already registered as a customcommand
         if await self.db(ctx.guild).commands.get_raw(command, default=None):
@@ -112,6 +112,7 @@ class CommandObj:
             "created_at": self.get_now(),
             "editors": [],
             "response": response,
+            "mod" : mod
         }
         await self.db(ctx.guild).commands.set_raw(command, value=ccinfo)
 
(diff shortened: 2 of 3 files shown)


Jackenmen:
How about creating [context parameters](https://red-discordbot.readthedocs.io/en/v3-develop/cog_customcom.html#context-parameters) (working only for mod custom commands of course) for this instead of requiring the user to ping everyone or here to create a command in the first place?
So like {everyone} would be replaced with `@everyone` and {here} with `@here`

mikeshardmind:
I don't see this as worth adding to the bot. Red is intended to support every general case, while allowing cogs to support the extended ones. Many servers do not want additional ways in which everyone can be mentioned, and I'd personally argue if you have a cc to mention everyone/here, you have larger issues.

Edit: And even if this were to be added, I find the defaults incredibly lax. The default perms on this should be at a serverowner level. everyone pings ruin large servers, and mods are frequently added by servers which would never give that permission out. While permissions can relax this, the default on this needs to be strict.

Mystard:
> How about creating [context parameters](https://red-discordbot.readthedocs.io/en/v3-develop/cog_customcom.html#context-parameters) (working only for mod custom commands of course) for this instead of requiring the user to ping everyone or here to create a command in the first place?
> So like {everyone} would be replaced with `@everyone` and {here} with `@here`

Think I mentioned something similar in #2518, but I like this new command approach as well.


> I don't see this as worth adding to the bot. Red is intended to support every general case, while allowing cogs to support the extended ones. Many servers do not want additional ways in which everyone can be mentioned, and I'd personally argue if you have a cc to mention everyone/here, you have larger issues.

Well even if this command was admin only, I expect those admins can use `@everyone` anyways? Nobody is forcing servers to add commands that mention here or everyone. People can edit their notification settings and suppress here and everyone if they really need to. I will argue that this should be a feature, since admins can probably use those mentions anyways. Why not give them the convenience of doing it with a CC?

If servers really have such bad admins that they can't trust them with the ability to create and use CCs that can ping `@here` and `@everyone`, then imo they can use permissions to filter this out.

**Edit to respond to Sinbad's edit:** I agree, this should default to admin, and then servers can add permission to whatever role they see fit.

>mods are frequently added by servers which would never give that permission out.

Edit2: I'm not very familiar with checks, but would it be possible to add a check for the "mention everyone" permission? So these could only be used if the user running the commands has mention everyone anyways?

mikeshardmind:
Yes, and if the mods and admins can use @ everyone, they probably should be doing so with consideration involved, not just tossing a pre-existing message with it. While changing that to be more restrictive is an absolute minimum from me, I'd prefer this not be added to the core bot at all. I don't think a moderation bot should come packaged with what I would consider as bad moderation practices.

mikeshardmind:
Okay, so there was some discussion of this on discord since what I wrote.

Here's what I would need to see an in implementation of this if it is to be added to prevent potential issues with this interacting with larger servers moderation needs in a poor way.

1. This command needs to be enabled to be used. It is toggled on/off per guild by another command which is guild owner only. The default is off.

2. The default check for the command when enabled is guildowner or already has permissions to mention everyone (overridable by permissions cog)


This can also be written easily as a 3rd party extension of the core custom commands cog.

```py
from redbot.cogs.customcommands import CustomCommands as _Red_CC
from redbot.core.utils import commands, checks

class CustomCommands(_Red_CC):
    # override anything / add anything
    # stays up to date with other changes here automatically.

```

This version would take precedence for load if installed as overrides of core extensions are considered to be a deliberate user choice.

Malarne:
Just added a toggle to that new command, to enable or disable everything:
- if toggled off, the `[p]cc add mod` command won't appear in help
- if toggled on, it'll appear, and work just fine
- if toggled off again, it'll disappear and the existing ccs created using `[p]cc add mod` will be blocked

I think it is kinda nice to have that feature in core, so people can use it if they need or not, it's still really simple, i don't think like a 3rd party cog is really needed to allow that kind of ccs to be created, anyway it's just my opinion.

Malarne:
After some discussion on #testing yesterday, i ended up adding a new decorator to commands, allowing that command to have a cost, set by the cog author.
If the command errors out, the cost if refunded.
Would allow more minigames linked to economy, also i didn't locked the cost to positive amount, so it also allow to reward the author of the command if the cost is set to a negative amount
(i didn't know how to make 2 seperate PRs with same fork, don't yell at me i still need to learn gh aha)

EgonSpengler:
Regarding multiple pull requests, the easiest method I find is to have a separate branch for each feature you're working on, based on upstream/V3/develop, then you can have a PR from each branch.

    git remote add upstream https://github.com/Cog-Creators/Red-DiscordBot
    git checkout -b my-feature upstream/V3/develop
    <do work>
    git push -u origin my-feature

mikeshardmind (dismissed):
This needs to be split into seperate PRs.

Malarne:
this PR is now only for ccs, i did the change and forgot to say it :x

Tobotimus:
Branch was deleted? :( Will leave the PR open for discussion

Flame442:
As a result of the discussion for #4633 and the fact that the branch for this PR was deleted, we are no longer interested in this PR.

PR #2616, opened 2019-04-25
[CustomCom] Update command generator to work with dpy update

tekulvw:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -466,7 +466,7 @@ class CustomCommands(commands.Cog):
             return
 
         # wrap the command here so it won't register with the bot
-        fake_cc = commands.Command(self.cc_callback)
+        fake_cc = commands.command(name=ctx.invoked_with)(self.cc_callback)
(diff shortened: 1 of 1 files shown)


redbot/cogs/customcom/customcom.py
```diff
@@ -466,7 +466,7 @@ def __init__(self, bot):
             return
 
         # wrap the command here so it won't register with the bot
-        fake_cc = commands.Command(ctx.invoked_with, self.cc_callback)
+        fake_cc = commands.Command(self.cc_callback)
```
mikeshardmind:
```suggestion
        fake_cc = commands.command(name=ctx.invoked_with)(self.cc_callback)
```
Manual invocation of the decorator should be less fragile. I'll admit I'm not super happy about dynamic command creation without attaching them to the bot. There's a lot of poorly defined behavior going on here, and most of it has no guarantees of non-breakage.

mikeshardmind (approved):

PR #2621, opened 2019-04-25
[Commands] Fix built in check decorator order affecting permissions

tekulvw:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

There is an issue in the existing **dev** codebase where the following does not respect permissions:

```py
@checks.is_owner()
@commands.command()
async def command(ctx):
    pass
```
However, if the check and command decorators are flipped in order, everything works as intended. The issues is caused by the rewritten commands extension from d.py "copying" command objects on cog creation and wasn't keeping the `requires` object during this copy function.

## This ONLY affects the latest DEV code
3.0.2 is not impacted by this problem.


mikeshardmind (approved):
:ghost:

PR #2623, opened 2019-04-25
Add @bank.cost decorator

Malarne:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x ] New feature

Add a new decorator @bank.cost(X) allowing a command to have a credits cost.
Note that there is no check that X is positive, so it also allows a command to reward credits to someone.

### DIFF
--- a/redbot/cogs/bank/bank.py
+++ b/redbot/cogs/bank/bank.py
@@ -124,4 +124,13 @@ class Bank(commands.Cog):
         await bank.set_currency_name(name, ctx.guild)
         await ctx.send(_("Currency name has been set to: {name}").format(name=name))
 
+    async def on_command_error(self, ctx, exception):
+        author = ctx.author
+        command = ctx.command.qualified_name
+        data = await bank._conf.member(author).last_command_cost()
+        if data["command"] != command:
+            return
+        else:
+            await bank.deposit_credits(author, data["cost"])
+
     # ENDSECTION
--- a/redbot/core/bank.py
+++ b/redbot/core/bank.py
@@ -3,7 +3,7 @@ from typing import Union, List, Optional
 
 import discord
 
-from . import Config, errors
+from . import Config, errors, commands
 
 __all__ = [
     "MAX_BALANCE",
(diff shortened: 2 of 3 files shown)


redbot/cogs/bank/bank.py
```diff
@@ -124,4 +124,13 @@ def __init__(self, bot: Red):
         await bank.set_currency_name(name, ctx.guild)
         await ctx.send(_("Currency name has been set to: {name}").format(name=name))
 
+    async def on_command_error(self, ctx, exception):
+        author = ctx.author
+        command = ctx.command.qualified_name
+        data = await bank._conf.member(author).last_command_cost()
```
mikeshardmind, line 130: This isn't a safe method of handling. async code makes no guarantees about the order this is handled compared to what another command does.

redbot/core/bank.py
```diff
@@ -37,7 +37,12 @@
 
 _DEFAULT_GUILD = {"bank_name": "Twentysix bank", "currency": "credits", "default_balance": 100}
 
-_DEFAULT_MEMBER = {"name": "", "balance": 0, "created_at": 0}
+_DEFAULT_MEMBER = {
+    "name": "",
+    "balance": 0,
+    "created_at": 0,
+    "last_command_cost": {"command": None, "cost": 0},
```
mikeshardmind, line 44: see above comment, but not a suitable solution here.

redbot/core/bank.py
```diff
@@ -669,3 +677,20 @@ def _invalid_amount(amount: int) -> bool:
...
+                last_data = {"command": str(ctx.command.qualified_name), "cost": amount}
+                await _conf.member(ctx.author).last_command_cost.set(last_data)
+                return True
+            else:
+                await ctx.send("You don't have enough money to use that command !")
+                return False
+        except:
+            return True
+
+    return commands.check(predicate)
```
mikeshardmind, line 696: This can't be a check decorator. This will subtract money for every use of help per usage of this decorator. Checks should not have interaction consequences beyond the current invocation being valid or not.

mikeshardmind (changes requested):
The idea for such a decorator is solid, the implementation unfortunately will not work here.

mikeshardmind:
Are you still interested in doing this? I think the concept is good, so if not, this should be closed and converted to a feature request for someone else to work on.

Malarne:
yup, needs to take time and think about how to make the changes you asked, may look into it this week :)

PR #2628, opened 2019-04-29
[Core] Help Redesign

mikeshardmind:
- Creates a new Red specific help formatting paradigm.
- Creates and uses a new Red specific help command
- Modifies each of the following to use them:

  - redbot.core.commands.Context
  - redbot.core.bot.RedBase

## This includes some behavior changes:

- By default, commands which are not found will not raise an error.
  - This may still be shown with fuzzy help
- Invoking help for a command you can't run will fail siltently

### DIFF
--- a/docs/changelog_3_1_0.rst
+++ b/docs/changelog_3_1_0.rst
@@ -37,6 +37,7 @@ Core
  * ``[p]set locale`` now only accepts actual locales (`#2553`_)
  * ``[p]listlocales`` now displays ``en-US`` (`#2553`_)
  * ``redbot --version`` will now give you current version of Red (`#2567`_)
+ * Redesign help and related formatter (`#2628`_)
  * Default locale changed from ``en`` to ``en-US`` (`#2642`_)
 
 ------
@@ -148,5 +149,6 @@ Utility Functions
 .. _#2605: https://github.com/Cog-Creators/Red-DiscordBot/pull/2605
 .. _#2606: https://github.com/Cog-Creators/Red-DiscordBot/pull/2606
 .. _#2620: https://github.com/Cog-Creators/Red-DiscordBot/pull/2620
+.. _#2628: https://github.com/Cog-Creators/Red-DiscordBot/pull/2628
 .. _#2639: https://github.com/Cog-Creators/Red-DiscordBot/pull/2639
 .. _#2642: https://github.com/Cog-Creators/Red-DiscordBot/pull/2642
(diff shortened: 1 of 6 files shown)


DevilXD:
Can there be a setting / command where one could set automatic deletion of help messages? Something along the lines of this `help_autodelete` cog for v2: https://cogs.red/cogs/ZeLarpMaster/ZeCogs/help_autodelete/, where a help response message is deleted along with the context message.

It does a really good job at cleaning up stray help messages people sometimes leave in not-so-frequently-visited channels.

mikeshardmind:
@DevilXD not in the initial conversion, but restructuring the help formatter like this is going to open up more easily adding such settings.

Stonedestroyer:
One request for this help-formatter would be an option either in permissions or in the help-formatter itself that you can decide to hide so people can not do help on commands they do not have permission on. The current iteration of help formatter does this but only on categories and overall menu and not specific single commands. 

From PR: https://github.com/Cog-Creators/Red-DiscordBot/pull/2583 .

mikeshardmind:
@stonedestroyer this should be the default behavior after the initial changes, with a setting exposed for this later on. Current work on this is intended to continue supporting settings we already have, while picking good defaults for remaining things for now, but in a way this can be changed later with additional settings.

After some further thought since discussion in your other PR, making the behavior consistent there by default is the right choice, and this is the best time to include that sort of change to the default behavior.

mikeshardmind:
This is ready for review. 

There are a couple small behavior changes here regarding commands which users can't run or are hidden, (as well as similar changes surrounding commands not found), but these are only intended as defaults. Settings for these can be added, and the plan to allow more easily registered 3rd party formatters is still planned as far as I'm concerned.

The behavior changes are all centered around help being quiet by default when the user should have no direct knowledge of a command from the bot in that context.

Stonedestroyer:
Not sure if I messed something up but getting
`   raise AttributeError("'{}' is not a valid registered Group or value.".format(item))
AttributeError: 'max_pages_in_guild' is not a valid registered Group or value.`

Edit: Disregard this, works on a clean venv.

mikeshardmind:
I need to fix something with the full bot help on this later, I seem to have overlooked a potential issue with it.

mikeshardmind:
PR is ready for use, fixed the bot specific issue which was overlooked. 

This is now failing on a docs issue unrelated to this PR. discord.py removed a build from RTD in removing the rewrite branch.

tekulvw (approved):

PR #2631, opened 2019-04-29
[Core_commands] Fix some missing i18n strings

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

There was some strings was missing i18n in core_commands, this PR fix that.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -38,7 +38,7 @@ __all__ = ["Core"]
 
 log = logging.getLogger("red")
 
-OWNER_DISCLAIMER = (
+OWNER_DISCLAIMER = _(
     "⚠ **Only** the person who is hosting Red should be "
     "owner. **This has SERIOUS security implications. The "
     "owner can access any data that is present on the host "
@@ -276,7 +276,7 @@ class Core(commands.Cog, CoreLogic):
     @commands.command(hidden=True)
     async def ping(self, ctx: commands.Context):
         """Pong."""
-        await ctx.send("Pong.")
+        await ctx.send(_("Pong."))
 
     @commands.command()
     async def info(self, ctx: commands.Context):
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -38,7 +38,7 @@
 
 log = logging.getLogger("red")
 
-OWNER_DISCLAIMER = (
+OWNER_DISCLAIMER = _(
```
mikeshardmind: This needs a small bit of handling above it to prevent translating this during import.

redbot/core/core_commands.py
```diff
@@ -276,7 +276,7 @@ class Core(commands.Cog, CoreLogic):
     @commands.command(hidden=True)
     async def ping(self, ctx: commands.Context):
         """Pong."""
-        await ctx.send("Pong.")
+        await ctx.send(_("Pong."))
```
mikeshardmind: This particular string needs to not go through i18n. If you want to suggest a different string such as the unicode character for a ping pong paddle, that's fine, but it has already been decided that this command specifically needs to do as little as possible to make it a reliable way to verify responsiveness. An additional comment can be added about this for future viewers if this is done.

mikeshardmind (changes requested):
Thank you for this. The more i18n support we get cleaned up in the bot the better. There are a few things here which need changes, but overall this looks good.

redbot/core/core_commands.py
PredaaA: Woops. Fixed, I have deplaced it directly to the command.

redbot/core/core_commands.py
PredaaA: Yes I can understand that. I put it back as it was before.

mikeshardmind (approved):
This is an improvement from missing coverage entirely. Looking over it, this is ready to go. I think there's still more to be done overall with i18n, but this is a nice chunk of progress on it.

PR #2643, opened 2019-05-03
Fixed missing imports.

Kowlin:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature


mikeshardmind (approved):
Side note: We need to add a missing imports test. This is something which really should have been caught by tests.

PR #2654, opened 2019-05-08
Config locks

Tobotimus:
This provides a way to lock by individual value, at any level of the config hierarchy desired.

Uses a weakref dictionary to maintain locks across multiple `Value` objects whilst they're simultaneously alive.

This also updates the `async with config.foo()` context manager to use the `foo` value's lock, unless a new `acquire_lock` kwarg is set to `False` for some edge case where acquiring the lock is undesirable.


mikeshardmind (changes requested):
This needs some work...

image

(This has a registered channel : `channels={}`, was being lazy)
(gotta love how the `__getattr__` continues to bite us)

However, there's another issue as well with this:

image

mikeshardmind (approved):
I've not found any other issues. I also looked through (did not manually test each of) the other places we already use config as a context manager to see if I could spot anywhere this might turn into a deadlock which wasn't before (though this would have indicated that section was problematic for the things this aims to help mitigate) and didn't spot any.

Definitely want to see if we can get this onto some of our larger ambassador bots without issue for testing if we have willing ~~victims~~ participants

PR #2657, opened 2019-05-10
[Config] Driver locks

tekulvw:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This PR introduces the `.atomic()` method on all value objects to be used as a context manager for atomic operations.

```py
async with config.data1.atomic() as data1:
    val = await data1()
    await data1.set(val + 1)
```

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -71,6 +71,33 @@ class _ValueCtxManager(Awaitable[_T], AsyncContextManager[_T]):
             await self.value_obj.set(self.raw_value)
 
 
+class _ValueAtomicManager:
+    def __init__(self, value_obj: "Value"):
+        self.value_obj = value_obj
+
+    async def __aenter__(self):
+        await self.value_obj.driver.lock_mgr.acquire_atomic(self.value_obj.identifier_data)
+        return self
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        self.value_obj.driver.lock_mgr.release_atomic(self.value_obj.identifier_data)
+
+    async def _get(self, default=...):
+        try:
+            ret = await self.value_obj.driver.get(self.value_obj.identifier_data, atomic=True)
+        except KeyError:
+            return default if default is not ... else self.value_obj.default
+        return ret
+
+    def __call__(self, default=...):
+        return self._get(default=default)
+
+    async def set(self, value):
+        if isinstance(value, dict):
+            value = _str_key_dict(value)
+        await self.value_obj.driver.set(self.value_obj.identifier_data, value=value, atomic=True)
+
+
 class Value:
     """A singular "value" of data.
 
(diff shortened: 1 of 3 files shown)


Tobotimus:
I'm obviously a bit biased (because of my alternative PR #2654) but I have some qualms with this implementation. There may be some advantages of this one which I'm not aware of, but for now, this is what I've got:
- This implementation is reinventing the wheel and adding more boilerplate code. There's a lot of code here and I'm not really sure what half of it does (esp. in LockManager).
  - It looks like the point is to do reference counting to keep the lock alive whilst it's acquired - why not just use weakref?
  - I'm not sure why locks need to be managed by the driver at all. I can see that the lock is acquired during a get or a set when `atomic=False`, whereas the lock should already be acquired and will be released later when `atomic=True`. But the only situation where this may be useful is if the cog dev is using `atomic()` in some places but not others, and this shouldn't be a situation we support or encourage, because it's bad practice, and it further complicates the design and muddles the critical region(s).
- The name "atomic" is a bit misleading. Atomic can have a few meanings: namely when referring to concurrency, atomic means only one thread (or task, in this case) can enter the critical region at any point in time. However it has a more prevalent meaning when referring to data transactions - atomicity as in ACID - where the transaction can end in either two states of fail or succeed, and failing ensures the underlying data is not modified. There is no guarantee of atomicity as in ACID from this implementation.
- The new context manager is not a very ergonomic interface to me. Asynchronously entering the context manager, then awaiting the context manager, seems a bit weird. Asyncio locks are already context managers, and it's very obvious what they do. I think we should just use them directly.

Advantages of #2654:
- Allows lock objects to be accessed directly. This means using them with `async with` *isn't* compulsory, as sometimes context managers are not viable.
- Using the existing `async with conf.value() as value:` implies the acquisition of this lock (pending a kwarg which is yet to be added). This will solve many issues where this is used without cog creators having to do a thing. It won't have a performance impact for the vast majority of cases.
- I think its operation is more intuitive. The cog creator doesn't need to learn how to use any new object or interface, in most cases it'll either require no extra code or knowledge, or simply the insertion of `async with conf.value.get_lock():`.

mikeshardmind:
I agree with Toby's comments about using the word atomic like this in any storage abstraction. Best to avoid any potential misunderstanding which we would have reason to be aware of.

I'll need to look more in depth at both designs before I can comment on the remaining parts.

mikeshardmind:
With the lock implementation at the config level added in #2654 , is this still something we want available at the driver level for other things, or is the expectation that anything we need access to should be handled via config, not the drivers?

Tobotimus:
I think given we have #2654, that's the interface we settled on. Locks being used at the driver level for future "compound" operations (e.g. inc and toggle) can accept the config lock as a kwarg and conditionally acquire it.

PR #2664, opened 2019-05-14
[Core] Added co-owners to outdated message warnings

Kowlin:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Since Co-owners are acting as another owner (Especially in the case of Team applications) they should receive the warnings of an outdated bot on start up.

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -104,7 +104,9 @@ def main():
     log.debug("Data Path: %s", data_manager._base_data_path())
     log.debug("Storage Type: %s", data_manager.storage_type())
 
-    red = Red(cli_flags=cli_flags, description=description, dm_help=None)
+    red = Red(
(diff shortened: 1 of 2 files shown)


mikeshardmind (changes requested):
With this attempting to send to multiple owners, we need to supress any errors sent so that the event doesn't die if say, you try to DM the team owner (403 expected) or any specific owner has DMs closed.

Additionally, this should use `get_user`, not `fetch_user`, since you can't send DMs without a shared server anyway.

Kowlin:
The reason behind the changes for ``core/__main__.py`` are to accommodate the need for a mutual server. There are cases whereby the bot will **not** fetch a guild cache due to its size, this is seen in mega guilds. This will result in a longer boot up cycle, but its a small trade-off down the line for not only this change but also other cogs that need a up to date guild cache with everyone in it from the start.

mikeshardmind (approved):

PR #2690, opened 2019-05-16
[Downloader] Fix problem with copying directory tree.

### LINKED ISSUE #2685: [Downloader] Cog isn't installed properly after uninstalling
Cog files fail to copy during installing if installation process for that cog was done before in current Red session, e.g:
```
[p]cog install SauriCogs cookies
[p]cog uninstall cookies
[p]cog install SauriCogs cookies
```

This issue is connected to `distutils.dir_util.copy_tree` behavior, which saves all created dirs to `distutils.dir_util._path_created`. There are few ways of solving this:

1. A little hacky way - clear `distutils.dir_util._path_created` before using `distutils.dir_util.copy_tree`
```py
distutils.dir_util._path_created = {}
```
2. Use `shutil.copytree` function, this will fail, if directory already exists, so that's not really an option. Option to change that behavior is going to be available in Python 3.8, but that's long way, so this option can't be used.
3. Use `distutils.dir_util.remove_tree` during removal of cog instead of `shutil.rmdir`
This option is less error-proof, because there can still be something left in `distutils.dir_util._path_created` and this will also cause more troubles for people who want to manually remove a cog or just change some directories in it.

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

fix #2685
This might not be perfect solution as we have to clear `distutils.dir_util._path_created` dict, but it seems like the best way to fix this, if we don't want to reimplement `distutils.dir_util.copy_tree`.

### DIFF
--- a/redbot/cogs/downloader/installable.py
+++ b/redbot/cogs/downloader/installable.py
@@ -114,6 +114,7 @@ class Installable(RepoJSONMixin):
         if self._location.is_file():
             copy_func = shutil.copy2
         else:
+            distutils.dir_util._path_created = {}
             copy_func = distutils.dir_util.copy_tree
 
         # noinspection PyBroadException
--- a/redbot/cogs/downloader/installable.py
+++ b/redbot/cogs/downloader/installable.py
@@ -114,6 +114,7 @@ class Installable(RepoJSONMixin):
         if self._location.is_file():
             copy_func = shutil.copy2
         else:
+            # clear copy_tree's cache to make sure missing directories are created (GH-2690)
             distutils.dir_util._path_created = {}
             copy_func = distutils.dir_util.copy_tree
 


mikeshardmind:
I'd like to explore this a bit more before using this particular fix to ensure resetting `distutils.dir_util._path_created` won't have extra side effects which aren't acceptable.

Additionally, shutil.copy_tree doesn't directly call any platform specific code, and we can copy the 3.8 implementation of it into `redbot.core.utils` with proper attribution if this fix does have undesirable side effects.

(see [bpo-20849](https://github.com/python/cpython/commit/9e00d9e88fbf943987e4771c753f5ca8f794103e) for more details)

Tobotimus:
When moving a directory tree in Python comes down to messing with the internals of distutils... Makes me wonder whether it's worth looking elsewhere (3rd party) or implementing it ourselves

Jackenmen:
Okay, while I understand why we may not want to mess with internals just to copy directory directory tree, it seemed as a better option than having to copy updated implementation of shutil.copy_tree from Python 3.8 just for 

Before crating this PR I checked to make sure, that `_path_created` is not used for anything else than checking if path was created previously by dir_util and that variable is only used in 2 places: `mkpath` which check if path was created before trying to create it again (which is only put there to eliminate redundant calls and that will still happen as we only clear `_path_created` before using `copy_tree`) and `remove_tree` which removes paths from this cache after removing a particular directory.

Not sure if it makes it any better for you, that even Python's tests clear `_path_created` in tests running multiple times to make sure, that dir_util's cache won't cause any issues.
https://github.com/python/cpython/blob/3.7/Lib/test/libregrtest/refleak.py#L201

I know it makes a code look a little obscure, but I still believe this solution is better than pretty much duplicating code of what is already available in Python.

mikeshardmind (changes requested):
I'm not super happy about this, but I can't find any reason for which this is unsafe to use.

I'd prefer if you added a comment linking to the issue *and* this PR next to that line prior to merge, but I'm fine with this being the solution used.

Jackenmen:
Is this fine?
```
# clear copy_tree's cache to make sure missing directories are created (GH-2690)
```

mikeshardmind:
Yep, just something to prevent people from scratching heads at this later

Jackenmen:
Okay, I added it.

mikeshardmind (approved):

PR #2693, opened 2019-05-17
[Core] Strip commas

Zephyrkul:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Strips commas from cogs in load, unload, and reload. Allows for easy copy-pasting from the cogs command.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -529,6 +529,7 @@ class Core(commands.Cog, CoreLogic):
         """Loads packages"""
         if not cogs:
             return await ctx.send_help()
+        cogs = tuple(map(lambda cog: cog.rstrip(","), cogs))
         async with ctx.typing():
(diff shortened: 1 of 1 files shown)


mikeshardmind (commented):
I'm not sure this is the best way to handle this.

simply making the converter used `lambda s: s.rstrip(",")` for cogs would handle this.

That has it's own downsides with type checking. I'll wait for more discussion on this, but if we're going to commonly allow comma separated values to be stripped automatically, it's worth discussing a standard way to handle it.

Zephyrkul:
I generally prefer only to use function typehint converters when it's syntactically clear / easily readable, which lambda typehints are not. Even then my own mypy complains when that happens so I've been using them less and less. That said if you prefer to handle this the way you personally prefer to, I will close this and instead open an issue.

mikeshardmind (approved):
per discussion in discord, this one being a simple case doesn't require some of the planned things towards more useful type hints in red.

PR #2694, opened 2019-05-17
[Config] Adds more channel type specificity

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This is breaking, despite a comment about not changing something in it to prevent (other) breakage.

This should not be merged prior to the cut off for micro version in the 3.1.x series.


redbot/core/config.py
```diff
@@ -1022,6 +1097,57 @@ def custom(self, group_identifier: str, *identifiers: str):
...
+        values which have not yet been set.
+
+        Returns
+        -------
+        dict
+            A dictionary in the form {`int`: `dict`} mapping
+            :code:`CHANNEL_ID -> data`.
+
+        """
+        return await self._all_from_scope(self.VOICE_CHANNEL)
```
Tobotimus:
```suggestion
        return await self._all_from_scope(self.TEXT_CHANNEL)
```

Tobotimus (changes requested):

redbot/core/config.py
```diff
@@ -893,6 +920,54 @@ def channel(self, channel: discord.TextChannel) -> Group:
...
+
+        """
+        return self._get_base_group(self.VOICE_CHANNEL, str(channel.id))
+
+    def category_channel(self, channel: discord.CategoryChannel) -> Group:
+        """Returns a `Group` for the given channel.
+
+        Parameters
+        ----------
+        channel : `discord.TextChannel`
```
Tobotimus:
```suggestion
        channel : `discord.CategoryChannel`
```

Tobotimus (changes requested):

mikeshardmind:
Closing this, for now. Will be easier to re-implement when time is correct with the other large changes to config.

PR #2698, opened 2019-05-18
[CustomCom] Fix KeyError with space use.

### LINKED ISSUE #2697: CustomCommand on_message listener error
# Other bugs

The CustomCommands cog throws an error into the console if a message processed in the `on_message` listener happens to start with a prefix and a space following right after.

1. Have CC loaded
2. `[p]cc add simple bugtest This command is for bug testing only!`
3. `[p] test` (notice the space after the prefix)

Haven't tested if this is still the case without any CC added. Traceback: https://i.imgur.com/SQdk8ot.png

Python 3.7.1, Red 3.1.1, D.py 1.0.1

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

fixes #2697

### DIFF
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -1,3 +1,4 @@
+import contextlib
 import re
 import random
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


mikeshardmind:
closing in favor of #2739 which more clearly shows what the failure condition actually is being handled.

PR #2699, opened 2019-05-18
[Streams] Fix NameError

### LINKED ISSUE #2696: Undefined variables in #2679
in #2679 
redbot/cogs/streams/streams.py line 160 has 2 undefined variables
**self**.bot and **channel**.guild

This method is static so it doesn't have access to self.
channel is never defined in the method either probably just a typo and the correct one should be "ctx" instead of "channel "from what i understand

mikeshardmind:
- fixes #2696
 - My fault for just looking at the github diff and seeing the logic fix
 on this one, without verifying the name validity in surrounding
 context. (see my review in #2679)

### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -128,8 +128,7 @@ class Streams(commands.Cog):
         stream = PicartoStream(name=channel_name)
         await self.check_online(ctx, stream)
 
-    @staticmethod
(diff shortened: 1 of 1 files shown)


palmtree5 (approved):
TBH I should've noticed said issues when I made the PR lol

PR #2706, opened 2019-05-19
[Alias] Properly track whitespace for use

### LINKED ISSUE #2704: [V3 Alias] New lines don't get passed to target command
# Command bugs

#### Command name
   [p]alias

   alias

   When using an alias, new lines get passed on to the target command.

#### What actually happened?
   It removes the new lines and passes multiple lines as a single line to the target command.

   Should be pretty self explanatory, but here's a screencap [dead link]
   Also, although I'm not sure if it matters, the bug occurred on my v3.1.1 bot

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Actually tracks the whitespace used instead of assuming it's all spaces, later restores the tracked whitespace.

fixes #​2704


Olaroll:
The alias no longer works if it starts with a line break. I'm no programmer though, so I don't really know why that is.
EDIT: it does work if there's a space after the alias, then a line break.

DiscordPTB_2019-05-19_11-30-39
DiscordPTB_2019-05-19_11-45-13

mikeshardmind:
I've pushed a fix for that, but at this point I'm convinced that while this fixes the issue in the short term, it's not a sufficient long term solution.

Relying on discord.py's string view like this precludes the ability to have input to an alias such as 

```
[p]somealias "this alias has no args and invokes something with consume-rest", "But this will still fail where the command wouldn't as a result of the comma after the prior quote"
```

mikeshardmind:
I'm still not happy with this as a solution for this. I think the better option might be to rewrite alias and customcom to not exhibit any of the odd issues related to this. Should probably freeze this unless someone sees an immediate need for only a partial fix on this edge case.

Tobotimus:
Considering putting this in as a hotfix for 3.1.3. Perhaps a rewrite of the alias internals is on the cards for 3.2.

PR #2726, opened 2019-05-23
[Docs] Mac/Windows autostart docs

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Adds in Mac/Windows autostart docs. Gonna need to have someone actually test the instructions in the mac ones though seeing as I don't have a mac and came up with that plist based on reading docs


palmtree5:
Ok, NSSM docs have been added, though so far it only covers installing NSSM via Chocolatey and then the actual setup of the service. Service setup is written to be done in Powershell, not cmd (there's four spots that use syntax for Powershell, three of them I know for sure how to get the same behavior in cmd, the fourth (which involves using the output of one command as an argument in another) I'm not sure on replicating in cmd)

docs/autostart_win.rst
```diff
@@ -0,0 +1,84 @@
...
+
+----------------------
+Setting up the service
+----------------------
+
+To create the service, run the following command:
+
+.. code-block:: none
+
+    nssm install redbot $(Get-Command redbot).Definition <instance name> --no-prompt
```
Jackenmen, line 53: For this to work correctly with venv install, we should tell the user to activate venv before running this command.

docs/autostart_win.rst
```diff
@@ -0,0 +1,84 @@
...
+==================================
+Setting up auto-restart on Windows
+==================================
+
+-------------------------------
+Downloading NSSM via Chocolatey
+-------------------------------
+
+.. note:: This assumes you already have chocolatey installed. 
+          If not, you may install it or follow the manual instructions for getting NSSM
```
Vexed01: missing full stop
```suggestion
          If not, you may install it or follow the manual instructions for getting NSSM.
```

docs/autostart_win.rst
```diff
@@ -0,0 +1,84 @@
...
+    nssm set redbot AppStdout $Env:USERPROFILE\red-stdout.log
+    nssm set redbot AppStderr $Env:USERPROFILE\red-stderr.log
+    nssm set redbot AppNoConsole 1
+
+The above commands change a number of settings for the service including:
+- Ensuring that when ``[p]shutdown`` is run, the service doesn't attempt to restart
+- Ensuring that if Control-C is pressed for the process, nssm waits 10 seconds
+- Setting a 15 second delay for service restart
+- Setting a couple of log files for the service
+- Disabling the console window
```
Vexed01: When `sphiux-build` is run, this doesn't format correctly:
image
For reference here's my `.html` file:
[#2726 html files.zip](https://github.com/Cog-Creators/Red-DiscordBot/files/3827095/2726.html.files.zip)

Vexed01 (commented):
Just spotted a few errors 😄

docs/autostart_win.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Then run the following set of commands one at a time:
+
+.. code-block:: none
+
+    nssm set redbot AppExit 0 Exit
+    nssm set redbot AppStopMethodConsole 10000
+    nssm set redbot AppRestartDelay 15000
+    nssm set redbot AppStdout $Env:USERPROFILE\red-stdout.log
+    nssm set redbot AppStderr $Env:USERPROFILE\red-stderr.log
```
Jackenmen, line 65: I think it would be better to use one log file for both stdout and stderr (nssm support that), it's easier to read like this.

docs/autostart_win.rst
```diff
@@ -0,0 +1,105 @@
...
+
+- Setting a couple of log files for the service
+
+- Disabling the console window
+
+Now that the service has been configured, run the following command:
+
+.. code-block:: none
+
+    nssm set redbot ObjectName $Env:COMPUTERNAME\$Env:USERNAME <password>
```
Jackenmen, line 84: I personally wouldn't want to put my password in cmd/powershell as part of a command (one of the reason being it saving in command history), I think there should at least be mentioned an alternative way to set that password through `services.msc`.

Another issue user may run into is having no password on the account - he won't be able to run service on an account with a blank password as it could be considered a security issue, see: <https://stackoverflow.com/a/1048691>
While we probably shouldn't provide solution for it, it might be worth mentioning in the document.

docs/autostart_win.rst
```diff
@@ -0,0 +1,105 @@
...
+
+.. code-block:: none
+
+    choco install nssm -y
+
+------------------------
+Manually installing NSSM
+------------------------
+
+Go to `<https://nssm.cc/download>`_ and download the latest release. Once the download
```
Jackenmen, line 27: It might be good to point at latest **pre** release (or at least pre-release build 2.2.4-101) as otherwise users may have problems with starting service on Windows 10 1703+. Technically we avoid this issue with `AppNoConsole=1` but choco already installs that build anyway and this build also comes with handy `nssm dump` command that may help make certain things easier for Support Squad.
Update: From only 2 support cases, I already found `nssm dump` very valuable debug command that allows me to help people with nssm.

Jackenmen (changes requested):

Jackenmen:
One more thing about log files - I want to mention that there's also an option in nssm to rotate log files which might be worth enabling to avoid getting one very big log file, but I'm leaving that to reviewer of this PR.
I haven't looked closely at file rotation settings, so I'm not sure if there's an option to set max amount of those files before the old ones get removed.

docs/autostart_win.rst
Jackenmen: A follow up on this, both powershell and cmd have certain characters that have to be escaped, currently provided solution assumes that user's password won't contain such characters. I think that a better solution here would be using something that will prompt for password, e.g:
`nssm set redbot ObjectName $Env:COMPUTERNAME\$Env:USERNAME $(Read-Host "Password")`

This one would still keep the password visible (as `input("Password: ")` does in Python), but it would at least always get the password right.

There is an option to replace password characters with asterixes in `Read-Host` by using `-AsSecureString` but it would result in this awfully wrong command (or at least I couldn't find a way to make this shorter in powershell):
`nssm set redbot ObjectName $Env:COMPUTERNAME\$Env:USERNAME $([System.Net.NetworkCredential]::new("", $(Read-Host -AsSecureString "Password")).Password)`

palmtree5: yeah, probably gonna take some testing on this one

mikeshardmind:
This is being deferred until further work with making this user-friendly.

Drapersniper:
We should also update this to use `python -O -m redbot` instead of just `redbot` (to go in line with the rest of the docs)

aikaterna:
If we have any other Mac users willing to test the plist file I would appreciate feedback as well. When I used it back when originally testing this, I recieved Error 2's while trying to have it autostart (usually representative of "file not found", but could also be something else, even though I pointed it at the correct redbot executable). The only other testing I have seen related to this in the Red server was a user in support at https://discord.com/channels/133049272517001216/387398816317440000/668614855397212170 who recieved a permissions error while trying to run the sudo command - not sure if it was a password issue or a true perms issue. Iirc I didn't run into permissions trouble when I tried this in Dec 19 and again in Februrary but it's something to be aware of.

palmtree5:
Superseded by #4081 and #4082 (splitting the OSes up)

PR #2728, opened 2019-05-23
Add debug info command.

Stonedestroyer:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

This adds a debug command that works with both embeds and codeblocks.
I refrained from adding translation due to the fact that it's a debug command.

Thanks to Sinbad for some help with this and Palm.


mikeshardmind (approved):
Thanks!

PR #2731, opened 2019-05-24
Add respectable aliases for consistency

NeuroAssassin:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Before, there were a lot of different commands where some were using `del`, some using `delete` and others using `remove` (in the same context that `del` or `delete` could be).  This PR adds all of them to each of the commands (so that you can use `del`, `delete` or `remove` in the proper places), so that there is consistency between the commands.


mikeshardmind (changes requested):
Can you switch some of these so that all of the names are the same and all the aliases are the same? I'm not picky on which (though it should be one of the ones fully spelled out I guess) but this will keep the interface consistent in help too while still allowing the others.

mikeshardmind (approved):

PR #2729, opened 2019-05-24
Fixes some issues with API help commands

Flame442:
### Type
- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

### Description of the changes
Lots of general formatting bugs and clarity issues.

PR #2733, opened 2019-05-25
[Audio] Clarity changes for the API info commands

Flame442:
### Type
- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes the same issues fixed in #2729 but for Audio API commands.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -592,15 +592,15 @@ class Audio(commands.Cog):
     async def spotifyapi(self, ctx):
         """Instructions to set the Spotify API tokens."""
         message = _(
-            f"1. Go to Spotify developers and log in with your Spotify account\n"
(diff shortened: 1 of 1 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -592,15 +592,15 @@ def _restart_connect(self):
     async def spotifyapi(self, ctx):
         """Instructions to set the Spotify API tokens."""
         message = _(
-            f"1. Go to Spotify developers and log in with your Spotify account\n"
+            f"1. Go to Spotify developers and log in with your Spotify account.\n"
```
mikeshardmind:
```suggestion
            "1. Go to Spotify developers and log in with your Spotify account.\n"
```

mikeshardmind (changes requested):
There's an f-string hiding inside there with no reason other than to break i18n. Can be merged once that's sorted.

mikeshardmind (approved):
Thanks!

PR #2734, opened 2019-05-26
[p]slowmode

DevilXD:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This PR adds a `[p]slowmode [interval=0]` command, for easy slowmode management on a channel. This lets anyone with `Manage Channels` permission or a `Mod` role to change the slowmode setting.

Could be changed to require `0` to disable instead of having it there by default, this is up for discussion.

### DIFF
--- a/redbot/cogs/mod/slowmode.py
+++ b/redbot/cogs/mod/slowmode.py
@@ -23,9 +23,7 @@ class Slowmode(MixinMeta):
         You can specify the unit using `s`, `m` or `h` as suffix.
         Use without parameters or set to 0 to disable.
         """
-        if interval is None:
-            interval = 0
-        else:
+        if interval:
             match = re.match(r"(?i)(\d{1,5})([a-zA-Z])?", interval)
             if match:
                 interval = int(match.group(1))
@@ -36,6 +34,8 @@ class Slowmode(MixinMeta):
             if not match or not 0 <= interval <= 21600:
                 await ctx.send(_("Interval must be between 0 seconds and 6 hours!"))
                 return
+        else:
+            interval = 0
         await ctx.channel.edit(slowmode_delay=interval)
         if interval > 0:
             interval = humanize_timedelta(seconds=interval)
--- a/redbot/cogs/mod/slowmode.py
+++ b/redbot/cogs/mod/slowmode.py
@@ -1,6 +1,8 @@
-import discord
-from redbot.core import commands, i18n, checks
+import re
 from .abc import MixinMeta
+from datetime import timedelta
+from redbot.core import commands, i18n, checks
+from redbot.core.utils.chat_formatting import humanize_timedelta
 
 _ = i18n.Translator("Mod", __file__)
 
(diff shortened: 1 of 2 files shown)


redbot/cogs/mod/slowmode.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+
+    @commands.command()
+    @commands.guild_only()
+    @commands.bot_has_permissions(manage_channels=True)
+    @checks.mod_or_permissions(manage_channels=True)
+    async def slowmode(self, ctx, interval: int = 0):
+        """Change channel's slowmode setting.
+
+        Interval can be anything from 0 to 21600 seconds.
```
skeith: can you add human time format converter?

seconds, minutes, hours etc

DevilXD: As the input parameter? Or for the `Slowmode interval is now {interval} seconds.` feedback message?

skeith: both

Jackenmen:
I don't believe it's for us to decide, if mods should be able to use this and I think this should be changed so only people with admin role or manage channels permission can use it and just leave it to Red users to decide if they want to allow it to mods role (which they can do by using permissions cog)

DevilXD:
This last commit comes with a simple time converter per request, that will convert input like `5`, `5s`, `5m` or `5h` to appropriate time interval. `humanized_timedelta` from `utils` is now used to pretty-print the slowmode interval set. Addressing jack's concerns, I've also bumped the decorator to `admin_or_permissions`, with the intention that it could be overwritten with permissions later if needed as suggested.

A proper, fully fledged time interval converter could be used here, but I believe this is beyond this simple PR. Such converter could be really useful to everyone using Red in general, not only here in the Mod cog, and could be incorporated here at a later date when available.

redbot/cogs/mod/slowmode.py
DevilXD: Should be resolved with the recent commit.

mikeshardmind:
Assuming this works properly (It looks solid, but I won't have a chance to test until later) I'm fine with merging prior to having a generalized timedelta converter in core and adding it and switching this to it later.

For anyone who goes to add that, the converter should likely be generalized in such a way in which valid units can be passed as well as an upper and lower bound on the time so that all of that is handled in the conversion step.

mikeshardmind (changes requested):
This works great for an initial addition. I do have a small change request here:

image

If this is going to use a simplistic handling prior to there being a generalized converter, this should throw an error to not cause confusion, rather than allow it to succeed with an ambiguous meaning.

a tweak to the regex used, or using `re.fullmatch` should be reasonably sufficient for this on the short term until the generalized converter is done.

DevilXD:
Right, I was using `match` instead of `search` already,  but forgot that it still requires an explicit line end. Should be as easy as adding `$` at the end of the regex pattern.

redbot/cogs/mod/slowmode.py
```diff
@@ -0,0 +1,44 @@
...
+                    ) * interval
+            if not match or not 0 <= interval <= 21600:
+                await ctx.send(_("Interval must be between 0 seconds and 6 hours!"))
+                return
+        else:
+            interval = 0
+        await ctx.channel.edit(slowmode_delay=interval)
+        if interval > 0:
+            interval = humanize_timedelta(seconds=interval)
+            await ctx.send(_(f"Slowmode interval is now {interval}."))
```
Tobotimus: Please use `_("message").format()` rather than f-strings for user-facing strings. f-strings fundamentally can't work with gettext due to order of evaluation.
```suggestion
            await ctx.send(_("Slowmode interval is now {interval}.").format(interval=interval))
```

redbot/cogs/mod/slowmode.py
```diff
@@ -0,0 +1,44 @@
...
+        """
+        if interval:
+            match = re.match(r"(?i)(\d{1,5})([a-zA-Z])?$", interval)
+            if match:
+                interval = int(match.group(1))
+                if match.group(2) is not None:
+                    interval = {"s": 1, "m": 60, "h": 60 * 60}.get(
+                        match.group(2).lower(), -1
+                    ) * interval
+            if not match or not 0 <= interval <= 21600:
```
mikeshardmind: These negative conditions should be handled separately. 

To a user, getting this feedback as a result of this input isn't clear what they did wrong. Sending the help on a non match is fine here.

image

mikeshardmind (dismissed):

redbot/cogs/mod/slowmode.py
DevilXD: Right, switched it to the format method, slightly shorter than your version - hopefully that's alright.

redbot/cogs/mod/slowmode.py
DevilXD: This should be resolved once #2736 will be accepted, so I can use it here.

DevilXD:
This should be ready to go now.

Tobotimus (approved):
Thanks for your contribution! Works well :)

PR #2736, opened 2019-05-28
Add a timedelta converter

DiscordLiz:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

This reuses a lot of logic from @mikeshardmind 's scheduler cog with permission

It adds a timedelta converter
It keeps it generalized as requested
It keeps the function available for non converter use as requested

### DIFF
--- a/redbot/core/commands/converter.py
+++ b/redbot/core/commands/converter.py
@@ -45,12 +45,16 @@ TIME_RE = re.compile(TIME_RE_STRING, re.I)
 
 def parse_timedelta(
     argument: str,
+    *,
     maximum: Optional[timedelta] = None,
     minimum: Optional[timedelta] = None,
     allowed_units: Optional[List[str]] = None,
 ) -> Optional[timedelta]:
     """
     This converts a user provided string into a timedelta
+
+    The units should be in order from largest to smallest. 
+    This works with or without whitespace.
 
     Parameters
     ----------
@@ -60,7 +64,7 @@ def parse_timedelta(
         If provided, any parsed value higher than this will raise an exception
     minimum : Optional[timedelta]
         If provided, any parsed value lower than this will raise an exception
-    allowed_untis : Optional[List[str]]
+    allowed_units : Optional[List[str]]
         If provided, you can constrain a user to expressing the amount of time
         in specific units. The units you can chose to provide are the same as the
         parser understands. `weeks` `days` `hours` `minutes` `seconds`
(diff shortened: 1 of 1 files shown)


redbot/core/commands/converter.py
```diff
@@ -1,23 +1,108 @@
...
+
+    Parameters
+    ----------
+    argument : str
+        The user provided input
+    maximum : Optional[timedelta]
+        If provided, any parsed value higher than this will raise an exception
+    minimum : Optional[timedelta]
+        If provided, any parsed value lower than this will raise an exception
+    allowed_untis : Optional[List[str]]
```
mikeshardmind:
```suggestion
    allowed_units : Optional[List[str]]
```

redbot/core/commands/converter.py
```diff
@@ -116,3 +201,44 @@ class ValidatedConverter(DictConverter, metaclass=PartialMeta):
...
+    minimum: Optional[timedelta] = None,
+    allowed_units: Optional[List[str]] = None,
+) -> type:
+    """
+    See `parse_timedelta` for more information
+    """
+
+    class PartialMeta(type(TimedeltaConverter)):
+        __call__ = functools.partialmethod(
+            type(DictConverter).__call__, allowed_units, minimum, maximum
```
mikeshardmind: you've put these in the wrong order for the signature of the class

mikeshardmind (changes requested):
There are a couple of things which need fixing, beyond that I think the arguments other than the user input here are better suited as keyword only in all instances for the timedelta conversion stuff.

redbot/core/commands/converter.py
```diff
@@ -116,3 +205,81 @@ class ValidatedConverter(DictConverter, metaclass=PartialMeta):
...
+        If provided, any parsed value higher than this will raise an exception
+    minimum : Optional[timedelta]
+        If provided, any parsed value lower than this will raise an exception
+    allowed_units : Optional[List[str]]
+        If provided, you can constrain a user to expressing the amount of time
+        in specific units. The units you can chose to provide are the same as the
+        parser understands. `weeks` `days` `hours` `minutes` `seconds`
+    """
+
+    def __init__(self, *, minimum=None, maximum=None, allowed_units=None):
```
DevilXD, line 230: Hmm, what about adding something along the lines of a `default_unit`, that would be used when the argument parsed would be an integer only?

This would be really helpful for other cogs where passing just a number would default to the unit specified, while passing numbers with units would be converted as it currently is now.

The argument could easily be checked with `argument.isdigit()` to determine if it should be parsed with `default_unit` or not.

Finally, setting `default_unit` to `None` (probably by default) would disable this additional functionality altogether.

DevilXD (changes requested):

redbot/core/commands/converter.py
DiscordLiz: Is there any benefit to this beyond not using a union converter? I don't think guessing or assuming units in a converter designed to accept multiple is a good idea.

mikeshardmind: I somewhat agree, except in the case of being guaranteed a specific type rather either using an int converter or needing to write a custom converter on top of this for that case.

I'm not completely made up on this either way, and think that while it's worth discussion, it's not worth delaying the addition since it can be added non-breakingly later, and in the mean time there are other options for this specific case.

mikeshardmind (approved):
This is ready to go, there may be more to add to this later, but it's in a good place to get people using it for now.

PR #2739, opened 2019-05-29
Fixed KeyError on specific message edge-case

DevilXD:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

After some deep testing, it turns out that when `get_raw` gets passed in an empty string, it returns the same thing `self.db(message.guild).get_raw("commands")` would return. This seems to be an underlying config issue with empty string handling, but this isn't a place to discuss this I guess.

Instead of all of this importing in #2698, my suggestion would be to just check if the `command` passed is an empty string, and if so - `raise NotFound()`
Appears to be working great with fixing #2697 just fine. A comment could be left to indicate the specific case this check checks for.


DevilXD:
So, in short - this fixes #2697, and is a simpler alternative to #2698

mikeshardmind (approved):

PR #2746, opened 2019-05-31
Adds support for non interactive use

DiscordLiz:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

Adds assume_yes to context
Changes cleanup's 100+ check
Changes cog update.

This prevents 3rd party cogs such as @mikeshardmind 's scheduler from needing to patch methods, and just add full support for it. If a scheduler ever is included in Red, this means there will already be a supported method for allowing commands which usually require confirmation.

3rd party cogs can be supported by things like a scheduler without the scheduler needing to be aware of those cogs.

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -33,6 +33,9 @@ class Cleanup(commands.Cog):
         Tries its best to cleanup after itself if the response is positive.
         """
 
+        if ctx.assume_yes:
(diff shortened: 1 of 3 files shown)


mikeshardmind (approved):
:+1: This is a nice addition. While I'm not sure a core scheduler is really feasible unless we reconsider flag use in core commands (prior discussion a while back), this still opens up a lot of possibilities.

PR #2753, opened 2019-06-02
[Commands] Added optional default_unit to the TimedeltaConverter

DevilXD:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This PR adds `default_unit` to the TimedeltaConverter, that let's you specify the default unit to use in cases where user would provide an integer input only. Let's consider this example:

`async def mute(self, ctx, user: discord.Member, minutes: int = 10)`

The command above, made using a standard base and discord converters, presumably lets you mute server members for the amount of time specified by `minutes`. Now, if one would like to mute for couple of hours or days, and generally have more granular and easy to use input, they could use the new TimedeltaConverter to achieve this:

`async def mute(self, ctx, user: discord.Member, time: TimedeltaConverter = timedelta(minutes=10))`

Now, because of this change, doing something like `[p]mute Member 20` will no longer work due to the converter expecting a unit specified right after. With this PR, a `default_unit` can be passed into the converter like so:

`async def mute(self, ctx, user: discord.Member, time: TimedeltaConverter(default_unit="minutes") = timedelta(minutes=10))`

..., making it parse integer-only input as a time duration as well, using the unit specified. This provides backwards compatibility with every command that used seconds, minutes, hours, days or weeks as the input unit, while also allowing to specify the time more precisely than it was possible.

I'm not that good with docs, so any improvement suggestions are welcome. 

Note on the implementation: `.isdigit()` specification ensures that `int()` should never fail to convert the string provided, so should be safe to use here.

### DIFF
--- a/redbot/core/commands/converter.py
+++ b/redbot/core/commands/converter.py
@@ -224,7 +224,7 @@ class TimedeltaConverter(dpy_commands.Converter):
     allowed_units : Optional[List[str]]
         If provided, you can constrain a user to expressing the amount of time
         in specific units. The units you can chose to provide are the same as the
-        parser understands. `weeks` `days` `hours` `minutes` `seconds`
+        parser understands: `weeks` `days` `hours` `minutes` `seconds`
     default_unit : Optional[str]
         If provided, it will additionally try to match integer-only input into
         a timedelta, using the unit specified. Same units as in `allowed_units`
@@ -254,6 +254,7 @@ class TimedeltaConverter(dpy_commands.Converter):
 
 def get_timedelta_converter(
     *,
+    default_unit: Optional[str] = None,
     maximum: Optional[timedelta] = None,
     minimum: Optional[timedelta] = None,
     allowed_units: Optional[List[str]] = None,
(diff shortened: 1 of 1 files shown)


redbot/core/commands/converter.py
```diff
@@ -225,17 +225,28 @@ class TimedeltaConverter(dpy_commands.Converter):
...
         self.allowed_units = allowed_units
+        self.default_unit = default_unit
         self.minimum = minimum
         self.maximum = maximum
 
     async def convert(self, ctx: "Context", argument: str) -> timedelta:
-        delta = parse_timedelta(
-            argument, minimum=self.minimum, maximum=self.maximum, allowed_units=self.allowed_units
-        )
+        if self.default_unit and argument.isdigit():
```
mikeshardmind:
```suggestion
        if self.default_unit and argument.isdecimal():
```

`isdigit` isn't safe for this.

for why, consider `³` . (In case of font display issues, the equivalent escaped string is `\N{SUPERSCRIPT THREE}`)

This passes both `isnumeric` and `isdigit` but will raise a ValueError if you attempt to make it an int.

mikeshardmind (changes requested):
Looks mostly fine, but needs a small adjustment before merge. Thinking on it, we may have the mixup I'm mentioning here in other places as well, and it may be a good idea to check for it.

DevilXD:
Right, didn't spot this one. I think I also may have this mixup in other places too :P
Updated per request, should be ready to go now =)

mikeshardmind (changes requested):
I should have noticed this before, but didn't. Sorry about tossing on something extra here again.

`get_timedelta_converter` also needs to be updated for the extra optional keyword arg.

mikeshardmind (approved):

PR #2754, opened 2019-06-02
[Core] Pass exceptions on Cog Load to user.

### LINKED ISSUE #2752: [3.1.2][Core] Conflicting commands empty error message
# Command bugs

#### Command name

`[p]load`

Core

When a cog with a conflicting command is loaded, it should show a related error message

#### What actually happened?

No specific error is shown, however only the wrapping "This package could not be loaded for the following reason:" error seems to be shown.

1) Load a cog with a conflicting command with an already loaded cog

Example

Tried force-reinstalling Red with --no-cache-dir included, tried re-adding repo, reinstalling cog etc.

Bakersbakebread:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Passed exception to fix #2752

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -262,7 +262,7 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):
         except Exception as e:
             self._remove_module_references(lib.__name__)
             self._call_module_finalizers(lib, name)
-            raise errors.CogLoadError() from e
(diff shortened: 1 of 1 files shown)


mikeshardmind (approved):
Fixes it in a sufficient enough way for now. I want to take a closer look at some of how we're handling this later, but for now, this should be fine.

PR #2755, opened 2019-06-02
[Audio] Remove blacklisted architectures

aikaterna:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

With the new Lavalink.jar we're distributing we shouldn't need to blacklist any architectures any more, but the list should be kept available in case we need it in the future.


mikeshardmind (approved):
If any particular arch does become a problem again, we can always add them back in until we have a solution. I agree on keeping the ability easy to re-add here.

Issue #2721, opened 2019-05-21
[core] New method for commands.Requires


mikeshardmind:
# Feature request

#### Select the type of feature you are requesting:

- [ ] Cog
- [ ] Command
- [X] API functionality

#### Describe your requested feature

add a "partial_verify" method which accepts a new partial_context object which has similar attributes to a full context object, but can be safely manually created, and is expected to not always have all ids. (The typical use case would likely still include all)

This would expose a bit more info for things like what's being worked on with RPC for a future possible web panel, without requiring manually stepping through the states to get the same info.

This would not be suitable for use in place of checks and should be documented as such. This should only return back a permission state. (as this would not be safe to run the full checks on, only get back the permission state from)

mikeshardmind:
There are better options for this which don't belong in the requires class to be explored at a point where it would enable something rather than "just be nice to have"

PR #2757, opened 2019-06-03
[Mod] Improved Slowmode

Kowlin:
Updated slowmode with an optional channel argument.

### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/cogs/mod/slowmode.py
+++ b/redbot/cogs/mod/slowmode.py
@@ -1,6 +1,8 @@
 import re
+import discord
 from .abc import MixinMeta
 from datetime import timedelta
+from typing import Optional
 from redbot.core import commands, i18n, checks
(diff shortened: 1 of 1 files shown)


mikeshardmind (changes requested):
This isn't safe for the check used. If this is to be added, we need something similar to the voice perm check allowing voice mutes since perms are granular by channel.

Kowlin:
Hmmm, right, the fact that we don't have those at the moment is becoming a real pain in the ass :eyes:

mikeshardmind:
Probably worth implementing the more general thing in #2721 prior to this to allow it to be done in a better way with less duplicated code everywhere we want to use such a pattern.

Kowlin:
Ye I do agree, I'm gonna go ahead and close this until later then.

PR #2761, opened 2019-06-05
[Bank API] Add cost decorator

mikeshardmind:
### Type

- [ ] Bugfix
- [x] Enhancement
- [X] New feature

Alternative to #2623 which wraps the actual callback. (discussed with original PR author in discord)


Flame442 (dismissed):
This is good for merge, some things I'd like to point out however:

-The deco works with group commands, however it still takes money when the group command is used with no subcommands. The following code can prevent money from being deducted on a group command with no subcommands:
```py
    raise bank.AbortPurchase
```

-It might be useful to make it a bit more clear that `AbortPurchase` is used to refund **without the user seeing an error** (I originally questioned why that error was there, not realizing it would be caught. Of course, I could just be a dumbass...).

-L711 could use a period.

mikeshardmind:
I've gone ahead and added that period, as well as been slightly more wordy in the docstring to clarify that `AbortPurchase` is used for control flow, not for user facing messages.

Flame442 (dismissed):

Tobotimus:
I've made a PR to your branch with some minor tweaks I'd like before merging this PR :)

Tobotimus (approved):

PR #2766, opened 2019-06-09
 [ModLog] Use custom scopes for ModLog Config

Tobotimus:
Modlog is currently the biggest culprit for seriously large documents in the MongoDB backend, since it stores all cases as nested dicts in the guild scope. So, for example, on the Fortnite server, the guild document for @Kowlin's bot has exceeded 8MB. 

This PR gives each case its own document. It also does the same for casetypes. Not only does it remove the possibility of the document exceeding the maximum size in MongoDB, it's also just more efficient for all backends.

#### Config Bug
These changes highlighted a bug in Config, which I've fixed here but will probably open a new PR for: when a custom scope uses more than one primary key, but does this with a partial primary key: `conf.custom(SCOPE, primary_key_1).all()`, the defaults will be mixed in with the top-level of the returned dict, even though the actual documents are nested further in.

#### General ModLog improvements
Fixed a bunch of type-hints, and also added more support for when an object related to a case (user, moderator, channel etc.) can't be found (because it was deleted or something rather)

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -116,7 +116,6 @@ def main():
     red.add_cog(CogManagerUI())
     if cli_flags.dev:
         red.add_cog(Dev())
-    loop = asyncio.get_event_loop()
     # noinspection PyProtectedMember
     loop.run_until_complete(modlog._init())
     # noinspection PyProtectedMember
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -113,11 +113,11 @@ def main():
     red.add_cog(CogManagerUI())
     if cli_flags.dev:
         red.add_cog(Dev())
+    loop = asyncio.get_event_loop()
     # noinspection PyProtectedMember
-    modlog._init()
+    loop.run_until_complete(modlog._init())
     # noinspection PyProtectedMember
     bank._init()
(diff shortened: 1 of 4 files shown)


Kowlin:
PR seems to work, however I think that there is a migration failur, it cannot reference to old mod cases within the ``casesfor`` command, or continue its old count.

Tobotimus:
Thanks, will take a look at it later this afternoon.

Tobotimus:
This PR now depends on #2795

redbot/__main__.py
```diff
@@ -116,8 +116,9 @@ def main():
     red.add_cog(CogManagerUI())
     if cli_flags.dev:
         red.add_cog(Dev())
+    loop = asyncio.get_event_loop()
```
mikeshardmind: Already assigned in line 110, no need to refetch the loop here.

Tobotimus: Oops! Thanks, that one on line 110 was added in a merge from V3/develop

Kowlin (approved):
PR is working as intended. Data is migrated over to custom handler and deleted from the ``modlog.guild`` as it should. Pulling back on old cases works, and continuity in modlog is working as intended.

PR #2770, opened 2019-06-13
[V3 Core] Add checks to [p]command

NeuroAssassin:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Now makes sure that when running `[p]command disable`, you are not disabling the `[p]command` command.  In addition, it makes sure the user running the `[p]command <> server` command can actually run the command they are disabling/enabling.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1912,6 +1912,12 @@ class Core(commands.Cog, CoreLogic):
             )
             return
 
+        if self.command_manager in command_obj.parents or self.command_manager == command_obj:
+            await ctx.send(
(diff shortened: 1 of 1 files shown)


mikeshardmind (changes requested):
Checking if the person can run it shouldn't happen IMO. This removes the possiblity for the bot owner to manage enabled guild only commands from DMs. If the goal of this is to avoid a guild owner locking an owner command, check the requires object for the required permission level. Beyond that, there's no reason someone able to run a disable should be unable to disable commands.

NeuroAssassin:
Just pushed a commit to address the requested changes.

mikeshardmind (approved):
Thanks!

PR #2774, opened 2019-06-16
[V3 Repo Manager] Sanitize Repo.url

Zephyrkul:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

If URL basic auth is present in a repo's URL, remove it.

### DIFF
--- a/redbot/cogs/downloader/repo_manager.py
+++ b/redbot/cogs/downloader/repo_manager.py
@@ -11,6 +11,7 @@ from subprocess import run as sp_run, PIPE
 from string import Formatter
 from sys import executable
 from typing import List, Tuple, Iterable, MutableMapping, Union, Optional
+from urllib.parse import urlsplit
 
 from redbot.core import data_manager, commands
 from redbot.core.utils import safe_delete
@@ -57,7 +58,7 @@ class Repo(RepoJSONMixin):
         available_modules: Tuple[Installable] = (),
         loop: asyncio.AbstractEventLoop = None,
     ):
-        self.url = url
+        self._url = url
         self.branch = branch
 
         self.name = name
@@ -206,13 +207,13 @@ class Repo(RepoJSONMixin):
         if self.branch is not None:
             p = await self._run(
                 ProcessFormatter().format(
-                    self.GIT_CLONE, branch=self.branch, url=self.url, folder=self.folder_path
+                    self.GIT_CLONE, branch=self.branch, url=self._url, folder=self.folder_path
                 )
             )
         else:
             p = await self._run(
                 ProcessFormatter().format(
-                    self.GIT_CLONE_NO_BRANCH, url=self.url, folder=self.folder_path
+                    self.GIT_CLONE_NO_BRANCH, url=self._url, folder=self.folder_path
                 )
             )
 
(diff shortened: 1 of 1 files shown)


mikeshardmind (changes requested):
This is an API change to do like this. Add a sanitized url property instead of changing the attribute to a property.

Use that in the places which grab it and present it in client.

Zephyrkul:
I think this API change is worth it. I don't think Repo should potentially expose URL auth through its public attributes, nor should future code changes have to remember to specifically call the sanitized version. Auth like this should be only for backend and not exposed to public attributes. If this does actually break something in the backend of downloader, feel free to let me know.

Tobotimus:
I'm with Zeph here... Besides, this API is undocumented and should only be used internally. We Red developers really need to settle on what APIs we consider public and shouldn't be broken in micro releases, and which APIs are flexible. Because the more we consider public, the harder everything is to improve. The costs of considering it public often far outweigh the benifits - I think this is an example of that.

Tobotimus:
I don't think Downloader functions were documented for the benefit of 3rd party devs. I think the documentation was put there to help core developers maintain and improve Downloader, but it managed to end up on RTD anyway. I'd really like to reconsider the APIs we show on RTD and consider public in the next minor release, and perhaps move the docs to a new "Red Developer Docs", or even just leave them as comments in the code. Other candidates for this would be things like the drivers for config, and some parts of data manager and cog manager.

mikeshardmind:
For better or for worse right now, this *is* a documented part of the API.

It's also possible someone is currently using this part of the API to build something which gets info about added repos which we do not. Changing this would would break something like that.

I don't think it's a hard ask for developers to use the correct attributes, especially if we document them, so I don't think it's worth changing a documented API to do this.

Tobotimus:
I know the `Repo` class is documented, the `url` attribute is not though

mikeshardmind:
Even so, I'd still disagree with changing this. It's not that we don't want the URL accessible outside the class, but that we don't want to leak secrets. Add a `clean_url` property which has this functionality, and we've now mirrored other similar in use APIs in a more clear manner.

All this does is make someone have to follow around some code to see why this is done this way, as neither of these are now documented, and we would be using both (even if only internally)

mikeshardmind:
(and such a change should probably include documenting both, the currently existing one with a note to note that it's unsafe to send to discord and a ref to `clean_url` for such)

Zephyrkul:
I will close this PR in favor of such an approach by a different PR, then.

Zephyrkul:
If I may ask... is there any particular reason we *want* this unsanitized URL available as a public API? Do any core cogs use it outside of the Repo class? Would these cogs require the auth? Should we allow access to the auth to 3rd party cogs if we're keeping this as part of the API? Is there any reason these cogs would need the unsanitized URL over the sanitized URL, and if so, is there any reason that they wouldn't just use the Repo class itself and let it handle the unsanitized URL? Is it worth the risk of malicious 3rd party code?

mikeshardmind:
I thought my last comments on this were pretty clear, but I guess I'll clear it up.

Having a `._url` and `.url`, neither of which are documented and both would need to be used is not okay, even from just an internal maintaining perspective.

Names should be meaningful. 

As for changing the API here to not allow it to be accessed externally, I'm actually not in favor of it. I see no good reason to hide access to it. 

> Is it worth the risk of malicious 3rd party code?

If you load code, the assumption is you trust the code to do what it is supposed to. Hiding this attribute doesn't remove the ability for malicious code someone loaded to use it, it just creates more things which aren't documented.  There's no such thing as private in python, it's only private by convention.

On the other hand, to prevent it from being accidentally leaked to discord, creating a `.clean_url` property (and documenting it, and the `.url` attribute) creates more maintainable code, and helps with pieces of code which communicate with users in discord.

PR #2783, opened 2019-06-21
[Core] Multiple mod admin roles

DiscordLiz:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This allows multiple mod and admin roles to be used.

I've updated everywhere which uses it directly.

If cog developers are expected to be allowed to interact with the bot's settings directly, this is also a breaking change.

### DIFF
--- a/redbot/core/commands/requires.py
+++ b/redbot/core/commands/requires.py
@@ -126,13 +126,13 @@ class PrivilegeLevel(enum.IntEnum):
         # The following is simply an optimised way to check if the user has the
         # admin or mod role.
         guild_settings = ctx.bot.db.guild(ctx.guild)
-        admin_role_id = await guild_settings.admin_role()
-        mod_role_id = await guild_settings.mod_role()
+        admin_role_ids = set(await guild_settings.admin_role())
+        mod_role_ids = set(await guild_settings.mod_role())
         is_mod = False
         for role in ctx.author.roles:
-            if role.id == admin_role_id:
+            if role.id in admin_role_ids:
                 return cls.ADMIN
-            elif role.id == mod_role_id:
+            elif role.id in mod_role_ids:
                 is_mod = True
         if is_mod:
             return cls.MOD
--- a/redbot/core/commands/requires.py
+++ b/redbot/core/commands/requires.py
@@ -127,12 +127,12 @@ class PrivilegeLevel(enum.IntEnum):
         # admin or mod role.
         guild_settings = ctx.bot.db.guild(ctx.guild)
 
-        snowflakes = ctx.author._roles  # DEP-WARN
+        member_snowflakes = ctx.author._roles  # DEP-WARN
         for snowflake in await guild_settings.admin_role():
-            if snowflakes.has(snowflake):  # DEP-WARN
+            if member_snowflakes.has(snowflake):  # DEP-WARN
                 return cls.ADMIN
         for snowflake in await guild_settings.mod_role():
-            if snowflakes.has(snowflake):  # DEP-WARN
+            if member_snowflakes.has(snowflake):  # DEP-WARN
                 return cls.MOD
 
         return cls.NONE
--- a/redbot/core/commands/requires.py
+++ b/redbot/core/commands/requires.py
@@ -126,16 +126,14 @@ class PrivilegeLevel(enum.IntEnum):
         # The following is simply an optimised way to check if the user has the
         # admin or mod role.
         guild_settings = ctx.bot.db.guild(ctx.guild)
-        admin_role_ids = set(await guild_settings.admin_role())
-        mod_role_ids = set(await guild_settings.mod_role())
-        is_mod = False
-        for role in ctx.author.roles:
-            if role.id in admin_role_ids:
+
+        snowflakes = ctx.author._roles  # DEP-WARN
+        for snowflake in await guild_settings.admin_role():
+            if snowflakes.has(snowflake):  # DEP-WARN
                 return cls.ADMIN
-            elif role.id in mod_role_ids:
-                is_mod = True
-        if is_mod:
-            return cls.MOD
+        for snowflake in await guild_settings.mod_role():
+            if snowflakes.has(snowflake):  # DEP-WARN
+                return cls.MOD
 
         return cls.NONE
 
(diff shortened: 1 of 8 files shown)


mikeshardmind:
I don't think we ever publicly documented the bot's db. I'm inclined to say that even if this may break someone, they should be using the exposed documented functions instead of this. I'm happy to review it later, and open to discussing delaying this for a breaking release if someone disagrees with that stance.

mikeshardmind:
A note for anyone going to test this: Since it modifies the bot's db including the types of things in it, I suggest testing with a brand new instance or a copy of a prior one.

redbot/core/core_commands.py
```diff
@@ -739,18 +738,46 @@ class Core(commands.Cog, CoreLogic):
...
     @commands.guild_only()
-    async def adminrole(self, ctx: commands.Context, *, role: discord.Role):
-        """Sets the admin role for this server"""
-        await ctx.bot.db.guild(ctx.guild).admin_role.set(role.id)
-        await ctx.send(_("The admin role for this guild has been set."))
+    async def addadminrole(self, ctx: commands.Context, *, role: discord.Role):
+        """ Adds an admin role for this guild """
+        async with ctx.bot.db.guild(ctx.guild).admin_role() as roles:
+            if role.id in roles:
+                return await ctx.send(_("This is already an admin role."))
```
Flame442: This line and L756 probably should be "That role is already a[n] <admin/mod> role.".

redbot/core/core_commands.py
```diff
@@ -739,18 +738,46 @@ class Core(commands.Cog, CoreLogic):
...
+            roles.append(role.id)
+        await ctx.send(_("That role was added as an admin role for this guild."))
```
Flame442: This line and L758, "That role was added" doesn't sound very good, maybe a simple "Role added successfully."?

redbot/core/core_commands.py
```diff
@@ -739,18 +738,46 @@ class Core(commands.Cog, CoreLogic):
...
+        await ctx.send(_("That role was added as a mod role for this guild."))
+
+    @_set.command(aliases=["remadmindrole"])
+    @checks.guildowner()
+    @commands.guild_only()
+    async def removeadminrole(self, ctx: commands.Context, *, role: discord.Role):
+        """ Removes an admin role for this guild """
+        async with ctx.bot.db.guild(ctx.guild).admin_role() as roles:
+            if role.id not in roles:
+                return await ctx.send(_("This was not an admin role."))
```
Flame442: This line and L778 probably should be "That role was already not a[n] <admin/mod> role".

Flame442 (changes requested):
I haven't had a chance to run this on a bot yet, however I looked over the strings a bit and have some changes to suggest.

-All doc strings: The spaces in between the text and the quotes should be removed for consistency with the rest of red
-All doc strings: Missing periods at the end.
-I remember a PR at some point added consistency to all of the del/delete/rem/remove commands by aliasing all of that, `[p]set remove<mod/admin>role` might need to get the same treatment.

redbot/core/core_commands.py
DiscordLiz: I went with something different here, let me know if it sounds better to you.

redbot/core/commands/requires.py
```diff
@@ -126,13 +126,13 @@ class PrivilegeLevel(enum.IntEnum):
         # The following is simply an optimised way to check if the user has the
         # admin or mod role.
         guild_settings = ctx.bot.db.guild(ctx.guild)
-        admin_role_id = await guild_settings.admin_role()
-        mod_role_id = await guild_settings.mod_role()
+        admin_role_ids = set(await guild_settings.admin_role())
```
mikeshardmind: If you could change this section to do the role lookup using `ctx.author._roles.has(role_id)` and not convert to a set, short circuiting on admin roles first, That would be great. Additionally, please add `# DEP-WARN` to each line using this. We'd like to get the most performance possible out of this section, but we do need to take caution with the internals here.

redbot/core/core_commands.py
```diff
@@ -705,15 +705,14 @@ class Core(commands.Cog, CoreLogic):
...
+                admin_role_ids = await ctx.bot.db.guild(ctx.guild).admin_role()
+                admin_roles = [r for r in guild.roles if r.id in admin_role_ids]
+                admin_roles_str = humanize_list(admin_roles) if admin_roles else "Not Set."
+                mod_roles = [r for r in guild.roles if r.id in admin_role_ids]
+                mod_roles_str = humanize_list(mod_roles) if mod_roles else "Not Set."
                 prefixes = await ctx.bot.db.guild(ctx.guild).prefix()
-                guild_settings = _("Admin role: {admin}\nMod role: {mod}\n").format(
-                    admin=admin_role, mod=mod_role
+                guild_settings = _("Admin roles: {admin}\nMod roles: {mod}\n").format(
+                    admin=admin_roles_str, mod=mod_roles_str
```
mikeshardmind, line 718: This section looks fine at a glance, but it warrants adding pagination to the send below to prevent something really stupid with large number of roles.

mikeshardmind (changes requested):
I need to do a bit more actual testing with this (not a full review), but it's what I've noticed so far. 

Definitely liking most of what I'm seeing here so far.

redbot/core/commands/requires.py
```diff
@@ -126,16 +126,14 @@ class PrivilegeLevel(enum.IntEnum):
         # The following is simply an optimised way to check if the user has the
         # admin or mod role.
         guild_settings = ctx.bot.db.guild(ctx.guild)
-        admin_role_id = await guild_settings.admin_role()
-        mod_role_id = await guild_settings.mod_role()
-        is_mod = False
-        for role in ctx.author.roles:
-            if role.id == admin_role_id:
+
+        snowflakes = ctx.author._roles  # DEP-WARN
```
mikeshardmind: I don't mean to nitpick too much here, but seeing it side by side with how you named the other instance of this, the other form is much more legible.

```suggestion
        member_snowflakes = ctx.author._roles  # DEP-WARN
```

mikeshardmind: I don't mean to nitpick too much here, but seeing it side by side with how you named the other instance of this, the other form is much more legible.

```suggestion
        member_snowflakes = ctx.author._roles  # DEP-WARN
```

mikeshardmind (changes requested):

Flame442 (dismissed):
While I still don't love the "this" in "This role is already an <admin/mod> role.", the changes I requested have been addressed.

mikeshardmind (dismissed):
Thanks @DiscordLiz for the quick follow up on issues discussed both here and in Discord.

:tada:

mikeshardmind (approved):

PR #2787, opened 2019-06-22
[Audio] Add equalizer

aikaterna:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

An audio equalizer (`[p]eq`) has been added, thanks to Devoxin#0001's work on his own music player. I am usually reluctant to add commands that add more than roughly 5 emojis at once because of the time it takes applying and removing reactions and how that can lead to rate limiting on large bots that are very active, but this concern shouldn't apply to most of the users and it seemed one of the best options for user interface in regards to setting so many varied functions. For the folks that don't like interacting with reactions, there is also a command to use to set the bands directly instead.

* 15 different bands can be adjusted in increments of 0.1, from -0.25 (completely removing the frequency) to 1.0 (0.25 is doubling it). 
* Bands to adjust include 25, 40, 63, 100, 160, 250, 400, 630, 1k, 1.6k, 2.5k, 4k, 6.3k, 10k, and 16k Hz.
* [p]eq subcommands include: `save` to save the current eq, `list` to list saved eq settings, `delete` to remove a setting, `load` to load in a saved eq set, and `set` for setting bands directly
* _clear_react has been modified to remove the bot's control reactions one by one when it lacks permissions to clear all reactions instead of leaving the reactions on the message.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -30,11 +30,12 @@ from redbot.core.utils.menus import (
 )
 from redbot.core.utils.predicates import MessagePredicate, ReactionPredicate
 from urllib.parse import urlparse
+from .equalizer import Equalizer
 from .manager import ServerManager
 
 _ = Translator("Audio", __file__)
 
-__version__ = "0.0.9"
+__version__ = "0.0.10"
 __author__ = ["aikaterna"]
 
 log = logging.getLogger("red.audio")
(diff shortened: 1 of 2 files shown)


mikeshardmind (approved):
Might be worth modifying the eq set command to use the dict converter if people get tired of adding a band at a time for an easyish way to add multiple at once, but the initial work here is great, and that can be added on later.

PR #2790, opened 2019-06-22
[Help] Prevent spamming when a user blocks the bot

MeatyChunks:
Currently the bot sends an error message for each page of help, this should make it only send once.


mikeshardmind (approved):
Thanks!

PR #2791, opened 2019-06-22
Added a warning about modifying the PATH variable

DevilXD:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature
- [x] Docs only

Modifying the PATH variable on Windows usually requires a restart to take an effect. Even with the option checked and a completely new command prompt, it might say that the `python` executable couldn't be found.

This PR adds a warning about this possibility, guiding the user to restart their computer before proceeding.

### DIFF
--- a/docs/install_windows.rst
+++ b/docs/install_windows.rst
@@ -39,8 +39,6 @@ Manually installing dependencies
 .. note:: Please make sure that the box to add Python to PATH is CHECKED, otherwise
           you may run into issues when trying to run Red.
 
-.. attention:: A restart might be required for the PATH changes to take an effect.
-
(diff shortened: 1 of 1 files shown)


Tobotimus:
In my experience simply restarting the command prompt or doing `refreshenv` updates the path. Has this not been the case for you?

mikeshardmind:
`refreshenv` is only available to those who use the chocolatey install. This isn't a windows builtin and this section is for the manual install.

Tobotimus:
Hah, I didn't know that, I just assumed it was built in

Tobotimus (approved):

PR #2794, opened 2019-06-23
[Filter] Make name filtering behavior consistent

### LINKED ISSUE #2740: Filtering issue

Someone changed his name to Night and it filtered the name because the word Nig is on the filter list. so basically it shouldn't filter the persons name because it was Night and not Nig.

Now this issue is solved in the filtering system when it comes to proper chat in a channel, where someone can say for example assistance but not ass when ass is on the filter list. but assistance consisting the word ass is possible. (like i stated that works correctly)

however if you do the same for nicknames (!filter names) it will filter the name if a word consists of the word that is in the filter list. so Nig is filtered and night is not. however changing the name to Night will filter that name because Nig is in Night.

Hope that made sense.

For nickname Night not be filtered because that actual word is not on the filter list.

#### What actually happened?

The word night got filtered because the word Nig which is part of Night is on the filter list.

1. !filter add Nig
2. !filter names (to enable the filtering names option)
3. Try to change your name to Night
4. See how the bot changes your name because nig is on the list and part of the word night.

mikeshardmind:
### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

Resolves #2740 

  - Uses the updated filter check
  - This is also a performance gain on large servers with filter names
  enabled.

### DIFF
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -448,13 +448,12 @@ class Filter(commands.Cog):
         if not await self.settings.guild(member.guild).filter_names():
             return
 
-        word_list = await self.settings.guild(member.guild).filter()
-        for w in word_list:
-            if w in member.display_name.lower():
-                name_to_use = await self.settings.guild(member.guild).filter_default_name()
-                reason = _("Filtered nickname") if member.nick else _("Filtered name")
-                try:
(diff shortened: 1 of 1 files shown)


Tobotimus (approved):
Tested and this works :+1:

PR #2795, opened 2019-06-23
Various Config and Mongo Driver fixes

Tobotimus:
- Fixes defaults being mixed into custom groups above the document level when doing `Group.all()`
- Fixes `Config.clear_all()` with Mongo driver
- Fixes `Group.set()` with Mongo driver on custom groups above the document level
- Fixes `IdentifierData.custom_group_data` being set to the wrong thing in `BaseDriver.import/export_data` (although this was an inconsequential bug)

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -1,16 +1,14 @@
 import logging
 import collections
 from copy import deepcopy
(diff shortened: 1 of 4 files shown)


mikeshardmind (approved):
:+1:

PR #2796, opened 2019-06-24
Kill JsonIO and fix JSON driver caching issues

### LINKED ISSUE #2793: [Config] Cache is set to new object even if object isn't JSON serializable
# Other bugs

Set object that isn't JSON serializable.
```py
config.someval.set({231, 321})
```

Get exception about non-serializable object and Config's cache to stay intact.

#### What actually happened?

Config's cache got updated with a non-serializable object.

Assuming `config` contains `Config` instance, use this:
```py
try:
    await config.someval.set({231, 321})
except TypeError as e:
    print(str(e))
    print("yeah, we all know it's not JSON serializable, shut up")
print(await config.someval())
```

You should get this printed:
```py
Object of type set is not JSON serializable
yeah, we all know, it's not JSON serializable, shut up
{321, 231}
```

Tobotimus:
I admit I got a bit carried away and killed my least favourite module in all of Red!

This PR fixes #2793 with the solution described in my comment on that issue.

The reasons for killing JsonIO were mainly:
- The JSON driver should be managing its lock, rather than the a JsonIO class (this is really the one which defeats the purpose of JsonIO)
- It's an unnecessary level of abstraction
- All of the names of the `_threadsafe` methods were misleading

Other changes include:
- Removing a `deepcopy()` on all of a cog's data every time a JSON is saved (oof)
- Using `pickle.loads(pickle.dumps())` over `deepcopy()` everywhere in Config and the JSON driver


Tobotimus:
I've updated the header of this PR with more info :smiley:

mikeshardmind (approved):
:+1:

PR #2798, opened 2019-06-24
Fix duplicate commands in fuzzy help

### LINKED ISSUE #2797: [Core] Fuzzy command search - commands show up multiple times if they have aliases
# Other bugs

Misspell a command with aliases (as in, coded aliases, not the ones from alias cog) with fuzzy command search enabled.

Fuzzy command search to show proper command once.

#### What actually happened?

Fuzzy command search showed the proper command twice.

Use these commands with fuzzy command search enabled (misspell in second command intended)
```
[p]load general
[p]stopwatdc
```
Get this embed as a response:
image

Tobotimus:
Resolves #2797.

The `commands` argument of `fuzzy_command_search` should now take a `set` of `commands.Command` objects instead of some other sequence.

### DIFF
--- a/redbot/core/commands/help.py
+++ b/redbot/core/commands/help.py
@@ -450,7 +450,7 @@ class RedHelpFormatter:
         """
         Sends an error, fuzzy help, or stays quiet based on settings
         """
-        coms = [c async for c in self.help_filter_func(ctx, ctx.bot.walk_commands())]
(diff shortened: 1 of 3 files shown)


Tobotimus:
This now also makes sure the fuzzy command search uses the help command's filter function, even when fuzzy results are shown after simply mistyping a command name after the prefix.

mikeshardmind (approved):

PR #2809, opened 2019-06-27
[Utils] Improved class-based menu utilities

Tobotimus:
I did a thing. All of these additions are best understood in the docs, which I have built [here](https://red-discordbot-toby.readthedocs.io/en/v3-better_menu/framework_utils.html#module-redbot.core.utils.menus).

The old menu remains, however its usage in core is limited to the audio cog, for now. I've decided to leave audio alone this time around :smile:

~~The docs build depends on #2806.~~


redbot/core/commands/help.py
```diff
@@ -556,13 +555,7 @@ def parse_command(ctx, help_for: str):
...
                         )
         else:
-            # Specifically ensuring the menu's message is sent prior to returning
-            m = await (ctx.send(embed=pages[0]) if embed else ctx.send(pages[0]))
-            c = menus.DEFAULT_CONTROLS if len(pages) > 1 else {"\N{CROSS MARK}": menus.close_menu}
-            # Allow other things to happen during menu timeout/interaction.
-            asyncio.create_task(menus.menu(ctx, pages, c, message=m))
-            # menu needs reactions added manually since we fed it a messsage
-            menus.start_adding_reactions(m, c.keys())
+            await menus.PagedMenu.send_and_return(ctx, pages=pages)
```
mikeshardmind, line 678: i need to look closer at this part before merge

redbot/core/utils/menus.py
```diff
@@ -1,29 +1,1262 @@
...
+        arrows_always: bool = False,
+        **kwargs,
+    ) -> None:
+        self._pages = list(pages)
+        self._cur_page = first_page
+        self._footer_text = footer_text
+
+        if arrows_always is False and len(self._pages) == 1:
+            initial_emojis = ["❌"]
+        else:
```
mikeshardmind: logic here doesnt appear to pass the eye test. the if doesnt check that there arent emojis in constructer. this should determine the default. This should also probably use kwargs.pop with the use of super below

mikeshardmind (commented):
not a full review, just looking at parts right now based on a change to help as a result, im seeing a few things which dont look correct right now, but need to look in more depth to be sure arent.

redbot/core/utils/menus.py
Tobotimus: Thanks for pointing this out - should be fixed up now.

Tobotimus:
~~I think I need to do some minor tweaks on inheritance of handlers and the way that affects the initial reaction order. Will look into it soon.~~ Fixed

NeuroAssassin:
I found a slight docs issue:

image
In the line where the role is added to the user (`await member.add_roles(role, "Adding reactrole")`), the string is not casted to the reason argument, and is therefore interpreted by discord.py that it is a role, resulting in that eventually erroring and the embed not being deleted when the role is added.

Stonedestroyer:
I have this code,

```py
        async def fetchanime(user, selection):
            nonlocal number
            number = selection
            await self.id_lookup(ctx, cmd, number)

        menu_instance = await PagedOptionsMenu.send_and_wait(
            ctx,
            timeout=30,
            pagenum_in_footer=1,
            footer_text=f"Requested by: {str(ctx.author)}",
            options=animelist,
            callback=fetchanime,
        )
```
I noticed a problem where someone spams a button two times my event will trigger multiple times even though menu should exit and just take first event, is there any way I can avoid this or is this as designed? I have not touched exit_on_selection. 

The only thing I edited is _after_timeout to pass rather than delete the menu on timeout.

redbot/core/utils/menus.py
```diff
@@ -1,29 +1,1247 @@
...
+
+    def __init__(
+        self,
+        *,
+        ctx: Optional[commands.Context] = None,
+        channel: Optional[discord.abc.Messageable] = None,
+        bot: Optional["Red"] = None,
+        controller_ids: Optional[Set[int]] = None,
+        initial_emojis: Optional[Sequence[str]] = None,
+        timeout: float = 30.0,
```
TrustyJAID: To keep the typing this should be `Optional[float]` since we can also pass `None` as a valid type.

I am also certain this will be a big topic for discussion as it already has been. Technically changing the default behavior to delete the message after 30 seconds of inactivity breaks all current cog creators expectations. While the default could be changed to `None` I think leaving it forces cog creators to take a look at their code and simply add the optional timeout kwarg wherever they use menu if they want to change it. That said this will also break regular users expectations if they're not aware of the code and I could see them being upset that the bots messages are automatically deleting when the cog creator has not updated to use this new menu system yet.

redbot/cogs/economy/leaderboard_menu.py
```diff
@@ -0,0 +1,111 @@
...
+
+        if self._cur_page == len(self._pages) - 1:
+            self._pages.append(await self._format_page())
+
+        await super().next_page(payload=payload)
+
+    async def _format_page(self) -> str:
+        accounts = self._accounts[self._cur_rank : self._cur_rank + 10]
+
+        bal_colwidth = len(str(accounts[0][1]["balance"])) + 5
```
TrustyJAID, line 65: If there's exactly 10 people in the leaderboard this ends up with a IndexError when trying to view more pages.

redbot/core/utils/menus.py
Tobotimus: Yeah, I've talked to Draper about this and I will add another kwarg which changes the timeout action. The default will probably be clearing reactions (or removing the bot's reactions if permissions are missing), and the message deletion will probably be an option too.

This might slightly break expectations with the old menu, but I think it's an improvement. When the menu times out, it should give *some* sort of indication that the menu's no longer active. Removing reactions should be enough.

Drapersniper:
image

Something i found with `PagedOptionsMenu`  is that the Arrows are always avaliable, even when there's only 1 page, from the docs my understanding is that `arrows_always` should be off by default, this is the code.
```py
await PagedOptionsMenu.send_and_wait(
            ctx,
            timeout=60,
            pagenum_in_footer=True,
            options=folder_page_list,
            title=title,
            embed=True,
            footer_text=footer,
        )
```

DevilXD:
Will this find its way into the v3.2 release?

Flame442:
Most likely not as it is still a work in progress. It isn't a breaking change so it could make it in 3.2.1.

Stonedestroyer:
The documentation and understanding how to subclass them are very easy. One issue I found, correct me if I am wrong but the old menu wouldn't delete itself after the timeout where this one deletes itself. 

What I have done is I have added a delete_menu bool that is by default false and can be set to true to delete the menu after timeout. There is some scenarios where you would want a reaction timeout for GC purposes but would want the menu to still stay.

Stonedestroyer:
pagenum_in_footer is documented but not implemented.

Stonedestroyer:
Setting options_per_page=20 and then spawning the menu results in an error 
```py
    raise Forbidden(r, data)
discord.errors.Forbidden: 403 Forbidden (error code: 30010): Maximum number of reactions reached (20)
```
This is due to it not taking in account the default reactions it adds as well.

Drapersniper:
propose we close this in favour of #4039

Drapersniper:
No backlash, #4039 has been merged.

So will be closing this PR, in favour of using d.py menus doing forward.

If you haven't seen d.py menus <https://github.com/Cog-Creators/Red-DiscordBot/pull/4062> have some examples as well as <https://ptb.discordapp.com/channels/133049272517001216/133251234164375552/731246605113753630> and <https://github.com/Rapptz/discord-ext-menus>

PR #2812, opened 2019-06-27
[Audio] Fix config set in shuffle and repeat commands

### LINKED ISSUE #2811: [Audio] Shuffle command shows error, but changes the setting anyway
# Command bugs

#### Command name

`[p]shuffle`

Audio

Not sure, but I'm sure that this error should mean that the setting shouldn't be changed.

#### What actually happened?

The setting got changed despite the error

Use these commands with a user that can't instaskip (and with DJ role disabled) and see that the "Shuffle" setting gets changed despite the error:
```
[p]audioset settings
[p]shuffle
[p]audioset settings
```
Here's screenshot:
image

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Resolves #2811 and also the same thing in repeat command.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -2828,12 +2828,13 @@ class Audio(commands.Cog):
                 return await self._embed_msg(
                     ctx, _("You must be in the voice channel to toggle repeat.")
                 )
+
         repeat = await self.config.guild(ctx.guild).repeat()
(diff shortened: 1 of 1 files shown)


Jackenmen:
_data_check doesn't do its thing, when you moved set commands under it

PredaaA:
@jack1142 Fixed on the last commit. (Thanks to aika)

Tobotimus (approved):

PR #2813, opened 2019-06-28
[Audio] Update equalizer permissions

aikaterna:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Changed the equalizer permissions to be more in line with the rest of Audio's permission mechanics. The DJ role is now used to restrict access to the eq instead of using a mod/admin bot role or the administrator guild permission.

* All users will be able to manage the eq if DJ role is off and there are no restrictions on subcommand use.
* If the DJ role is on, and the user does not have the role, they will only be able to view the current eq.
* Deleting eq presets now follows the same permissions style as playlist deletion. If the person is a bot mod/admin, bot owner, or other generally exempt user, they will be able to delete any eq setting. There is now an author field on eq settings, and if a pleb tries to delete someone else's setting they will not be able to.
* If someone was using the dev version before these changes were made and they have saved eq settings, there are a few lines to handle that previous eq data storage file arrangement. Since there is no author, the mods/admins will need to be the ones to delete those settings as they will be the only user with authorization to do so.
* The eq list command output has changed to show author and has a slightly different format.

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -840,44 +840,63 @@ class Audio(commands.Cog):
     @commands.guild_only()
     @commands.cooldown(1, 15, discord.ext.commands.BucketType.guild)
     @commands.bot_has_permissions(embed_links=True, add_reactions=True)
-    @checks.mod_or_permissions(administrator=True)
     async def eq(self, ctx):
         """Equalizer management."""
         if not self._player_check(ctx):
             return await self._embed_msg(ctx, _("Nothing playing."))
+        dj_enabled = await self.config.guild(ctx.guild).dj_enabled()
         player = lavalink.get_player(ctx.guild.id)
         eq = player.fetch("eq", Equalizer())
         reactions = ["◀", "⬅", "⏫", "🔼", "🔽", "⏬", "➡", "▶", "⏺", "ℹ"]
         await self._eq_msg_clear(player.fetch("eq_message"))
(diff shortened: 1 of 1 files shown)


PredaaA (approved):
I just tested this PR, all changes that are listed in the description works as intended.

NIXC:
Had a look through as well, couldn't see anything out of the ordinary. Looks good to me.

redbot/cogs/audio/audio.py
```diff
@@ -840,44 +840,63 @@ def _restart_connect(self):
...
                 pass
+        else:
+            for reaction in reactions:
+                try:
+                    await eq_message.add_reaction(reaction)
+                except discord.errors.NotFound:
+                    pass
 
-        await self._eq_interact(ctx, player, eq, eq_message, 0)
+        eq_msg_with_reacts = await ctx.fetch_message(eq_message.id)
```
Tobotimus, line 866: Hmm, I'm not sure that this extra API call is necessary, what issues have you found when you use the cached message object?

Tobotimus (commented):
I've just got one question for now :)

redbot/cogs/audio/audio.py
aikaterna: The eq visualizer requires having the message with the reactions or it enters a loop.

Tobotimus (approved):
Alright, I'm happy with this then :+1:

Issue #2800, opened 2019-06-24
[Warnings] Option to disable DMing warns, option to warn in guild channel


Jackenmen:
# Feature request

#### Select the type of feature you are requesting:

- [ ] Cog
- [x] Command
- [ ] API functionality

#### Describe your requested feature

I want to request 2 things in this issue, first is an option to disable sending warnings in DM to warned user and second is option to warn a user (i.e. ping them) in set guild channel and/or just warn that user in channel where the `[p]warn` command was issued.

DevilXD:
> option to warn a user (i.e. ping them) in set guild channel and/or just warn that user in channel where the [p]warn command was issued.

Could you please clarify on this part? It's kinda hard to understand what the feature should be from this description.

Is it a channel where all warnings should be going to? If so, the warnings cog adds a modlog type you can enable to log those into the modlog channel already.

Jackenmen:
But modlog is generally private, it doesn't "ping" and you can't have multiple modlog channels, I just want the bot to say it warned a user in the channel where the `[p]warn` command was issued OR in a set channel. To be more specific, it should do `ctx.send(f"{user.mention} was warned for spamming")` (or `channel.send` if it's set)

DevilXD:
Right, so separate channel which can be made public. Gotcha.

PR #2814, opened 2019-06-28
Features requested for warning

Ianardo-DiCaprio:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This adds commands requested in #2800. It allows to toggle warns being sent to DM's, allows to toggle warns being sent to a channel on a server and has a command to set the channel the warns get sent to, if no channel is set it gets sent to the channel the [p]warn command was used in.


redbot/cogs/warnings/warnings.py
```diff
@@ -24,7 +25,14 @@
 class Warnings(commands.Cog):
     """Warn misbehaving users and take automated actions."""
 
-    default_guild = {"actions": [], "reasons": {}, "allow_custom_reasons": False}
+    default_guild = {
+        "actions": [],
+        "reasons": {},
+        "allow_custom_reasons": False,
+        "toggle_dm": True,
```
NeuroAssassin, line 31: IMO, this should default to False so that guild owners/admins can choose whether or not for this to be enabled.

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,49 @@ def __init__(self, bot: Red):
...
             await ctx.send(_("Custom reasons have been disabled."))
 
+    @warningset.command()
+    @commands.guild_only()
+    async def toggledm(self, ctx: commands.Context):
+        """Toggle warns being sent to the users DM's"""
+        guild = ctx.guild
+        toggle = await self.config.guild(guild).toggle_dm()
+        await self.config.guild(guild).toggle_dm.set(not toggle)
+        toggle = await self.config.guild(ctx.guild).toggle_dm()
```
NeuroAssassin: You have the toggle variable already, so you should probably use `not toggle` instead of making another Config call.

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,49 @@ def __init__(self, bot: Red):
...
+        if toggle == True:
+            await ctx.send("Warns will now be sent to users DM's")
+        else:
+            await ctx.send("Warns will not be sent to users DM's")
+
+    @warningset.command()
+    @commands.guild_only()
+    async def warnchannel(self, ctx: commands.Context, channel=None):
```
NeuroAssassin: You can use type casting for this to make it require a `discord.TextChannel`, so then you don't have to convert it later.

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,49 @@ def __init__(self, bot: Red):
...
+            return await ctx.send("Couldn't find that channel.")
+
+    @warningset.command()
+    @commands.guild_only()
+    async def togglechannel(self, ctx: commands.Context):
+        """Toggle if the warn should be sent to a channel"""
+        guild = ctx.guild
+        toggle = await self.config.guild(guild).toggle_channel()
+        await self.config.guild(guild).toggle_channel.set(not toggle)
+        toggle = await self.config.guild(ctx.guild).toggle_channel()
```
NeuroAssassin: Same thing with above, you can use `not toggle` instead of calling Config again.

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +353,41 @@ def __init__(self, bot: Red):
...
-                _("You have received a warning in {guild_name}.").format(
-                    guild_name=ctx.guild.name
-                ),
-                embed=em,
-            )
-        except discord.HTTPException:
-            pass
+            warn_channel = await self.config.guild(ctx.guild).warn_channel()
+            if warn_channel == None:
+                channel == ctx.channel
```
NeuroAssassin: Is this supposed to be a `=`?

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +353,41 @@ def __init__(self, bot: Red):
...
+            else:
+                channel == warn_channel
```
NeuroAssassin: Is this supposed to be a `=`?

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +353,41 @@ def __init__(self, bot: Red):
...
+                await channel.send(
+                    _("{user} has been warned.").format(user=user.mention), embed=em
+                )
+        if dm and toggle_channel == False:
+            ctx.send("You have warned the user but you have DM and channel warns toggled off.")
```
NeuroAssassin: Forgot an await here.

NeuroAssassin (changes requested):
I've noticed a couple things while taking a quick look at your PR.  I've made a list of them below.

redbot/cogs/warnings/warnings.py
Jackenmen: I agree, though warnings cog was acting as if it was set to `True`, so this would change that behavior and it should be included in changelog, if that's going to be changed to `False` as default.

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +346,43 @@ def __init__(self, bot: Red):
...
-            await user.send(
-                _("You have received a warning in {guild_name}.").format(
-                    guild_name=ctx.guild.name
-                ),
-                embed=em,
+            warn_channel = await self.config.guild(ctx.guild).warn_channel()
+            if warn_channel == None:
+                channel = ctx.channel
+            else:
+                channel == warn_channel
```
Jackenmen: Second `==` still not changed to `=`, so I'm mentioning in case you didn't see.

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +346,43 @@ def __init__(self, bot: Red):
...
+                await channel.send(
+                    _("{user} has been warned.").format(user=user.mention), embed=em
+                )
+        if dm and toggle_channel == False:
```
Jackenmen: I would remove this whole section and go back to how it was handled in this line, because there's no need to inform the user each time that DM and channel warns are toggled off:
https://github.com/Cog-Creators/Red-DiscordBot/pull/2814/files#diff-bd66023271557007376afcf648f3b3ffL341

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +346,43 @@ def __init__(self, bot: Red):
 
         await warning_points_add_check(self.config, ctx, user, current_point_count)
         try:
+            dm = await self.config.guild(ctx.guild).toggle_dm()
+            if dm == True:
```
Jackenmen: You can change it to just:
```suggestion
            if dm:
```

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +346,43 @@ def __init__(self, bot: Red):
...
+                )
+                em.add_field(name=_("Points"), value=str(reason_type["points"]))
+                await user.send(
+                    _("You have received a warning in {guild_name}.").format(
+                        guild_name=ctx.guild.name
+                    ),
+                    embed=em,
+                )
+                await ctx.send(_("User {user} has been warned in their DM's.").format(user=user))
+        except discord.HTTPException:
```
jack1142: This `try` block is too wide, currently it may catch `discord.HTTPException` for `ctx.send()`

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +346,43 @@ def __init__(self, bot: Red):
...
                 description=reason_type["description"],
             )
             em.add_field(name=_("Points"), value=str(reason_type["points"]))
-            await user.send(
-                _("You have received a warning in {guild_name}.").format(
-                    guild_name=ctx.guild.name
-                ),
-                embed=em,
+            warn_channel = await self.config.guild(ctx.guild).warn_channel()
+            if warn_channel == None:
```
```suggestion
            if warn_channel is None:
```
You should just use identity check, because it's singleton.

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +346,43 @@ def __init__(self, bot: Red):
...
+                    ),
+                    embed=em,
+                )
+                await ctx.send(_("User {user} has been warned in their DM's.").format(user=user))
+        except discord.HTTPException:
+            await ctx.send(
+                "The warning couldn't be sent to the user because the user has DM's from server members turned off."
+            )
+        toggle_channel = await self.config.guild(ctx.guild).toggle_channel()
+        if toggle_channel == True:
```
Jackenmen: You can change it to just:
```suggestion
        if toggle_channel:
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +83,43 @@ def __init__(self, bot: Red):
...
 
+    @warningset.command()
+    @commands.guild_only()
+    async def toggledm(self, ctx: commands.Context):
+        """Toggle warns being sent to the users DM's"""
+        guild = ctx.guild
+        toggle = await self.config.guild(guild).toggle_dm()
+        await self.config.guild(guild).toggle_dm.set(not toggle)
+        toggle = not toggle
+        if toggle == True:
```
Jackenmen: You can change it to just:
```suggestion
        if toggle:
```
Or even not assign line above and just do `if not toggle:`

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +83,43 @@ def __init__(self, bot: Red):
...
+            await ctx.send("Warns will now be sent to users DM's")
+        else:
+            await ctx.send("Warns will not be sent to users DM's")
+
+    @warningset.command()
+    @commands.guild_only()
+    async def warnchannel(self, ctx: commands.Context, channel: discord.TextChannel):
+        """Set the channel that warns get set to"""
+        guild = ctx.guild
+        try:
```
Jackenmen: This `try` block doesn't seem to be needed anymore.

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +83,43 @@ def __init__(self, bot: Red):
...
+
+    @warningset.command()
+    @commands.guild_only()
+    async def togglechannel(self, ctx: commands.Context):
+        """Toggle if the warn should be sent to a channel"""
+        guild = ctx.guild
+        toggle = await self.config.guild(guild).toggle_channel()
+        await self.config.guild(guild).toggle_channel.set(not toggle)
+        toggle = not toggle
+        if toggle == True:
```
Jackenmen: You can change it to just:
```suggestion
        if toggle:
```
Or even not assign line above and just do `if not toggle:`

Jackenmen (changes requested):

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +342,39 @@ def __init__(self, bot: Red):
...
-                embed=em,
-            )
-        except discord.HTTPException:
-            pass
+            warn_channel = await self.config.guild(ctx.guild).warn_channel()
+            if warn_channel is None:
+                channel = ctx.channel
+            else:
+                channel = warn_channel
+                await channel.send(
```
Jackenmen: I think this could work nicely if you would unindent this `channel.send()` once, so that it gets sent, when warn channel is not set too (see my proposed addition in comment to `[p]warningset warnchannel` for additional context)

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +342,39 @@ def __init__(self, bot: Red):
...
+                    ),
+                    embed=em,
+                )
+        except discord.HTTPException:
+            await ctx.send(
+                "The warning couldn't be sent to the user because the user has DM's from server members turned off."
+            )
+        await ctx.send(_("User {user} has been warned in their DM's.").format(user=user))
+        toggle_channel = await self.config.guild(ctx.guild).toggle_channel()
+        if toggle_channel:
```
Jackenmen, line 372: You should add some response if toggle is set to disabled (as in add `else` block after this `if`), something like there was before, for example:
```py
else:
     await ctx.send(_("User {user} has been warned.").format(user=user))
```

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +342,39 @@ def __init__(self, bot: Red):
...
+                    _("You have received a warning in {guild_name}.").format(
+                        guild_name=ctx.guild.name
+                    ),
+                    embed=em,
+                )
+        except discord.HTTPException:
+            await ctx.send(
+                "The warning couldn't be sent to the user because the user has DM's from server members turned off."
+            )
+        await ctx.send(_("User {user} has been warned in their DM's.").format(user=user))
```
Jackenmen: In my opinion, this message is unnecessary (as user that used this command should already know what settings he has), but if you want to keep it, put this in `else` block and check if DMs are enabled (cause otherwise, this message will be sent even if the warning couldn't be sent to the user or when sending DMs is disabled in cog's settings):
```suggestion
        else:
            if dm:
                await ctx.send(_("User {user} has been warned in their DMs.").format(user=user))
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +83,39 @@ def __init__(self, bot: Red):
...
+        await self.config.guild(guild).toggle_dm.set(not toggle)
+        toggle = not toggle
+        if toggle:
+            await ctx.send("Warns will now be sent to users DM's")
+        else:
+            await ctx.send("Warns will not be sent to users DM's")
+
+    @warningset.command()
+    @commands.guild_only()
+    async def warnchannel(self, ctx: commands.Context, channel: discord.TextChannel):
```
Jackenmen: I think that it should be allowed to reset this setting (as in, set it to `None` when channel is not provided), so that bot would use `ctx.channel` in `[p]warn` command, when `togglechannel` is enabled and this is not set.

Jackenmen (changes requested):

redbot/cogs/warnings/warnings.py
Ianardo-DiCaprio: I feel like it should as there is quite a big chance that not every moderator in a server would know exactly how the bot is set up and may not be familiar with the bot and what it means if it says nothing after a warn has been given.

Jackenmen:
I forgot, you should use `_("string")` format for strings, so that they will be available for translations. For most of those that miss it, it should be as simple as adding `_(` before and `)` after the string, but in the place where you used f-string, you will have to change it to regular string and use `.format()`, like in other places where translation strings are used.

Issue #2649, opened 2019-05-07
[Mod Enhancement] DM user with Kick, Ban, and Softban Reason


Skyridr:
# Feature request
- [x] Enhancement

From what I can gather, V3 has no feature for the bot to DM the user with the kick/ban reason before they get actually kicked. Seeing RED v3 is also a moderation bot (seeing kick/ban is a core feature), I'd like to suggest this as an enhancement.

If by any chance the user does not allow any DM from the bot and or from any users, as per Slime#3160 words, `pass DM silently`. If a user is being kicked/banned from a server, it would be nice that the user gets to know why and not be stumped if they can no longer see the server icon on their server list or unable to join the server in question. Regardless if they are aware of their actions or not.

It would also be nice that it would be optional, so people can choose to use this function if they desire to do so. Not sure if warn DM the user with the exact warn reason, but if that's not the case, would be nice if this would apply to the warn command as well.

PR #2815, opened 2019-06-28
[MOD] Features requested for kick/ban.

Ianardo-DiCaprio:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Added features to kick and ban requested in #2649. This will now send an embed to the user when they are kicked/banned from a server detailing if it was a kick or ban, from what server, <s>the moderator that issued the kick/ban</s> and the reason for the kick/ban. Also blacked.


yamikaitou:
Please allow this feature to be optional (default off)

redbot/cogs/mod/kickban.py
```diff
@@ -179,6 +179,20 @@ class KickBanMixin(MixinMeta):
             return
         audit_reason = get_audit_reason(author, reason)
         try:
+            if reason == None:
```
NeuroAssassin, line 177: Comparison with `None`s should be using `is`, not `==`.

redbot/cogs/mod/kickban.py
```diff
@@ -179,6 +179,20 @@ class KickBanMixin(MixinMeta):
...
+                reason = "No reason was given."
+            else:
+                reason = reason
```
NeuroAssassin, line 180: You don't need the `else` here for redefining the reason to itself.

redbot/cogs/mod/kickban.py
```diff
@@ -179,6 +179,20 @@ class KickBanMixin(MixinMeta):
...
+            self == self.bot.get_cog("Mod")
```
NeuroAssassin, line 181: Is this supposed to be a `=`?  And if so, `self` is already the Mod cog, so this wouldn't be needed.

redbot/cogs/mod/kickban.py
```diff
@@ -218,7 +232,22 @@ class KickBanMixin(MixinMeta):
 
         If days is not a number, it's treated as the first word of the reason.
         Minimum 0 days, maximum 7. Defaults to 0."""
-
+        author = ctx.author
+        guild = ctx.guild
+        if reason == None:
```
NeuroAssassin, line 242: Should use `is` instead of `==` when comparing with `None`.

redbot/cogs/mod/kickban.py
```diff
@@ -218,7 +232,22 @@ class KickBanMixin(MixinMeta):
...
+            reason = "No reason was given."
+        else:
+            reason = reason
```
NeuroAssassin, line 245: You don't need the `else` here for redefining `reason`.

redbot/cogs/mod/kickban.py
```diff
@@ -218,7 +232,22 @@ class KickBanMixin(MixinMeta):
...
+        self == self.bot.get_cog("Mod")
```
NeuroAssassin, line 246: Same thing as above, if its supposed to be a `=`, then it isn't needed as `self` is already the Mod cog.

redbot/cogs/mod/kickban.py
```diff
@@ -179,6 +179,20 @@ class KickBanMixin(MixinMeta):
             return
         audit_reason = get_audit_reason(author, reason)
         try:
+            if reason == None:
+                reason = "No reason was given."
+            else:
+                reason = reason
+            self == self.bot.get_cog("Mod")
+            toggle = await self.settings.guild(guild).toggle_dm()
+            if toggle == True:
```
NeuroAssassin: This can be changed to just `if toggle:`.

redbot/cogs/mod/kickban.py
```diff
@@ -218,7 +232,22 @@ class KickBanMixin(MixinMeta):
...
-
+        author = ctx.author
+        guild = ctx.guild
+        if reason == None:
+            reason = "No reason was given."
+        else:
+            reason = reason
+        self == self.bot.get_cog("Mod")
+        toggle = await self.settings.guild(guild).toggle_dm()
+        if toggle == True:
```
NeuroAssassin: Can be changed to just `if toggle:`.

NeuroAssassin (changes requested):
I've made a list of things could use a few improvements/changes.  Here's the list:

redbot/cogs/mod/kickban.py
```diff
@@ -218,7 +232,22 @@ class KickBanMixin(MixinMeta):
...
+            with contextlib.suppress(discord.HTTPException):
+                em = discord.Embed(
+                    title=_("**You have been banned from {guild}.**").format(guild=guild)
+                )
+                em.add_field(name=_("**Moderator**"), value=author.name, inline=False)
```
mikeshardmind: Showing the reason here is fine. Showing the moderator is not unless there is another setting for this.

redbot/cogs/mod/settings.py
```diff
@@ -199,3 +203,20 @@ class ModSettings(MixinMeta):
                     command=f"{ctx.prefix}unban"
                 )
             )
+
+    @modset.command()
+    @commands.guild_only()
+    async def toggledm(self, ctx: commands.Context):
```
mikeshardmind: For consistency with other core commands with configuration, this should accept a bool for setting it directly. If you'd like it to act as a toggle when that bool is not provided, that is fine.

mikeshardmind (changes requested):
In addition to some feedback another contributor pointed out, there are a few issues of design here.

redbot/cogs/mod/kickban.py
Ianardo-DiCaprio: Would you prefer no moderator or an extra setting for it?

Issue #2441, opened 2019-02-10
[Mod] Set default days of messages to delete for [p]ban


BurritoWrapped:
# Feature request

#### Select the type of feature you are requesting:

- [ ] Cog
- [x] Command
- [ ] API functionality

#### Describe your requested feature

- Ability to set the default days of messages purged when using the ban command.  This would be able to replicate what Dyno does with regards to banning.

Zephyrkul:
The ban command allows you to specify how many days worth of messages you want to delete along with the ban. See `[p]help ban` (replace `[p]` with your prefix) for details.

BurritoWrapped:
What I’m looking for is a feature to set the default days of messages to delete.

Zephyrkul:
While not a direct solution, #2455 would allow you to set an alias for banning with a specific number of days.

Jackenmen:
Added in #2930

PR #2816, opened 2019-06-28
[MOD] Added defaultdays command.

Ianardo-DiCaprio:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Added a `[p]modset defaultdays` command requested in #2441


redbot/cogs/mod/kickban.py
```diff
@@ -210,15 +210,19 @@ class KickBanMixin(MixinMeta):
...
     ):
         """Ban a user from this server and optionally delete days of messages.
 
         If days is not a number, it's treated as the first word of the reason.
         Minimum 0 days, maximum 7. Defaults to 0."""
-
+        guild = ctx.guild
+        if days == None:
+            days = await self.settings.guild(guild).default_days()
+        if days == 0:
```
Jackenmen: This doesn't seem to have any effect, you check if it's equal to 0 and then set it to 0 if it is

redbot/cogs/mod/kickban.py
```diff
@@ -210,15 +210,19 @@ class KickBanMixin(MixinMeta):
...
         *,
         reason: str = None,
     ):
         """Ban a user from this server and optionally delete days of messages.
 
         If days is not a number, it's treated as the first word of the reason.
         Minimum 0 days, maximum 7. Defaults to 0."""
-
+        guild = ctx.guild
+        if days == None:
```
Jackenmen: You should use `if days is None` instead, because there's only one `None` object and identity check will be faster.

redbot/cogs/mod/settings.py
```diff
@@ -199,3 +199,17 @@ class ModSettings(MixinMeta):
...
             )
+
+    @modset.command()
+    @commands.guild_only()
+    async def defaultdays(self, ctx: commands.Context, days: int = 0):
+        """Set the default days worth of message to be deleted when a user is banned.
+        The maximum amount of days is 7.
+        """
+        guild = ctx.guild
+        await self.settings.guild(guild).default_days.set(days)
```
Jackenmen, line 215: You have to check if days value is between 0 and 7 here

Jackenmen (changes requested):

Jackenmen:
It might also be good to update the docstring of `[p]ban` command, so that it indicates that the default is whatever is set as default with `[p]modset defaultdays` (no idea how to put it in words)

redbot/cogs/mod/kickban.py
Ianardo-DiCaprio: While I had it as if days ==None:
if 0 was said in the command for days it was still taking the default days, although after changing if days == None: to if days is None: fixed that and I've removed that part.

redbot/cogs/mod/kickban.py
Ianardo-DiCaprio: Done.

redbot/cogs/mod/settings.py
Ianardo-DiCaprio: Made it so if the number is lower than 0 it sets it to 0, if the number is larger than 7 it sets it to 7.

redbot/cogs/mod/settings.py
```diff
@@ -199,3 +199,19 @@ class ModSettings(MixinMeta):
...
+
+    @modset.command()
+    @commands.guild_only()
+    async def defaultdays(self, ctx: commands.Context, days: int = 0):
+        """Set the default days worth of message to be deleted when a user is banned.
+        The max amount of days is 7.
+        """
+        guild = ctx.guild
+        if days is not 0:
+            days = min(max(days, 0), 7)
```
Jackenmen: I think this should be changed to:
```py
if not (0 <= days <= 7):
    return await ctx.send(_("Invalid days. Must be between 0 and 7."))
```
So that it just doesn't change it, if the value is incorrect.

Jackenmen:
This looks nice, but you should probably add info about this setting when `[p]modset` command is issued, here:
https://github.com/Cog-Creators/Red-DiscordBot/blob/073ae325e9cd9d829166d6a0e080735dc9117cc1/redbot/cogs/mod/settings.py#L19-L21

Something along these lines should be good (someone can correct me, if these string aren't clear enough):
```py
default_days = await self.settings.guild(guild).default_days()
msg += _("Delete messages on ban (default): {num_days}\n").format(
    num_days=_("from last {num} days").format(num=default_days)
    if default_days > 0
    else _("No")
)
```

redbot/cogs/mod/kickban.py
```diff
@@ -210,15 +210,18 @@ class KickBanMixin(MixinMeta):
...
         *,
         reason: str = None,
     ):
         """Ban a user from this server and optionally delete days of messages.
 
         If days is not a number, it's treated as the first word of the reason.
-        Minimum 0 days, maximum 7. Defaults to 0."""
-
+        Minimum 0 days, maximum 7. If defaultdays is set and days not used 
+        in command, the defaultdays will be used."""
```
DevilXD: >If defaultdays is set and days not used in command, the defaultdays will be used.

This could use some rewording. Maybe something along the lines of:
```
If not specified, defaultdays will be used instead.
```

redbot/cogs/mod/settings.py
```diff
@@ -26,6 +26,7 @@ class ModSettings(MixinMeta):
             respect_hierarchy = await self.settings.guild(guild).respect_hierarchy()
             delete_delay = await self.settings.guild(guild).delete_delay()
             reinvite_on_unban = await self.settings.guild(guild).reinvite_on_unban()
+            default_days = await self.settings.guild(guild).default_days()
```
DevilXD: I know this may not be the best PR for it, but fetching the entire guild settings at once via `self.settings.guild(guild).all()` and then unpacking each variable from a simple dict will be less backend-straining than fetching six individual settings that are normally right next to each other.

redbot/cogs/mod/settings.py
```diff
@@ -199,3 +201,19 @@ class ModSettings(MixinMeta):
                     command=f"{ctx.prefix}unban"
                 )
             )
+
+    @modset.command()
+    @commands.guild_only()
+    async def defaultdays(self, ctx: commands.Context, days: int = 0):
+        """Set the default days worth of message to be deleted when a user is banned.
```
DevilXD:
```suggestion
        """Set the default days worth of messages to be deleted when a user is banned.
```

redbot/cogs/mod/settings.py
```diff
@@ -199,3 +201,19 @@ class ModSettings(MixinMeta):
...
+        The max amount of days is 7.
```
DevilXD:
```suggestion
        The amount of days has to be between 0 and 7.
```

DevilXD (changes requested):
Just some minor description changes/fixes and a small efficiency improvement.

redbot/cogs/mod/kickban.py
```diff
@@ -205,15 +205,18 @@ class KickBanMixin(MixinMeta):
...
         *,
         reason: str = None,
     ):
         """Ban a user from this server and optionally delete days of messages.
 
         If days is not a number, it's treated as the first word of the reason.
-        Minimum 0 days, maximum 7. Defaults to 0."""
-
+        Minimum 0 days, maximum 7. If not specified, and defaultdays is set,
+        defaultdays will be used instead."""
```
DevilXD:
```suggestion
        it will be used instead."""
```
Let's not repeat the same thing twice.

DevilXD (changes requested):
Other than this, everything looks fine now.

redbot/cogs/mod/kickban.py
Jackenmen: I would change the whole "If not specified, and defaultdays is set, defaultdays will be used instead." to "If not specified, defaultdays will be used.", because it is always set to something.

DevilXD (dismissed):
Looks ready to go for me 👌

redbot/cogs/mod/mod.py
```diff
@@ -67,7 +68,7 @@ def __init__(self, bot: Red):
 
         self.settings = Config.get_conf(self, 4961522000, force_registration=True)
         self.settings.register_global(**self.default_global_settings)
-        self.settings.register_guild(**self.default_guild_settings)
+        self.settings.register_guild(**self.default_guild_settings, force_registration=True)
```
Jackenmen:
```suggestion
        self.settings.register_guild(**self.default_guild_settings)
```
this shouldn't be here

PR #2767, opened 2019-06-11
[Core] Fix user output on cog load/reload

mikeshardmind:
### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

### Description of the changes
  - Properly fixes the load/reload exception handling
  - Fixes some i18n use here to not make assumptions about other
  languages pluralization rules.

PR #2817, opened 2019-06-28
[Core] Don't change type of exception during cog load

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This will prevent `[p]load` from suppressing tracebacks for all exceptions made by cog during load.
It still allows devs to use `redbot.core.errors.CogLoadError` if they want to show the "This package could not be loaded for the following reason" message, but now it doesn't throw at user all kinds of exceptions and doesn't swallow the traceback in console.


Jackenmen:
Let's just forget about it...
#2767

PR #2819, opened 2019-06-30
Add uvloop as Posix+CPython dependency and tweak new Make recipes

Tobotimus:
These are the changes: 
- uvloop is a dependency on non-Windows CPython systems
- `make setupenv` renamed to `make newenv`
- `make syncenv` added to sync local venv to current dependencies
- `dev-requirements.txt` moved into `tools` directory


mikeshardmind (approved):
:+1: from me

PR #2833, opened 2019-07-03
[Audio] Add a local cache to reduce API calls

Drapersniper:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

This PR adds a local cache to the Audio cog

This cache will be loaded in memory on cog load and will only write to memory on Cog unload/bot crash
This is to avoid killing your back-end with reads and write operations

This will allows the user to play Spotify songs that have been played in the past even if they run out of their YouTube API quota


Drapersniper:
Yikes my bad .... Sorry for the last commit .. wrong branch ...

PR #2834, opened 2019-07-03
[Context] Adds react_quietly method to context

Drapersniper:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

(Based from yesterdays conversation in support server)

This allows cog creators to use this to set Emojis to be added to messages

This will work similarly to 'tick' However it will have a positional argument.

`react_quietly(True)`: Adds a Tick to message
`react_quietly(False)`: Adds a Cross to message
`react_quietly(Union[discord.Emoji, discord.Reaction, discord.PartialEmoji, str])` : Adds specified Emoji

### DIFF
--- a/redbot/core/commands/context.py
+++ b/redbot/core/commands/context.py
@@ -1,6 +1,6 @@
 import asyncio
 import contextlib
-from typing import Iterable, List
+from typing import Iterable, List, Union
 import discord
 from discord.ext import commands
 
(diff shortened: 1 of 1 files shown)


mikeshardmind (changes requested):
I'm not in favor of true/false support here for specific emojis. 

`ctx.tick()` exists for simple success. an `X` is more ambiguous and should be communicated to users what failed.

the other part will need a more thorough review, but that particular portion is not appropriate for Red's Context.

Drapersniper:
Removed bool support from it

redbot/core/commands/context.py
```diff
@@ -86,6 +86,24 @@ def __init__(self, **attrs):
...
+    ) -> bool:
+        """Adds a reaction to to the command message.
+        Returns
+        -------
+        bool
+            :code:`True` if adding the reaction succeeded.
+        """
+        try:
+            await self.message.add_reaction(reaction)
+        except (discord.HTTPException, discord.NotFound, discord.Forbidden):
```
mikeshardmind: The other two here inherit from this one
```suggestion
        except discord.HTTPException:
```

redbot/core/commands/context.py
```diff
@@ -86,6 +86,24 @@ def __init__(self, **attrs):
...
+            return False
+        except discord.InvalidArgument:
```
mikeshardmind: I'd prefer we not catch this one and allow it to automatically propagate back to cog devs (this shouldn't ever occur in a well designed cog, but I see how it can occur)

mikeshardmind (changes requested):
Hey, thanks for both the contribution and for handling the bool part already, here's the remaining parts now that I've had time to look in depth. There's not much in terms of needed changes here.

Drapersniper:
Thanks for the feedback, didn't actually know `NotFound` and `Forbidden` inherit from `HTTPException`, thanks!

Removed the former 2 and also stopped catching `InvalidArgument` so that it can be propagated back to devs

mikeshardmind (approved):
Thanks for the quick turnaround on changes!

PR #2836, opened 2019-07-04
[Utils] adds humanize_number method to chat formatting

Drapersniper:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

This PR adds a new method to chat formatting that allows formatting int and float based on bot locale.

e.g. 1456425612 will now be shows as 1,456,425,612 on en_US

I've applied this method to all mentions of currency I came across (Audio, Bank, Economy, Trivia)


redbot/core/utils/chat_formatting.py
```diff
@@ -429,3 +430,44 @@ def humanize_timedelta(
...
+    ----------
+    val : Union[int, float]
+        The int to be formatted.
+
+    Returns
+    -------
+    str
+        locale aware formated number.
+    """
+    return format_decimal(val, locale=_convertlocale(get_locale()))
```
Tobotimus: There's a chance the locale being used isn't actually known to Babel, in which case it will throw a `babel.core.UnknownLocaleError` - when this happens, simply default to `en_US` :smiley:

Tobotimus (changes requested):
This looks great! Just one change being requested for now :smile:

I'll also add that I think `humanize_decimal` might be a better name for the util, considering it works for any decimal number (int or float)

redbot/core/utils/chat_formatting.py
Drapersniper: Just for wider visibility,

I changed the code up a little bit.

But basically it uses the parses provided by babel (no need to reinvent the wheel) 
It first tries to parse the red locale, if it fails with `ValueError`, then it tries to get the closest match, if  it  fails (`ValueError` or `TypeError` or `babel.core.UnknownLocaleError`) or can't get a close match then it defaults to  `en_US`

Red's locale and Babel supported locale is then cached and only updated when Red's locale is changed
This avoid calling the babel parser and Try/Except blocks every single time we call `humanize_int`

Tried with about 100 different locale formats and not once it was defaulted to `en_US` :D 

Note to self: Stop trying to reinvent the wheel

Drapersniper:
Thanks @Tobotimus for the suggestions via DM made this a lot more useful!

I may or many not have only seen your mention of `humanize_decimal` its true, but decimal may confuse less familiar devs ? maybe `humanize_number` ?  (however that kinda implies that string numbers also work, im not agaisn't trying to convert a number to int/float if i fails raise it back to user) i'll let you confirm the desired aproach here.

skeith:
I'd say `humanize_digit` since it's actually called digit grouping separator.

Tobotimus:
I think `humanize_number` is best, I guess it's implied that the number will be formatted as a decimal :)

There shouldn't be any need to handle string inputs, simply state the argument type in the docs and users will figure it out.

`humanize_digit` is confusing to me, because a digit refers to a single character in a number

changelog.d/2836.feature.rst
```diff
@@ -0,0 +1 @@
+New :meth:`humanize_number` in :module:`redbot.core.utils.chat_formatting` function to convert numbers into text which respect locale.
```
Tobotimus: Should be ``:func:`humanize_number` ``, `:meth:` is for methods.
```suggestion
New :func:`humanize_number` in :module:`redbot.core.utils.chat_formatting` function to convert numbers into text which respect locale.
```

changelog.d/2836.misc.rst
```diff
@@ -0,0 +1 @@
+New :meth:`humanize_number` is used throughout the bot.
```
Tobotimus:
```suggestion
New :func:`humanize_number` is used throughout the bot.
```

redbot/cogs/cleanup/cleanup.py
```diff
@@ -152,7 +155,7 @@ def check(m):
         to_delete.append(ctx.message)
 
         reason = "{}({}) deleted {} messages containing '{}' in channel {}.".format(
-            author.name, author.id, len(to_delete), text, channel.id
+            author.name, author.id, humanize_number(len(to_delete)), text, channel.id
```
Tobotimus: This one should probably use `babel.format_decimal` directly with en_US as the locale, since the message will always be in English, we probably shouldn't put a number formatted for a different locale in there. Especially when we have context-specific locale :)

redbot/cogs/cleanup/cleanup.py
```diff
@@ -208,7 +211,14 @@ def check(m):
         reason = (
             "{}({}) deleted {} messages "
             " made by {}({}) in channel {}."
-            "".format(author.name, author.id, len(to_delete), member or "???", _id, channel.name)
+            "".format(
+                author.name,
+                author.id,
+                humanize_number(len(to_delete)),
```
Tobotimus: Same as previous comment

redbot/cogs/cleanup/cleanup.py
```diff
@@ -240,7 +250,7 @@ def check(m):
         )
 
         reason = "{}({}) deleted {} messages in channel {}.".format(
-            author.name, author.id, len(to_delete), channel.name
+            author.name, author.id, humanize_number(len(to_delete)), channel.name
```
Tobotimus: Same here

redbot/cogs/cleanup/cleanup.py
```diff
@@ -277,7 +287,7 @@ def check(m):
         to_delete.append(ctx.message)
 
         reason = "{}({}) deleted {} messages in channel {}.".format(
-            author.name, author.id, len(to_delete), channel.name
+            author.name, author.id, humanize_number(len(to_delete)), channel.name
```
Tobotimus: Same here

redbot/cogs/cleanup/cleanup.py
```diff
@@ -319,7 +329,7 @@ def check(m):
         )
         to_delete.append(ctx.message)
         reason = "{}({}) deleted {} messages in channel {}.".format(
-            author.name, author.id, len(to_delete), channel.name
+            author.name, author.id, humanize_number(len(to_delete)), channel.name
```
Tobotimus: Same here

redbot/cogs/cleanup/cleanup.py
```diff
@@ -420,7 +430,7 @@ def check(m):
         reason = (
             "{}({}) deleted {} "
             " command messages in channel {}."
-            "".format(author.name, author.id, len(to_delete), channel.name)
+            "".format(author.name, author.id, humanize_number(len(to_delete)), channel.name)
```
Tobotimus: Same here

redbot/cogs/cleanup/cleanup.py
```diff
@@ -500,7 +510,7 @@ def check(m):
         reason = (
             "{}({}) deleted {} messages "
             "sent by the bot in {}."
-            "".format(author.name, author.id, len(to_delete), channel_name)
+            "".format(author.name, author.id, humanize_number(len(to_delete)), channel_name)
```
Tobotimus: Same here

redbot/cogs/economy/economy.py
```diff
@@ -284,9 +289,9 @@ def __init__(self, bot: Red):
                     ).format(
                         author=author,
                         currency=credits_name,
-                        amount=await self.config.PAYDAY_CREDITS(),
-                        new_balance=await bank.get_balance(author),
-                        pos=pos,
+                        amount=humanize_number(await self.config.PAYDAY_CREDITS()),
+                        new_balance=humanize_number(await bank.get_balance(author)),
+                        pos=humanize_number(pos) if pos else pos,
```
Tobotimus, line 297: When can pos be zero? If there is a legitimate case for this, please add a comment explaining it :)

redbot/cogs/economy/economy.py
```diff
@@ -331,9 +338,9 @@ def __init__(self, bot: Red):
                     ).format(
                         author=author,
                         currency=credits_name,
-                        amount=credit_amount,
-                        new_balance=await bank.get_balance(author),
-                        pos=pos,
+                        amount=humanize_number(credit_amount),
+                        new_balance=humanize_number(await bank.get_balance(author)),
+                        pos=humanize_number(pos) if pos else pos,
```
Tobotimus, line 346: Same as previous comment

redbot/cogs/economy/economy.py
```diff
@@ -384,14 +391,14 @@ def __init__(self, bot: Red):
                 if await ctx.bot.is_owner(ctx.author):
                     user_id = f"({str(acc[0])})"
                 name = f"{acc[1]['name']} {user_id}"
-            balance = acc[1]["balance"]
+            balance = humanize_number(acc[1]["balance"])
 
             if acc[0] != author.id:
-                temp_msg += f"{f'{pos}.': <{pound_len+2}} {balance: <{bal_len + 5}} {name}\n"
+                temp_msg += f"{f'{humanize_number(pos)}.': <{pound_len+2}} {balance: <{bal_len + 5}} {name}\n"
```
Tobotimus: Line is a bit too long, could you split it over two lines? Black doesn't split up strings to get under the max line length

redbot/core/modlog.py
```diff
@@ -149,7 +150,9 @@ def __init__(
         """
         casetype = await get_casetype(self.action_type)
         title = "{}".format(
-            _("Case #{} | {} {}").format(self.case_number, casetype.case_str, casetype.image)
+            _("Case #{} | {} {}").format(
+                humanize_number(self.case_number), casetype.case_str, casetype.image
```
Tobotimus: I think in this circumstance we shouldn't use `humanize_number`. We use case numbers as input to other commands, this makes it slightly more annoying for users to copy+paste the case number

Tobotimus (changes requested):
Just a few small changes for now :)

redbot/cogs/economy/economy.py
Drapersniper: in `get_leaderboard_position` there is a `return None` so there is a chance of it being None rather than 0 , it seems to use `discord.utils.find` which can return None... Doubt it could ever happen from a quick glance at the code, but since it's in the source function just wanted to handle it.

redbot/cogs/cleanup/cleanup.py
Drapersniper: I've added the `override_locale` arg to `humanize_number` this is to facilitate the overriding of the current bots locale while still being able to use `humanize_number` (So devs don't need to import babel and figure out it and do all the error handling we done here)

Drapersniper (commented):
I've also added the Override to en_US on a few errors and and log messages that you didn't mention on the review, let me know if you'd prefer reverting it.

redbot/core/bank.py
```diff
@@ -242,12 +242,19 @@ def _invalid_amount(amount: int) -> bool:
     if not isinstance(amount, int):
         raise TypeError("Withdrawal amount must be of type int, not {}.".format(type(amount)))
     if _invalid_amount(amount):
-        raise ValueError("Invalid withdrawal amount {} < 0".format(humanize_number(amount)))
+        raise ValueError(
```
Drapersniper, line 241: Wasn't too sure if this gets translated to users locale down the pipe never used the bot with a non en locale.
If it doesn't it make sense to override these to en_US, but if its translated later then this should be removed

redbot/core/bank.py
```diff
@@ -242,12 +242,19 @@ def _invalid_amount(amount: int) -> bool:
...
+            "Invalid withdrawal amount {} < 0".format(
+                humanize_number(amount, override_locale="en_US")
+            )
+        )
 
     bal = await get_balance(member)
     if amount > bal:
         raise ValueError(
-            "Insufficient funds {} > {}".format(humanize_number(amount), humanize_number(bal))
+            "Insufficient funds {} > {}".format(
```
Drapersniper, line 250: Same as above

redbot/core/bank.py
```diff
@@ -279,7 +286,11 @@ def _invalid_amount(amount: int) -> bool:
     if not isinstance(amount, int):
         raise TypeError("Deposit amount must be of type int, not {}.".format(type(amount)))
     if _invalid_amount(amount):
-        raise ValueError("Invalid deposit amount {} <= 0".format(humanize_number(amount)))
+        raise ValueError(
```
Drapersniper, line 285: Same as above

redbot/core/bank.py
```diff
@@ -313,7 +324,11 @@ def _invalid_amount(amount: int) -> bool:
     if not isinstance(amount, int):
         raise TypeError("Transfer amount must be of type int, not {}.".format(type(amount)))
     if _invalid_amount(amount):
-        raise ValueError("Invalid transfer amount {} <= 0".format(humanize_number(amount)))
+        raise ValueError(
```
Drapersniper, line 326: Same as above

redbot/core/bank.py
Tobotimus: That's all good, you made the right call :+1: Even if these messages are sent to the user, they're not translated, so using `humanize_number` with a custom locale wouldn't make sense.

redbot/cogs/economy/economy.py
Tobotimus: Ah no problem :+1:

redbot/core/i18n.py
```diff
@@ -160,6 +171,32 @@ def _add_translation(self, untranslated, translated):
...
+            try:
+                # Try to find a babel locale that's close to the one used by red
+                babel_locale = Locale(Locale.negotiate([red_locale], supported_locales, sep="-"))
+            except (ValueError, TypeError, babel.core.UnknownLocaleError):
+                # If we fail to get a close match we will then default to "en_US"
+                babel_locale = Locale("en", "US")
+    return babel_locale
+
+
+def get_babel_locale(locale: Optional[str] = None) -> babel.core.Locale:
```
Tobotimus, line 194: Could you add a docstring for this function? And make sure it's in the i18n docs :)

Tobotimus (changes requested):

redbot/core/i18n.py
Drapersniper: image
Hows this?

Tobotimus (approved):

PR #2837, opened 2019-07-04
[Audio] Add a local cache to reduce YouTube API calls

Drapersniper:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

This PR adds a local cache to the Audio cog

This cache will be loaded in memory on cog load and will only write to memory on Cog unload/bot crash
This is to avoid killing your back-end with reads and write operations

This will allows the user to play Spotify songs that have been played in the past even if they run out of their YouTube API quota

Sorry for the repeated PR... was trying to fix the mess of a fork I ended up with when I created the original PR from v3/develop ... Note kids ... when pushing a PR double check you're in the correct branch ....


Stylite:
I looked this over and I have two questions. Why are you not using config for the cache and why doesn't the cache include the song link?

Drapersniper:
The cache is a dict for {song name as given by Spotify : YouTube url}

Reason config is not used is because this file can get stupidly large, so I don’t want to kill people’s backend with reads/writes so it only writes once the bot is shutdown/audio is unloaded. And cache is loaded to memory on load 

Specially when #2861 is merged as that will have a [p]playlist update command  and allow users to create playlists

Drapersniper:
Ah I misunderstood the question from @Stylite, He was asking why the Spotify URL isn't cached.

This is for a few reasons:

1. Reduce Cache size (This has the potential to have millions of entries since the cache is never wiped)
1. Spotify doesn't have hard limits on it's quota AFAIK
1. Doing so would require a full rework of `Audio._spotify_playlist()`

I could update this PR to cache Spotify API calls too but the issue I see here is that we use the following end points:

- `https://api.spotify.com/v1/tracks`
- `https://api.spotify.com/v1/albums`
-  `https://api.spotify.com/v1/playlists`

All of which returns multiple links and with #2861 we would only want to cache entries from `https://api.spotify.com/v1/tracks`  and this cache would become even larger than the current PR cache so I'd likely split it into 2 files

@aikaterna  If this is something you think we should cache it let me know and I'll update the PR before you review it, if so what information should be cached ? First song returned so if the user used a specific Link it will always return the same track regardless of what that link may return in the future?

mikeshardmind:
This needs to write more frequently. While I understand the concerns on backends, there are other solutions such as a max cache size with age out, uing sqlite3 in the cog data dir (etc)

Drapersniper:
Okay so I done #2890 which uses SQL caching instead of a JSON to Memory (although a big slower, negligible really), this allows rooms to cache other things such as Lavalink track info to which could potentially allow for more tracks to be loaded in a short period without Lavalink timing out and throwing an error when loading multiple songs back to back or large playlists.

I would recommend we review that one and merge that instead of this one, but leaving this one open as a `simple` cache option in case we decide that 2890 is not suitable to core.

Drapersniper:
Superseded by #2904

PR #2840, opened 2019-07-07
[i18n] Untrack .po files, upload translations on commit, download on release

Tobotimus:
So I think we should try to automate this whole thing - uploading and downloading translations.

Uploading on commit means that crowdin will always be translating up-to-date strings, so more coverage should be achieved at release.

Downloading on release also means the translations will be the most up-to-date available.

We should perhaps communicate a bit better with translators, when we plan on doing a release (maybe about a week's notice or more if possible) and we can give a sort-of "grace period" where we will try not to change too many strings leading up to the release.


Tobotimus:
Needs a rethink

PR #2842, opened 2019-07-08
[Audio] Check for player when not connected

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

The changes introduced with #2812 needed to have a player check before attempting to send the values to the player. Using [p]shuffle or [p]repeat right now while the bot is not connected will result in an error otherwise.


Tobotimus (approved):
:+1:

PR #2849, opened 2019-07-10
[Utils] Text to file

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This is a short util that will convert a long string into a `discord.File` object, useful to bypass the character limit. Nothing is wrote on the disk, `io.BytesIO` is used instead.

### Example

```py
from redbot.core.utils.chat_formatting import text_to_file

@commands.command()
async def getchannelmsg(self, amount: str, channel: discord.TextChannel = None):
    text = ""
    async for message in channel.history(limit=amount):
        text += f"{message.author}: {message.content}"
    if len(text) > 2000:
        await ctx.send(file=text_to_file(text))
    else:
        await ctx.send(text)
```


Jackenmen:
Please use `StringIO` instead as you're using just text in this file, not bytes content.
Also, if other decide `BytesIO` is fine, `bytes` requires encoding passed to it, for example `bytes(text, encoding="utf-8")`

Zephyrkul:
iirc, discord.File requires the use of BytesIO for in-memory streams.

Jackenmen:
discord.File support any file object, I also tested it specifically with `StringIO` before I posted.

laggron42:
I tested with StringIO and it just returned an empty file.
image
^ following the python tutorial on how to use it, you can see the "0 bytes"

And I tested it without the encoding, it works fine.

Zephyrkul:
https://discordpy.readthedocs.io/en/stable/api.html#file

The type requires [BufferedIOBase](https://docs.python.org/3/library/io.html#io.BufferedIOBase). StringIO does not inherit from this base class.

redbot/core/utils/chat_formatting.py
```diff
@@ -429,3 +430,27 @@ def humanize_timedelta(
             strings.append(f"{period_value} {unit}")
 
     return ", ".join(strings)
+
+
+def text_to_file(text: str, filename: str = "file.txt", *, spoiler: bool = False):
```
Tobotimus: Would like an `encoding: str = "utf-8"` kwarg here :)

redbot/core/utils/chat_formatting.py
```diff
@@ -429,3 +430,27 @@ def humanize_timedelta(
...
+        The name of the file sent. Defaults to ``file.txt``.
+    spoiler: bool
+        Whether the attachment is a spoiler. Defaults to ``False``.
+
+    Returns
+    -------
+    discord.File
+
+    """
+    file = BytesIO(bytes(text))
```
Tobotimus: Prefer:
```suggestion
    file = BytesIO(text.encode(encoding))
```

Tobotimus (changes requested):
Thanks, this is a pretty handy util :)

DevilXD:
@retke I had the same problem when I wanted to send a file to the chat long time ago, and this is what was explained to me: https://i.imgur.com/nJnT4Wt.png
Your example should very much work after adding a `seek(0)` right before sending.

As for BytesIO/StringIO usage, one of the things I was trying to send in the chat as StringIO simply failed with some weird encoding error, that went away after switching to BytesIO with UTF-8 encoding set, so I'd recommend that. It was quite complicated mess of markdown-rich reddit post generator if this gives you any context to what could've happened. `.encode()` uses UTF-8 by default.

redbot/core/utils/chat_formatting.py
```diff
@@ -429,3 +430,29 @@ def humanize_timedelta(
...
 
     return ", ".join(strings)
+
+
+def text_to_file(
+    text: str, filename: str = "file.txt", *, spoiler: bool = False, encoding: str = "utf-8"
+):
+    """Prepares text to be sent as a file on Discord, without character limit.
+
+    This writes text into a bytes object that can be used for the ``file`` or ``files`` parameters\
```
Tobotimus: Do not escape line-endings, anywhere in Python, except when using `textwrap.dedent`

Simply limit docstring line length to about 80 characters, use normal newlines, and sphinx / whichever IDE you're using will render the newlines as spaces where appropriate.

redbot/core/utils/chat_formatting.py
```diff
@@ -429,3 +430,29 @@ def humanize_timedelta(
...
+    of :func:`discord.abc.send`.
```
Tobotimus:
```suggestion
    of :meth:`discord.abc.Messageable.send`.
```

redbot/core/utils/chat_formatting.py
```diff
@@ -429,3 +430,29 @@ def humanize_timedelta(
...
+    text: str
+        The text to put in your file.
+    filename: str
+        The name of the file sent. Defaults to ``file.txt``.
+    spoiler: bool
+        Whether the attachment is a spoiler. Defaults to ``False``.
+
+    Returns
+    -------
+    discord.File
```
Tobotimus, line 454: Best to give a description under ``Returns``, even if it is obvious or trivial. I've had cases where Sphinx has whinged because it was missing. If it doesn't do it now, it could well do it in a future version of Sphinx.

Tobotimus (changes requested):

laggron42:
Should be ready now.

Tobotimus (approved):

PR #2853, opened 2019-07-12
[Core] Cog load fixes

mikeshardmind:
- Fixes an issue observed on a live bot with a race condition with
  permissions

  This includes a hack of a fix to get the best possible compromise
  between an out of cycle breaking change, and a must fix issue fixed
  ASAP

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

PR #2854, opened 2019-07-12
[Core] Cog load fixes

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Per discussion in discord, we may have another way to fix the race condition part which won't result in breaking. Other fixes from #2853 have been split out to here.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -408,9 +408,7 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):  # pylint: d
                 f"the cog to adhere to this requirement."
             )
         if cog.__cog_name__ in self.cogs:
-            raise RuntimeError(
-                f"There is already a cog named {cog.__cog_name__} loaded."
-            ) from None
+            raise RuntimeError(f"There is already a cog named {cog.__cog_name__} loaded.")
         if not hasattr(cog, "requires"):
             commands.Cog.__init__(cog)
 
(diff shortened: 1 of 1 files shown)


redbot/core/bot.py
```diff
@@ -407,32 +408,46 @@ def add_cog(self, cog: commands.Cog):
...
+                        "http://red-discordbot.readthedocs.io/en/v3-develop/framework_commands.html"
+                    )
+            super().add_cog(cog)
+            self.dispatch("cog_add", cog)
+            for command in cog.__cog_commands__:
+                self.dispatch("command_add", command)
+        except Exception:
+            for hook in added_hooks:
+                with contextlib.suppress(ValueError):
+                    self.remove_permissions_hook(hook)
```
Tobotimus, line 445: Not sure about `contextlib.suppress` here. A `ValueError` *shouldn't* ever be raised, suppressing it is more likely to be silencing something which has actually gone wrong.

Tobotimus (dismissed):

redbot/core/bot.py
```diff
@@ -407,32 +407,52 @@ def add_cog(self, cog: commands.Cog):
                 f"not inherit from the commands.Cog base class. The cog author must update "
                 f"the cog to adhere to this requirement."
             )
+        if cog.__cog_name__ in self.cogs:
+            raise RuntimeError(
+                f"There is already a cog named {cog.__cog_name__} loaded."
+            ) from None
```
Tobotimus: `from None` is unnecessary here

Tobotimus (approved):

PR #2857, opened 2019-07-13
Make Requires.verify() wait until rules are loaded

Tobotimus:
This gives every `Requires` object an `asyncio.Event` object to wait on until all of its rules are loaded.

Also ensures `Requires` objects are reset when unloaded, particularly in case a `Command` object manages to stay in memory between cog unload and load, and its permissions rules change between those events.

Also, this PR re-ordered some of the event loop policy stuff, because it was required that the event loop policy be set before creating any `Requires` objects. This may or may not have an effect on other `get_event_loop()` calls elsewhere (either in our code, a dependency's, or asyncio's). Either way, these effects would be a *correction*, and any bugs that arise from it are likely to have been occurring silently beforehand.

Also sets the event loop *policy* on Windows to `WindowsProactorEventLoopPolicy` instead of directly setting the loop object. The reason this wasn't done originally was because the policy was only made available in Python 3.7.0 (see the last dot point [here](https://docs.python.org/3/whatsnew/3.7.html#asyncio)).

This PR is an alternative to #2853.

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -10,6 +10,22 @@ import sys
 
 import discord
 
+# Set the event loop policies here so any subsequent `get_event_loop()`
+# calls, in particular those as a result of the following imports,
+# return the correct loop object.
+if sys.platform == "win32":
+    asyncio.set_event_loop_policy(asyncio.WindowsProactorEventLoopPolicy())
+else:
+    # Let's not force this dependency, uvloop is much faster on cpython
+    if sys.implementation.name == "cpython":
+        try:
+            import uvloop
+        except ImportError:
+            uvloop = None
+            pass
+        else:
+            asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
+
 import redbot.logging
 from redbot.core.bot import Red, ExitCodes
 from redbot.core.cog_manager import CogManagerUI
@@ -21,18 +37,6 @@ from redbot.core.dev_commands import Dev
 from redbot.core import __version__, modlog, bank, data_manager
 from signal import SIGTERM
 
-# Let's not force this dependency, uvloop is much faster on cpython
-if sys.implementation.name == "cpython":
-    try:
-        import uvloop
-    except ImportError:
-        uvloop = None
-        pass
-    else:
-        asyncio.set_event_loop_policy(uvloop.EventLoopPolicy())
-
-if sys.platform == "win32":
-    asyncio.set_event_loop(asyncio.ProactorEventLoop())
 
 log = logging.getLogger("red.main")
 
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -15,16 +15,15 @@ import discord
 # return the correct loop object.
 if sys.platform == "win32":
     asyncio.set_event_loop_policy(asyncio.WindowsProactorEventLoopPolicy())
-else:
+elif sys.implementation.name == "cpython":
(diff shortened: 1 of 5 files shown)


redbot/__main__.py
```diff
@@ -10,6 +10,22 @@
...
 import discord
 
+# Set the event loop policies here so any subsequent `get_event_loop()`
+# calls, in particular those as a result of the following imports,
+# return the correct loop object.
+if sys.platform == "win32":
+    asyncio.set_event_loop_policy(asyncio.WindowsProactorEventLoopPolicy())
+else:
+    # Let's not force this dependency, uvloop is much faster on cpython
+    if sys.implementation.name == "cpython":
```
mikeshardmind: Mostly nitpicking, but I'd prefer this converted to an `elif` for the indentation to properly represent the options as "on windows, not on windows, but on cpython, something else.

redbot/cogs/permissions/__init__.py
```diff
@@ -8,6 +8,7 @@
     # the permissions commands themselves have rules added.
```
mikeshardmind: Remove or update the outdated comment

redbot/cogs/permissions/__init__.py
```diff
@@ -8,6 +8,7 @@
...
     # Automatic listeners being added in add_cog happen in arbitrary
     # order, so we want to circumvent that.
-    bot.add_listener(cog.red_cog_added, "on_cog_add")
-    bot.add_listener(cog.red_command_added, "on_command_add")
+    await cog.on_cog_add(cog)
+    for command in cog.walk_commands():
```
mikeshardmind: This needs to be swapped, `walk_commands` includes aliases, and exists for other purposes (mainly so that cogs mirror the bot method from discord.py)
```suggestion
    for command in cog.__cog_commands__:
```

redbot/core/bot.py
```diff
@@ -452,6 +443,29 @@ def add_cog(self, cog: commands.Cog):
...
+            raise RuntimeError("Commands must be instances of `redbot.core.commands.Command`")
+
+        super().add_command(command)
+
+        permissions_not_loaded = "permissions" not in self.extensions
+        self.dispatch("command_add", command)
+        if permissions_not_loaded:
+            command.requires.ready_event.set()
+        if isinstance(command, commands.Group):
+            for subcommand in command.walk_commands():
```
mikeshardmind: walk commands isn't safe here either, see prior.

redbot/core/bot.py
```diff
@@ -452,6 +443,29 @@ def add_cog(self, cog: commands.Cog):
...
+                self.dispatch("command_add", subcommand)
+                if permissions_not_loaded:
+                    command.requires.ready_event.set()
+
+    def remove_command(self, name: str) -> None:
+        command = super().remove_command(name)
+        command.requires.reset()
+        if isinstance(command, commands.Group):
+            for subcommand in command.walk_commands():
```
mikeshardmind: don't use walk commands

mikeshardmind (dismissed):
There's some issues here (see review comments), but this is mostly fine and overall a better option for fixing this.

redbot/cogs/permissions/__init__.py
Tobotimus: Ah I see. I chose `walk_commands()` because it's a public+documented interface, turns out it's not documented very well :D

mikeshardmind (approved):

PR #2865, opened 2019-07-14
[Permissions] Send help on missing argument

### LINKED ISSUE #2851: [Permissions] Adding a permission for something that doesn't exist returns no error
# Command bugs

#### Command name

`[p]permissions addglobalrule`

Permissions

The bot says the thing I'm trying to add a rule for something that doesn't exist.

#### What actually happened?

The `discord.Greedy` converter evaluates to 0 things converted, the `for` loop just doesn't run for an empty list (so nothing happens), and the bot says "Rule added."

 `[p]permissions addglobalrule allow choose <put_any_random_string_here>`

This might be a case for other commands in the cog as well (any command that uses the Greedy converter).

Tobotimus:
Resolves #2851. ~~Bit of a hotfix, may look at not using a Greedy converter soon.~~ Greedy converter replaced with varargs

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -294,6 +294,9 @@ class Permissions(commands.Cog):
 
         `<who_or_what>` is one or more users, channels or roles the rule is for.
         """
+        if not who_or_what:
+            await ctx.send_help()
+            return
         for w in who_or_what:
             await self._add_rule(
                 rule=cast(bool, allow_or_deny),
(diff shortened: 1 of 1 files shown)


redbot/cogs/permissions/permissions.py
```diff
@@ -294,6 +294,9 @@ def __init__(self, bot: Red):
 
         `<who_or_what>` is one or more users, channels or roles the rule is for.
         """
+        if not who_or_what:
+            await ctx.send_help()
```
mikeshardmind, line 298: This can possibly be made more clear later, but as this is more a hotfix for a last minute addition to 3.1.3, we can worry about improving the UX on this later.

mikeshardmind (approved):

PR #2866, opened 2019-07-14
Mute Rewrite

DiscordLiz:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature

Work in progress.


mikeshardmind:
I know this isn't finished, but since this is a ground up rewrite of how red handles mutes with you as the sole author, put your name in the codeowners file for the cog prior to merge.

redbot/core/utils/discord_helpers.py
```diff
@@ -0,0 +1,149 @@
...
+    .. versionadded: 3.2.0
+    """
+
+    def __init__(self, **data: int):
+        self.allows_added = data.pop("allows_added", 0)
+        self.allows_removed = data.pop("allows_removed", 0)
+        self.denies_added = data.pop("denies_added", 0)
+        self.denies_removed = data.pop("denies_removed", 0)
+
+        if (self.allows_added & self.denies_added) or (self.allows_removed & self.denies_removed):
```
mikeshardmind: Needs checks for allow_add/remove and deny add/remove intersections as well.

mikeshardmind:
Just a quick question based on a commit note
```
Put uuids back in the mute records

 - I forgot that the plan is to support custom overwrites.
 - This means there can be two temp mutes with different expiring diffs
 under some circumstances
```
Would it be better to have a way to retroactively update mutes to the new settings instead of this + needing to detect additionally which overwrites would have been applied if not for already existing from a prior mute?

DiscordLiz:
That would be a lot easier. I'll do that instead.

DiscordLiz:
I am still working on this, but I won't be pushing progress commits anymore until I am done with it.

mikeshardmind:
What's the status on this? Can this be ready for review by end of week, or should we look into other means of including this? (either delaying, working with additional contributors to handle the remaining work, or incorporating a subset of the changes adding additional features later)

DiscordLiz:
Yes, it can be ready by end of week.

GamerGirl1995:
:( and was going to be ready..... by the end of the week. sike its not :( - i was looking forward to this pr.

DiscordLiz:
> :( and was going to be ready..... by the end of the week. sike its not :( - i was looking forward to this pr.

@GamerGirl1995 If you think it was that easy and that it was just dropped for no reason, do it yourself then. Take the code that's here for a headstart if you want.

It was only going to be by the end of the week before my life flipped upside down, so I'm sure *you* can wrap up the remaining work on it no problem right? 

It won't be a problem for *you* to make whatever changes are needed to meet every partially conflicting goal before whenever the next release is at this point right?

GamerGirl1995:
> > :( and was going to be ready..... by the end of the week. sike its not :( - i was looking forward to this pr.
> 
> @GamerGirl1995 If you think it was that easy and that it was just dropped for no reason, do it yourself then. Take the code that's here for a headstart if you want.
> 
> It was only going to be by the end of the week before my life flipped upside down, so I'm sure _you_ can wrap up the remaining work on it no problem right?
> 
> It won't be a problem for _you_ to make whatever changes are needed to meet every partially conflicting goal before whenever the next release is at this point right?

no need to get salty over that message jeez. i'm sorry your life is shitty. i'm sure that counts for more people though. if i wasn't handicapped i would have finished the code sure. eitherway good luck with life hope things start looking up for you soon.

mikeshardmind:
We appreciate the time people take when they can. The features here are still planned, but won't be a part of 3.2.0 (we may see these in 3.2.x)

Comments about a particular contributor's ability to meet deadlines aren't really useful here. Life comes first, and people contribute with their own time as they are available to do so.

Should it be a recurring issue, we might need to speak with a contributor about what projects they take on, but this is not the case here, nor would it be appropriate for public discussion.

PR #2876, opened 2019-07-15
[CustomCom] Fix broken custom commands

Tobotimus:


mikeshardmind (approved):
This fixes the issue, but I still don't think this is an appropriate use of command objects. I'll be opening an issue for things we need to fix up in the code base with this and similar issues. For now, this is good enough for a hotfix release.

Tobotimus:
Gonna admin-merge this because for some reason GitHub doesn't think the build is finished 

image

PR #2877, opened 2019-07-15
[ModLog] Fix get_case() and get_casetype()

Tobotimus:
This fixes `[p]reason` and `[p]case` with cases that were created after 3.1.3.


mikeshardmind (changes requested):
This will cause a regression. (We had tracebacks in support for cases not having the id) We need to fix the root cause. (the id missing from the stored dict)

mikeshardmind (approved):
We're good here now.

PR #2881, opened 2019-07-16
[Commands] `cls` parameter can now be passed to group like in d.py

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

d.py allows that behavior since 1.1.0 release


changelog.d/2881.misc.rst
```diff
@@ -0,0 +1 @@
+[Commands] Allow passing cls in the ``group()`` decorator
```
Tobotimus: We're able to cross-reference things in the changelog, which makes it easier to know what you've changed :smiley:

Also, I don't think we're really doing the `[Category]` thing in our changelog, the categories are already sort of as defined as we'd like them to be :)
```suggestion
Allow passing cls in the :func:`redbot.core.commands.group()` decorator
```

mikeshardmind (approved):

PR #2884, opened 2019-07-17
Slowmode now accepts integer only inputs as seconds

DevilXD:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

The `[p]slowmode` command now accepts integer only values as seconds.
Instead of doing `[p]slowmode 10s` one can do `[p]slowmode 10` for the same effect as well.


DevilXD:
Hmm, so apparently Travis failed on docs only because of the new changelog stuff. Note that this PR doesn't even touch docs, so hopefully it should be good regardless .-.

mikeshardmind:
Needa a changelog entry. I think Github should have prompted that the contribution guidelines were updated.since.your last PR.

DevilXD:
Didn't know this. Should be alright now I guess

mikeshardmind (approved):
👍

PR #2890, opened 2019-07-18
[Audio] Toggable Cache for Spotify, Youtube Data and Lavalink API (+ No longer wait for full playlist to load before start playing tracks)

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

This it's an alternative and improved version of #2837 

This will store responses from the YouTube Data API and the Spotify API into a SQL database and it will check if they exist in the database before going ahead and making a new API call.

TL:DR of this is that API calls for a specific song are only made once, and every future query is from the database.

Note: This still makes Spotify API calls for Playlists and Albums, since these can change, but it will store the tracks from both in the database and auto update a track if its information is out of date.


Drapersniper:
Superseded by #2904

PR #2899, opened 2019-07-23
[Docs] Add a copy button to code blocks

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This includes the [sphinx_copybutton](https://github.com/choldgraf/sphinx-copybutton/) extension to the docs with modified CSS rules to work with our theme (thanks to @Eznel).

### DIFF
--- a/docs/_static/copybutton.css
+++ b/docs/_static/copybutton.css
@@ -1,0 +1,62 @@
+/*
+ CSS rules for the copy button modified to work on the theme "sphinx_rtd_theme"
+ Original: https://github.com/choldgraf/sphinx-copybutton/blob/master/sphinx_copybutton/_static/copybutton.css
+*/
+
+/* Copy buttons */
+a.copybtn {
+    position: absolute;
+    top: 2px;
+    right: 2px;
+    width: 1.6em;
+    height: 1em;
+    padding: .3em;
+	opacity: .3;
+    transition: opacity 0.5s;
+}
+
+div.highlight  {
+    position: relative;
+}
+
+.highlight:hover .copybtn {
+	opacity: 1;
+}
+
+/**
+ * A minimal CSS-only tooltip copied from:
+ *   https://codepen.io/mildrenben/pen/rVBrpK
+ *
+ * To use, write HTML like the following:
+ *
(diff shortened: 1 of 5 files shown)


Drapersniper:
One thing to keep in mind is if a block has multiple commands it should either:

- Be split into multiple blocks
- Add the Platform supported `&&`'s  to join the command into 1 

Can see users pressing copy and trying to run a command they copied, only have 1 command run and then they go to Support asking for help as they can't set it up.

laggron42:
@Drapersniper I know that, on Linux and macOS, when you paste multiple lines, it paste the first line, executes it, then paste the next one, but the last line will still need a return, so I'll go for `&&`. Not sure what Windows needs tho.

mikeshardmind (commented):
This is a no from me.

This makes more than just what makes sense to have a copy button have one.

It also encourages what is a bad user training pattern. Namely, trusting the browser as  clipboard source for shell commands.

I'm more inclined to add a warning not to copy paste here than to add a button encouraging it.

Bakersbakebread:
The reality is that a vast majority of users will actually copy the commands to the clipboard, providing this small UX isn't a matter of user training.

I'd suggest making the commands explicit if it can't be copied in 'bulk'. Namely, one copy button per line, this would train the user to follow the order instead of a block.

mikeshardmind:
@Bakersbakebread Even if you disagree about it being a user training issue, it is one. While it may be a lost cause to an extent, it only is because of a critical mass of people who either don't care, don't understand the impact, or already consider it a lost cause because of the prior two. People trade convenience for doing things correctly way too frequently.

I'm also against making a per line code block and adding an extra dependency (multiple actually, when you look at the chain and realize we'd also be fetching some additional js stuff in docs) for just adding a javascript copy button when ctrl+C (command+C for our OSX users) works and could be argued to at least be relatively more safe if the user blocks javascript access to the clipboard (there's a setting for this in many browsers)

Jackenmen:
Hmm, how about a button that would just select all text in code block? It would prevent issue of not selecting whole command when copying, but it would not modify user's clipboard.

PR #2901, opened 2019-07-24
[ModLog API] Add default casetypes, remove need for a specific auditlog action

### LINKED ISSUE #2897: [Modlog API] Fundamental change to modlog API
# Feature request

- [ ] Cog
- [ ] Command
- [X] API functionality

The modlog API should not auto-detect events caused by the bot.

- This would allow the simplification of interacting with the modlog API.
- This would make interacting with the modlog API easier, but required for certain actions.
- This would not remove auto-detecting bans made by moderators through the discord interface rather than the bot

### A practical case and rationale

While working on #2866, I noticed there is no clean way to create a case type for multiple audit log entries. In wanting to allow role mutes to be supplemented with permission overwrites as needed, I would need a case type which supported multiple audit log actions. 

Rather than change the modlog API to support this, removing the action type all together and just not detecting actions the bot took allows cleaner ease of use.

### Downsides

This does mean that someone can't do a ban with a reason and have an audit log entry automatically generated, they would need to create a case.

### Mitigating the impact of downsides

If we provide a few pre-registered case types which cover generic use already registered and support their use, the downside becomes one extra import and one extra line per mod action.

### Breakage

This would remove a key from the dicts used in case registration.
This would be a change in expected behavior if cogs are not updated for this.

DiscordLiz:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This is doing multiple things.

1. Modlog entries are no longer reliant on the mod cog being loaded.
2. This adds some default, preregistered generic casetypes for cog creator use.
3. This removes the `audit_type` parameter from CaseTypes
4. This fixes the double post to modlog for bans which are registered by cogs other than red's mod cog
  - The choice of fix was to require case usage for all bot actions intended to be caught.
  - Red now only detects actions made by other users directly in the interface.
  - This can be improved if https://github.com/discordapp/discord-api-docs/issues/575 is implemented in a way which doesn't wait on the auditlog to appear or if events are ever changed to include the auditlog info with them.

resolves #2897


Jackenmen:
I know it's a draft and all that, but wouldn't it be better if the case could be first created without a mod so that it shows up in modlog and later add a mod if we manage to find it in audit log? Yes, I know there is a `[p]reason` command that requires moderator to be available in case, but at least the ban will still show up in modlog, if we don't find audit log entry.

DiscordLiz:
No.
This still detects and records non-bot ban actions for servers that care about a full record, but doesn't log actions taken by the bot unless a cog creates a case. This prevents the awkward issues with double case entries for 3rd party cogs which issue bans. I also moved all the listeners out of the mod cog so that modlog can function with or without mod loaded.

redbot/core/modlog.py
```diff
@@ -51,12 +57,94 @@
...
+                entry = await guild.audit_logs(
+                    action=discord.AuditLogAction.ban, before=before, after=after
+                ).find(lambda e: e.target.id == member.id)
+            except discord.Forbidden:
+                break
+            except discord.HTTPException:
+                pass
+            else:
+                if entry:
+                    if entry.mod.id == guild.me.id:
```
TrustyJAID: This should be `entry.user.id`.

DiscordLiz: I've fixed that and the other instance of using the wrong attribute.

redbot/core/modlog.py
```diff
@@ -51,12 +57,94 @@
...
+
+        mod, reason, date = None, None, None
+        attempts = 0
+
+        while mod is None and attempts < 12:  # wait up to an hour to find a matching case
+            attempts += 1
+            try:
+                entry = await guild.audit_logs(
+                    action=discord.AuditLogAction.ban, before=before, after=after
+                ).find(lambda e: e.target.id == member.id)
```
mikeshardmind: This before/after condition needs to exist inside the find as well. (Try fetching the auditlogs manually to see why)
(on both listeners)

redbot/core/modlog.py
```diff
@@ -51,12 +57,94 @@
     _conf.register_custom(_CASETYPES)
     _conf.register_custom(_CASES)
     await _migrate_config(from_version=await _conf.schema_version(), to_version=_SCHEMA_VERSION)
+    await register_casetypes(all_generics)
+
+    async def on_member_ban(guild: discord.Guild, member: discord.Member):
+
```
mikeshardmind, line 63: Can you add a 10 second sleep here (and on the other) to help make the servers without a large audit log delay not wait as long?

redbot/core/modlog.py
```diff
@@ -51,12 +57,94 @@
...
+            except discord.Forbidden:
+                break
+            except discord.HTTPException:
+                pass
+            else:
+                if entry:
+                    if entry.user.id == guild.me.id:
+                        # Don't create modlog entires for the bot's own unbans, cogs do this.
+                        return
+                    mod, reason, date = entry.user, entry.reason, entry.created_at
```
mikeshardmind: needs a break after this (and other) to prevent sleeping again before making the case

mikeshardmind (changes requested):
A few remaining things before this is ready for use.

redbot/core/modlog.py
DiscordLiz: I went for a slight restructure of the loop instead.

mikeshardmind (approved):
Thanks for following up on reviews, This seems ready for use now.

PR #2908, opened 2019-07-30
[ModLog] Optimise get_next_case_number()

Tobotimus:
We've noticed that this is taking an obscenely long time on very large servers.

### DIFF
--- a/redbot/core/modlog.py
+++ b/redbot/core/modlog.py
@@ -598,6 +598,27 @@ async def get_case(case_number: int, guild: discord.Guild, bot: Red) -> Case:
     return await Case.from_json(mod_channel, bot, case_number, case)
 
 
+async def get_latest_case(guild: discord.Guild, bot: Red) -> Optional[Case]:
+    """Get the latest case for the specified guild.
+
+    Parameters
+    ----------
+    guild : discord.Guild
+        The guild to get the latest case for.
+    bot : Red
+        The bot object.
+
+    Returns
+    -------
+    Optional[Case]
+        The latest case object. `None` if it the guild has no cases.
+
+    """
+    case_number = await _conf.guild(guild).latest_case_number()
+    if case_number:
+        return await get_case(case_number, guild, bot)
+
+
 async def get_all_cases(guild: discord.Guild, bot: Red) -> List[Case]:
     """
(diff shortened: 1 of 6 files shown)


redbot/core/modlog.py
```diff
@@ -174,6 +174,14 @@
...
         await _conf.schema_version.set(3)
 
+    if from_version < 4 <= to_version:
+        # set latest_case_number
+        for guild_id, cases in (await _conf.custom(_CASES).all()).items():
+            if cases:
+                await _conf.guild(
+                    cast(discord.Guild, discord.Object(id=guild_id))
+                ).latest_case_number.set(max(map(int, cases.keys())))
+
```
mikeshardmind, line 183: you need to set the schema version to 4 in this block

mikeshardmind (changes requested):

redbot/core/modlog.py
Tobotimus: Good point, forgot about that :+1:

redbot/core/modlog.py
```diff
@@ -763,6 +767,7 @@ def from_json(cls, name: str, data: dict, **kwargs):
         message=None,
     )
     await _conf.custom(_CASES, str(guild.id), str(next_case_number)).set(case.to_json())
```
mikeshardmind: This needs to be inside a lock on the value for case numbers. We are fetching, awaiting something else, then incrementing here.

mikeshardmind (dismissed):
found something else...

redbot/core/modlog.py
Tobotimus: Good point

mikeshardmind (changes requested):
one last change from me after some consideration.

Let's go ahead and make `get_next_case_number` private and remove it from `__all__`

There's not a reason to expose this function outside of modlog, and the next release is one which allows for removing this from the public API.

Tobotimus:
@mikeshardmind I've actually gone ahead and removed the function all together. It was only used in one place, and that place is kind of easier to understand (w.r.t. lock usage) if it just uses config directly.

mikeshardmind (dismissed):

mikeshardmind (approved):

PR #2920, opened 2019-08-06
[Docs] Update linux install docs, redo venv docs

### LINKED ISSUE #2558: [Docs] incorrect pre-requirements for Fedora
the [docs](https://red-discordbot.readthedocs.io/en/v3-develop/install_linux_mac.html#centos-7-fedora-and-rhel) state that the following commands need to be run on Fedora and possibly RHEL

I'm going to adress every command on its own

> yum -y groupinstall development

the development group group is called `Development Tools` but its questionable if all those tools are really needed for this bot, if they are needed please correct me in a comment

> yum -y install https://centos7.iuscommunity.org/ius-release.rpm

the repo is designed for CentOS 7 and adding it to Fedora introduces a bunch of new packages which are not really needed and just slow down dnf (yum on Fedora is a dnf wrapper)

> sudo yum install zlib-devel bzip2 bzip2-devel readline-devel sqlite sqlite-devel \
openssl-devel xz xz-devel libffi-devel git2u java-1.8.0-openjdk

everything is fine here except git2u
Fedora already includes a recent version of git (git version 2.20.1 as of writing) so there is no reason at all to install the ius repo

I'd be willing to make a seperate pre-requirements list for Fedora needed

Tobotimus:
Some of our pre-req installation docs needed updating on Linux - this adds new sections for Fedora Linux and Debian/Raspbian Buster, and also removes some unnecessary pre-requirements from other distributions. These have all been tested on fresh VPSes, installing Red both in venvs and with --user, and they all seem to work.

Also, apparently the venv docs were too scary before. These changes try to make it clear that it's easier to use than users may think.

This also includes a little note to stop users from accidentally installing Python with pyenv after installing pre-requirements on Ubuntu.

Resolves #2558.


Jackenmen:
https://github.com/Cog-Creators/Red-DiscordBot/blob/2d99b403de8e7950f0935bb42cdf5f9c8a2b86d3/docs/install_linux_mac.rst#L261
It might not be clear for the user what means adding a flag, so perhaps we could improve this wording too?

mikeshardmind:
Note on why only outside venv: using `py -3.7` is venv-incompatible. py doesn't appear to recognize being in a venv

mikeshardmind (approved):

PR #2921, opened 2019-08-06
Stop saving JSON files with indents

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Stop saving json files with 4 indents, this will significantly reduce file size and improve `.set()` performance for config saves when using the JSON driver

### DIFF
--- a/changelog.d/2921.enhancement.rst
+++ b/changelog.d/2921.enhancement.rst
@@ -1,1 +1,0 @@
-JSON files are now stored without indentation.
--- a/changelog.d/2921.enhancement.rst
+++ b/changelog.d/2921.enhancement.rst
@@ -1,0 +1,1 @@
+JSON files are now stored without indentation.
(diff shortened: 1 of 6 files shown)


Tobotimus (changes requested):
I think we can keep the indents for the instance config files (in backup functions / data manager). This file is usually tiny and is rarely written to. It's harmless to have it indented.

So the indents only need to be removed in `red_json.py` :)

Drapersniper:
Got too happy nuking indents xD reverted and only applied to the json driver :+1:

changelog.d/2921.enhancement.rst
```diff
@@ -0,0 +1 @@
+JSON files are now stored without indentation.
```
Tobotimus: Maybe also just add why this was done, i.e. lower disk usage and better performance.

Also, the filename should be "enhance" rather than "enhancement"

Drapersniper:
Done, still not 100% clear on how much detail should be added to change logs.

Tobotimus:
@Drapersniper When writing changelogs, simply consider a user or a cog creator reading them. If they simply see "JSON files are now saved without indents", they might consider that annoying if they aren't aware of the space efficiency/performance gain :)

Tobotimus (approved):

PR #2924, opened 2019-08-08
Fixed small docstring inconsistency

DevilXD:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Noticed a small inconsistency where the docs say about raising a `KeyError`, but the class actually raises a `TypeError`. This PR changes the docs to match the code (rather than change the code to match the docs, which could break things).


changelog.d/2924.bugfix.rst
```diff
@@ -0,0 +1 @@
+Fixed Requires docs inconsistencies
```
Flame442: Since both `_IntKeyDict` and `_RulesDict` are private functions which do not show up in the docs, the changelog should be named `2924.misc.rst`. It falls under "any changes which don't have a user facing change, and don't belong in the changelog for users".

Flame442 (changes requested):

Cog-CreatorsBot (changes requested):

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

PR #2925, opened 2019-08-08
[Bank] Check recipient balance before completing transfer 

Drapersniper:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

### Description of the changes
Fix for #2923

PR #2926, opened 2019-08-08
[Bank] Allow bank managers to set the maximum allowed balance in the bank

Drapersniper:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

`[p]bankset maxbal` can be used to set the maximum bank balance

NOTE: This removes the `MAX_BALANCE` constant from the `bank` module

To get the maximum allowed balance use the new `bank.get_max_balance()` method


Drapersniper:
Lets merge this one AFTER #2925, these 2 conflict, once 2925 is merged I will update this one accordingly.

Drapersniper:
Conflicts have been resolved, this should  be ready to be tested/merged now

Tobotimus:
The upper limit on the balance was put in place to prevent an integer overflow on MongoDB. Specifically, this means the upper limit should always be 2^64 - 1 (although it's probably 2^32 - 1 on 32-bit systems running Mongo, Mongo isn't recommended on 32-bit systems)

mikeshardmind:
@Tobotimus The limit is still bound to that here as an upper limit (not made clear in the description)

redbot/cogs/economy/economy.py
```diff
@@ -625,7 +627,7 @@ def __init__(self, bot: Red):
     async def paydayamount(self, ctx: commands.Context, creds: int):
         """Set the amount earned each payday."""
         guild = ctx.guild
-        if creds <= 0 or creds > bank.MAX_BALANCE:
+        if creds <= 0 or creds > await bank.get_max_balance(ctx.guild):
             await ctx.send(_("Har har so funny."))
```
Flame442: This should probably be changed to an actual error message now that it is easier to accidentally set the payday amount to be higher than the max balance (along with L649)
<not a full review, I just wanted to point this out>

changelog.d/2926.breaking.1.rst
```diff
@@ -0,0 +1 @@
+Removed :cons:`bank.MAX_BALANCE`, use :meth:`bank.get_max_balance()` from now.
```
Flame442: You only need to have the <count> "should multiple entries for the same issue number **and type** be necessary."

redbot/core/bank.py
```diff
@@ -617,6 +627,74 @@ def _invalid_amount(amount: int) -> bool:
...
+            "Amount must be greater than zero and less than {max:,}.".format(max=_MAX_BALANCE)
+        )
+
+    if await is_global():
+        await _conf.max_balance.set(amount)
+    elif guild is not None:
+        await _conf.guild(guild).max_balance.set(amount)
+    else:
+        raise RuntimeError(
+            "Guild must be provided if setting the currency name of a guild-specific bank."
```
Flame442: Should be "..if setting the **maximum balance** of..."

redbot/cogs/economy/economy.py
```diff
@@ -550,14 +550,16 @@ def __init__(self, bot: Red):
                         "Slot cooldown: {slot_time}\n"
                         "Payday amount: {payday_amount}\n"
                         "Payday cooldown: {payday_time}\n"
-                        "Amount given at account registration: {register_amount}"
+                        "Amount given at account registration: {register_amount}\n"
+                        "Maximum allowed balance: {maximum_bal}"
```
Flame442, line 576: Should this be under economyset or bankset? The command to change it is under bankset.

redbot/cogs/bank/bank.py
```diff
@@ -128,4 +128,11 @@ def __init__(self, bot: Red):
         await bank.set_currency_name(name, ctx.guild)
         await ctx.send(_("Currency name has been set to: {name}").format(name=name))
 
+    @bankset.command(name="maxbal")
+    @check_global_setting_guildowner()
+    async def bankset_maxbal(self, ctx: commands.Context, *, amount: int):
+        """Set the maximum balance a user can get."""
+        await bank.set_max_balance(amount, ctx.guild)
```
Flame442: You raise a ValueError if the balance is invalid, however you do not catch this error in this command.

redbot/core/bank.py
```diff
@@ -617,6 +627,74 @@ def _invalid_amount(amount: int) -> bool:
...
+    if await is_global():
+        return await _conf.max_balance()
+    elif guild is not None:
+        return await _conf.guild(guild).max_balance()
+    else:
+        raise RuntimeError("Guild must be provided.")
+
+
+async def set_max_balance(amount: int, guild: discord.Guild = None) -> int:
+    """Set the currency name for the bank.
```
Flame442: Should be "Set the maximum balance for the bank."

redbot/core/bank.py
```diff
@@ -617,6 +627,74 @@ def _invalid_amount(amount: int) -> bool:
...
+
+    Parameters
+    ----------
+    amount : int
+        The new maximum balance.
+    guild : `discord.Guild`, optional
+        The guild to set the currency name for (required if bank is
```
Flame442: Should be "...max balance for..."

redbot/core/bank.py
```diff
@@ -617,6 +627,74 @@ def _invalid_amount(amount: int) -> bool:
     return name
 
 
+async def get_max_balance(guild: discord.Guild = None) -> int:
+    """Get the max balance for the bank.
+
+    Parameters
+    ----------
+    guild : `discord.Guild`, optional
+        The guild to get the currency name for (required if bank is
```
Flame442: Should be "...max balance for..."

redbot/core/bank.py
```diff
@@ -617,6 +627,74 @@ def _invalid_amount(amount: int) -> bool:
...
+    -------
+    int
+        The new maximum balance.
+
+    Raises
+    ------
+    RuntimeError
+        If the bank is guild-specific and guild was not provided.
+    ValueError
+        If the amount is invalid.
```
Flame442: You might want to be a bit more clear about what qualifies as "invalid".

redbot/core/bank.py
```diff
@@ -617,6 +627,74 @@ def _invalid_amount(amount: int) -> bool:
...
+
+    """
+    if not (0 <= amount <= _MAX_BALANCE):
+        raise ValueError(
+            "Amount must be greater than zero and less than {max:,}.".format(max=_MAX_BALANCE)
```
Flame442: This says it must be greater than zero, but the code says it must be greater or equal to zero.

Flame442 (changes requested):
The default bank balance can be higher than the max balance.

Cog-CreatorsBot (changes requested):

redbot/cogs/economy/economy.py
Drapersniper: I've added it to both now , i think given that economyset  sets the payday values etc its nice to haver it there.

redbot/core/bank.py
```diff
@@ -766,6 +768,10 @@ def _invalid_amount(amount: int) -> bool:
 
     """
     amount = int(amount)
+    max_bal = await get_max_balance(guild)
```
Flame442, line 770: I think raising an error would be better than silently changing the value. Make sure you also change the command to respond to whatever error is raised.

Flame442 (changes requested):
Looking much better, the only other thing is that the payday and starting balance values can be set higher than the max balance if the max balance is changed to lower than either of those values. It isn't really an issue since the balance will be set to where it should be at some point anyway, however it is a small inconsistency. If you can think of a good way to fix it, fix it. If not, don't stress about it.

redbot/core/bank.py
Drapersniper: > Looking much better, the only other thing is that the payday and starting balance values can be set higher than the max balance if the max balance is changed to lower than either of those values. It isn't really an issue since the balance will be set to where it should be at some point anyway, however it is a small inconsistency. If you can think of a good way to fix it, fix it. If not, don't stress about it.

As discussed in discord
We have 2 options:

1.  Set it the payday amount and initial balance amount  with the `bankset maxbal` command if the old value is higher than the current max

1.  Change the values in  `bankset` and `economyset` help message to the max value if they are higher than the current max value

My issue with 1 is that since this is in the Bank cog it shouldn't edit settings for the Economy Cog, with 2 it's technically False values

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

PR #2928, opened 2019-08-11
[Core Commands] remove set owner and set token

mikeshardmind:
Neither of these commands encourages good security, and their downsides are not comparable to the risks in keeping them.


palmtree5 (dismissed):

palmtree5 (approved):

changelog.d/2928.removal.rst
```diff
@@ -0,0 +1 @@
+The ``set owner`` and ``set token`` commands have been removed in favor of managing server side.
```
Jackenmen, line 1: Since we can call Discord's guild "servers" too, I found this changelog entry a little confusing when I first read it. Perhaps you could change it to "in favor of managing it through command line"? 
Edit: I guess this might not make sense currently that we can't actually do it through command line, so I don't know actually, maybe you should ignore my suggested change, though the current sentence is still a little confusing in my opinion 😄
```suggestion
The ``set owner`` and ``set token`` commands have been removed in favor of managing it through command line. 
```

mikeshardmind: I'm electing to leave it as-is for now. It is technically possible to change later if needed, but this is clear enough for this particular case.

PR #2929, opened 2019-08-12
[Warnings] Add setting for warn channel and DMing warns toggle

### LINKED ISSUE #2800: [Warnings] Option to disable DMing warns, option to warn in guild channel

Ianardo-DiCaprio:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This adds commands requested in #2800. It allows to toggle warns being sent to DM's, allows to toggle warns being sent to a channel on a server and has a command to set the channel the warns get sent to, if no channel is set it gets sent to the channel the [p]warn command was used in.


redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +83,45 @@ def __init__(self, bot: Red):
...
             await ctx.send(_("Custom reasons have been disabled."))
 
+    @warningset.command()
+    @commands.guild_only()
+    async def toggledm(self, ctx: commands.Context):
+        """Toggle warns being sent to the users DM's"""
+        guild = ctx.guild
+        toggle = await self.config.guild(guild).toggle_dm()
+        await self.config.guild(guild).toggle_dm.set(not toggle)
+        toggle = not toggle
```
DevilXD: This `not toggle` should either be in the if statement itself, or this line removed and the send messages swapped.

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +349,44 @@ def __init__(self, bot: Red):
...
+                await user.send(
+                    _("You have received a warning in {guild_name}.").format(
+                        guild_name=ctx.guild.name
+                    ),
+                    embed=em,
+                )
+        except discord.HTTPException:
+            await ctx.send(
+                _(
+                    "The warning couldn't be sent to the user because the user has DM's from server members turned off."
```
DevilXD: This really should be just suppressed, if they don't care about mod actions, that's entirely on them.

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +349,44 @@ def __init__(self, bot: Red):
...
-                embed=em,
+            warn_channel = self.bot.get_channel(await self.config.guild(guild).warn_channel())
+            if warn_channel:
+                channel = warn_channel
+            else:
+                channel = ctx.channel
+            await channel.send(_("{user} has been warned.").format(user=user.mention), embed=em)
+        if not dm and not toggle_channel:
+            await ctx.send(
+                _("You have warned the user but you have channel and DM warns turned off.")
```
DevilXD: I don't get the point of this. One shouldn't enforce those, if someone likes to have both of them disabled and still use warns, this should be possible without getting an annoying message on every usage.

DevilXD (changes requested):
Just couple of changes.

Also, from the original issue:
> (...) option to warn a user (i.e. ping them) in set guild channel and/or just warn that user in channel where the [p]warn command was issued.

This could be slightly hard to understand, at least for me. I've commented on the original issue, but basically - there already is an option to log warnings into the modlog channel. A second channel with warnings only could be made public tho, so I guess there is some logic behind having it like this ¯\\\_(ツ)\_/¯

redbot/cogs/warnings/warnings.py
```diff
@@ -301,19 +348,37 @@ def __init__(self, bot: Red):
...
+                    description=reason_type["description"],
+                )
+                em.add_field(name=_("Points"), value=str(reason_type["points"]))
+                await user.send(
+                    _("You have received a warning in {guild_name}.").format(
+                        guild_name=ctx.guild.name
+                    ),
+                    embed=em,
+                )
+        except discord.HTTPException:
```
DevilXD: Since the purpose of this is to suppress the exception, I recommend `contextlib.suppress()` and making sure the only thing inside the suppress block is the `await user.send(...)` part.

DevilXD (changes requested):

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,44 @@ def __init__(self, bot: Red):
...
+    @commands.guild_only()
+    async def toggledm(self, ctx: commands.Context):
+        """Toggle warns being sent to the users DM's"""
+        guild = ctx.guild
+        toggle = not await self.config.guild(guild).toggle_dm()
+        await self.config.guild(guild).toggle_dm.set(toggle)
+        if toggle:
+            await ctx.send(_("Warns will now be sent to users DM's"))
+        else:
+            await ctx.send(_("Warns will not be sent to users DM's"))
```
```suggestion
            await ctx.send(_("Warnings will no longer be sent to users in DMs."))
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,44 @@ def __init__(self, bot: Red):
...
 
+    @warningset.command()
+    @commands.guild_only()
+    async def toggledm(self, ctx: commands.Context):
+        """Toggle warns being sent to the users DM's"""
+        guild = ctx.guild
+        toggle = not await self.config.guild(guild).toggle_dm()
+        await self.config.guild(guild).toggle_dm.set(toggle)
+        if toggle:
+            await ctx.send(_("Warns will now be sent to users DM's"))
```
```suggestion
            await ctx.send(_("Warnings will now be sent to users in DMs."))
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,44 @@ def __init__(self, bot: Red):
         else:
             await ctx.send(_("Custom reasons have been disabled."))
 
+    @warningset.command()
+    @commands.guild_only()
+    async def toggledm(self, ctx: commands.Context):
+        """Toggle warns being sent to the users DM's"""
```
```suggestion
        """Toggle whether warnings should be sent to users in DMs."""
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,44 @@ def __init__(self, bot: Red):
...
+        await self.config.guild(guild).toggle_dm.set(toggle)
+        if toggle:
+            await ctx.send(_("Warns will now be sent to users DM's"))
+        else:
+            await ctx.send(_("Warns will not be sent to users DM's"))
+
+    @warningset.command()
+    @commands.guild_only()
+    async def warnchannel(self, ctx: commands.Context, channel: discord.TextChannel = None):
+        """Set the channel that warns get set to"""
```
```suggestion
        """Set the channel where warnings should be sent to.

        Leave empty to use the channel `[p]warn` command was called in.
        """
```

redbot/cogs/warnings/warnings.py
```diff
@@ -300,20 +348,36 @@ def __init__(self, bot: Red):
...
+            em = discord.Embed(
+                title=_("Warning from {user}").format(user=ctx.author),
+                description=reason_type["description"],
             )
-        except discord.HTTPException:
-            pass
+            em.add_field(name=_("Points"), value=str(reason_type["points"]))
+            warn_channel = self.bot.get_channel(await self.config.guild(guild).warn_channel())
+            if warn_channel:
+                channel = warn_channel
```
Jackenmen, line 386: If message about warning is sent in different channel, it would be good to still know that command worked from the channel command was sent in, so adding tick here as well would help that.
```suggestion
                channel = warn_channel
                await ctx.tick()
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,44 @@ def __init__(self, bot: Red):
...
+        """Set the channel that warns get set to"""
+        guild = ctx.guild
+        if channel:
+            await self.config.guild(guild).warn_channel.set(channel.id)
+            await ctx.send(
+                _("The warn channel has been set to {channel}").format(channel=channel.mention)
+            )
+        else:
+            await self.config.guild(guild).warn_channel.set(channel)
+            await ctx.send(_("The warn channel has been reset."))
```
```suggestion
            await ctx.send(_("Warnings will now be sent in the channel command was used in."))
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,44 @@ def __init__(self, bot: Red):
...
+
+    @warningset.command()
+    @commands.guild_only()
+    async def togglechannel(self, ctx: commands.Context):
+        """Toggle if the warn should be sent to a channel"""
```
```suggestion
        """
        Toggle if warnings should be sent to a channel set with `[p]warningset warnchannel`
        """
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,44 @@ def __init__(self, bot: Red):
...
+        guild = ctx.guild
+        toggle = await self.config.guild(guild).toggle_channel()
+        await self.config.guild(guild).toggle_channel.set(not toggle)
+        if not toggle:
+            await ctx.send(_("Warns will now be sent to channels"))
```
```suggestion
            await ctx.send(_("Warnings will now be sent to set warn channel."))
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,44 @@ def __init__(self, bot: Red):
...
+        else:
+            await ctx.send(_("Warns will not be sent to channels"))
```
```suggestion
            await ctx.send(_("Warns will no longer be sent in the channel."))
```

changelog.d/warnings/2929.enhance.rst.txt
```diff
@@ -0,0 +1,2 @@
+Added features to mod to be able to toggle warns being sent to both DM's and channel.
+Added a command to set the channel the warns get sent to. (defaults to the channel the warn was issued in).
```
Jackenmen: This file should be called `2929.enhance.rst` not `2929.enhance.rst.txt`

Jackenmen (changes requested):
Looks good, I know it looks like a lot of review comments, but these are all just minor string improvements =)

changelog.d/warnings/2929.enhance.rst.txt
Jackenmen: I saw you marked this as resolved, but from the looks of it you renamed wrong file ;)

changelog.d/warnings/2929.enhance.rst
```diff
@@ -0,0 +1,2 @@
+Added features to mod to be able to toggle warns being sent to both DM's and channel.
```
Drapersniper: Dont take it as a definitive thing but according to https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/.github/CONTRIBUTING.md#47-using-towncrier

This should be `2929.feature.rst`

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,49 @@ def __init__(self, bot: Red):
...
+    async def warnchannel(self, ctx: commands.Context, channel: discord.TextChannel = None):
+        """Set the channel where warnings should be sent to.
+
+        Leave empty to use the channel `[p]warn` command was called in.
+        """
+        guild = ctx.guild
+        if channel:
+            await self.config.guild(guild).warn_channel.set(channel.id)
+            await ctx.send(
+                _("The warn channel has been set to {channel}").format(channel=channel.mention)
```
Drapersniper: This should be:
```suggestion
                _("The warn channel has been set to {channel}.").format(channel=channel.mention)
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,49 @@ def __init__(self, bot: Red):
...
+            )
+        else:
+            await self.config.guild(guild).warn_channel.set(channel)
+            await ctx.send(_("Warnings will now be sent in the channel command was used in."))
+
+    @warningset.command()
+    @commands.guild_only()
+    async def togglechannel(self, ctx: commands.Context):
+        """
+        Toggle if warnings should be sent to a channel set with `[p]warningset warnchannel`
```
Drapersniper: This should be:
```suggestion
        Toggle if warnings should be sent to a channel set with `[p]warningset warnchannel`.
```

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,49 @@ def __init__(self, bot: Red):
...
+        """
+        guild = ctx.guild
+        toggle = await self.config.guild(guild).toggle_channel()
+        await self.config.guild(guild).toggle_channel.set(not toggle)
+        if not toggle:
+            await ctx.send(_("Warnings will now be sent to set warn channel."))
```
Drapersniper: As a user i'd expect this to show a name/mention of the channel
`            await ctx.send(_("Warnings will now be sent to set warn channel."))`

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,49 @@ def __init__(self, bot: Red):
...
+        else:
+            await ctx.send(_("Warns will no longer be sent in the channel."))
```
Drapersniper: As a user i'd expect this to show a name/mention of the channel
`await ctx.send(_("Warns will no longer be sent in the channel."))`

redbot/cogs/warnings/warnings.py
```diff
@@ -1,6 +1,7 @@
 from collections import namedtuple
 from typing import Union, Optional
 
+import contextlib
```
Drapersniper: Note really a necesity but...

```
from collections import namedtuple
from typing import Union, Optional

import contextlib
import discord
```
should be 
```
import contextlib
from collections import namedtuple
from typing import Union, Optional

import discord
```

This one is just a nitpick for import organization 😈

redbot/cogs/warnings/warnings.py
```diff
@@ -300,20 +353,37 @@ def __init__(self, bot: Red):
         await member_settings.total_points.set(current_point_count)
 
         await warning_points_add_check(self.config, ctx, user, current_point_count)
-        try:
+        dm = await self.config.guild(ctx.guild).toggle_dm()
+        if dm:
```
Drapersniper, line 363: The DM feature .. are we expecting it to only send it via DM or send DM as an aditional thing?

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,49 @@ def __init__(self, bot: Red):
...
 
+    @warningset.command()
+    @commands.guild_only()
+    async def toggledm(self, ctx: commands.Context):
+        """Toggle whether warnings should be sent to users in DMs."""
+        guild = ctx.guild
+        toggle = not await self.config.guild(guild).toggle_dm()
+        await self.config.guild(guild).toggle_dm.set(toggle)
+        if toggle:
+            await ctx.send(_("Warnings will now be sent to users in DMs."))
```
Drapersniper: I know it's not ideal but we shouldn't say this in the affirmative, perhaps it should be changed to `"I will try to send warning to user in DMs."` since users can for example block the bot - it's not guaranteed that warned users will get a warning.
Same applied to the toggle off message.

Drapersniper (changes requested):

redbot/cogs/warnings/warnings.py
Jackenmen: We expect it to send to DM as an additional thing, this is a separate setting.

redbot/cogs/warnings/warnings.py
```diff
@@ -300,20 +353,37 @@ def __init__(self, bot: Red):
...
-        except discord.HTTPException:
-            pass
+            em.add_field(name=_("Points"), value=str(reason_type["points"]))
+            warn_channel = self.bot.get_channel(await self.config.guild(guild).warn_channel())
+            if warn_channel:
+                channel = warn_channel
+                await ctx.tick()
+            else:
+                channel = ctx.channel
+            await channel.send(_("{user} has been warned.").format(user=user.mention), embed=em)
```
Jackenmen: Sinbad pointed out that this needs handling for cases when bot can't send the message in set channel.
Preferred way of doing this would be first checking if bot has permission to send messages in the channel and also catching any HTTP exception that occurs when sending the message.
```suggestion
            if channel.permissions_for(guild.me).send_messages:
                with contextlib.suppress(discord.HTTPException):
                    await channel.send(
                        _("{user} has been warned.").format(user=user.mention), embed=em
                    )
```

Jackenmen (changes requested):
One thing I didn't notice in first review.
Once this and UX stuff pointed out by Draper gets handled, this should be ready to go.

FixedThink:
Would it be possible to inform the author about the DM not being sent if it's enabled? 
If the DM fails, the kick/ban will still succeed, but the bot will send a message like `"This user did not receive your case as they have either blocked the bot or disabled DMs."` I think this will be much better then just sending a DM without any sort of feedback as of what happened to it.

The reason for this is that (to me at least) it tends to happen that someone *says* they have never been informed about why they're kicked/muted/whatever, but that there is no way for a staff member to actually verify whether that's true or whether they're 🙈. In other words it would ease a lot of moderation arguments and such, *especially* with warnings.

Jackenmen:
@FixedThink It would be possible but in my opinion this shouldn't be part of this PR, I would advice making a separate issue with a request for this enhancement.

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,54 @@ def __init__(self, bot: Red):
...
+        await self.config.guild(guild).toggle_channel.set(not toggle)
+        channel = await self.config.guild(guild).warn_channel()
+        if not toggle:
+            await ctx.send(
+                _("Warnings will now be sent to {channel}.").format(channel=channel.mention)
+            )
+        else:
+            await ctx.send(
+                _("Warns will no longer be sent to {channel}.").format(channel=channel.mention)
+            )
```
Jackenmen: `channel.mention` won't work here because:
a) `channel` variable isn't a channel object
b) warn channel might not be set at all - bot should then say it will send warnings in the channel command was used in

Jackenmen (changes requested):
This should be the only needed change before this can be approved.

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,59 @@ def __init__(self, bot: Red):
...
+        await self.config.guild(guild).toggle_channel.set(not toggle)
+        channel = self.bot.get_channel(await self.config.guild(guild).warn_channel())
+        if not toggle:
+            if channel:
+                await ctx.send(
+                    _("Warnings will now be sent to {channel}.").format(channel=channel.mention)
+                )
+            else:
+                await ctx.send(
+                    _("Toggle channel has been enabled but a channel hasn't been set.").format(
```
Jackenmen: This should say that warnings will now be sent to the channel command was used in, instead of saying that channel hasn't been set.
And no need to use `.format` if there's nothing to be replaced.

Jackenmen (changes requested):

redbot/cogs/warnings/warnings.py
```diff
@@ -76,6 +84,55 @@ def __init__(self, bot: Red):
...
+        toggle = await self.config.guild(guild).toggle_channel()
+        await self.config.guild(guild).toggle_channel.set(not toggle)
+        channel = self.bot.get_channel(await self.config.guild(guild).warn_channel())
+        if not toggle:
+            if channel:
+                await ctx.send(
+                    _("Warnings will now be sent to {channel}.").format(channel=channel.mention)
+                )
+            else:
+                await ctx.send(_("Warnings will now be sent to this channel."))
```
Jackenmen: Like I said, it should say that warnings will be sent to the channel warn command was used in, similarly to what's said in line 114. Current string isn't true.

Jackenmen (changes requested):

Jackenmen (approved):

PR #2931, opened 2019-08-12
[Mod] Features requested for kick/ban.

Ianardo-DiCaprio:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Added features to kick and ban requested in #2649 . This will now send an embed to the user when they are kicked/banned from a server detailing if it was a kick or ban, from what server and the reason for the kick/ban.


redbot/cogs/mod/kickban.py
```diff
@@ -174,6 +174,30 @@ class KickBanMixin(MixinMeta):
             return
         audit_reason = get_audit_reason(author, reason)
         try:
+            if reason == None:
+                reason = "No reason was given."
+            else:
+                reason = reason
+            self == self.bot.get_cog("Mod")
+            toggle = await self.settings.guild(guild).toggle_dm()
+            try:
```
DevilXD: This entire try-except is unnecessary if you're already using suppress inside.

redbot/cogs/mod/kickban.py
```diff
@@ -174,6 +174,30 @@ class KickBanMixin(MixinMeta):
             return
         audit_reason = get_audit_reason(author, reason)
         try:
```
DevilXD, line 188: This try shouldn't really be here if you're sending the message before the kick, since there might be a situation where the message will send successfully, and then the kick itself will fail.
Since there already is a role hierarchy check on lines 163-171, and proper decorators for bot and user permissions are used, the `except` case with `discord.Forbidden` should be removed since it will never have a chance to run after the changes (if I'm wrong on this one, please correct me - hierarchy and permissions should be the only cause of Forbidden here, and these are covered).

In addition, the `try` statement should be moved to right before the kick happens, to make sure it's guarding the right / intended piece of code.

redbot/cogs/mod/kickban.py
```diff
@@ -213,11 +237,29 @@ class KickBanMixin(MixinMeta):
...
         Minimum 0 days, maximum 7. Defaults to 0."""
+        author = ctx.author
+        guild = ctx.guild
+        if reason == None:
+            reason = "No reason was given."
+        else:
+            reason = reason
+        self == self.bot.get_cog("Mod")
+        toggle = await self.settings.guild(guild).toggle_dm()
+        try:
```
DevilXD: Same deal as before, unnecessary try-except with suppress already inside.

DevilXD (changes requested):

redbot/cogs/mod/kickban.py
Jackenmen: Unless I'm getting it wrong, kick can't fail because the command has a `@commands.bot_has_permissions(kick_members=True)` check.

DevilXD (dismissed):
Looks good to me now 👌

redbot/cogs/mod/kickban.py
```diff
@@ -173,6 +173,18 @@ class KickBanMixin(MixinMeta):
             await ctx.send(_("I cannot do that due to discord hierarchy rules"))
             return
         audit_reason = get_audit_reason(author, reason)
+        if reason == None:
```
Jackenmen: `None` is singleton so you should use identity check
```suggestion
        if reason is None:
```

redbot/cogs/mod/kickban.py
```diff
@@ -173,6 +173,18 @@ class KickBanMixin(MixinMeta):
...
+            reason = "No reason was given."
+        self == self.bot.get_cog("Mod")
+        toggle = await self.settings.guild(guild).toggle_dm()
+        if toggle == True:
```
Jackenmen: `True` is singleton so you should either use identity check or since it's a bool, you can just use `if toggle:`
```suggestion
        if toggle:
```

redbot/cogs/mod/kickban.py
```diff
@@ -213,11 +225,24 @@ class KickBanMixin(MixinMeta):
 
         If days is not a number, it's treated as the first word of the reason.
         Minimum 0 days, maximum 7. Defaults to 0."""
+        author = ctx.author
+        guild = ctx.guild
+        if reason == None:
+            reason = "No reason was given."
+        self == self.bot.get_cog("Mod")
+        toggle = await self.settings.guild(guild).toggle_dm()
+        if toggle == True:
```
Jackenmen: `True` is singleton so you should either use identity check or since it's a bool, you can just use `if toggle:`
```suggestion
        if toggle:
```

redbot/cogs/mod/kickban.py
```diff
@@ -213,11 +225,24 @@ class KickBanMixin(MixinMeta):
 
         If days is not a number, it's treated as the first word of the reason.
         Minimum 0 days, maximum 7. Defaults to 0."""
+        author = ctx.author
+        guild = ctx.guild
+        if reason == None:
```
Jackenmen: `None` is singleton so you should use identity check
```suggestion
        if reason is None:
```

Jackenmen (commented):
Few small irks

Ianardo-DiCaprio:
Are you fucking kidding me.... Fuck conflicted merges....

Drapersniper:
👀  You built this off my Music PR branch? may be worth mentioning that in the future just in case this was merged before the audio PR is

PR #2933, opened 2019-08-13
[Converters] Always raise BadArgument with parse_timedelta

laggron42:
### Type

- [x] Bugfix (I guess)
- [ ] Enhancement
- [ ] New feature

When using `redbot.core.commands.converter.parse_timedelta`, if the format given is incorrect, the converter doesn't raise `BadArgument` but returns `None` instead. This behaviour is a problem when using `typing.Optional` or `commands.Greedy` since the commands module doesn't know the converstion failed.

The case that made me create this pull request is the following: I'm trying to make a mute command with an optional duration (`[p]mymute <members> [time] <reason>`) using `typing.Optional`
```py
from typing import Optional
from redbot.core.commands.converter import parse_timedelta

@commands.command()
async def mymute(self, ctx, member: discord.Member, duration: Optional[parse_timedelta], *, reason: str):
    pass
```
Here if the user tries this: `[p]mymute @Laggron a potato reason`, the length of the mute will be undefined, as desired, but the reason will be "potato reason" (instead of "a potato reason"). By returning `None` instead of raising `BadArgument`, the commands module will consider that the convertion succeeded, therefore removing the first word of the reason.

### DIFF
--- a/redbot/core/commands/converter.py
+++ b/redbot/core/commands/converter.py
@@ -49,7 +49,7 @@ def parse_timedelta(
     maximum: Optional[timedelta] = None,
     minimum: Optional[timedelta] = None,
     allowed_units: Optional[List[str]] = None,
-) -> Optional[timedelta]:
+) -> timedelta:
     """
     This converts a user provided string into a timedelta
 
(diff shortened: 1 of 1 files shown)


mikeshardmind:
Use the converter interface instead, this is a functional interface and is not suitable for use as a converter.

PR #2934, opened 2019-08-13
Begin work on reducing Red's potential bad API calls

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Adds a new exception type for our use. 
Changes the potential exceptions raised by `context.send` (Breaking)
Logs unavoidable 403s raised by `context.send`, `context.tick`, `context.react_quietly`

see <https://discordapp.com/channels/81384788765712384/381871767846780928/610143340238733322> (discord API server announcement) for details on why now on this.

Even without this though, we should be reducing the bad calls we might make.

### DIFF
--- a/changelog.d/2934.breaking.rst
+++ b/changelog.d/2934.breaking.rst
@@ -1,0 +1,1 @@
+Red's context.send may raise a different new exception type in an effort to reduce bad API calls.
--- a/changelog.d/2934.enhance.rst
+++ b/changelog.d/2934.enhance.rst
@@ -1,0 +1,1 @@
+Red logs some instances of bad API calls. This may be enhanced further later.
--- a/changelog.d/2934.feature.rst
+++ b/changelog.d/2934.feature.rst
@@ -1,0 +1,1 @@
+New exception type added for handling prevented bad API calls
--- a/redbot/core/commands/context.py
+++ b/redbot/core/commands/context.py
@@ -80,7 +80,9 @@ class Context(commands.Context):
                     log.warning(f"User id: {self.author.id} was (unsuccessfully) sent a DM")
                 raise
         else:
-            raise PreventedAPIException()
+            raise PreventedAPIException(
+                f"Missing `send_messages` in channel with id {self.channel.id}"
+            )
 
     async def send_help(self, command=None):
         """ Send the command help message. """
--- a/redbot/core/commands/context.py
+++ b/redbot/core/commands/context.py
@@ -1,9 +1,12 @@
 import asyncio
 import contextlib
+import logging
 from typing import Iterable, List, Union
+
 import discord
 from discord.ext import commands
 
+from ..errors import PreventedAPIException
 from .requires import PermState
 from ..utils.chat_formatting import box
 from ..utils.predicates import MessagePredicate
@@ -12,6 +15,8 @@ from ..utils import common_filters
 TICK = "\N{WHITE HEAVY CHECK MARK}"
 
 __all__ = ["Context"]
+
+log = logging.getLogger("red")
 
 
 class Context(commands.Context):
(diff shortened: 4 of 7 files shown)


mikeshardmind:
This should be good for review. I added a little bit to help cog devs track it down if running in debug. Beyond that, if this exception goes unhandled, it really should do nothing since we've already determined we'd be trying to be doing something we don't have perms for.

Rapptz:
This seems like a waste. Jake and I have discussed this heavily internally and you're more likely to get banned by 429s than 403s when sending messages which makes the permission check rather strange and inconsistent with other send(...) methods for no good reason.  Not to mention the library handles the 429 case already so you're not going to get banned.

Everyone who got banned from 403 spams got it from spamming audit logs. If you want to investigate 403s then do so there IMO.

mikeshardmind:
> Everyone who got banned from 403 spams got it from spamming audit logs. If you want to investigate 403s then do so there IMO.

This is an initial part of making the entire bot look before leap with regard to discord's API. Anything we can do to reduce the API impact of Red is a good thing. The only reason I started with context is because it also helps out with 3rd party code which may not be as well written as the core bot.

Just because the example case which caused this announcement to happen was a bot spamming audit logs without perms doesn't mean other pathological cases shouldn't be prevented.

> This seems like a waste. Jake and I have discussed this heavily internally and you're more likely to get banned by 429s than 403s when sending messages which makes the permission check rather strange and inconsistent with other send(...) methods for no good reason. 

The official announcement is pretty clear. There's no stated difference between the bad calls. you get only so many in an hour. If for some reason the official announcement is incorrect and you have knowledge of that, I'd suggest you talk with Jake again and have him clarify it.

DevilXD:
Could this exception inherit from `discord.Forbidden`? Since it's infeasible to track 429 or 401 here, this exception will only be used to check for correct permissions before trying to do the action, and both - trying to do the action and failing because of missing permissions - and - trying to do the action and failing on the permission check - should cause a `discord.Forbidden` (or some subclass of it ofc) to be raised. This would also make it a non-breaking change.

mikeshardmind:
I don't really like the idea of inheriting from discord.Forbidden. It has attributes which we should not be populating, and removing those attributes but faking the subclass is a liskov violation. Even if we were to populate them, we would need to populate them with a compatible type for it to be compatible. An example of some of these attributes is actually visible in the PR, currently being used to log unavoidable 403s separately so that bot owners have a means to handle someone causing those frequently (may be enhanced with some automatic things later as well)

mikeshardmind:
Following up on Danny's comment about the audit logs, I actually found a place where it's possible we hit the audit logs with 403s, though the conditions are a little contrived. It's been addressed in a commit here.

Following up with the exception type, I've gone ahead and put (in it's own commit for easy removal) what I think that should look like if we absolutely insist on inheriting from discord.Forbidden here, but I'm not entirely convinced doing so is a good idea. This includes manual creation of a ClientResponse object

Method for manually creating the ClientResponse object safely is a slight variation on what aiohttp itself uses for it's tests.

mikeshardmind:
Exception type has been reverted.

Per discussion in discord, we're gonna hope we can get some more info prior to 3.2

There has been no official assurance made that 403s which occur as a result of malicious actors (eg. user blocking bot, using command which reacts) wont cause issues.

Without such assurances, it would not be difficult to have 17 bad requests per second caused by a handful of userbots if someone wanted to take down a public instance (as of the enhanced limits)

However, with **official** assurances that this is not how discord will handle it, we can simplify a lot of this. 

This is on hold for now, but should be included in 3.2 is we don't have those assurances by then.

mikeshardmind:
There's a comment from Discord Jake in the reddit thread which was linked in that announcement and assurances that they will handle the abuse case should it arise as an issue. 

With that in mind, I'm closing this. Other changes may still be prudent, but large changes here shouldn't be needed.

#2944 seems like a less drastic middle ground on addressing the issue, though it too might be unneeded.

PR #2939, opened 2019-08-18
Bump dependencies (no more SSL errors)

Tobotimus:
Also made the Makefile work slightly nicer with other tools, e.g. IDE run configurations, by allowing the python executable to be set as an env var.


mikeshardmind (approved):

mikeshardmind (changes requested):
Conflicting currently..

PR #2943, opened 2019-08-21
[Mod] Fix modset deletedelay

### LINKED ISSUE #2942: Modset deletedelay behavior, 3.1.5
# Other bugs
This bug happens with the modset's deletedelay feature enabled, I have submitted about this on a previous version which was 3.0.2 on python 3.6.8 @ [#2705](https://github.com/Cog-Creators/Red-DiscordBot/issues/2705) 

Just normally chat with friends in a public server, though some users happen t use the prefix [p] in front of their messages and it just deletes.

Ignored message as it's not a command

#### What actually happened?

The bot seems to registers any messages with the prefix as a command, so it just deletes any messagess as `[p]whateverMessage is this`

Red's Version: 3.1.5
Python Version: 3.7.3

1. enable `[p]modset delaydelete X`
2. type whatever message with the prefix in the front `[p]whateverhere`
3. auto deleted?

Example screenshots:
https://i.shikashi.me/BYX0o
https://i.shikashi.me/ewyny

Note: I have tried this with cogs unloaded and loaded. (Looks like a core behavior)

DiscordLiz:
fixes #2942

### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Was broken for a while. Didn't seem broken due to another bug fixed in #2768 masking the issue.

### DIFF
--- a/changelog.d/mod/2924.fix.rst
+++ b/changelog.d/mod/2924.fix.rst
@@ -1,0 +1,1 @@
+``modset deletedelay`` doesn't delete non-commands.
--- a/redbot/cogs/mod/movetocore.py
+++ b/redbot/cogs/mod/movetocore.py
@@ -21,10 +21,12 @@ class MoveToCore(MixinMeta):
     async def on_command_completion(self, ctx: commands.Context):
         await self._delete_delay(ctx)
 
-    # noinspection PyUnusedLocal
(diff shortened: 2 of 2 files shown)


mikeshardmind (approved):
Thanks!

PR #2944, opened 2019-08-21
Alternative fix for reaction issue

DiscordLiz:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature

Removes `Context.react_quietly()`
Changes `Context.tick()` to just send a message instead.

This is mostly in replacement of #2934 

I don't think the changes in #2934 are the right direction for addressing the problem.

The `Context.send` change in that PR also seems like a non-issue if people check for permissions before doing things. The only case where it wouldn't be is DMs and if someone can make the context channel a DM channel, then the bot should be able to return the DM.

### DIFF
--- a/changelog.d/2944.removal.rst
+++ b/changelog.d/2944.removal.rst
@@ -1,0 +1,1 @@
+Removed ``Context.react_quietly``
--- a/redbot/core/commands/context.py
+++ b/redbot/core/commands/context.py
@@ -1,6 +1,6 @@
 import asyncio
 import contextlib
-from typing import Iterable, List, Union
+from typing import Iterable, List, Union, Optional
 import discord
 from discord.ext import commands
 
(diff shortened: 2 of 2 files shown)


DiscordLiz:
This doesn't do anything about `ctx.send`. I don't think anything needs to happen there.

Tobotimus:
I'd prefer if we used a reaction by default and send the message only if perms are missing. We should use cached permissions where possible to prevent 403s. In this case, maybe return `True` if a reaction was sent and return the message when a message was sent.

`react_quietly` doesn't need to be removed IMO, just make it check for `add_reaction` perms first. This'll prevent the vast majority of 403s. It's a utility, and if we really were hellbent on removing it (which I'm not), it should go through a deprecation cycle first.

I'm aware of some concerns with regards to users blocking the bot and then abusing the 403 limit. My stance on this initially was, if it's possible, we should probably try and account for it. But given that Jake has basically said "don't worry about that" (although I understand @mikeshardmind interpreted it differently), I'm happy to let it slide, and I'd also consider the sudden fear of adding reactions overzealous.

mikeshardmind:
To clarify my concerns with it.

The comments from Jake can be summed up as

1. The bans for 403 thresholds are automatic
2. They have no means to manually unban, you have to wait the hour.
3. You can't check in advance in some cases around users blocking a bot.
4. They are unwilling to exclude the cases you can't know about *unless* it becomes an issue.
5. If it becomes an issue they will handle it.

This set of premises leads to a conclusion of someone can get bots kicked off of discord for hours at a time, and if they don't at least log the 403s, they won't even have recourse when speaking with discord about it.

While Jake did say it would be handled, he also shrugged off the possibility for abuse, which frankly annoyed me more than a little here. It's not the first time discord has taken the "it's just a theoretical abuse case" stance with something. (and in other cases, it went on to be abused)

__

With that said, if we are willing to trust discord at face value on handling it, then we shouldn't be removing this, but we should be logging failures and checking what we can before doing anything with it.

If you'd like to make those changes with regards to this, that's fine, otherwise someone else can ensure it is included.

PR #2960, opened 2019-08-28
[Permissions] Make who_or_what a required arg in syntax

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

At the moment the syntax says who_or_what is not required in the command.
The docstring conflicts this, saying that who_or_what is needed.
Actual behaviour also reflects this, with who_or_what being required.
This affects `addglobalrule`, `addguildrule`, `removegloalrule` and `removeguildrule`.


mikeshardmind (changes requested):
This is not an appropriate fix for this. This breaks allowing multiple rules being set at once.

Vexed01:
This is beyond my ability. When I opened this PR, I wasn't aware multiple rules *could* be set at once.

This has moved into an issue, #2991

PR #2961, opened 2019-08-28
Move some commands from Economy cog to Bank cog

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Move `[p]bank` from Economy to Bank
Move `[p]leaderboard` from Economy to Bank
Move `[p]economyset registerbalance` to `[p]bankset initialbalance` 

Rename `[p]bank balance` to `[p]balance check`
Rename `[p]bank set` to `[p]balance set`
Rename `[p]bank transfer` to `[p]balance transfer`

This one builds on #2926  I'll mark this one ready for review once that one has been merged


Drapersniper:
Marking as ready for review ... but need to fix conflicts, I’ll hopefully have time to later today if not tomorrow!

Drapersniper:
Conflicts have been resolved, ready for review

Jackenmen:
Unless `[p]balance` and `[p]balance Username/id` will work (as in show balance), this change makes this IMO even less user-friendly. I think `[p]balance check` just isn't really something user would think of using at first.

Aaand with `[p]bank balance` you can at least alias it to `[p]balance` which is other reason why I wouldn't want this command to be changed `[p] balance check`.

PR #2965, opened 2019-08-29
Allow keeping data in `redbot-setup delete`

### LINKED ISSUE #2962: [Setup] `redbot-setup delete` should allow to keep Red's data within database
# Feature request

Deleting instance from Red shouldn't require deleting Red's data from database - especially that it doesn't require that for JSON backend.

Tobotimus:
Closes #2962.

### DIFF
--- a/changelog.d/2962.enhance.rst
+++ b/changelog.d/2962.enhance.rst
@@ -1,0 +1,1 @@
+```redbot-setup delete`` now has the option to leave Red's data untouched on database backends.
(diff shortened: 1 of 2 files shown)


Jackenmen (commented):
This works as expected, but I feel like it might be a little unclear how deleting instance's data and instance's entire datapath differ when using JSON instance.
Edit: Another thing that I thought of, do you think we should tell the user that database has to be running if they want to delete data from it? 🤔

mikeshardmind (approved):

PR #2966, opened 2019-08-29
Fix issues with embed_requested()

DJtheRedstoner:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Remove if statement that causes embed_requested() not to always return the correct user set value.

### DIFF
--- a/changelog.d/2966.bugfix.rst
+++ b/changelog.d/2966.bugfix.rst
@@ -1,0 +1,1 @@
+Make embedset user only affect DM's
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -203,16 +203,12 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):  # pylint: d
         bool
             :code:`True` if an embed is requested
         """
-        if isinstance(channel, discord.abc.PrivateChannel) or (
-            command and command == self.get_command("help")
-        ):
-            user_setting = await self.db.user(user).embeds()
-            if user_setting is not None:
-                return user_setting
-        else:
-            guild_setting = await self.db.guild(channel.guild).embeds()
-            if guild_setting is not None:
-                return guild_setting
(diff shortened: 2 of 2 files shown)


Tobotimus (dismissed):
This will break `embed_requested()` in DMs, because PrivateChannels don't have a guild attribute.

The current behaviour is that the guild setting takes priority over the user setting, and as far as I can tell, this behaviour is intended. Perhaps we should open up a discussion on whether the embed setting should be personalised, or whether the guild setting should take priority?

DJtheRedstoner:
What does need to be fixed is that the user setting should always affect help but only affects help run with [p]help and not help sent by other commands

mikeshardmind:
> What does need to be fixed is that the user setting should always affect help but only affects help run with [p]help and not help sent by other commands

I can't replicate the described issue. `ctx.send_help()` sent from a command actually invokes the same exact help logic as running `[p]help commandname` as a command with respect to the embed setting.  (they both call: `ctx.bot.send_help_for` with the context preserved)

DJtheRedstoner:
Screen Shot 2019-08-29 at 7 10 46 PM

It only works in a guild channel.

mikeshardmind:
As Toby said,
> The current behaviour is that the guild setting takes priority over the user setting, and as far as I can tell, this behaviour is intended. Perhaps we should open up a discussion on whether the embed setting should be personalised, or whether the guild setting should take priority?



Above image shows this, but these changes don't adequately handle this.

I'm also not sure we should change the priority here. Guild owners have a reason to expect help to look consistent within their own server.

DJtheRedstoner:
I did not realize that the intended behaviour was guild having priority over user, but it does not make sense to me why user should only have priority over guild in help (and there IS a bug with this).

Is the intended priority Guild > User > Global?

My suggestion is to make the user setting ONLY apply in DM's

mikeshardmind (approved):
Much more acceptable way to handle this.

PR #2967, opened 2019-08-30
Rename `bot.db` as `bot._config`

mikeshardmind:
- Continues work towards strong version guarantees
  - Added methods for cog use for a few things which were previously
  only accessible via direct access.
  - Retained private use in a few internal use locations, though most
  methods were updated away from this.
  - Updated documentation for shared api token users

### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

### DIFF
--- a/changelog.d/2967.breaking.rst
+++ b/changelog.d/2967.breaking.rst
@@ -1,1 +1,11 @@
 The main bot config is no longer directly accessible to cogs. New methods have been added for use where this is concerned.
+New methods for this include
+
+  - ``bot.get_shared_api_tokens``
+  - ``bot.get_shared_api_tokens``
+  - ``bot.get_embed_color``
+  - ``bot.get_embed_colour``
+  - ``bot.get_admin_roles``
+  - ``bot.get_admin_role_ids``
+  - ``bot.get_mod_roles``
+  - ``bot.get_mod_role_ids``
--- a/changelog.d/2967.breaking.rst
+++ b/changelog.d/2967.breaking.rst
@@ -2,7 +2,7 @@ The main bot config is no longer directly accessible to cogs. New methods have b
 New methods for this include
 
   - ``bot.get_shared_api_tokens``
-  - ``bot.get_shared_api_tokens``
+  - ``bot.set_shared_api_tokens``
   - ``bot.get_embed_color``
   - ``bot.get_embed_colour``
   - ``bot.get_admin_roles``
--- a/changelog.d/2967.breaking.rst
+++ b/changelog.d/2967.breaking.rst
@@ -1,0 +1,1 @@
+The main bot config is no longer directly accessible to cogs. New methods have been added for use where this is concerned.
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -332,7 +332,7 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):  # pylint: d
         This will not clear existing values not specified.
         """
 
-        async with self._config.api.tokens.get_attr(service_name)() as method_abuse:
+        async with self._config.api_tokens.get_attr(service_name)() as method_abuse:
             method_abuse.update(**tokens)
 
     async def get_context(self, message, *, cls=commands.Context):
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -6,7 +6,7 @@ from collections import Counter
 from enum import Enum
 from importlib.machinery import ModuleSpec
 from pathlib import Path
-from typing import Optional, Union, List
+from typing import Optional, Union, List, Dict
 
 import discord
 from discord.ext.commands import when_mentioned_or
@@ -36,11 +36,11 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):  # pylint: d
 
     def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
         self._shutdown_mode = ExitCodes.CRITICAL
-        self.db = Config.get_core_conf(force_registration=True)
+        self._config = Config.get_core_conf(force_registration=False)
         self._co_owners = cli_flags.co_owner
         self.rpc_enabled = cli_flags.rpc
         self._last_exception = None
-        self.db.register_global(
+        self._config.register_global(
             token=None,
             prefix=[],
             packages=[],
@@ -69,7 +69,7 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):  # pylint: d
             schema_version=0,
         )
 
-        self.db.register_guild(
+        self._config.register_guild(
             prefix=[],
             whitelist=[],
             blacklist=[],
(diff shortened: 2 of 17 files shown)


changelog.d/2967.breaking.rst
```diff
@@ -0,0 +1 @@
+The main bot config is no longer directly accessible to cogs. New methods have been added for use where this is concerned.
```
Tobotimus: Should probably explicitly add mentions to new functions in the changelog for those who need to account for the breaking changes, especially the API key stuff

mikeshardmind:
A note to anyone impacted by this. `bot._config` is as safe to use as `bot.db` We never intended for the data structure here to be rigid, and it can and will change as we need it to. 

That's not to say you should find and replace to update, but that if you do, you aren't losing any protections. 

I'd recommend moving to the new methods which do fall under our version guarantees.

If you have a specific need for something in the global config which we have not exposed a method for which would be protected under version guarantees, we can accommodate most reasonable requests around this.

redbot/core/bot.py
```diff
@@ -134,24 +134,50 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
...
+            and await self._config.guild(guild).use_bot_color()
+            and not isinstance(location, discord.Member)
+        ):
+            return guild.me.color
+
+        return self._color
+
+    get_embed_colour = get_embed_color
+
+    async def _maybe_update_config(self):
```
Flame442, line 163: This method is currently shown in the docs <https://red-discordbot.readthedocs.io/en/latest/framework_bot.html#redbot.core.bot.RedBase.maybe_update_config>. Should it be added to the breaking changes now that it is private?

changelog.d/2967.breaking.rst
```diff
@@ -0,0 +1,11 @@
...
+New methods for this include
+
+  - ``bot.get_shared_api_tokens``
+  - ``bot.get_shared_api_tokens``
```
Flame442: One of these needs to be `set_shared_api_tokens`.

Flame442 (changes requested):
Trying to launch a bot with this PR raises this error:
```py
  File "<Python>\Python37-32\lib\runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "<Python>\Python37-32\lib\runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "<path_to_venv>\lib\site-packages\redbot\__main__.py", line 182, in <module>
    main()
  File "<path_to_venv>\lib\site-packages\redbot\__main__.py", line 134, in main
    loop.run_until_complete(_get_prefix_and_token(red, tmp_data))
  File "<Python>\Python37-32\lib\asyncio\base_events.py", line 573, in run_until_complete
    return future.result()
  File "<path_to_venv>\lib\site-packages\redbot\__main__.py", line 55, in _get_prefix_and_token
    indict["token"] = await red.db.token()
AttributeError: 'Red' object has no attribute 'db'
```

`redbot/__main__.py` lines 55 and 56 both have `bot.db`.

Cog-CreatorsBot (changes requested):

redbot/core/bot.py
mikeshardmind: I dont think so. Just because it was documented doesnt mean it was ever intended for use externally (part of why some of these changes are being made)

No 3rd party code could ever run this function safely before.

redbot/core/bot.py
```diff
@@ -117,10 +117,10 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
         if "command_not_found" not in kwargs:
             kwargs["command_not_found"] = "Command {} not found.\n{}"
 
-        self.counter = Counter()
-        self.uptime = None
-        self.checked_time_accuracy = None
-        self.color = discord.Embed.Empty  # This is needed or color ends up 0x000000
```
Flame442, line 123: `self.color` is also on line 52, should that also be renamed to `self._color`? The embed color seems to default to gray (the `color=None` color) and `get_embed_color` returns `Embed.Empty`.

redbot/core/bot.py
```diff
@@ -203,47 +229,112 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
...
                     return True
-            for snowflake in await self.db.guild(member.guild).mod_role():
+            for snowflake in await self._config.guild(member.guild).mod_role():
                 if member_snowflakes.has(snowflake):  # DEP-WARN
                     return True
         except AttributeError:  # someone passed a webhook to this
             pass
         return False
 
+    async def get_admin_roles(self, guild: discord.Guild) -> List[discord.Role]:
```
Flame442, line 275: I don't know if you want to add any input validation to the new funcs, but (for example) passing `None` to the `guild` param of this func makes an ugly error a cog creator *might* not understand.

redbot/core/bot.py
```diff
@@ -203,47 +229,112 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
...
             return True
         return await super().is_owner(user)
 
-    async def is_admin(self, member: discord.Member):
+    async def is_admin(self, member: discord.Member) -> bool:
         """Checks if a member is an admin of their guild."""
         try:
             member_snowflakes = member._roles  # DEP-WARN
-            for snowflake in await self.db.guild(member.guild).admin_role():
+            for snowflake in await self._config.guild(member.guild).admin_role():
```
Flame442, line 254: Can/should the new methods be used here (and in `is_mod()`) instead?

redbot/core/bot.py
```diff
@@ -203,47 +229,112 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
...
+
+        Returns
+        -------
+        Dict[str, str]
+            A Mapping of token names to tokens.
+            This mapping exists because some services have multiple tokens.
+        """
+        return await self._config.api_tokens.get_raw(service_name, default={})
+
+    async def set_shared_api_tokens(self, service_name: str, **tokens: str):
```
Flame442, line 325: This seems to always throw `AttributeError: 'Value' object has no attribute 'tokens'` for line 335
```py
  File "<venv>\lib\site-packages\redbot\core\dev_commands.py", line 182, in _eval
    result = await func()
  File "<string>", line 2, in func
  File "<venv>\lib\site-packages\redbot\core\bot.py", line 335, in set_shared_api_tokens
    async with self._config.api.tokens.get_attr(service_name)() as method_abuse:
AttributeError: 'Value' object has no attribute 'tokens'
```

Flame442 (changes requested):
All of the other logic appears good, but I can't test every single change. The only places where it is more than a find & replace are the image cog and other cogs w/ API keys. I don't use these features, so I don't have keys to test with.

redbot/core/bot.py
mikeshardmind: uhh no.... this is the definition for one of those new methods.

ED: no, just also allowed to use private attrs. 

Making the methods is to discourage direct config access from cog creators to allow us to change thigns non-breakingly, not to avoid needing to update our own code as appropriate.

redbot/core/bot.py
mikeshardmind: The type hinting really should be enough here. At most, I'd add a more comprehensive docstring as any input validation slows down normal use of this.

redbot/core/bot.py
mikeshardmind: Great. (fixed)

redbot/core/bot.py
mikeshardmind: line 52 is part of a config declaration, not an instance attribute here attribute.

There were some other places in red's code where we used that attribute directly which I needed to go find.

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

PR #2974, opened 2019-09-01
Fixes the strings of `[p]set usebotcolor`

Flame442:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Right now, the help text and response of `[p]set usebotcolor` say the opposite of what actually happens. This fixes the strings to have accurate descriptions of what they are doing.

Wording the help text was pretty hard, so I am open to suggestions on wording.

### DIFF
--- a/changelog.d/2974.bugfix.rst
+++ b/changelog.d/2974.bugfix.rst
@@ -1,0 +1,1 @@
+Fixed the help text and response of `[p]set usebotcolor` to accurately reflect what the command is doing.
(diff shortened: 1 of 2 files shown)


mikeshardmind (dismissed):

mikeshardmind (approved):
Was going to suggest fixing the command feedback string for i18n here, but that can be done in a separate PR as it's an unrelated change other than it being in the same command.

PR #2977, opened 2019-09-02
Reserves command names

### LINKED ISSUE #2973: Reserve [p]cancel
# Feature request

- [ ] Cog
- [ ] Command
- [X] API functionality

Reserve a no-op `[p]cancel` command, which is currently used by the built-in [cancelled predicate](https://red-discordbot.readthedocs.io/en/stable/framework_utils.html#redbot.core.utils.predicates.MessagePredicate.cancelled), in the core cog, with a valid docstring, in order to prevent surprising behavior from other third-party cogs registering that command.

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Resolves #2973 

  - Currently, only reserving ``cancel``
  - This should only impact matching command names
  - This only impacts top level commands (and groups), not subcommands.
  - This also checks aliases
  - This makes cc and alias use the new module constant with info about
  this
  - Module constant is available for use by 3rd party cogs which may
  dynamically create responses.

### Notes

Adding anything to this constant should constitute a breaking change, as it may break 3rd party cogs, and can't be fixed by unloading the cog with a conflicting command (there wont be one)

The constant is exposed for programmatic use, and should not break cogs which use it for responses, but could still break if we need to reserve a name in use by a cog for non-user generated commands and/or responses.

This is used alternatively to adding no-op commands for a a few reasons.

1. No-op commands will still be shown in help, even if disabled and hidden based on user settings for help
2. No additional dispatches.
3. Prevents someone from replacing a dummy command with a real one.

The current use case for this is our use of `[p]cancel` in MessagePredicate.

This pattern should be used sparingly. Adding to this constant needs to be done with caution and thought behind why we want to add to it.

### DIFF
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -102,9 +102,10 @@ class CommandObj:
     async def create(self, ctx: commands.Context, command: str, *, response):
         """Create a custom command"""
         # Check if this command is already registered as a customcommand
-        if command in commands.RESERVED_COMMAND_NAMES or await self.db(
-            ctx.guild
-        ).commands.get_raw(command, default=None):
+        if (
+            await self.db(ctx.guild).commands.get_raw(command, default=None)
+            or command in commands.RESERVED_COMMAND_NAMES
+        ):
             raise AlreadyExists()
         # test to raise
         ctx.cog.prepare_args(response if isinstance(response, str) else response[0])
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -102,7 +102,9 @@ class CommandObj:
     async def create(self, ctx: commands.Context, command: str, *, response):
         """Create a custom command"""
         # Check if this command is already registered as a customcommand
-        if await self.db(ctx.guild).commands.get_raw(command, default=None):
+        if command in commands.RESERVED_COMMAND_QUALNAMES or await self.db(
+            ctx.guild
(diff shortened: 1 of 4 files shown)


redbot/cogs/customcom/customcom.py
```diff
@@ -102,7 +102,10 @@ def get_now() -> str:
     async def create(self, ctx: commands.Context, command: str, *, response):
         """Create a custom command"""
         # Check if this command is already registered as a customcommand
-        if await self.db(ctx.guild).commands.get_raw(command, default=None):
+        if (
+            await self.db(ctx.guild).commands.get_raw(command, default=None)
+            or command in commands.RESERVED_COMMAND_NAMES
```
Flame442: The message send from this is "This command already exists. Use `[p]customcom edit` to edit it." rather than "There already exists a bot command with the same name.". Where the error is handled probably needs a check as well.

Flame442 (changes requested):
Everything else works fine.

Cog-CreatorsBot (changes requested):

redbot/cogs/customcom/customcom.py
mikeshardmind: Should be good to go now, moved the check in CC and added one which was missing.

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

PR #2978, opened 2019-09-03
Code Quality changes

Drapersniper:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Still haven't touched the following (Note, errors its just me not having pytest locally)
image


mikeshardmind:
This has a lot of changes that are more noise than benefit.

There are also a lot of good changes. Typos in strings fixed up, as an example.

A lot of this seems motivated by Pycharm's inspections, which they can be useful when looking at problems, pycharm is often overzealous in its messages.

This also breaks i18n in a few places by changing something without having context on why it was done.

I think this PR is trying to do a bit too much at once, and that not all of what it is trying to do is going to be beneficial.

### i18n details

The following pattern is intentional
```py

_ = lambda s: s
some_module_level_string = _("Something")
_ = translator(...)
```
We don't want the module level string translated at definition, but we still want it to be picked up by tools which use `_("a string")` as a detection for strings intended for inclusion in messages.pot

### Why not use noinspection statements?

Things such as `#noinspection PyCharmError` really should not be enforced in the code base, but in a local editor config, as this is a Pycharm specific pattern, and would require all contributors to be using Pycharm to maintain consistently. 

Okay, but then why are `# pylint: disable=...` statements fair game?

There are a couple places where we do use `# pylint: disable=pylint-error-name` however, this is with using pylint as part of the CI process to help quickly prevent simple typo based errors from being overlooked. 

pylint runs as a lightweight CLI tool, and the project config for pylint is already preventing anything less serious than an error to be shown. 

Anyone can add this without needing to be using a specific editor to even know they would need to do so, and we've taken steps to ensure that the majority of the time, it shouldn't be needed.

Drapersniper:
Okay cool I’ll revert the pycharm specific things and translations, sounds like I misunderstood how translations work here, though if we passed the string to the translation function they would be translated regardless.

Drapersniper:
Closing becayse Github hates me ..

Tobotimus:
@Drapersniper the main tips for i18n are:
- Only ever pass plain or raw string literals to a function named `_()`. No variables/constants, no f-strings, no bytestrings, and `.format()` the string *outside* the function call. If you need to pass a variable/constant to a translator function, the function should be given an alias other than `_`, so that redgettext (the script which extracts strings and makes the message catalogues) can warn about places where a mistake might have been made (e.g. `_("format string {}".format(abc))` is a common accidental mistake)
- Strings must be translated at runtime - this means strings defined at the module level need to be passed into the translator function later, usually during the command logic. However, because strings only get extracted by redgettext if the *string literal* is enclosed by `_()`, you need to create a dummy function (which just returns the string provided) called `_()` to mark the strings for extraction, without translating it.

Issue #2520, opened 2019-03-15
[RFC] Typing Red


mikeshardmind:
I'd like to see Red be fully typed, or at least to type all of the public APIs.

Note: I do consider all public methods and attributes of core cogs as part of the public API given that it is currently neccessary to interact with other cogs for the modlog to not misbehave, as well as it being documented that `bot.get_cog` is an appropriate way to share resources.

### Pros

- Better IDE support
- Quicker detection of errors
- Better maintainability

### Cons

- Adds more things for people to need to keep correct.
- There are some places in Red where a fully correct type hint is not easily added without extra dependencies (non-homogeonous dicts)

There are also places in Red where the typehints are currently incorrect.

For example, consider the below:

```py
# https://github.com/Cog-Creators/Red-DiscordBot/blob/94c3a2feddb8c987f66727cd1efe7c1336d5e9cb/redbot/core/config.py#L877-L894
    def custom(self, group_identifier: str, *identifiers: str):
        """Returns a `Group` for the given custom group.

        Parameters
        ----------
        group_identifier : str
            Used to identify the custom group.
        identifiers : str
            The attributes necessary to uniquely identify an entry in the
            custom group. These are casted to `str` for you.

        Returns
        -------
        `Group <redbot.core.config.Group>`
            The custom group's Group object.

        """
        return self._get_base_group(str(group_identifier), *map(str, identifiers))
```
This function goes out of it's way to support anything castable to a string for user convieniece, so the type here is wrong, and many IDEs typehint support will complain about passing a non string.

Casting to a string before-hand is also possible, but that defeats the purpose of the convienience added here.

Switching the typehint to use something along the lines of the below is correct for the duck typing here, without being overly restrictive on what can be passed to the function. (see PEP 544)
```
from typing import Protocol

class SupportsStr(Protocol):
    def __str__(self) -> str:
        ...
```

mikeshardmind:
Closing this, it's still an ongoing goal to improve, but we don't have any metrics for this we're using as a lot of our typing complexity is at the discord.py boundary and not easily resolved with discord.py supporting as low as python3.5.x

PR #2980, opened 2019-09-04
Code Quality changes

Drapersniper:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

It was easier making a new PR than reverting changes ...
This is a variant of #2978 but it doesn't have the PyCharm inspections hints, and it doesn't break l18n 👀 
This one also catches a few typos and mispelling that the original did not.

This PR essentially organises imports alphabetically into 3 groups.

Built-ins
3rd Party
Package

It also removes unused variables, convert lamda assignment to def's to PEP8 stops complaining and some other minor tweaks


mikeshardmind:
For other people coming across some of this later, as some discussion has happened in discord.

We should avoid using `_` as a placeholder var name anywhere in red's codebase due to interactions with gettext and i18n use.

It's fine to rename something to `_loop_counter` or something else which indicates to tools that the var name isn't important.

redbot/cogs/mod/mutes.py
```diff
@@ -114,7 +116,7 @@ class MuteMixin(MixinMeta):
                 channel=None,
             )
         except RuntimeError as e:
-            await ctx.send(e)
+            await ctx.send(str(e))
```
mikeshardmind: This adds an unneeded overhead to each case of it. Better to fix the typhehint on `ctx.send` see #2520 for more details on things needed around this.

redbot/cogs/mod/names.py
```diff
@@ -131,10 +132,12 @@ class ModInfo(MixinMeta):
             # In embed.fields.2.value: Must be 1024 or fewer in length.
             if len(role_str) > 1024:
                 # Alternative string building time.
-                # This is not the most optimal, but if you're hitting this, you are losing more time
+                # This is not the most optimal,
+                # but if you're hitting this, you are losing more time
                 # to every single check running on users than the occasional user info invoke
-                # We don't start by building this way, since the number of times we hit this should be
-                # infintesimally small compared to when we don't across all uses of Red.
+                # We don't start by building this way,
```
mikeshardmind: Why is this PR adjusting whitespace in comments outside of a line-length goal?

redbot/core/commands/context.py
```diff
@@ -28,7 +29,7 @@ def __init__(self, **attrs):
         self.permission_state: PermState = PermState.NORMAL
 
     async def send(self, content=None, **kwargs):
-        """Sends a message to the destination with the content given.
+        r"""Sends a message to the destination with the content given.
```
mikeshardmind: Why?

redbot/cogs/warnings/warnings.py
```diff
@@ -361,7 +360,7 @@ def __init__(self, bot: Red):
                 )
 
             try:
-                userid: int = user.id
+                user.id
```
mikeshardmind: No. This should be changed to use `hasattr` rather than try/except if we aren't using `userid` below

redbot/cogs/audio/audio.py
```diff
@@ -467,7 +469,8 @@ def _restart_connect(self):
         info_msg = _(
             "This setting is only for bot owners to set a localtracks folder location "
             "if the Lavalink.jar is being ran from outside of the Audio data directory.\n"
-            "In the example below, the full path for 'ParentDirectory' must be passed to this command.\n"
+            "In the example below, the full path for "
```
mikeshardmind: Why?

redbot/cogs/audio/audio.py
```diff
@@ -570,7 +575,8 @@ def _restart_connect(self):
         """Toggle the domain restriction on Audio.
 
         When toggled off, users will be able to play songs from non-commercial websites and links.
-        When toggled on, users are restricted to YouTube, SoundCloud, Mixer, Vimeo, Twitch, and Bandcamp links."""
+        When toggled on, users are restricted to YouTube, SoundCloud, \
```
mikeshardmind: why? (and also, dont use `\` line continuations here)

mikeshardmind (changes requested):
Some of these changes aren't ready to go.

I'm still mixed on accepting the import sorting without also making it a requirement going forward.
Doing this as is just increases the size of diffs every so often as we resort.

There's a lot of whitespace changes here which don't seem like positives.

Still seems like too much of this is being done to satisfy tools we haven't and probably won't standardize on.

redbot/core/commands/context.py
Tobotimus: `\*` down in the docstring below is an invalid escape character, and is technically deprecated since Python 3.6. Sometimes people escape `**kwargs` in docstrings with `\*\*kwargs` (same thing with `*args` -> `\*args`), thinking that sphinx will start bold/emphasis formatting. But sphinx actually handles this, and suggests omitting the backslashes for varargs: https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html

So, **TL;DR**, don't use `r"""` docstrings containing `\*\*kwargs`, use plain unicode `"""` docstrings containing `**kwargs`.

redbot/cogs/audio/audio.py
Drapersniper: This is due to line length being longer than 99 characters, one bellow it’s the same but the backlash bellow is so that when this its printed to a user it isn’t forced into a new line

redbot/core/commands/context.py
mikeshardmind: My "why" was more, why not fix the root issue here, just not expressing it as such.

redbot/cogs/audio/audio.py
mikeshardmind: Didn't notice the line length on this one was that long, mildly surprised black didn't complain about it before given that it was already inside parens.

redbot/cogs/mod/names.py
Drapersniper: Same here like length was longer than 99, for some reason black didn’t complain for some of these but IDE flagged them as being too long (usually 100/110 characters

redbot/cogs/audio/audio.py
Tobotimus: Yeah, this'll get rendered like this in the help command:
```
When toggled on, users are restricted to YouTube, SoundCloud,         Mixer, Vimeo, Twitch, and Bandcamp links.
```

I always get slightly annoyed with how docstrings end up looking in the help command - sticking to the 99-char line length but also considering where the line breaks will appear in the help embed is frustrating.

My solution would be to rewrap/unwrap each paragraph in the help formatter. Code-block help formatter should rewrap, whilst embed help formatter should unwrap. This means a single newline would usually treated as a space (we can make exceptions, possibly for numbered/unordered lists) and double newlines are preserved. Still not sure about this though.

mikeshardmind: Let's keep the docstring free of line continuations here.

I'll open an issue + handle this case for 3.2 in how we handle help.

Drapersniper:
Just for clarity, this should be fully up to date with the review from Sinbad and Toby.

mikeshardmind (changes requested):
I'm noticing some added `# isort: ignore` statements here.

For this to be included like this, if we're going to sort imports, it really should be added to both the makefile recipe (for reformatting) and the style test. Otherwise, these comments have no place in our code base.

I've already stated my dislike of mixing isort usage with black in discord, but if you have a set of settings and automatable steps which will produce consistent results, that shouldn't be a blocker to this.

I am going to insist that this be automated to keep the diff sizes down rather than something that just gets changed every so often.

Drapersniper:
> I'm noticing some added `# isort: ignore` statements here.
> 
> For this to be included like this, if we're going to sort imports, it really should be added to both the makefile recipe (for reformatting) and the style test. Otherwise, these comments have no place in our code base.
> 
> I've already stated my dislike of mixing isort usage with black in discord, but if you have a set of settings and automatable steps which will produce consistent results, that shouldn't be a blocker to this.
> 
> I am going to insist that this be automated to keep the diff sizes down rather than something that just gets changed every so often.

ermm ... this was unintentional ... i missed that the isort stuff was on this branch, let me remove them for now, there's 1 PR pending on the isort repo that after it's merged we should be able to fully automate to where black and isort can be used together consistently

Drapersniper:
Okay so Implemented both iSort and Black into tests, this is using iSort settings reccomended by the official Black docs.

All test are passing and no compatibility issues.

I would say if we want go ahead and use iSort in the project for further formatting consistency this PR is ready and could be merged after 3.2 (to reduce the number of conflicts just priort to a major release)

.travis.yml
```diff
@@ -30,10 +30,10 @@ jobs:
     # These jobs only occur on tag creation if the prior ones succeed
     - stage: PyPi Deployment
       if: tag IS present
-      python: 3.7.2
+      python: 3.7.5
```
Kowlin: Lets not change deployment versions... they are currently working on black magic and I don't want anything breaking with them.

.travis.yml
```diff
@@ -46,10 +46,10 @@ jobs:
             tags: true
     - stage: Crowdin Deployment
       if: tag IS present
-      python: 3.7.2
+      python: 3.7.5
```
Kowlin: Same over here.

.travis.yml
```diff
@@ -30,10 +30,10 @@ jobs:
     # These jobs only occur on tag creation if the prior ones succeed
     - stage: PyPi Deployment
       if: tag IS present
-      python: 3.7.2
+      python: 3.7.5
       env:
         - DEPLOYING=true
-        - TOXENV=py36
+        - TOXENV=py37
```
Kowlin: Same over here.

.travis.yml
```diff
@@ -46,10 +46,10 @@ jobs:
             tags: true
     - stage: Crowdin Deployment
       if: tag IS present
-      python: 3.7.2
+      python: 3.7.5
       env:
         - DEPLOYING=true
-        - TOXENV=py36
+        - TOXENV=py37
```
Kowlin: Same over here.

Kowlin (changes requested):

.travis.yml
Drapersniper: WUT ... I though I updated these to 3.7.5 .... *sigh* 
noted i'll ensure its 3.7.5 and py37

Drapersniper: oh wait im dumb .... shouldn't at least the TOXENV be py37, since the bot no longer supports python3.6 so running the test agaisnt 3.7 would be more realist and find bugs caused by an specific version of python?

.pre-commit-config.yaml
```diff
@@ -0,0 +1,36 @@
+default_language_version:
```
mikeshardmind: Adding this pre-commit config like this is a strong No from me. It requires we maintain settings in multiple places. We have configurations for these already, we have tests that will fail if someone didn't comply, and we have an easy way to run anything which should be automated. IDEs will format with black using our settings already just by setting black as the formatting provider. pre-commit runs in an isolated environment each time, ignoring our central settings. 

As an example of the trouble this can cause, this has args which mismatch the ones we are using for one of the tools here, and you just created it. The fewer places we duplicate settings, the better.

This just isn't suited to the project.

.isort.cfg
```diff
@@ -0,0 +1,19 @@
...
+multi_line_output=  3
+indent = '    '
+combine_as_imports = True
+case_sensitive = True
+order_by_type = True
+lines_between_types = 1
+force_grid_wrap = 0
+sections = FUTURE,STDLIB,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
+default_section = THIRDPARTY
+import_heading_stdlib = Standard Library
```
mikeshardmind: All of these headings are just visual noise. If isort is being enforced, they don't do anything useful, isort will sort it out anyway.

mikeshardmind (changes requested):
I haven't reviewed this in full. I'm still skeptical of a lot of the changes here at a surface level.

I had a few specific comments, but also I don't think it's appropriate here to be adding 
`# -*- coding: utf-8 -*-` to the top of each python file.

This is a python3.7+ project. Every python file is treated as utf-8 by default as of python3.
Encoding specifications are only neccessary when they differe from the expected, It's just noise here and doesn't even draw contrast in any files where we might need a different encoding because all of the files are utf-8.

.pre-commit-config.yaml
Drapersniper: Are we sure thats the case when I was running the pre-commit hook localy based on this file it was using the settings from the setup.cfg, the individual files and the .ini file 
Reason I added here was something user could easily run to process all the formatting  in 1 command, and ensure everything will pass the Style check while applying the extra style stuff thats black and isort compatible thats in the pre-commit hooks (like json formatting etc)

.isort.cfg
Drapersniper: So yeah this came up on discord before, yes they are noise but ... it helps people who are less familiar with the ordering  prior to running isort. So this is more educational than anything happy to remove but  from the discord talks they would be here to explain the ordering so people  learn  it rather than just having a tools "randomly" sort things priort to commiting.

Drapersniper:
> I haven't reviewed this in full. I'm still skeptical of a lot of the changes here at a surface level.
> 
> I had a few specific comments, but also I don't think it's appropriate here to be adding
> `# -*- coding: utf-8 -*-` to the top of each python file.
> 
> This is a python3.7+ project. Every python file is treated as utf-8 by default as of python3.
> Encoding specifications are only neccessary when they differe from the expected, It's just noise here and doesn't even draw contrast in any files where we might need a different encoding because all of the files are utf-8.

So this is the case for people with a real IDE, but in the past when using a non python IDE i did hit issues when editing on Windows  and handling emojis  and the emojis breaking

Kowlin:
That's the fault of your IDE and not the project. Visual studio code on Windows has fine emoji support.

.pre-commit-config.yaml
mikeshardmind: `make reformat` already handles the style bit. And yes, pre-commit creates a virtualenv specifically for all of this, using the versions specified here, not in setup.cfg and using the args specified here over the ones in our configuration files, which makes this another place we would need to update multiple things rather than just let people be responsible for their own development environments and catch it at the tests.

.isort.cfg
mikeshardmind: Info like this belongs in contributing.md, not in every source code file. This isn't a critical comment that needs visibility everywhere something is in use.

mikeshardmind:
Marking this is blocked for 2 reasons.

1. Let's wait on isort 5.0.0, it's supposed to be getting some extra stuff to make it cooperate with black a bit better.

2. This still appears to be touching files that were generated, not written (localization messages) which I'm sure isn't in our interests to have multiple tools changing.

PR #2997, opened 2019-09-12
Fixed crash originated in `bank.set_balance` 

Drapersniper:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixed crash seen in DM when calling `bank.set_balance`  which was introduced in #2926


Flame442:
If this is fixing a bug introduced in dev, the changelog should probably be `misc`.

Flame442 (changes requested):
The two functions affected by this change need to have their docstrings and typehints updated accordingly. Both now accept `Union[discord.Member, discord.User]` and both now raise `RuntimeError` (from the `get_`s they use) when a user is passed and the bank is server based.

Cog-CreatorsBot (changes requested):

Drapersniper:
All sorted, thanks for the review Flame!

Flame442:
The wording of the RuntimeError's description "guild was not provided" makes it seems like there should be a way to pass a guild to the functions. This should be changed to something that references how a user was used and a member is required.

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

PR #2998, opened 2019-09-13
[Docs] Git Install Updates, note about launcher deprecation, capitalise some words

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Git updated the title they gave to the add to PATH option during setup for Windows from "Run Git from the Windows Command Prompt" to "Git from the command line and also from 3rd-party software", as we found out in [#support](https://discordapp.com/channels/133049272517001216/387398816317440000/622096446333911061). I have also checked this is correct.
This PR also corrects some capitalisation and a note about launcher deprecation.

### DIFF
--- a/changelog.d/2998.docs.rst
+++ b/changelog.d/2998.docs.rst
@@ -1,1 +1,1 @@
-Update Git PATH install (Windows) and capitalise some words
+Update Git PATH install (Windows), capitalise some words, don't mention to launcher
--- a/changelog.d/2998.docs.rst
+++ b/changelog.d/2998.docs.rst
@@ -1,0 +1,1 @@
+Update Git PATH install (Windows) and capitalise some words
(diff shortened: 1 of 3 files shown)


Drapersniper:
"Please note the launcher will be deprecated soon." if this is indeed being added here, shoudln't we add it to the  Mac and Linux docs ? given that this change will not be specific to windows

Vexed01:
> "Please note the launcher will be deprecated soon." if this is indeed being added here, shoudln't we add it to the Mac and Linux docs ? given that this change will not be specific to windows

Good point, thanks for pointing out what my small brain cannot comprehend.

mikeshardmind (changes requested):
While we *are* looking to remove the launcher, it's better to guide users away from it by not directing them to use it than to say "It's deprecated" when we haven't even decided on if removing it is in the cards, or if we should just be minimizing the code which lives in it.

Vexed01:
I _think_ this is what you mean, you haven't clearly said exactly what changes you want.

mikeshardmind (approved):

PR #2976, opened 2019-09-01
More privatization, and some error helpers

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

### Description of the changes

This makes a lot more things private. Continued from #2967, fixes #2984
Adds public methods for various things.

Below is a brief summary of things available elsewhere, though this
particular set of changes may warrant a detailed section in the release notes.

 - bot.db.locale -> redbot.core.i18n.get_locale
   - Note: This one already existed.
 - bot.db.help -> redbot.core.commands.help.HelpSettings
 - bot db whitelist/blacklist? -> bot.allowed_by_whitelist_blacklist
   - This has also been made a single canonical function for this purpose including check usage
 - bot color? -> bot.get_embed_color/bot.get_embed_colour
 - bot.id.api_tokens? ->

   - bot.get_shared_api_tokens
   - bot.set_shared_api_tokens
   - bot.remove_shared_api_tokens

 -bot.db.prefix -> bot.get_valid_prefixes
   - (Note: This is a wrapper around bot.get_prefix)

 Other changes include
  - removing `bot.counter` as it was never used anywhere
  - Adding properties with helpful error messages for moved and renamed things
  - making bot.uptime a property with an error on set
  - adding a migration to the bot config for shared_api_tokens
  - fixing a couple prior missed bot.db -> bot._config uses

PR #3003, opened 2019-09-17
Replaces more `.db` -> `._config`

Flame442:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Two `.db` -> `._config`s were missed because they are called `red.db` rather than `bot.db`. This PR replaces them accordingly.


mikeshardmind:
These changes are already in #2976

PR #3004, opened 2019-09-18
Fixes the logic of MessagePredicate.greater and MessagePredicate.less

Flame442:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Currently, any valid int will cause `MessagePredicate.greater` and `MessagePredicate.less` to pass due to some incorrect boolean logic. `bool or bool and bool` is interpreted as `bool or (bool and bool)`, which causes the statement `valid_int(m) or valid_float(m) and float(m.content) > value` to allow any valid int to be accepted. This PR adds explicit parenthesis to ensure that only values greater or less than the given value are accepted.


mikeshardmind (approved):
:+1:

PR #3009, opened 2019-09-21
fix uptime for uptime of less than a second

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Yes, I'm aware this is an edge case.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -327,9 +327,10 @@ class Core(commands.Cog, CoreLogic):
         """Shows Red's uptime"""
         since = ctx.bot.uptime.strftime("%Y-%m-%d %H:%M:%S")
         delta = datetime.datetime.utcnow() - self.bot.uptime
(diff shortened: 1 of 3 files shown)


redbot/core/core_commands.py
```diff
@@ -326,9 +326,10 @@ class Core(commands.Cog, CoreLogic):
         """Shows Red's uptime"""
         since = ctx.bot.uptime.strftime("%Y-%m-%d %H:%M:%S")
```
Flame442, line 328: Should be `bot._uptime` now

Flame442 (changes requested):

Cog-CreatorsBot (changes requested):

redbot/core/core_commands.py
mikeshardmind: No, this one was left available as a property which only errors if you try to assign to it.

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

PR #3011, opened 2019-09-22
[Core] Search for a cog in core cogs path first

Jackenmen:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Seems to make more sense to prevent shadowing core cogs by the ones in external cog paths.


mikeshardmind:
Allowing 3rd party replacements of core cogs is entirely intentional. The order here is to allow more user freedom. It is the user's responsibility to only replace things they intend to replace.

PR #3012, opened 2019-09-24
Quotes in helpset tagline for clarity

Vexed01:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

Add quotation marks to helpset tagline's response so two consecutive full stops don't appear. This will prevent Red replying
> I'm a wonderful tagline!.

which may create confusion over as to whether a full stop was automatically inserted. Instead, this PR makes it
> The tagline has been set to "I'm a wonderful tagline!".

> The tagline has been set to: I'm a wonderful tagline!

### DIFF
--- a/changelog.d/3010.bugfix.rst
+++ b/changelog.d/3010.bugfix.rst
@@ -1,0 +1,1 @@
+Add quotation marks to helpset tagline's response so two consecutive full stops don't appear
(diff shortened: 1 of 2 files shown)


mikeshardmind (changes requested):
As the user *just* specified the tagline, I think it is safer to just tell them it's been set, without parroting back what they set it to.

Vexed01:
Should be ready for merging now 😄

mikeshardmind (approved):

PR #3013, opened 2019-09-24
[CustomCom] Use simple by default

### LINKED ISSUE #1767: [V3 CustomCom] Allow simple by default
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality

When inputting a command in the Customcom cog it should default to simple, Looking at it from where custom commands are used everywhere (twitch/discords/etc) this would be best instead of confusing the user and having them add extra steps.

Flame442:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

The group command `[p]cc create` can now be used to create simple CCs without specifying "simple".

closes #1767


redbot/cogs/customcom/customcom.py
```diff
@@ -204,15 +204,16 @@ def __init__(self, bot):
...
     @checks.mod_or_permissions(administrator=True)
-    async def cc_create(self, ctx: commands.Context):
+    async def cc_create(self, ctx: commands.Context, command: str.lower, *, text: str):
         """Create custom commands.
 
+        If a type is not specified, a simple CC will be created.
         CCs can be enhanced with arguments, see the guide
         [here](https://red-discordbot.readthedocs.io/en/v3-develop/cog_customcom.html).
         """
-        pass
```
mikeshardmind, line 215: Needs additional logic for invoking help now.

redbot/cogs/customcom/customcom.py
```diff
@@ -204,15 +204,16 @@ def __init__(self, bot):
...
+        await ctx.invoke(self.cc_create_simple, command=command, text=text)
```
mikeshardmind, line 216: `ctx.invoke` doesn't check permissions, but it's probably fine here specifically...

mikeshardmind (changes requested):

redbot/cogs/customcom/customcom.py
Flame442: I believe the group command has the same permissions as the subcommand, so that check should be fine.

redbot/cogs/customcom/customcom.py
Flame442: From my testing the help works fine. The group command is invoked when no subcommand could be found and the group command has required args, so if proper syntax isn't used it sends the help message. If you have a specific instance that doesn't work, I can take a look at that.

redbot/cogs/customcom/customcom.py
mikeshardmind: Technically speaking, because permissions can be customized this is never a safe assumption. The only case here where this would be problematic is if non-simple ccs were allowed, while simple ones were denied though.

redbot/cogs/customcom/customcom.py
mikeshardmind: You're right, I read through too quickly and didn't notice the required arg.

mikeshardmind (approved):

PR #3021, opened 2019-10-01
[Docs] Update Cog Creation guide install

yamikaitou:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Updates the Cog Creation guide to use the Stable version of Red instead of the Dev version (also removes the usage of the test egg)


mikeshardmind (changes requested):
-1 from me on these changes. Would prefer to elaborate on why the development version may be useful to cog creators than not include it in the guide aimed at that group.

yamikaitou:
Good point, I'll see if I can rewrite it decently to detail both later

docs/guide_cog_creation.rst
```diff
@@ -18,9 +18,19 @@ Getting started
...
 
 To start off, be sure that you have installed Python 3.7.
-Open a terminal or command prompt and type :code:`pip install -U git+https://github.com/Cog-Creators/Red-DiscordBot@V3/develop#egg=redbot[test]`
-(note that if you get an error with this, try again but put :code:`python -m` in front of the command
-This will install the latest version of V3.
+Next, you need to decide if you want to develop against the Stable or Develop version of Red.
+Depending on what your goal is should help determine which version you need.
+
+.. attention:: 
+    The Develop version may have changes on it that break compatibility with the Stable version and other cogs.
```
mikeshardmind:
```suggestion
    The Develop version may have changes on it which break compatibility with the Stable version and other cogs.
```

mikeshardmind (changes requested):
The wording change is a slight nitpick.

Optionally, it might be worth tossing a link to [our cookiecutter](https://github.com/Cog-Creators/cog-cookiecutter) into the additional resources section while working on the cog creation guide, but I'll leave whether or not you think that should be included in this update up to you.

docs/guide_cog_creation.rst
yamikaitou: Updated and added

mikeshardmind (approved):

Issue #2804, opened 2019-06-26
[Informational & RFC] 3.2 Plans


mikeshardmind:
We've had some internal talks about some of this, but the following is currently on my radar for the next minor version. There's not as much here as some might think, but as we're looking at a faster development cycle, the minor version really do signify when we have one or more changes which can't feasibly be added without a breaking change.

* There's some breaking changes planned for config to add things to it for safety and ease of use.
  - While these are breaking, we're on the very very low end of breaking here, more details to come, this may not impact most users

* Restructuring our docs, marking things we don't want to make guarantees about as private, and adding a page laying out in detail the commitments we intend on what is considered breaking. (requires some code restructure, but this one is mostly policy and docs for going forward)

  - By removing some things from our guarantees based on what we intend for cog dev use, we are empowered to change other things more flexibly to still prioritize users in future upgrades while informing devs what will and wont be supported.

* Modlog redesign

  - Needs to happen in a minor version, let's do it here.

* Killing `redbot-launcher`

* Make less mocked objects needed for efficient config usage (see below)

There will almost certainly be other things should they be ready in the same time window, but 3.2 doesn't need to be as drawn out. It just needs to provide the changes allowing a 3.2.x

As always, if there are other breakages to consider with a good reason for them, it's worth discussing, and this is even the time when it would be relevant.

palmtree5:
I'm gonna throw #2433 in as a 3.2 target as at this point, while its primary focus was a new feature, it's actually turned into a bit of an overhaul that would probably be better suited for a minor release as opposed to a micro

mikeshardmind:
addition to the list, killing `redbot-launcher` in favor of more non interactive opts and commands in discord

mikeshardmind:
Looking to allow some more direct access in config without needing to mock objects in 3.2 as well (something like, but not necessarily identical to the below signature)
```py
def member(self, member: Optional[discord.Member]=..., guild_id=None, member_id=None) -> Group:
```

PR #3022, opened 2019-10-02
Add support for accessing config by ids

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Works toward a goal in #2804 

Adds a few methods for accessing config by ids without mocked objects.

This is technically a breaking change should anyone be using one of the following names as a key in a global scope:

- member_from_ids
- guild_from_id
- user_from_id
- channel_from_id
- role_from_id

### DIFF
--- a/changelog.d/2804.feature.1.rst
+++ b/changelog.d/2804.feature.1.rst
@@ -1,0 +1,1 @@
+Adds methods to Config for accessing things by id without mocked objects
--- a/changelog.d/2804.feature.1.rst
+++ b/changelog.d/2804.feature.1.rst
@@ -1,1 +1,9 @@
 Adds methods to Config for accessing things by id without mocked objects
+
+  - Config.guild_from_id
+  - Config.user_from_id
+  - Config.role_from_id
+  - Config.channel_from_id
+  - Config.member_from_ids
+    - This one requires multiple ids, one for the guild, one for the user
+    - Consequence of discord's object model
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -913,6 +913,22 @@ class Config:
             config=self,
         )
 
+    def guild_from_id(self, guild_id: int) -> Group:
+        """Returns a `Group` for the given guild id.
+
+        Parameters
+        ----------
+        guild_id : int
+            A guild id.
+
+        Returns
+        -------
+        `Group <redbot.core.config.Group>`
+            The guild's Group object.
+
+        """
+        return self._get_base_group(self.GUILD, str(guild_id))
+
     def guild(self, guild: discord.Guild) -> Group:
         """Returns a `Group` for the given guild.
 
@@ -928,6 +944,24 @@ class Config:
 
         """
         return self._get_base_group(self.GUILD, str(guild.id))
+
+    def channel_from_id(self, channel_id: int) -> Group:
+        """Returns a `Group` for the given channel id.
+
+        This does not discriminate between text and voice channels.
+
+        Parameters
+        ----------
+        channel_id : int
+            A channel id.
+
+        Returns
+        -------
+        `Group <redbot.core.config.Group>`
+            The channel's Group object.
+
+        """
+        return self._get_base_group(self.CHANNEL, str(channel_id))
 
     def channel(self, channel: discord.TextChannel) -> Group:
         """Returns a `Group` for the given channel.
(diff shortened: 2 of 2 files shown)


DevilXD:
Wouldn't it be better to just modify the current groups to include the IDs as kwargs? Something like:

```py
def guild(self, guild: discord.Guild = None, *, guild_id: int = None) -> Group:
    if guild_id is None:
        guild_id = guild.id
    return self._get_base_group(self.GUILD, str(guild_id))
```

This approach could be enhanced with checks that'd make sure only one of the arguments was passed - however, in the above case it's not really necessary, as passing both will prioritize `guild_id` kwarg, and not passing anything will fail with `AttributeError` on the `guild.id` part.

This approach could be applied to `guild`, `user`, `channel` and `role` scopes.
For the special-cased `member` scope, same approach really:

```py
def member(
    self,
    member: discord.Member = None,
    *,
    guild_id: int = None,
    member_id: int = None
) -> Group:
    if guild_id is None:
        guild_id = member.guild.id
    if member_id is None:
        member_id = member.id
    return self._get_base_group(self.MEMBER, str(guild_id), str(member_id))
```

Same as above apply, with an exception of when passing in only one of the kwargs and a member object at the same time, the missing value would be substituted with the one from the member. This may have "hacky", but possibly beneficial applications:

```py
member: discord.Member
guild_id: int
other_member_id: int

# Changing the guild scope for the member passed
self.config.member(member, guild_id=guild_id)

# Targeting different member while retaining the original guild scope
self.config.member(member, member_id=other_member_id)
```

Overall, I think this is cleaner than adding five additional config methods, with an addition of making this a non-breaking change as well.

mikeshardmind:
I don't agree that this is cleaner than additional config methods, but I can appreciate not wanting more breaking changes. There's another approach I discussed before which I could swap this to. 

I would prefer if config hadn't been using `__getattr__` in such a way while allowing global values to be attached directly to config without needing to explicit ask for a scope as it strongly limits what we can do here without breaking changes, but that ship has long since sailed.

mikeshardmind:
This is the other potential implementation. I'm still not convinced it is better, but see it and a brief overview of (some of) the considerations with it.

```py
    def member(
        self, 
        member: Optional[discord.Member] = ...,
        *,
        guild_id: Optional[int] = None,
        member_id: Optional[int] = None,
    ) -> Group:
        if member is ...:
            identifiers = str(guild_id), str(member_id)
        else:
            identifiers = str(member.guild.id), str(member.id)
        return self._get_base_group(self.MEMBER, *identifiers)
```

It is intentionally doing as little checking for the user as possible to minimize the overhead, while also retaining prior behavior for the large majority of existing uses.

It doesn't default to `None` for a member object intentionally as well here, as there are plenty of places where people are currently using `guild.get_member` and passing that to config without any checking, and being fine with the exception terminating things (such as in events)

Changing this to have different behavior for None here would be a larger breaking change than adding methods would.

DevilXD:
Looks alright to me, good point about `None` from `guild.get_member` there. I just think that adding kwargs to the existing methods would both - be a non-breaking change, and just work better syntax-wise and ease of use in the long run.

mikeshardmind:
I'm still not convinced that adding more to existing methods vs adding methods is correct here. In terms of maintaining various code, I'd rather config be slightly more verbose about the intent if it means all code using it is also easier to maintain. I'm leaving the implementation as is, the considerations for breakage remain either way, just in different ways.

Flame442 (changes requested):
Everything works fine, however I don't like the gotcha from `member_from_ids` being the only plural.

Cog-CreatorsBot (changes requested):

mikeshardmind:
@Flame442 I don't feel this is a gotcha, just a natural consequence of the inability to identify a member from a singular id and accurate naming.

I'm open to other naming if you have an idea which doesn't present an issue with plurality.

Flame442 (changes requested):
The changelog docs should probably state the new methods.
Although I would really like less confusing names for the methods, if no one can think of anything better they can stay.

mikeshardmind:
As we're now over a week without alternative name suggestion, I've gone and updated the changelog entry with the current names.

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

PR #3028, opened 2019-10-04
Add autostart documentation for venv users

### LINKED ISSUE #3005: [Docs] Auto-restart guide doesn't mention venvs - add info about `ExecStart`
Basically I think there should be something in auto-restart docs that would be similar to the command we use in #support:
image

jerbob:
Resolves #3005

- [x] Enhancement

Adds autostart documentation for Red users who installed it inside a virtual environment.

### DIFF
--- a/changelog.d/3005.docs.rst
+++ b/changelog.d/3005.docs.rst
@@ -1,0 +1,1 @@
+Adds autostart documentation for Red users who installed it inside a virtual environment.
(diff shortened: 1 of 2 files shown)


yamikaitou:
#2920 updated the wording of the venv guide to use redenv instead of path/to/venv. We should probably remain consistent throughout the docs

mikeshardmind (approved):
Thanks!

PR #3029, opened 2019-10-04
[Downloader] Catch OSErrors from invalid repo names

### LINKED ISSUE #2827: [Downloader] `[p]repo add` command doesn't disallow what's said in help.
# Command bugs

#### Command name

`[p]repo add`

Downloader

#### What were you expecting to happen and what in fact happen?

According to help, name of the repo should "only contain characters A-z, numbers and underscores", so when I used names with characters that aren't one of those to fail, but instead it turned out that I can freely use space, hyphen, dollar sign, etc.

Just try adding repo with name that contains some other characters, for example `test$2`

### LINKED ISSUE #2999: [V3 Downloader] Error from invalid characters in repo name
# Command bugs

#### Command name

`[p]repo add`

Downloader

A warning message that the repo name is invalid when naming a repo something with invalid characters (such as `:` in windows)

#### What actually happened?

Error:
```py
[2019-09-14 10:39:30] [ERROR] red: Exception in command 'repo add'
  File "<VENV>\lib\site-packages\redbot\cogs\downloader\downloader.py", line 233, in _repo_add
    repo = await self._repo_manager.add_repo(name=name, url=repo_url, branch=branch)
  File "<VENV>\lib\site-packages\redbot\cogs\downloader\repo_manager.py", line 586, in add_repo
    r = Repo(url=url, name=name, branch=branch, folder_path=self.repos_folder / name)
  File "<VENV>\lib\site-packages\redbot\cogs\downloader\repo_manager.py", line 66, in __init__
    self.folder_path.mkdir(parents=True, exist_ok=True)
  File "<Python>\lib\pathlib.py", line 1241, in mkdir
    self._accessor.mkdir(self, mode)
NotADirectoryError: [WinError 267] The directory name is invalid: '<datapath>\\cogs\\RepoManager\\repos\\:'
```

Run `[p]repo add : aaaaa`
Type "I agree" if prompted

Flame442:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

~~Prevents repo names from containing anything except for a-z, 0-9, and underscores.
This fix might need to happen at a lower level, this PR only addresses the command `[p]repo add`~~
~~Catches OSErrors from invalid repo names, pretty much only on Windows. Also recommends users only use safe characters in the docstring.~~
Prevents repo names from containing anything except for a-z, 0-9, underscores, and hyphens.

Fixes #2999, Fixes #2827

### DIFF
--- a/changelog.d/downloader/2827.bugfix.rst
+++ b/changelog.d/downloader/2827.bugfix.rst
@@ -1,1 +1,1 @@
-Invalid repo names no longer cause an error.
+Repo names can now only contain the characters listed in the help text (A-Z, 0-9, underscores, and hyphens).
--- a/changelog.d/downloader/2827.bugfix.rst
+++ b/changelog.d/downloader/2827.bugfix.rst
@@ -1,1 +1,1 @@
-Repo names can now only contain the characters listed in the help text (A-Z, 0-9, and underscores).
+Invalid repo names no longer cause an error.
--- a/changelog.d/downloader/2827.bugfix.rst
+++ b/changelog.d/downloader/2827.bugfix.rst
@@ -1,0 +1,1 @@
+Repo names can now only contain the characters listed in the help text (A-Z, 0-9, and underscores).
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -1,6 +1,7 @@
 import asyncio
 import contextlib
 import os
+import re
 import shutil
 import sys
 from pathlib import Path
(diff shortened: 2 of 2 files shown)


Jackenmen:
Technically this fixes the problem though when you're already doing something about this, I wonder if we could also add hyphens to set of allowed characters, especially that some repos already use those in repo names and it would seem logical to allow such names.

mikeshardmind:
I'm somewhat concerned that this is an overly restrictive set and that it would be better to detect when a file path is problematic to the host system and inform users rather than error out, and adjust the help text to match that, however I also see how this is not the easiset approach.

(On windows) There are other things which are invalid which that regex allows

```
AUX
NUL
CON
COM[1-9]
```

And rather than handling each possible failure case preventatively, it is likely more productive to just detect the error and ask the user to use something their filesystem and OS allow, and adjust the help text to match that.

mikeshardmind (approved):

PR #3031, opened 2019-10-07
Download LastSuccessful Build from FredBoat CI

Kio-td:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Changes the Github URL to the FredBoat TeamCity CI Service; Downloading the Latest Success.
No more outdated builds and updating numbers!

### DIFF
--- a/redbot/cogs/audio/manager.py
+++ b/redbot/cogs/audio/manager.py
@@ -16,8 +16,6 @@ from tqdm import tqdm
 from redbot.core import data_manager
 from .errors import LavalinkDownloadFailed
 
-JAR_VERSION = "3.2.1"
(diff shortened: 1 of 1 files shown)


Kio-td:
In the update, I forgot to put in the name; I just removed the variables that controlled the version.

Kowlin:
I'm pretty sure that there was a specific reason that we pin specific releases from another repo that we manage, Aik will know more about this

aikaterna:
We build our own Lavalink jars that include natives for unsupported CPU architectures like Mac OS, ARM, Win32 bit, Linux 32bit, and aarch devices. This change would break audio for a lot of our users and is not an acceptable change.

PR #3038, opened 2019-10-08
CogManager docstring cleanup, present user with their real prefix

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Clean up docstrings, removing unnecessary indents and showing users their actual prefix under `[p]removepath`.

### Why is this a draft PR?
I can't get [p] to use the actual prefix, if anyone else has a fix or if it works for you please say.


redbot/core/cog_manager.py
```diff
@@ -353,9 +349,8 @@ class CogManagerUI(commands.Cog):
     @commands.command()
     @checks.is_owner()
     async def removepath(self, ctx: commands.Context, path_number: int):
-        """
-        Removes a path from the available cog paths given the path_number
-            from !paths
+        """Removes a path from the available cog paths given the path_number
+        from [p]paths.
```
Vexed01, line 353: To be clear, it's this line that won't replace with an actual prefix.

Vexed01:
The above commit fixes this but slightly changes the wording, moving `[p]` to the second part of the docstring.
`[p]` isn't replaced in the first section of a docstring.

redbot/core/cog_manager.py
Vexed01: Changed, and in #3039.

mikeshardmind:
> 
> 
> The above commit fixes this but slightly changes the wording, moving `[p]` to the second part of the docstring.
> `[p]` isn't replaced in the first section of a docstring.

I'd prefer not moving this just to avoid a separate issue which will be fixed before the next release. While rewording strings often does improve things, doing it here is just to avoid a bug which will be fixed separately (see https://github.com/Cog-Creators/Red-DiscordBot/issues/3039#issuecomment-539669036 for details)

Vexed01:
It is still slightly reworded, however, I feel this improves clarity as "Removes a path from the available cog paths given the path_number from !paths" needs an extra word before "given".

Vexed01:

PR #3040, opened 2019-10-08
Show command aliases in help with setting to disable

PredaaA:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This PR just add back aliases in help formatter, I liked to have them before help rewritte.


redbot/core/commands/help.py
```diff
@@ -187,6 +187,11 @@ def get_default_tagline(ctx: Context):
         description = command.description or ""
         tagline = (await ctx.bot._config.help.tagline()) or self.get_default_tagline(ctx)
         signature = f"`Syntax: {ctx.clean_prefix}{command.qualified_name} {command.signature}`"
+        signature += (
+            f"\n`Alias{'es' if len(command.aliases) > 1 else ''}: {', '.join(command.aliases)}`"
```
mikeshardmind: This construction isn't i18n safe and cannot be used as-is.

mikeshardmind (dismissed):

redbot/core/commands/help.py
mikeshardmind: Ignore this for now, the i18n stuff wasn't fully handled and this can be adapted for i18n when it is.

mikeshardmind:
This needs to be temporarily frozen until some of the help pagination behavior is changed in 3.2.1

This is *not* a rejection of adding this.

Jackenmen:
The mentioned changes to the pagination have been done already so I'm unfreezing this.

@PredaaA can you resolve the conflicts and make the construction mentioned by Sinbad compatible with i18n?

PredaaA:
> The mentioned changes to the pagination have been done already so I'm unfreezing this.
> 
> @PredaaA can you resolve the conflicts and make the construction mentioned by Sinbad compatible with i18n?

Sure. I will do that later today.

Jackenmen (changes requested):
Could you add a setting for showing aliases in help? It can be enabled by default.

PredaaA:
> Could you add a setting for showing aliases in help? It can be enabled by default.

Done.👍

redbot/core/core_commands.py
```diff
@@ -1439,15 +1439,31 @@ async def helpset_showhidden(self, ctx: commands.Context, show_hidden: bool = No
...
+        This defaults to True.
+        Using this without a setting will toggle.
+        """
+        if show_aliases is None:
+            show_aliases = not await ctx.bot._config.help.show_aliases()
+        await ctx.bot._config.help.show_aliases.set(show_aliases)
+        if show_aliases:
+            await ctx.send(_("Help will not show commands aliases."))
+        else:
+            await ctx.send(_("Help will show commands aliases."))
```
Jackenmen: These two strings should be the other way around.
```suggestion
        if show_aliases:
            await ctx.send(_("Help will show commands aliases."))
        else:
            await ctx.send(_("Help will not show commands aliases."))
```

redbot/core/commands/help.py
```diff
@@ -212,6 +213,9 @@ async def format_command_help(
         signature = (
             f"`{T_('Syntax')}: {ctx.clean_prefix}{command.qualified_name} {command.signature}`"
         )
+        if help_settings.show_aliases and command.aliases:
+            alias_fmt = T_("Aliases") if len(command.aliases) > 1 else T_("Alias")
+            signature += f"\n`{alias_fmt}: {humanize_list(command.aliases)}`"
```
Jackenmen: This should probably show the prefix before the alias name and it also needs support for command groups (already discussed in Discord, just posting this here for completeness of review)
```suggestion
            signature += f"\n`{alias_fmt}: {humanize_list(command.aliases)}`"
```

Jackenmen: You will need to update handling for non-embed help but as discussed in Discord, this can wait until we decide on the formatting.

Jackenmen: We need some limit for amount of shown aliases as well as their max length.

Jackenmen (changes requested):

redbot/core/commands/help.py
```diff
@@ -209,9 +210,22 @@ async def format_command_help(
         description = command.description or ""
 
         tagline = (help_settings.tagline) or self.get_default_tagline(ctx)
-        signature = _(
-            "`Syntax: {ctx.clean_prefix}{command.qualified_name} {command.signature}`"
-        ).format(ctx=ctx, command=command)
+        signature = (
+            f"`{_('Syntax')}: {ctx.clean_prefix}{command.qualified_name} {command.signature}`"
+        )
```
Jackenmen: Something went wrong with conflict resolution here - these lines weren't changed in this PR but now they are.

Jackenmen (changes requested):
There's one issue with conflict resolution + the non-embed help does not work properly with this PR as mentioned in the previous [review comment](https://github.com/Cog-Creators/Red-DiscordBot/pull/3040#discussion_r418624582). I said that it can wait until we decide on the formatting (which we did), so you have 2 options:
1. Update help's formatting as discussed [here (link to discussion in #old-advanced-coding)](https://discordapp.com/channels/133049272517001216/251782448629809152/705816866479669358)
2. Leave the formatting as is and make separate PR for that (assuming you want to do it of course).

And after either of these, you need to handle the non-embed help properly: there shouldn't be backticks in the formatting of non-embed help. Example of how it looks now:
image

Jackenmen (approved):
It's time.

PR #3042, opened 2019-10-09
[Streams] Fix TypeError in TwitchStream class and change stream_alert function for Twitch

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This PR fix a TypeError in TwitchStream class when calling Twitch client_id from Red shared APIs tokens and also changed the stream_alert function for Twitch alerts to make it works with how TwitchStream class works now.

### DIFF
--- a/changelog.d/streams/3042.bugfix.rst
+++ b/changelog.d/streams/3042.bugfix.rst
@@ -1,1 +1,1 @@
-Fix a TypeError in TwitchStream class when calling Twitch client_id from Red shared APIs tokens.
+Fix a TypeError in TwitchStream class when calling Twitch client_id from Red shared APIs tokens and also changed the stream_alert function for Twitch alerts to make it works with how TwitchStream class works now.
--- a/changelog.d/streams/3042.bugfix.rst
+++ b/changelog.d/streams/3042.bugfix.rst
@@ -1,0 +1,1 @@
+Fix a TypeError in TwitchStream class when calling Twitch client_id from Red shared APIs tokens.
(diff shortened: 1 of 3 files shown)


PredaaA:
I just updated my PR for Twitch streams alert. The first commit was making impossible to set a Twitch alert and also alerts was no longer sent. I changed the `stream_alert` function in streams.py to make it get the client_id instead of the whole dict.

PredaaA:
This PR can't be merged for now because there's a new issue with it. Once a stream alert for a specific channel has been sent, they are no longer sent after. Also as what I seen when looking into this issue, the `_messages_cache` defined in [self.streams](https://github.com/Cog-Creators/Red-DiscordBot/blob/3723b4b1ea79990c8af1c26b93d744483757ee5e/redbot/cogs/streams/streams.py#L64) and set at [load_streams](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/streams/streams.py#L636) and [check_streams](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/streams/streams.py#L548) functions, is no longer cleared when a stream goes offline while it is the intention [here](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/streams/streams.py#L565).

mikeshardmind:
Thanks for the update. I'm marking this as blocked then, when it's ready let us know.

mikeshardmind:
Unblocking this, the TypeError is a fix, but the other noticed issue pre-exists and isn't caused by this, fix is good to go.

mikeshardmind (approved):

PR #3043, opened 2019-10-09
[Help] Fix [p] that was not replaced by prefix in name of embed fields

### LINKED ISSUE #3039: Bold part of docstrings do not have [p] replaced with used prefix
# Other bugs

Have `[p]` in the bold part of a docstring be replaced for the used prefix.

The `[p]` to be replaced with the used prefix.

#### What actually happened?

The `[p]` stayed as `[p]`.

1 Create a docstring that uses `[p]` in the bold (first line, before the double newline) part
2 load/reload the cog
3 Run `[p]help <command>`
4 See that `[p]` is not replaced with the used prefix

#### Other Details

image
In this image, the docstring is
```py
"""from [p]paths.

[p]
"""
```
This was found in relation with #3038

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Resolves #3039.

### DIFF
--- a/changelog.d/3043.bugfix.rst
+++ b/changelog.d/3043.bugfix.rst
@@ -1,0 +1,1 @@
+Fix [p] that was not replaced by prefix in name of embed fields in help formatter.
--- a/redbot/core/commands/help.py
+++ b/redbot/core/commands/help.py
@@ -204,7 +204,7 @@ class RedHelpFormatter:
 
             if command.help:
                 splitted = command.help.split("\n\n")
-                name = splitted[0]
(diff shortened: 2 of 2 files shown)


mikeshardmind:
Hey, thanks for the interest in helping with this one. Unfortunately, we already have an intended method of handling this change which is not compatible with this as a fix. (as mentioned in a comment in the response to the issue https://github.com/Cog-Creators/Red-DiscordBot/issues/3039#issuecomment-539669036 )

PR #3052, opened 2019-10-12
Makes bot.send_filtered return the message that is sent

Flame442:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Most messageables (`ctx.send`, `channel.send`) return the sent message for convenience. This PR makes `bot.send_filtered` follow this behavior.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -690,6 +690,10 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):  # pylint: d
         This should realistically only be used for responding using user provided
         input. (unfortunately, including usernames)
         Manually crafted messages which dont take any user input have no need of this
+        
+        Returns
+        -------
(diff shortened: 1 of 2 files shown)


redbot/core/bot.py
```diff
@@ -690,6 +690,10 @@ def remove_cog(self, cogname: str):
         This should realistically only be used for responding using user provided
         input. (unfortunately, including usernames)
         Manually crafted messages which dont take any user input have no need of this
+        
+        Returns
+        -------
+        discord.Message
```
Jackenmen, line 696: I think this should also have a description as without it sphinx will still show `Returns:` field anyway but it will be empty (and I also think it's better to have a description regardless of whether sphinx would show it at all or not):
image

```suggestion
        discord.Message
            The message that was sent.
```

mikeshardmind (approved):

PR #3053, opened 2019-10-12
[Docs] Change links to d.py docs to use pinned version instead of v1.0.1.

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

I grepped for `discordpy` to find all files that needed to be updated.

### DIFF
--- a/changelog.d/3053.docs.1.rst
+++ b/changelog.d/3053.docs.1.rst
@@ -1,0 +1,1 @@
+Discord.py docs links will now always use docs for currently used version of discord.py.
--- a/changelog.d/3053.docs.2.rst
+++ b/changelog.d/3053.docs.2.rst
@@ -1,0 +1,1 @@
+Add ``|DPY_VERSION|`` substitution that will automatically get replaced by current discord.py version.
--- a/changelog.d/3053.docs.rst
+++ b/changelog.d/3053.docs.rst
@@ -1,1 +1,0 @@
-Change links to d.py docs to use stable version instead of v1.0.1.
--- a/changelog.d/3053.docs.rst
+++ b/changelog.d/3053.docs.rst
@@ -1,0 +1,1 @@
+Change links to d.py docs to use stable version instead of v1.0.1.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -36,6 +36,7 @@ os.environ["BUILDING_DOCS"] = "1"
 # ones.
 extensions = [
     "sphinx.ext.autodoc",
+    "sphinx.ext.extlinks",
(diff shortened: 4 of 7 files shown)


mikeshardmind (changes requested):
We don't install stable, we pin a specific version for concerns with compatability with overriding certain behaviors in `discord.ext.commands` For most cases for our documentation, we should point to the current pinned version (which would still need to be updated here)

Drapersniper:
Even then 1.0.1 it’s outdated right? Iirc the correct version is 1.2.3 something ?

mikeshardmind:
> Even then 1.0.1 it’s outdated right? Iirc the correct version is 1.2.3 something ?

I was alluding to that with this portion
> we should point to the current pinned version (which would still need to be updated here)

It's just not necessarily fine to link to stable

Jackenmen:
I added `:dpy_docs:` role (using extlinks) that will always create links to docs of our d.py version and used it where needed. I also added `|DPY_VERSION|` substitution as someone might find a use for it in future too.

Yes, I could have used intersphinx for d.py links, but I figured extlinks are easier to use as you don't need to look up `objects.inv` for right label and name of document and can just use part of docs link:
```rst
#### this is how extlinks works
# https://discordpy.readthedocs.io/en/v1.2.3/discord.html#creating-a-bot-account
:dpy_docs:`this guide <discord.html#creating-a-bot-account>`
# https://discordpy.readthedocs.io/en/v1.2.3/migrating.html
:dpy_docs:`discord.py's migration guide <migrating.html>`

#### this is how intersphinx works for roughly the same links
# https://discordpy.readthedocs.io/en/v1.2.3/discord.html#discord-intro
:ref:`this guide <dpy:discord-intro>`
# https://discordpy.readthedocs.io/en/v1.2.3/migrating.html
:doc:`discord.py's migration guide <dpy:migrating>`
```

Another note, while this covers most of links to d.py docs, it can't cover all of them, especially the ones in help docstrings. This might be added as another substitution in #3063 or possibly in this (or a separate) PR after #3063 gets merged.

For anyone looking for this in a future, currently the one and only place that is not covered by this PR is [here](https://github.com/Cog-Creators/Red-DiscordBot/blob/759ca3ba7e58152a1a45c3d62d5af47ccbaf0c7e/redbot/core/core_commands.py#L910)

mikeshardmind (changes requested):
Looks good to go. A quick change to the changelog to reflect the new approach and I'll get this merged.

mikeshardmind (approved):

PR #3057, opened 2019-10-13
[Core] Fix error message about guild-only command and add dm-only error message

Jackenmen:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

"That command is not available in DMs." error message wasn't showing up because of wrong order of `isinstance` calls.

I also decided to add an error message for dm-only messages and figured there's no point in doing 2 separate PRs for a change in practically the same place.


mikeshardmind (approved):
:+1:

PR #3058, opened 2019-10-13
[audio] adds typing indicator to playlist dedupe

kennnyshiwa:
### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

This adds ctx.typing() to the playlist dedupe command to show an indication that the command is working as previously it wouldn't provide any indication and could cause confusion or for the command to be ran a few times at once

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -1604,6 +1604,10 @@ class Audio(commands.Cog):
             return await self._embed_msg(
                 ctx, _("You must be in the voice channel to bump a track.")
             )
+        # if player.shuffle:
+        #     return await self._embed_msg(
+        #         ctx, _("Can't bump a track while shuffle is enabled.")
+        #     )
         if dj_enabled:
             if not await self._can_instaskip(ctx, ctx.author):
                 return await self._embed_msg(ctx, _("You need the DJ role to bump tracks."))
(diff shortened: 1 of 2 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -3727,89 +3727,92 @@ def _restart_connect(self):
...
 
-        track_objects = playlist.tracks_obj
-        original_count = len(track_objects)
-        unique_tracks = set()
-        unique_tracks_add = unique_tracks.add
-        track_objects = [
-            x for x in track_objects if not (x in unique_tracks or unique_tracks_add(x))
-        ]
+            if not await self.can_manage_playlist(scope, playlist, ctx, author, guild):
+                return
```
Drapersniper, line 3764: Looks good to me.
Although imo the context manager should be added after this line.
All the heavy work is done after it so thats where it would be relevant, Also showing typing for a split second for users that don't have the perms would go against all the other perm checks in audio. but i'll let @aikaterna  have the final say in regards to the UX here

Drapersniper: Meh ignore me me here, this will make it clearer to the user  if typing starts at the very start of the command, specialy once there several playlists added

Drapersniper (approved):
LGTM

Cog-CreatorsBot (approved):
Approved by aikaterna#1393

PR #3060, opened 2019-10-15
[Core] Add `redbot --edit` cli flag (replacement for `[p]set owner&token`)

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

This PR adds `redbot --edit` cli flag that allows to edit instance name, token, owner and datapath. It's supposed to be a replacement for `[p]set owner` and `[p]set token` that were removed due to being potentially dangerous.

I used the occasion to unify all confirmation prompts in Red's cli commands and also fixed a small bug that allowed user to pass no arguments to `--co-owner` and `--load-cogs`

Flags that can be used with `redbot --edit`:
`--no-prompt` - disables everything interactive in the command, if it's not passed the command will ask for everything that wasn't passed with other flags
`--token` - can be used to change token
`--owner` - can be used to change owner
`--edit-instance-name` - can be used to change instance name
`--overwrite-existing-instance` - can be used with `--edit-instance-name` to allow Red to overwrite existing instance when changing name
`--edit-data-path` - can be used to change data path
`--copy-data` - can be used with `--edit-data-path` to make Red copy data from old data path to new one

This might be considered breaking as I did change function signature for `confirm()` and `interactive_config()` that are under `redbot.core.cli` module but I didn't make changelog entries for that and I'm waiting for response whether I should add changelog entries, make this change non-breaking in that regard or don't change anything at all.

### DIFF
--- a/changelog.d/3060.feature.rst
+++ b/changelog.d/3060.feature.rst
@@ -1,1 +1,1 @@
-Add `redbot --edit` cli flag that allows to edit instance name, token, owner and datapath.
+Add ``redbot --edit`` cli flag that allows to edit instance name, token, owner and datapath.
--- a/changelog.d/3060.feature.rst
+++ b/changelog.d/3060.feature.rst
@@ -1,1 +1,1 @@
-Add ``redbot --edit`` cli flag that allows to edit instance name, token, owner and datapath.
+Added ``redbot --edit`` cli flag that can be used to edit instance name, token, owner and datapath.
--- a/changelog.d/3060.feature.rst
+++ b/changelog.d/3060.feature.rst
@@ -1,0 +1,1 @@
+Add `redbot --edit` cli flag that allows to edit instance name, token, owner and datapath.
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -118,6 +118,12 @@ def edit_instance(red, cli_flags):
 
 def _edit_token(red, token, no_prompt):
     if token:
+        if not len(token) >= 50:
+            print(
+                "The provided token doesn't look a valid Discord bot token."
+                " Instance's token will remain unchanged.\n"
+            )
+            return
         red.loop.run_until_complete(red._config.token.set(token))
     elif not no_prompt and confirm("Would you like to change instance's token?", default=False):
         interactive_config(red, False, True, print_header=False)
@@ -149,6 +155,7 @@ def _edit_owner(red, owner, no_prompt):
                     continue
                 owner_id = int(owner_id)
                 red.loop.run_until_complete(red._config.owner.set(owner_id))
+                print("Owner updated.")
                 break
         else:
             print("Instance's owner will remain unchanged.")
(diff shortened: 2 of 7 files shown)


Malarne:
After a bit of testing, a keyboard interrupt throws this error:
https://media.discordapp.net/attachments/133251234164375552/635226798006861834/unknown.png

Jackenmen:
This is now ready for review, I updated the PR description accordingly.

changelog.d/3060.feature.rst
```diff
@@ -0,0 +1 @@
+Add ``redbot --edit`` cli flag that allows to edit instance name, token, owner and datapath.
```
Flame442: "allows to"

redbot/__main__.py
```diff
@@ -64,15 +74,148 @@ def list_instances():
...
+    if token:
+        red.loop.run_until_complete(red._config.token.set(token))
+    elif not no_prompt and confirm("Would you like to change instance's token?", default=False):
+        interactive_config(red, False, True, print_header=False)
+        print("Token updated.\n")
+
+
+def _edit_owner(red, owner, no_prompt):
+    if owner:
+        if not (15 <= len(str(owner)) <= 21):
```
Flame442, line 135: You check `owner_id.isdecimal()` when done interactively, but not when done with flags.

redbot/__main__.py
```diff
@@ -64,15 +74,148 @@ def list_instances():
...
+def _edit_instance_name(old_name, new_name, confirm_overwrite, no_prompt):
+    if new_name:
+        name = new_name
+        if name in _get_instance_names() and not confirm_overwrite:
+            name = old_name
+            print(
+                "An instance already exists with this name."
+                " Continuing would overwrite the existing instance config.\n"
+                "Instance name will remain unchanged.\n"
+                "If you want to replace existing instance,"
```
Flame442: This grammar isn't great.

redbot/__main__.py
```diff
@@ -64,15 +74,148 @@ def list_instances():
...
+    if old_name != name:
+        save_config(old_name, {}, remove=True)
+
+
+def _edit_token(red, token, no_prompt):
+    if token:
+        red.loop.run_until_complete(red._config.token.set(token))
+    elif not no_prompt and confirm("Would you like to change instance's token?", default=False):
+        interactive_config(red, False, True, print_header=False)
+        print("Token updated.\n")
```
Flame442, line 130: This appears to be the only one that actually outputs something to tell the user their change was successful.

redbot/__main__.py
```diff
@@ -64,15 +74,148 @@ def list_instances():
...
+    _edit_data_path(data, data_path, copy_data, no_prompt)
+
+    save_config(name, data)
+    if old_name != name:
+        save_config(old_name, {}, remove=True)
+
+
+def _edit_token(red, token, no_prompt):
+    if token:
+        red.loop.run_until_complete(red._config.token.set(token))
```
Flame442, line 127: There is no validation that the token works when setting with flags.

Flame442 (changes requested):

Cog-CreatorsBot (dismissed):

redbot/__main__.py
Jackenmen: That's non-issue cause cli-flags is already set to "int"

redbot/__main__.py
Jackenmen: Should I add output to all other or remove it from here?

Flame442: There should probably be confirmation on all of them.

Kowlin (approved):
Works fine, can't break it.

PR #3064, opened 2019-10-16
[Docs] Update console image in Getting Started guide for 3.2

Vexed01:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

Updates the `red-console.png` resource for the Getting Started docs to use a 3.2.0 launcher, specifically written for docs.
I've put the image and the code used to generate it in [this Gist](https://gist.github.com/Vexed01/9ef17f35973ca7e7365cdbea60435f17), as it may be useful for Cog Creators.

I think this should be given the 3.2.0 milestone, but *only merged nearer 3.2 release* as otherwise, the getting started docs will reference a version that does not exist yet, as getting started isn't on stable docs.

This currently requires #3070 to be merged, though if that is closed or if this wants to be merged sooner I can make it not include #3070.


Vexed01:
A thought I've just had:
Maybe this should simulate having no cogs loaded as users probably won't of loaded any yet as they won't of been told how in the gettings started guide. Should I change this?

Jackenmen:
If we go by this idea then it should also be in no servers.

Vexed01:

PR #3066, opened 2019-10-17
[Audio] Handle Missing SQL deps more gracefully

### LINKED ISSUE #3065: [Audio] Instructions how to fix "missing" `sqlite3` show when you actually don't have `databases`
# Other bugs

Load Audio with missing `databases` requirement.

I expected error about missing `databases` and just have Audio fail to load. It would also be fine if it would disable caching feature but told me the real cause.

#### What actually happened?

I got error about disabled caching feature (which is fine), but for instructions on fixing, it told me to google `ModuleNotFoundError: No module named '_sqlite3'` which won't help me at all when I don't have `databases` requirement.

1. Remove `databases` package
2. Try to load Audio

---

One could say that we can't account for someone missing requirements and that makes sense as long as user regularly gets the actual `ImportError` that occurred and thanks to this can know what's wrong. Getting error that is telling you that something unrelated is wrong just causes unnecessary confusion.

Drapersniper:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #3065


redbot/cogs/audio/apis.py
```diff
@@ -15,11 +16,15 @@
     from databases import Database
 
     HAS_SQL = True
-except ModuleNotFoundError:
+    _ERROR = None
+except (ModuleNotFoundError, ImportError) as err:
```
Jackenmen: `ModuleNotFoundError` is a subclass of `ImportError` so this is redundant in this way.
Although I'm not sure if there was was more reasoning behind catching all `ImportError`s here except for me mentioning it in the original issue. If that was the case then it's my bad that I put wrong name there as I actually meant `ModuleNotFoundError` 😄

redbot/cogs/audio/apis.py
```diff
@@ -15,11 +16,15 @@
...
+    _ERROR = "".join(traceback.format_exception_only(type(err), err)).strip()
+    _ERROR_ARGS = f"Arguments:\n{err.args!r}"
```
Jackenmen: I don't think this is needed here, it looks like a debug message but it's not referenced anywhere in the code (and the format it would use if it were sending this to user doesn't seem very user-friendly due to using repr, but I don't know what you had in mind :)).

Jackenmen (commented):
Works as expected, but I'm not sure if the generic `ImportError` should really be catched there as well.

redbot/cogs/audio/apis.py
Drapersniper: Well you see ... During testing with broken sqlite3 issues,  I was seeing import errors in addition to ModuleNotFound .. so though i'd mightawell handle both to  ensure if the user has a messed up SQL install it doesn't break audio, as per the convo in Discord.

Forgot about the enheritance here 🤦‍♂

redbot/cogs/audio/apis.py
Drapersniper: Yes this was debug my bad

Jackenmen (approved):
Looks good to me 👍

mikeshardmind (approved):

PR #3070, opened 2019-10-18
Show users the Getting Started guide in console when in no servers

Vexed01:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

In the console, users will now be given the Getting Started guide (I've linked what it will be when the guide is on stable) if their bot is in no servers.
Also removes a useless `print("")` that makes a newline where one is not needed and looks out of place.


Vexed01:

PR #3073, opened 2019-10-19
reorder some startup to prevent heartbeat issues

### LINKED ISSUE #3045: Bot hangs after the connection to Discord as it's loading the data files for very large bots or large data set
# Other bugs

On a very large bot, or one with a large data set, the bot hangs after the connection to Discord as it's loading the data files.
This can result in a lot of heartbeat blocked messages for up to ~10 minutes in my case (~45 seconds for @aikaterna bot, cat, and almost the same for @kennnyshiwa bot) and can also result in the bot being disconnected from Discord on startup.
In my case it results in 2-3 disconnects, and a lot of retries to launch all the shards, with sometimes one or two shards totally dead for 5mins or more when the bot is fully started.
For context when writing this issue, my bot have 6155 guilds, aika's bot 495, and kennny's bot is on 397 guilds.
I also tried to use another storage backend, MongoDB, but it was even worse, the bot was simply not able to connect to Discord.

Is there a way to rearrange Red's order of operations to `load the data -> connect to Discord all the shards -> load cogs -> tell the bot it's ready and then make the bot usable` or another process that would not block the bot.

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

resolves #3045


mikeshardmind:
Context on the neccessity of this change:

It brought a larger Red's time to responsiveness down from a 15-16 minute window, all the way down to under 2 minutes, most of which is spent syncing with the gateway.

A note:

While it wasn't safe before due to ordering concerns, it is now completely unsafe to use `bot.wait_until_ready` in cog startup. If a cog absolutely needs to bot to be ready before doing anything, it should use a background task which waits until ready and doesnt allow the cog to do anything until that is finished.

DevilXD:
Does this mean that using `bot.wait_until_ready()` inside the extension's async setup function, to wait until the cache is ready before adding the cog, will just make everything hang on startup?

mikeshardmind:
Yes, that's why this is marked as breaking

mikeshardmind:
I've gone ahead and added a timeout on loading extensions. The timeout is only in use during the bot's initial load not part of the normal method of adding extensions, and should prevent infinite hangs should a cog be doing this. I've not put this at a lower level because this is only added to prevent infinite hangs during startup.

It wont prevent a cog from doing `time.sleep(2**1000 * 60)`, but nothing was preventing that already. Timeout is handled using asyncio.wait_for, and is relatively generous at 30 seconds per cog.

mikeshardmind:
Last 4 commits address additional preexisting issues, the effects of which were amplified and made more noticeable by the startup changes.

Kowlin (approved):
Can confirm that this PR works, cannot confirm a performance increase due the fact that I don't have a bot with the needed capacity. maybe @PredaaA can share some of his thoughts on this.

PredaaA:
> Can confirm that this PR works, cannot confirm a performance increase due the fact that I don't have a bot with the needed capacity. maybe @PredaaA can share some of his thoughts on this.

Yes, in my case the startup takes now ~1min 30, instead of over 15 minutes of hell before. And all that in peace without getting any heartbeat blocked and any disconnects.

Drapersniper:
@PredaaA  did you not mention that you saw an issue with this where all commands in a cog stop working after reloading the cog?

PredaaA:
Huhh no I forgot about this one. But well, yeah, but we wasn't sure if it was related to this PR. Basically after a restart audio commands wasn't working.

kennnyshiwa:
As draper mentioned above, with this PR both @PredaaA and I have seen an odd issue where specifically with audio none of the commands will work, you can do the help for them, like ``[p]help play`` will work but not ``[p]play test`` or any other sub commands, reloading audio doesn't do anything and only way to fix is to restart the bot again in which audio commands start working

mikeshardmind:
PR included a handful of changes to audio to address those issues after they were brought up in discord.

PR #3075, opened 2019-10-21
Remove a large amount of `fetch_user` calls

### LINKED ISSUE #2964: Modlog casesfor
# Other bugs
For each modlog cases that are created, it's sent to the file or so that contains all the modlog information. So, while doing the [p]casesfor <user> command. When you have a large (currently our modlog case is around 68k, almost 69k) number of cases every time you do the casesfor it takes longer compared to someone who wouldn't have that large amount of cases. As we continue to grow with the number of cases, expecting the command process to take longer.
Essentially, possibly adjustments of modlog cases where it clusters per ID or something that'll improve the processing time.
#### What actually happened?
Since we have around 68k cases going onto 69k, every time we use the casesfor command, it'll take a bit to process since it needs to loop through the entire file for the specific ID that you're searching for.
Gain a large number of cases for your bot and test the casesfor command to see the speed of the process.

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Fetch user calls a pretty restricted API endpoint, slowing down a lot of things in the bot. Non bot-owners should not be able to generate these calls.

resolves #2964

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1693,8 +1693,8 @@ class Core(commands.Cog, CoreLogic):
         """
         uid = getattr(user, "id", user)
         async with ctx.bot._config.whitelist() as curr_list:
-            if user.id not in curr_list:
-                curr_list.append(user.id)
+            if uid not in curr_list:
+                curr_list.append(uid)
 
         await ctx.send(_("User added to whitelist."))
 
@@ -1720,9 +1720,9 @@ class Core(commands.Cog, CoreLogic):
         removed = False
         uid = getattr(user, "id", user)
         async with ctx.bot._config.whitelist() as curr_list:
-            if user.id in curr_list:
+            if uid in curr_list:
                 removed = True
-                curr_list.remove(user.id)
+                curr_list.remove(uid)
 
         if removed:
             await ctx.send(_("User has been removed from whitelist."))
@@ -1746,7 +1746,7 @@ class Core(commands.Cog, CoreLogic):
         pass
 
     @blacklist.command(name="add")
-    async def blacklist_add(self, ctx: commands.Context, *, user: discord.User):
+    async def blacklist_add(self, ctx: commands.Context, *, user: Union[discord.Member, int]):
         """
         Adds a user to the blacklist.
         """
@@ -1754,9 +1754,10 @@ class Core(commands.Cog, CoreLogic):
             await ctx.send(_("You cannot blacklist an owner!"))
             return
 
+        uid = getattr(user, "id", user)
         async with ctx.bot._config.blacklist() as curr_list:
-            if user.id not in curr_list:
-                curr_list.append(user.id)
+            if uid not in curr_list:
+                curr_list.append(uid)
 
         await ctx.send(_("User added to blacklist."))
 
@@ -1775,16 +1776,17 @@ class Core(commands.Cog, CoreLogic):
             await ctx.send(box(page))
 
     @blacklist.command(name="remove")
-    async def blacklist_remove(self, ctx: commands.Context, *, user: discord.User):
+    async def blacklist_remove(self, ctx: commands.Context, *, user: Union[discord.Member, int]):
         """
         Removes user from blacklist.
         """
         removed = False
 
+        uid = getattr(user, "id", user)
         async with ctx.bot._config.blacklist() as curr_list:
-            if user.id in curr_list:
+            if uid in curr_list:
                 removed = True
-                curr_list.remove(user.id)
+                curr_list.remove(uid)
 
         if removed:
             await ctx.send(_("User has been removed from blacklist."))
@@ -1810,12 +1812,16 @@ class Core(commands.Cog, CoreLogic):
 
     @localwhitelist.command(name="add")
     async def localwhitelist_add(
-        self, ctx: commands.Context, *, user_or_role: Union[discord.Member, discord.Role]
+        self, ctx: commands.Context, *, user_or_role: Union[discord.Member, discord.Role, int]
     ):
         """
         Adds a user or role to the whitelist.
         """
         user = isinstance(user_or_role, discord.Member)
+        if isinstance(user_or_role, int):
+            user_or_role = discord.Object(id=user_or_role)
+            user = True
+
         async with ctx.bot._config.guild(ctx.guild).whitelist() as curr_list:
             if user_or_role.id not in curr_list:
                 curr_list.append(user_or_role.id)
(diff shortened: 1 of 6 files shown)


Zephyrkul:
Is there any particular reason to leave all the `try / catch Exception` statements around get_user calls?

redbot/cogs/audio/converters.py
```diff
@@ -376,17 +376,13 @@ class ComplexScopeParser(commands.Converter):
                 to_user_raw = str(to_user_raw)
             if target_user is None:
                 member_converter = commands.MemberConverter()
-                user_converter = commands.UserConverter()
```
Drapersniper: Is there a reason why you decided to remove the user converter here? this is supposed to be able to interact with users both inside and outside a given server

Drapersniper: And on line 421

mikeshardmind: Yes, the user converter calls fetch_user

Drapersniper: ohhh since when :eyes:  fuck ... okay yeah i'll get the ID pr done today as this need to be replaced lmao

mikeshardmind: Huh, no, apparently it doesnt. I remember there being discussion about this, but....

mikeshardmind:
Handling some additional changes with the audio converters...

redbot/cogs/audio/converters.py
```diff
@@ -176,31 +162,15 @@ class ScopeParser(commands.Converter):
             target_user = None
             user_raw = " ".join(author).strip()
             if user_raw.isnumeric():
-                user_raw = int(user_raw)
-                try:
-                    target_user = ctx.bot.get_user(user_raw)
-                except Exception:
-                    target_user = None
-                user_raw = str(user_raw)
+                target_user = ctx.bot.get_user(int(user_raw))
```
Flame442: Running a command such as `[p]playlist start foo --author ½` when a playlist named `foo` exist causes a conversion error due to the fact that `½` is numeric but cannot be `int`ed.

redbot/cogs/audio/converters.py
```diff
@@ -368,31 +310,18 @@ class ComplexScopeParser(commands.Converter):
             target_user = None
             to_user_raw = " ".join(to_author).strip()
             if to_user_raw.isnumeric():
-                to_user_raw = int(to_user_raw)
-                try:
-                    source_user = ctx.bot.get_user(to_user_raw)
-                except Exception:
-                    source_user = None
-                to_user_raw = str(to_user_raw)
+                source_user = ctx.bot.get_user(int(to_user_raw))
```
Flame442: Same issue with `½` as before.

redbot/cogs/audio/converters.py
```diff
@@ -405,33 +334,18 @@ class ComplexScopeParser(commands.Converter):
             source_user = None
             from_user_raw = " ".join(from_author).strip()
             if from_user_raw.isnumeric():
-                from_user_raw = int(from_user_raw)
-                try:
-                    target_user = ctx.bot.get_user(from_user_raw)
-                except Exception:
-                    source_user = None
-                from_user_raw = str(from_user_raw)
+                source_user = ctx.bot.get_user(int(from_user_raw))
```
Flame442: Same issue with `½` as before.

Flame442 (changes requested):

redbot/cogs/audio/converters.py
Drapersniper: #3089 should  fix this, if you just add a try/except acount this  and make `target_user` None  on the except when 3089 is merged it will handle it more gracefully

redbot/cogs/audio/converters.py
mikeshardmind: Handled this (and other) by changing the pre-condition to `.isdigit` from `.isnumeric` as isdigit is both strict enough here, and matches anything we would want to guess might be a snowflake.

mikeshardmind:
(Will resolve conflicts later review can wait)

Drapersniper:
> (Will resolve conflicts later review can wait)

Just for simplicity sake, all the current conflicts are because we already worked on updating all the audio converters (Thanks Sinbad!) so all that this require is to accept all upstream changes.

mikeshardmind:
Conflict resolved by dropping the audio changes (which were resolved separately from this in #3089).

redbot/core/core_commands.py
```diff
@@ -1687,10 +1687,11 @@ class Core(commands.Cog, CoreLogic):
...
 
     @whitelist.command(name="add")
-    async def whitelist_add(self, ctx, user: discord.User):
+    async def whitelist_add(self, ctx, *, user: Union[discord.Member, int]):
         """
         Adds a user to the whitelist.
         """
+        uid = getattr(user, "id", user)
         async with ctx.bot._config.whitelist() as curr_list:
             if user.id not in curr_list:
```
Flame442: Should be `if uid not in curr_list:`. Also fix the `user.id` in the line below.

```py
  File "<venv>\lib\site-packages\discord\ext\commands\core.py", line 79, in wrapped
  File "<venv>\lib\site-packages\redbot\core\core_commands.py", line 1696, in whitelist_add
    if user.id not in curr_list:
AttributeError: 'int' object has no attribute 'id'
```

redbot/core/core_commands.py
```diff
@@ -1712,12 +1713,12 @@ class Core(commands.Cog, CoreLogic):
...
-    async def whitelist_remove(self, ctx: commands.Context, *, user: discord.User):
+    async def whitelist_remove(self, ctx: commands.Context, *, user: Union[discord.Member, int]):
         """
         Removes user from whitelist.
         """
         removed = False
-
+        uid = getattr(user, "id", user)
         async with ctx.bot._config.whitelist() as curr_list:
             if user.id in curr_list:
```
Flame442: Same issue as `whitelist add`.

redbot/core/core_commands.py
```diff
@@ -1687,10 +1687,11 @@ class Core(commands.Cog, CoreLogic):
         pass
 
     @whitelist.command(name="add")
-    async def whitelist_add(self, ctx, user: discord.User):
+    async def whitelist_add(self, ctx, *, user: Union[discord.Member, int]):
         """
         Adds a user to the whitelist.
         """
+        uid = getattr(user, "id", user)
```
Flame442, line 1694: Any reason why these changes aren't being added to `[p]blacklist` as well?

Flame442 (changes requested):
Found some small issues taking a closer look at this.

Cog-CreatorsBot (changes requested):

redbot/core/core_commands.py
mikeshardmind: Intended for both, updated.

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

PR #3079, opened 2019-10-22
Slightly reword installation guide to attempt to stop people using dev

Vexed01:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Make using the dev version more heavily worded in an attempt to stop people from using it as an normal end user.


docs/install_linux_mac.rst
```diff
@@ -286,8 +286,10 @@ Or, to install with PostgreSQL support:
 .. note::
 
   To install the development version, replace ``Red-DiscordBot`` in the above commands with the
-  following link:
-
+  link below. Please note that **dev is experimental and not intended for end users.** We will not
+  support anyone using the development version in any support channels. Downgrading after an
+  accidental install of dev is not possible without data loss.
```
Flame442: It might be better to say "not __always__ possible".

docs/install_linux_mac.rst
```diff
@@ -286,8 +286,10 @@ Or, to install with PostgreSQL support:
 .. note::
 
   To install the development version, replace ``Red-DiscordBot`` in the above commands with the
-  following link:
-
+  link below. Please note that **dev is experimental and not intended for end users.** We will not
```
Flame442: This sentence should be reworded to something along the lines of "The development version of the bot contains experimental changes. It is not intended for normal use." (all in bold)

docs/install_windows.rst
```diff
@@ -91,7 +91,9 @@ Installing Red
   .. note::
 
       To install the development version, replace ``Red-DiscordBot`` in the above commands with the
-      following link:
+      link below. Please note that **dev is experimental and not intended for end users.** We will not
+      support anyone using the development version in any support channels. Downgrading after an
+      accidental install of dev is not possible without data loss.
```
Flame442: Changes from above apply.

Flame442 (changes requested):

Cog-CreatorsBot (changes requested):

docs/install_linux_mac.rst
Drapersniper: Only replying cuz I’m nosey, I do agree but also disagree if we don’t want end users to use dev since it usually has breaking changes/incomplete changes etc and often leaves the user unable to downgrade, we shouldn’t imply that sometimes they will get support although it will happen we probably shouldn’t imply here, if we take an official stance that we will not help it should discourage people

mikeshardmind: I'd prefer the more accurate wording here. It doesn't change that we don't provide support for development versions, only make it clear what risks might be associated with using that version.

Vexed01:
I'm not happy with how long this is getting when built:


image


Should we just leave it as:
>To install the development version, replace ``Red-DiscordBot`` in the above commands with the
link below. **The development version of the bot contains experimental changes. It is not
intended for normal users.** We will not support you with any problems that may arise with the core bot or third party cogs, and downgrading to stable may cause data loss.

I think the above gets out the message in as few words as possible.

docs/install_windows.rst
```diff
@@ -91,9 +91,11 @@ Installing Red
...
 
       To install the development version, replace ``Red-DiscordBot`` in the above commands with the
-      link below. Please note that **dev is experimental and not intended for end users.** We will not
-      support anyone using the development version in any support channels. Downgrading after an
-      accidental install of dev is not possible without data loss.
+      link below. **The development version of the bot contains experimental changes. It is not
+      intended for normal users.** We will not support anyone using the development version in any
+      support channels. Using the develepment version may break third party cogs and not all core
+      commands may work. Downgrading after an accidental install of dev is not always possible
+      without data loss.
```
Drapersniper: Data loss is not the only issue here it could also cause unexpected crashes

Drapersniper:
> I'm not happy with how long this is getting when built:
> 
> image
> 
> Should we just leave it as:
> 
> > To install the development version, replace `Red-DiscordBot` in the above commands with the
> > link below. **The development version of the bot contains experimental changes. It is not
> > intended for normal users.** We will not support you with any problems that may arise with the core bot or third party cogs, and downgrading to stable may cause data loss.
> 
> I think the above gets out the message in as few words as possible.

I would say since it’s the docs it’s better to be explicit rather than try to make the message as short as possible, there no reason to try to shorter the docs after all their purpose is to educate the end user if the end user doesn’t read it because it’s too long it’s their fault

docs/install_linux_mac.rst
```diff
@@ -286,7 +286,11 @@ Or, to install with PostgreSQL support:
 .. note::
 
   To install the development version, replace ``Red-DiscordBot`` in the above commands with the
-  following link:
+  link below. **The development version of the bot contains experimental changes. It is not
+  intended for normal users.** We will not support anyone using the development version in any
+  support channels. Using the develepment version may break third party cogs and not all core
```
Flame442: Typo, "develepment".

docs/install_linux_mac.rst
```diff
@@ -286,7 +286,11 @@ Or, to install with PostgreSQL support:
...
+  commands may work. Downgrading after an accidental install of dev may cause data loss, crashes
```
Flame442: I don't think we should say "after an accidental install of dev", since even purposeful installs have these side effects. Maybe "Downgrading after installing the development version may..."

docs/install_windows.rst
```diff
@@ -91,7 +91,11 @@ Installing Red
   .. note::
 
       To install the development version, replace ``Red-DiscordBot`` in the above commands with the
-      following link:
+      link below. **The development version of the bot contains experimental changes. It is not
```
Flame442, line 94: Same as above.

Flame442 (changes requested):

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

Issue #2936, opened 2019-08-15
[Downloader] Issue with deleted repo/branch.


Ianardo-DiCaprio:
# Command bugs

#### Command name

Cog update

#### What cog is this command from?

Downloader/RepoManager

#### What were you expecting to happen?

For all cogs to update

#### What actually happened?

Got a command invoke error.

#### How can we reproduce this issue?

Have a repo added, where the repo or branch no longer exists.

I feel like this should be handled better, and pass that repo and carry on updating, instead of erroring out the whole command.

Flame442:
Fixed by #3080

PR #3080, opened 2019-10-22
[Downloader] Catch and handle erorr in update_all when target repository/branch is missing

wereii:
## Type

- [x] Bugfix #2936 
- [ ] Enhancement
- [ ] New feature

The error in [update_all_repos()](https://github.com/wereii/Red-DiscordBot/blob/ee14112411b606694c833ffded97461f8c4e8bf9/redbot/cogs/downloader/repo_manager.py#L650) is now properly catched and reacted upon, notifying user.
The channel context has to be passed to successfully notify the calling user, that one of the updated repositories has failed (but still continue updating others). If mixing the context into `repo_manager` is wrong decision then whole `update_all` logic needs to be done in [downloader](https://github.com/wereii/Red-DiscordBot/blob/ee14112411b606694c833ffded97461f8c4e8bf9/redbot/cogs/downloader/downloader.py#L399).

The other solution was to catch the error in [`downloader:_cog_update()`](https://github.com/wereii/Red-DiscordBot/blob/ee14112411b606694c833ffded97461f8c4e8bf9/redbot/cogs/downloader/downloader.py#L399) but this removes possibility to update other repositories until the broken one is removed or resolved.

Edit: wrong issue reference, wrone one was #2927, should be #2936


Jackenmen:
I think `RepoManager` should remain to be not reliant on bot. One of solutions could be returning tuple of successful and unsuccessful repos (similarly to what other methods in `Repo` do with `Installable` objects) and handle the rest in `_cog_update`.

wereii:
Sure thing, it would hinder testing. Let me rewrite it.

wereii:
Rewritten.

wereii:
Just heads up, seems I have referenced wrong Issue.

redbot/cogs/downloader/repo_manager.py
```diff
@@ -647,22 +647,31 @@ def get_all_repo_names(self) -> Tuple[str]:
...
+
+        failed = []
         ret = {}
-        for repo_name, _ in self._repos.items():
-            repo, (old, new) = (await self.update_repo(repo_name)).popitem()
+        for repo_name, repo_class in self._repos.items():
+            try:
+                repo, (old, new) = (await self.update_repo(repo_name)).popitem()
+            except errors.UpdateError as err:
+                log.error("Repository %s has failed update.", repo_class.url, exc_info=err)
```
Flame442: "has failed __to__ update".

redbot/cogs/downloader/downloader.py
```diff
@@ -402,8 +402,19 @@ def __init__(self, bot: Red):
 
         async with ctx.typing():
             if cog_name is None:
-                updated = await self._repo_manager.update_all_repos()
-
+                updated, failed = await self._repo_manager.update_all_repos()
+
+                for failed_repo in failed:
+                    await ctx.send(
```
Flame442: It probably isn't the best idea to send a new message for every single failed repo.

redbot/cogs/downloader/downloader.py
```diff
@@ -402,8 +402,19 @@ def __init__(self, bot: Red):
...
+                        _(
+                            "Repository `{repo_name}` {repo_url} on branch:`{branch}`"
+                            " Failed to update, check if it wasn't deleted."
```
Flame442: "check __that__ it wasn't".

redbot/cogs/downloader/downloader.py
```diff
@@ -402,8 +402,19 @@ def __init__(self, bot: Red):
...
         async with ctx.typing():
             if cog_name is None:
-                updated = await self._repo_manager.update_all_repos()
-
+                updated, failed = await self._repo_manager.update_all_repos()
+
+                for failed_repo in failed:
+                    await ctx.send(
+                        _(
+                            "Repository `{repo_name}` {repo_url} on branch:`{branch}`"
```
Flame442: This needs a space after the :

redbot/cogs/downloader/downloader.py
```diff
@@ -402,8 +402,19 @@ def __init__(self, bot: Red):
...
+                            " Failed to update, check if it wasn't deleted."
+                        ).format(
+                            repo_name=failed_repo.name,
+                            repo_url=failed_repo.url,
+                            branch=failed_repo.branch,
+                        )
+                    )
             else:
                 try:
                     updated = await self._repo_manager.update_repo(cog_name.repo_name)
```
Flame442: You need to add support for this when updating a specific repo.

Flame442 (changes requested):
Some handling for repos that are deleted or made private might need to be added to this PR at some point (there is currently some internal discussion).

Cog-CreatorsBot (changes requested):

wereii:
image

Would this be acceptable format?
All information about failed repositories is grouped into one message, one repo per line inside ` ``` ` tags. 
Might add another newline after the code block to visually divide fail message from other result states  (`All installed cogs are already up to date.`).

Drapersniper:
> image
> 
> Would this be acceptable format?

I’d personally  go with name + padding | branch + padding | url, that way the nice table formatting is less likely to break

wereii:
Definitely going to be padded. 
This is how it looks with `name:12` `branch:20` :
image

or with centred padding and header:
image

Drapersniper:
> This is how it looks with `name:12` `branch:20` :

I'd reccomend `name:20`,  this should support all approved repos at the moment without breaking the columns, I know for certain there's at least 1 cog name thats 18 characters

wereii:
image

Flame442:
The list of failed repos would not look good on mobile and gives more information than necessary. You can say what name, url, and branch fails in the logs, but the discord message should not include more than just the repo names (look to the cog loading messages for reference).

Jackenmen:
I forgot to post this here before but I haven't seen any move on this PR so I think it's fine 😄 

The handling for deleted/private repos that Flame talked about should look like this:
1. Set `GIT_TERMINAL_PROMPT=0` (we already do this, so you don't need to add anything for this one)
2. Unset `GIT_ASKPASS` (add `env.pop("GIT_ASKPASS", None)` in `Repo._run`)
3. For clone and pull command we need to add `-c credential.helper= -c core.askpass=` after `git`, e.g. `git -c credential.helper= -c core.askpass= clone --recurse-submodules {url} {folder}`

If something's unclear, let me know.

wereii:
Was a bit busy. Looks like there were some changes to repo/cog logic in downloader in #2571. Will first merge the error handling with that.

Flame442:
FYI, `redbot/cogs/downloader/repo_manager.py` is now failing the style check. It needs to be reformatted to comply with the standards of Black.

wereii:
I know, that was just merge with current v3/develop state, once I think this is ready I will drop a word there. You guys changed a lot, it actually impacted the stuff around error handling.

wereii:
image
Message looks like this. Something to consider is if it should also warn user about failing repos when cogs are updated.

About the design change: `repo_manager.update_all_repos` is replaced by `update_repos(repos: List[Repo])`.
Why ? With the new possibility to pick repos in command `[p]repos update` the way with list argument encapsulates both updating one, multiple or all repos in one function and allows for error catching in one place instead of two or three. 

edit: wording, typings

Drapersniper:
> image
> Message looks like this. Something to consider is if it should also warn user about failing repos when cogs are updated.

Thinking about it given that 1. Only the bot owner can run this, 2. This command will only be run infrequently dumping the failing repos to console instead of discord would be a better user experience here

Jackenmen:
> Something to consider is if it should also warn user about failing repos when cogs are updated.

I think that yes, it should. Most people only have to use `[p]cog update` when updating cogs, so this is the most important place where it should tell the user that the repo update wasn't successful. Otherwise, someone might say "yes, I ran `[p]cog update` and your cog still doesn't work" when in fact, it silently failed.

IMO, it should also still mention names of the repos that failed to update, yes, more detailed information (branch, url) belongs to logs but this one will give better user experience, e.g. when the repo that failed isn't related to the repo that you care about right now. (second edit, looks like this is already here and I'm just blind)

About the design change, I think it makes sense to do it, I would just add second changelog file (changelog.d/downloader/3080.misc.rst) mentioning this change. While stuff in `redbot.cogs.*` namespace isn't guaranteed, `RepoManager` is documented so if anyone uses it, they may want to know about this.

Edit: Added opinion on mentioning names of the repos.

redbot/cogs/downloader/downloader.py
```diff
@@ -499,24 +499,31 @@ def __init__(self, bot: Red):
...
-                    old, new = await repo.update()
-                    if old != new:
-                        updated.add(repo.name)
+            failed = []
+
+            updated_repos, failed = await self._repo_manager.update_repos(repos)
+            updated = {repo.name for repo in updated_repos}
+
+            if failed:
+                await ctx.send(
```
Flame442: Rather than sending two messages, use a message obj and append both failed and successful updates and send that single message at the end.

redbot/cogs/downloader/downloader.py
```diff
@@ -499,24 +499,31 @@ def __init__(self, bot: Red):
...
+                    _(
+                        "Some repositories failed to update. See logs for more information"
+                        " and check if the remote repository or branch wasn't deleted or moved."
```
Flame442: You might want to use clearer language for this (most users won't be familiar with git terms). Also, the grammar of this sentence is shaky ("if ... wasn't" should be "that ... wasn't").

redbot/cogs/downloader/repo_manager.py
```diff
@@ -1125,22 +1125,51 @@ def get_all_cogs(self) -> Tuple[Installable, ...]:
...
-        """Call `Repo.update` on all repositories.
+    async def update_repos(
+        self, upd_repo_names: List[Repo] = []
+    ) -> Tuple[Dict[Repo, Tuple[str, str]], List[str]]:
+        """Call `Repo.update` on selected repositories
+        or all if called without selection
+
+        Parameters
+        ----------
+        repo_names: List
```
Flame442: The parameter is called `repo_names` here but `upd_repo_names` above.

redbot/cogs/downloader/repo_manager.py
```diff
@@ -1125,22 +1125,51 @@ def get_all_cogs(self) -> Tuple[Installable, ...]:
         old, new = await repo.update()
         return (repo, (old, new))
 
-    async def update_all_repos(self) -> Dict[Repo, Tuple[str, str]]:
-        """Call `Repo.update` on all repositories.
+    async def update_repos(
+        self, upd_repo_names: List[Repo] = []
+    ) -> Tuple[Dict[Repo, Tuple[str, str]], List[str]]:
+        """Call `Repo.update` on selected repositories
+        or all if called without selection
```
Flame442: "or all if called without selection" could be clearer.

redbot/cogs/downloader/repo_manager.py
```diff
@@ -1125,22 +1125,51 @@ def get_all_cogs(self) -> Tuple[Installable, ...]:
         old, new = await repo.update()
         return (repo, (old, new))
 
-    async def update_all_repos(self) -> Dict[Repo, Tuple[str, str]]:
-        """Call `Repo.update` on all repositories.
+    async def update_repos(
+        self, upd_repo_names: List[Repo] = []
```
Flame442: The variable name being called `*repo_names` makes me think it should be `str` not `Repo`. Since this var handles `Repo` objects not names, it shouldn't be called `*repo_names`.

wereii:
Found another scope to this change, some util functions around cog updating (`downloader`) use directly `Repo.update` method. 
Basically, to avoid copying error catching logic that's already inside of `RepoManager.update_repos` I will have to change all of this into `self._repo_manager.update_repos([repo.name])`. 

I am also thinking of renaming the method to `safe_update_repos()` to make it distinct that it actually handles the fails and that's why it should be used instead of the direct `.update()`.
EDIT: It also returns the list of failed repos so I think the new name is better

Jackenmen:
**Note: I'm just a contributor, don't treat this as official review comment or any of the sort, it's just an opinion.**

> Found another scope to this change, some util functions around cog updating (`downloader`) use directly `Repo.update` method.
> Basically, to avoid copying error catching logic that's already inside of `RepoManager.update_repos` I will have to change all of this into `self._repo_manager.update_repos([repo.name])`.

👍 

> I am also thinking of renaming the method to `safe_update_repos()` to make it distinct that it actually handles the fails and that's why it should be used instead of the direct `.update()`.

I see why this could make sense, but personally I don't like adding `safe` word here, as it suggests that there might be "unsafe" method, and `Repo.update()` (or `RepoManager.update_repo()`) isn't unsafe, it just requires you to catch exception yourself. And it's done that way, because pythonic way is to raise exceptions on errors (speaking of that, `Repo.update()`'s docstring should probably mention it raises `UpdateError`, if you want to, you can add it there).
It's just that I would say that `update_repos()` is pretty much just a utility method - it returns updated repos that had any changes since last update and it returns repos that failed to update. I wouldn't really call it "safer" than `Repo.update()`.

mikeshardmind:
A name like `update_repos_with_error_handling` would avoid the safe vs unsafe potential confusion while suitably describing the distinction. That's not an official recommendation on what the actual name should be, but at least the direction of such a name.

Edit: Sorry, I actually misread that. (Thinking there would be an additional new method) leaving the name as is and documenting the behavior is fine.

redbot/cogs/downloader/repo_manager.py
mikeshardmind: Mutable default arguments are dangerous in python.
```suggestion
        self, upd_repo_names: Optional[List[Repo]] = None
```

It doesn't look like any other handling for this should be necessary as the empty check for this below was using container truthfulness

mikeshardmind (dismissed):
Overall the direction of these changes is a good improvement. I understand this isn't finished yet, but I did notice one small issue while looking it over.

redbot/cogs/downloader/downloader.py
```diff
@@ -499,24 +499,31 @@ def __init__(self, bot: Red):
...
 
             if updated:
                 message = _("Repo update completed successfully.")
                 message += _("\nUpdated: ") + humanize_list(tuple(map(inline, updated)))
-            elif repos is None:
+            elif not repos:
                 await ctx.send(_("All installed repos are already up to date."))
                 return
             else:
                 await ctx.send(_("These repos are already up to date."))
```
Flame442: Nitpicking, but there could only be one repo, so "these" wouldn't be correct.

Flame442 (changes requested):

wereii:
~~What about `catch_update_repos()~~ ? ツ 
EDIT: Didn't notice Flame's comments. Thanks for review I will change it accordingly.

redbot/cogs/downloader/downloader.py
wereii: I did this on purpose, if the user selects one repo and that repo fails, the repo is not propagated further (`updated`) and it _might_ cause the whole _global-ish_ message string to not be send.

In this instance if you scroll a bit down, [line 512](https://github.com/Cog-Creators/Red-DiscordBot/pull/3080/files/6254992be520c34efae25115f724d8e84b66ecd4#diff-ac6f0b0860ced6532b452a289250c2b6L512) the message is re-assigned, overriding anything before and in this particular if chain there are even conditions that would result in sending completely different message and ending the function there. 

Also, these fail messages would be fairly scarce?, as of now I don't consider it as significantly bigger traffic usage. Over all, the main point is to be sure the user is warned something is wrong under the hood, no matter what happens after it.

Jackenmen: Just change that re-assign to `+=` or if we want error to be at the bottom (similarly to rest of them), append it at the other side of string. In each case, some message is sent, just make sure that the error is appended to each one.

Jackenmen:
To keep this more visible, here's what I initially proposed as a good start for improvements in error strings:
Single repo case:
```Failed to update following repository: `reponame`
Repository's branch might have been removed or it's no longer accessible at set url. See logs for more information. ```
Multi repo case:
```Failed to update following repositories: `reponame`, `anotherreponame`
Repository's branch might have been removed or it's no longer accessible at set url. See logs for more information. ```

Meanwhile, I remembered that this may also happen in other conditions (some we might not be able to predict), one I can think of right now is not being able to fast-forward after owner of some repo force-pushing to their repo. So I think this message should make it clear to the user that they aren't only possible problems with the update, and neither of them has to be the problem that is the issue in their case.

It might be worth looking at what error codes git returns for those cases and if they are unique, use that. But since it's git we're talking about, they probably won't be unique, so instead we may want to parse error messages from command's output instead (possibly with some flag that would make this git command's output reliable).

File: redbot/cogs/downloader/repo_manager.py [hunk 12]
```diff
@@ -1119,21 +1122,22 @@ def get_all_cogs(self) -> Tuple[Installable, ...]:
...
             containing old and new commit hashes.
-
         """
         repo = self._repos[repo_name]
         old, new = await repo.update()
         return (repo, (old, new))
 
     async def update_repos(
-        self, upd_repo_names: List[Repo] = []
+        self, upd_repos: Optional[List[Repo]] = None
```
Jackenmen: IMO, the arg should just be called `repos`, the `upd_` prefix doesn't give any additional context for the argument.

redbot/cogs/downloader/downloader.py
```diff
@@ -1214,3 +1235,16 @@ def cog_name_from_instance(self, instance: object) -> str:
             msg = _("This command is not provided by a cog.")
 
         await ctx.send(box(msg))
+
+    @staticmethod
+    def format_failed_repos(failed: Iterable[str]) -> str:
+        """Format iterable of `Repo.names` into failed message"""
+
+        message = _("Failed to update following")
+        message += _(" repositories: ") if len(failed) > 1 else _(" repository: ")
```
Jackenmen: You should make this 2 separate strings ("Failed to update following repository: " and "Failed to update following repositories") as this isn't translation-friendly.

redbot/cogs/downloader/repo_manager.py
```diff
@@ -1119,28 +1122,58 @@ def get_all_cogs(self) -> Tuple[Installable, ...]:
...
+            except errors.UpdateError as err:
+                log.error(
+                    "Repository '%s' failed to update. URL: '%s' on branch '%s'",
+                    repo.name,
+                    repo.url,
+                    repo.branch,
+                    exc_info=err,
+                )
+
+                failed.append(repo.name)
```
Jackenmen, line 1170: This is just a suggestion, *I think* it might be more useful in future, if we're gonna add `Repo` objects here, not just names, but it's totally fine to leave it as is.

redbot/cogs/downloader/downloader.py
```diff
@@ -792,63 +808,66 @@ def __init__(self, bot: Red):
                         )
                     for page in pagify(msg):
                         await ctx.send(msg)
-                    return
```
Jackenmen, line 795: This `return` shouldn't be removed, it is supposed to fail preemptively - all the messages that get appended after this make no sense with this error, e.g.
```
Error: short sha1 `ab2c` is ambiguous. Possible candidates:
commit ab2cxyz - commit message
commit ab2cabc - other commit message
There were no cogs to check.
```
"There were no cogs to check." is not valid here.

redbot/cogs/downloader/downloader.py
```diff
@@ -792,63 +808,66 @@ def __init__(self, bot: Red):
...
-                            rev=rev, repo=repo
-                        )
-                    )
-                    return
-                await repo.checkout(commit)
-                cogs_to_check = await self._get_cogs_to_check(repos=[repo], cogs=cogs)
+                    message += _(
+                        "Error: there is no revision `{rev}` in repo `{repo.name}`"
+                    ).format(rev=rev, repo=repo)
+
```
Jackenmen: Same as above, the `return` shouldn't be removed + the message has to still be sent.

redbot/cogs/downloader/downloader.py
```diff
@@ -722,7 +720,7 @@ def __init__(self, bot: Red):
         Use `[p]cog update` to update cogs.
         """
         async with ctx.typing():
-            cogs_to_check = await self._get_cogs_to_check()
+            cogs_to_check, failed = await self._get_cogs_to_check()
```
Jackenmen, line 726: We should probably mention failed repos here too since the repo update happens when running this command too.

redbot/cogs/downloader/downloader.py
```diff
@@ -499,24 +499,22 @@ def __init__(self, bot: Red):
...
             updated: Set[str]
-            if not repos:
-                updated = {repo.name for repo in await self._repo_manager.update_all_repos()}
-            else:
-                updated = set()
-                for repo in repos:
-                    old, new = await repo.update()
-                    if old != new:
-                        updated.add(repo.name)
+            failed = []
```
Jackenmen: Just a nitpick, but this is unneeded piece of code.

Jackenmen (changes requested):
I haven't tried to run the bot with these changes, but while I was going through code I saw few things that should be changed.

redbot/cogs/downloader/repo_manager.py
wereii: That was there because there were already lots of repo, repos, Repo and alike used inside.
Tried to say "repos_to_update". Looking around, I should indeed turn it around and change the param to `repos` and if needed change the naming inside the func.

redbot/cogs/downloader/downloader.py
```diff
@@ -721,8 +718,10 @@ def __init__(self, bot: Red):
         This command doesn't update cogs, it only checks for updates.
         Use `[p]cog update` to update cogs.
         """
+        message = ""
```
Flame442: `message` appears to be defined in two places (here and L726).

redbot/cogs/downloader/downloader.py
```diff
@@ -794,61 +808,68 @@ def __init__(self, bot: Red):
...
                 except errors.UnknownRevision:
-                    await ctx.send(
-                        _("Error: there is no revision `{rev}` in repo `{repo.name}`").format(
-                            rev=rev, repo=repo
-                        )
-                    )
+                    message += _(
+                        "Error: there is no revision `{rev}` in repo `{repo.name}`"
+                    ).format(rev=rev, repo=repo)
                     return
```
Flame442: It doesn't appear that this text will ever be used.

redbot/cogs/downloader/downloader.py
```diff
@@ -1207,3 +1237,30 @@ def cog_name_from_instance(self, instance: object) -> str:
...
+        Returns
+        -------
+        str
+            formatted message
+        """
+
+        message = (
+            _("Failed to update following repositories:")
+            if len(failed) > 1
+            else _("Failed to update following repository:")
```
Flame442: Should be "Failed to update **the** following..."

redbot/cogs/downloader/downloader.py
```diff
@@ -1207,3 +1237,30 @@ def cog_name_from_instance(self, instance: object) -> str:
...
+        )
+        message += " " + humanize_list(tuple(map(inline, failed))) + "\n"
+        message += _(
+            "Repository's branch might have been removed or"
+            " it's no longer accessible at set url. See logs for more information."
```
Flame442: Should be reworded to "**The** repository's branch might have been removed or **the repository is** no longer accessible at **the** set url."

redbot/cogs/downloader/downloader.py
```diff
@@ -499,24 +499,21 @@ def __init__(self, bot: Red):
...
-                    if old != new:
-                        updated.add(repo.name)
+
+            updated_repos, failed = await self._repo_manager.update_repos(repos)
+            updated = {repo.name for repo in updated_repos}
 
             if updated:
                 message = _("Repo update completed successfully.")
                 message += _("\nUpdated: ") + humanize_list(tuple(map(inline, updated)))
             elif repos is None:
```
Jackenmen: This `if` is inaccessible, it isn't like that because of changes you did, but I'm mentioning it since you did fix it in one of previous versions of this PR.
```suggestion
             elif not repos:
```

Flame442 (changes requested):

redbot/cogs/downloader/downloader.py
```diff
@@ -794,61 +808,68 @@ def __init__(self, bot: Red):
...
-                        message = _("Provided cogs are already up to date with this revision.")
-                    else:
-                        message = _(
-                            "Cogs from provided repo are already up to date with this revision."
-                        )
+                cogs_to_update, libs_to_update = await self._available_updates(cogs_to_check)
+
+                updates_available = cogs_to_update or libs_to_update
+                cogs_to_update, filter_message = self._filter_incorrect_cogs(cogs_to_update)
+                message = ""
```
Jackenmen: This is unneeded code, you're already setting `message` to empty string at the top of this method, one of those places doesn't need it.

redbot/cogs/downloader/downloader.py
```diff
@@ -1024,23 +1045,32 @@ def _filter_incorrect_cogs(
...
+            # remove failed repos
+            repos = [repo for repo in repos if repo.name not in failed]
+
             for repo in repos:
                 if await repo.is_on_branch():
                     exit_to_commit = None
                 else:
                     exit_to_commit = repo.commit
-                await repo.update()
                 await repo.checkout(exit_to_commit)
```
Jackenmen: This loop without `repo.update()` in it doesn't serve its purpose.

BUT, truth to be told, I think that this loop shouldn't have been done like this in first place. The purpose of `repo.is_on_branch()` check and `repo.checkout()` later after `repo.update()` was done to checkout the repo back to revision it had before update IF it's not on branch - the reason this was even needed is the checkout that we do in `_cog_update_logic` method for `[p]cog updatetoversion` command (see L816).

Since this isn't (or at least to my best knowledge shouldn't be) needed when updating multiple repos, I think this loop should just be removed all together and special case (i.e. kwarg switch) for `[p]cog updatetoversion` should be added.

What I'm thinking of is this (sorry for basically writing this code for you but I just feel like it might be unclear what I'm exactly thinking of):
```diff
diff --git a/redbot/cogs/downloader/downloader.py b/redbot/cogs/downloader/downloader.py
index ee54d5f..33a0f45 100644
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -814,7 +814,9 @@ def __init__(self, bot: Red):
                     return
 
                 await repo.checkout(commit)
-                cogs_to_check, __ = await self._get_cogs_to_check(repos=[repo], cogs=cogs)
+                cogs_to_check, __ = await self._get_cogs_to_check(
+                    repos=[repo], cogs=cogs, update_repos=False
+                )
 
             else:
                 cogs_to_check, check_failed = await self._get_cogs_to_check(repos=repos, cogs=cogs)
@@ -1045,6 +1047,7 @@ def _filter_incorrect_cogs(
         *,
         repos: Optional[Iterable[Repo]] = None,
         cogs: Optional[Iterable[InstalledModule]] = None,
+        update_repos: bool = True,
     ) -> Tuple[Set[InstalledModule], List[str]]:
         if not (cogs or repos):
             __, failed = await self._repo_manager.update_repos()
@@ -1060,17 +1063,11 @@ def _filter_incorrect_cogs(
                 cogs = cast(Iterable[InstalledModule], cogs)
                 repos = {cog.repo for cog in cogs if cog.repo is not None}
 
-            __, failed = await self._repo_manager.update_repos(repos)
+            if update_repos:
+                __, failed = await self._repo_manager.update_repos(repos)
             # remove failed repos
             repos = [repo for repo in repos if repo.name not in failed]
 
-            for repo in repos:
-                if await repo.is_on_branch():
-                    exit_to_commit = None
-                else:
-                    exit_to_commit = repo.commit
-                await repo.checkout(exit_to_commit)
-
             if cogs:
                 cogs_to_check = {cog for cog in cogs if cog.repo is not None and cog.repo in repos}
             else:
```

redbot/cogs/downloader/repo_manager.py
```diff
@@ -1119,28 +1122,58 @@ def get_all_cogs(self) -> Tuple[Installable, ...]:
...
-
         """
         repo = self._repos[repo_name]
         old, new = await repo.update()
         return (repo, (old, new))
 
-    async def update_all_repos(self) -> Dict[Repo, Tuple[str, str]]:
-        """Call `Repo.update` on all repositories.
+    async def update_repos(
+        self, repos: Optional[List[Repo]] = None
```
Jackenmen: This type hint is incorrect, the function accepts any iterable (and is even called with different iterables than list already):
```suggestion
        self, repos: Optional[Iterable[Repo]] = None
```
Remember to also update parameter type in docstring.

redbot/cogs/downloader/downloader.py
```diff
@@ -1207,3 +1237,31 @@ def cog_name_from_instance(self, instance: object) -> str:
             msg = _("This command is not provided by a cog.")
 
         await ctx.send(box(msg))
+
+    @staticmethod
+    def format_failed_repos(failed: Iterable[str]) -> str:
```
Jackenmen: This typing is incorrect, the function needs the object to be iterable and sized - this type hint should probably be `Collection` (can be imported from `typing` module):
```suggestion
    def format_failed_repos(failed: Collection[str]) -> str:
```
Remember to also update parameter type in docstring.

Jackenmen (changes requested):

redbot/cogs/downloader/downloader.py
wereii: https://github.com/Cog-Creators/Red-DiscordBot/pull/3080#discussion_r349374222 ?

redbot/cogs/downloader/downloader.py
wereii: So, extend `_get_cogs_to_check` and `_filter_incorrect_cogs` with bool param `update_repos` and propagate it inside? 

edit: Just making sure I get the snippet, I can see the filter_incorrect function extended but you also used ` update_repos=False` when calling `_get_cogs_to_check` - at first I thought get_cogs_ calls internally filter_incorrect so you meant to propagate it to that function, but it doesn't seem like that.
So, just a switch for both of them if they should or shouldn't update - call `.update_repos()`.

Jackenmen: `_filter_incorrect_cogs` isn't involved, just the generated git diff shows wrong hunk header, sorry about that. `_filter_incorrect_cogs` doesn't even update repos so it's just about `_get_cogs_to_check` having `update_repos` param, that should be set to `False` when calling the function after checkout to commit.

redbot/cogs/downloader/downloader.py
Jackenmen: The comment you linked doesn't seem to be related to the issue I'm talking about here. If I'm correct, Flame was talking just about the string that is sent in `else` block (it isn't a part of changes in this PR, so I guess that's why he might have called it a nitpick) and I'm talking about inaccessible code in the `elif` above.

wereii: Before changes of this PR it looked like this:
`elif repos is None:`
Thru out the PR I have changed it to
`elif not repos:`
Flame commented that the `repos` can contain multiple repos or one (`ctx, *, repos`) I didn't notice that and changed it back to:
`elif repos is None`

Why should it be inaccessible ?

Flame442 (dismissed):
Small nitpick, `[p]repo update <one_repo>` outputs the message "These repos are already up to date." even though there is only one repo. Not enough for me to block this, but if you'd like to fix it, that would be nice. 

I'd like to get Jack's review too, and once he approves I'll mark this as approved.

redbot/cogs/downloader/downloader.py
```diff
@@ -1207,3 +1238,31 @@ def cog_name_from_instance(self, instance: object) -> str:
...
+
+        message = (
+            _("Failed to update the following repositories:")
+            if len(failed) > 1
+            else _("Failed to update the following repository:")
+        )
+        message += " " + humanize_list(tuple(map(inline, failed))) + "\n"
+        message += _(
+            "The repository's branch might have been removed or"
+            " the repository is no logner accessible at set url."
```
Jackenmen: typo
```suggestion
            " the repository is no longer accessible at set url."
```

Jackenmen (dismissed):
Looks good to me, just noticed one typo.

wereii:
~Fuh, typo in last commit comment, should be `up-to-date message`, should I reword it? Guess it will still get squashed.~
EDIT: Additionally I have overridden the typo fix with my later commit, rebased.

Jackenmen (approved):

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

Issue #3086, opened 2019-10-26
Warnings Cog [p]warnings command for Mods


Dav-Git:
# Command bugs

#### Command name

[p]warnings

#### What cog is this command from?

Warnings

#### What were you expecting to happen?

It would be good to allow people with the modrole to check the warnings for a user. Maybe as a togglable option? Or based on permissions handled by the Permissions Cog. If a user can run the [p]warnings command, that user should also be able to expect full functionality. If a User should not be able to run the [p]warnings command, just disallow it using the Permissions cog.

#### What actually happened?

Only admins can see the results of [p]warnings.

#### How can we reproduce this issue?

1.Enable warnings Cog
2.Set a modrole and adminrole
3.Set the permissions to allow the use of the Warnings Cog for the role you used as modrole
4.Warn someone
5.Try to use [p]warnings <User you just warned> with a user that is just a member of the modrole

mikeshardmind:
Red's permission model would require these two separate behaviors to be split into separate commands. A command anyone could run by default to view their own warnings, and a command which used a normal permissions override-able check (at the same admin default level) to view others.

Vexed01:
Also with default perms, only admins can warn people. Would this be changed as well so mods can warn people (which makes sense)?

mikeshardmind:
I don't see a reason to make the default more permissive as a part of this. Using permissions, people can make it less restrictive, and warnings (Depending on how configured) could allow a mod to take an action against a user they normally would be incapable of by spamming warns.

While this *is* more an issue of "how much do you trust your mods", the defaults shouldn't require trusting them as much as an admin, and can be handled with permissions.

mikeshardmind:
Duplicate of #2900

PR #3088, opened 2019-10-26
[Warnings] Allow the permissions cog to allow mods to check other user's warnings

### LINKED ISSUE #2900: [Warnings] Permissions for `[p]warnings` command can't be modified
# Command bugs

#### Command name

`[p]warnings`

Warnings

Permissions cog to be able to change required permissions for `[p]warnings` command - in this case, allow mods to use it.

#### What actually happened?

The command still couldn't be used by mods. Reason for this is that we use a `is_admin_or_superior` function inside the command instead of using a check. This was done like this to allow users to check their own warnings without any permissions, but because of this, there's no way to use the command for people without admin role, even when changing the required permissions with Permissions cog. I never had to check if args is given within a check before, so I'm not sure if it's possible to do using `commands.check`, but IMO Permissions cog should allow to change required permissions for this command.

`[p]permissions addserverrule allow warnings Mods`

Vexed01:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

This changes `[p]warnings` to only check other user's warnings and adds `[p]mywarnings` so that the Permissions cog can dictate who can check other's warnings, instead of an unchangeable `is_admin_or_superior` check.
This fixes the duplicate issue #3086.
Closes #2900.

### DIFF
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -12,7 +12,7 @@ from redbot.cogs.warnings.helpers import (
 from redbot.core import Config, checks, commands, modlog
 from redbot.core.bot import Red
 from redbot.core.i18n import Translator, cog_i18n
-from redbot.core.utils.mod import is_admin_or_superior
+from redbot.core.utils.mod import is_admin_or_superior, is_mod_or_superior
 from redbot.core.utils.chat_formatting import warning, pagify
 from redbot.core.utils.menus import menu, DEFAULT_CONTROLS
 
@@ -355,7 +355,7 @@ class Warnings(commands.Cog):
         if user is None:
             user = ctx.author
         else:
-            if not await is_admin_or_superior(self.bot, ctx.author):
+            if not await is_mod_or_superior(self.bot, ctx.author):
                 return await ctx.send(
                     warning(_("You are not allowed to check warnings for other users!"))
                 )
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -343,7 +343,7 @@ class Warnings(commands.Cog):
     @commands.command()
     @commands.guild_only()
     @checks.admin_or_permissions(ban_members=True)
-    async def warnings(self, ctx: commands.Context, user: Union[discord.Member, int] = None):
+    async def warnings(self, ctx: commands.Context, user: discord.Member):
         """List the warnings for the specified user."""
         msg = ""
         member_settings = self.config.member(user)
@@ -357,8 +357,6 @@ class Warnings(commands.Cog):
                         mod = discord.utils.get(
                             self.bot.get_all_members(), id=user_warnings[key]["mod"]
                         )
-                        if mod is None:
-                            mod = await self.bot.fetch_user(user_warnings[key]["mod"])
                     msg += _(
                         "{num_points} point warning {reason_name} issued by {user} for "
                         "{description}\n"
@@ -395,8 +393,6 @@ class Warnings(commands.Cog):
                         mod = discord.utils.get(
                             self.bot.get_all_members(), id=user_warnings[key]["mod"]
                         )
-                        if mod is None:
-                            mod = await self.bot.fetch_user(user_warnings[key]["mod"])
                     msg += _(
                         "{num_points} point warning {reason_name} issued by {user} for "
                         "{description}\n"
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -348,15 +348,17 @@ class Warnings(commands.Cog):
         """Check your own warnings."""
         if user is None:
             user = ctx.author
-        else: #TODO: remove this else statement in a release after 3.3.0 (will allow users time start using [p]warncheck)
+        else:  # TODO: remove this else statement in a release after 3.3.0 (will allow users time start using [p]warncheck)
             if not await is_admin_or_superior(self.bot, ctx.author):
                 return await ctx.send(
                     warning(_("You are not allowed to check warnings for other users!"))
                 )
             else:
-                return await ctx.send(_(
-                    "Checking other user's warnings has moved to `{prefix}warncheck`."
-                ).format(prefix=ctx.prefix))
+                return await ctx.send(
+                    _("Checking other user's warnings has moved to `{prefix}warncheck`.").format(
+                        prefix=ctx.prefix
+                    )
+                )
 
         msg = ""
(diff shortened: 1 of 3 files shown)


redbot/cogs/warnings/warnings.py
```diff
@@ -340,33 +340,57 @@ def __init__(self, bot: Red):
...
+            if not user_warnings.keys():  # no warnings for the user
+                await ctx.send(_("You have no warnings! Well done!"))
+            else:
+                for key in user_warnings.keys():
+                    mod = ctx.guild.get_member(user_warnings[key]["mod"])
+                    if mod is None:
+                        mod = discord.utils.get(
+                            self.bot.get_all_members(), id=user_warnings[key]["mod"]
+                        )
+                        if mod is None:
```
Drapersniper: Fetch_user should not be used here, since this command is usable but anyone, at this stage we should return a placeholder

redbot/cogs/warnings/warnings.py
```diff
@@ -340,33 +340,57 @@ def __init__(self, bot: Red):
             pass
         await ctx.send(_("User {user} has been warned.").format(user=user))
 
-    @commands.command()
+    @commands.command(usage="")
```
Drapersniper: Why is usage set to blank here?

redbot/cogs/warnings/warnings.py
```diff
@@ -340,33 +340,57 @@ def __init__(self, bot: Red):
...
+                        prefix=ctx.prefix
+                    )
+                )
 
-            try:
-                userid: int = user.id
-            except AttributeError:
-                userid: int = user
-                user = ctx.guild.get_member(userid)
-                user = user or namedtuple("Member", "id guild")(userid, ctx.guild)
```
Drapersniper, line 368: instead of doing a if/else check you could shorter is with a simpler “If user is not None” and as such there would be no  no need of the else

redbot/cogs/warnings/warnings.py
```diff
@@ -340,33 +340,57 @@ def __init__(self, bot: Red):
...
-        Omit `<user>` to see your own warnings.
-
-        Note that showing warnings for users other than yourself requires
-        appropriate permissions.
-        """
+        """Check your own warnings."""
         if user is None:
             user = ctx.author
-        else:
+        else:  # TODO: remove this else statement in a release after 3.3.0 (will allow users time start using [p]warncheck)
```
Drapersniper: Core cogs don’t have to respect the version guarantees and as such breaking changes can be done any time (just a FYI)

Vexed01: I assume you mean the whole else statement, not just the note?

mikeshardmind (changes requested):
I'm not going to do a line by line.

Leave the behavior of `warnings` being to check other users, the new command should be checking for your own. I'd name it `mywarnings` but there are plenty of good options for this.

Leaving the behavior of warnings to view others makes sense as this was the more common usage, and the one which the check was trying to enable.

There's also no need to retain compatibility handling, but if you do, the logic is much simpler in this way as you don't need to override usage, just handle a default.

Vexed01:
Hmm it didn't dismiss Draper's review as stale... Well I've marked it "resolved" as I've implemented Sinbad's review instead.

redbot/cogs/warnings/warnings.py
mikeshardmind: This is still relevant and should not use fetch_user

mikeshardmind (changes requested):

redbot/cogs/warnings/warnings.py
```diff
@@ -342,45 +342,57 @@ def __init__(self, bot: Red):
...
-        appropriate permissions.
-        """
-        if user is None:
-            user = ctx.author
-        else:
-            if not await is_admin_or_superior(self.bot, ctx.author):
-                return await ctx.send(
-                    warning(_("You are not allowed to check warnings for other users!"))
+    @checks.admin_or_permissions(ban_members=True)
+    async def warnings(self, ctx: commands.Context, user: discord.Member):
```
mikeshardmind: This should retain the existing ability to check users not in the guild by accepting an id still.

redbot/cogs/warnings/warnings.py
```diff
@@ -342,45 +342,57 @@ def __init__(self, bot: Red):
...
-        Note that showing warnings for users other than yourself requires
-        appropriate permissions.
-        """
-        if user is None:
-            user = ctx.author
-        else:
-            if not await is_admin_or_superior(self.bot, ctx.author):
-                return await ctx.send(
-                    warning(_("You are not allowed to check warnings for other users!"))
+    @checks.admin_or_permissions(ban_members=True)
```
mikeshardmind: This should just be `checks.admin()` matching the prior behavior handled inside the command.

redbot/cogs/warnings/warnings.py
```diff
@@ -342,45 +342,57 @@ def __init__(self, bot: Red):
...
+        member_settings = self.config.member(user)
+        async with member_settings.warnings() as user_warnings:
+            if not user_warnings.keys():  # no warnings for the user
+                await ctx.send(_("That user has no warnings!"))
+            else:
+                for key in user_warnings.keys():
+                    mod = ctx.guild.get_member(user_warnings[key]["mod"])
+                    if mod is None:
+                        mod = discord.utils.get(
+                            self.bot.get_all_members(), id=user_warnings[key]["mod"]
```
mikeshardmind: use `ctx.guild.get_member`

mikeshardmind (changes requested):
There's more changes needed on this.

redbot/cogs/warnings/warnings.py
```diff
@@ -342,45 +342,57 @@ def __init__(self, bot: Red):
...
+        member_settings = self.config.member(user)
+        async with member_settings.warnings() as user_warnings:
+            if not user_warnings.keys():  # no warnings for the user
+                await ctx.send(_("That user has no warnings!"))
+            else:
+                for key in user_warnings.keys():
+                    mod = ctx.guild.get_member(user_warnings[key]["mod"])
+                    if mod is None:
+                        mod = discord.utils.get(
+                            ctx.guild.get_member(), id=user_warnings[key]["mod"]
```
mikeshardmind, line 358: Not really what I meant by that...

My point wasn't to try and make `guild.get_member` work as part of utils.get (this is already properly handled above), but that while this was being changed, we should avoid leaking info from other guilds and not attempt to match outside the current guild.

Vexed01:

PR #3090, opened 2019-10-27
License info command

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

This works on some things we've had in discussion in discord. It's merely putting the technical parts of this forward for now. More is needed on this still.

### DIFF
--- a/changelog.d/3090.feature.rst
+++ b/changelog.d/3090.feature.rst
@@ -1,0 +1,1 @@
+adds a licenseinfo command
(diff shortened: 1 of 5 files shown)


Kowlin:
What we may wanna do in terms of listing is directly show the ``LICENCE.md`` in GitHub, as it does go out of its way to be more user friendly with an more human readable blurb on the top, which we wouldn't have to maintain or verify that it complies with the license.

mikeshardmind:
This does link to that. The problem is that all of the other licenses which we lean on will not fit in a single message, so we need to find a good way to show those to users.

Kowlin (dismissed):

Kowlin (dismissed):

Flame442 (approved):

Kowlin (approved):

PR #3091, opened 2019-10-28
[Image] Only send one Imgur result and update wording for setting API keys

### LINKED ISSUE #3044: imgur search Posts Three Images into Chat
When using the built in imgur search functions, it pulls three image search results from imgur/reddit and posts all three to Discord. When I looked at the code it looks like there's a slice that tells it to post all three. Is this intentional behavior? When the search succeeds on all three that's a huge wall of image posted to Discord chat as a single post. I can edit the slice to only use the first but I figured I'd ask here first in case this is a bug/error or if I'll need to edit this. Just seems like an insane default or I'm just doing something wrong on my end.

Slice in question (same code is used for each type of image search it looks like):

```
for r in results[:3]:
  msg += r["gifv"] if "gifv" in r else r["link"]
  msg += "\n"
```

Thanks!

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

The title says it all really... I've updated the wording for Imgur and GIPHY API keys to be more accurate.
Also made Imgur search only return one result, instead of three.
I tried not using a for loop... but I couldn't make it work without one. At least the fix works lol.
Closes #3044

### DIFF
--- a/changelog.d/image/.bugfix.rst
+++ b/changelog.d/image/.bugfix.rst
@@ -1,0 +1,1 @@
+Correct wording for setting Imgur API keys.
--- a/changelog.d/image/.bugfix.rst
+++ b/changelog.d/image/.bugfix.rst
@@ -1,1 +1,0 @@
-Correct wording for setting Imgur API keys.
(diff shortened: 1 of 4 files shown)


Vexed01:
Vexed is no longer active on Discord, and consequently on this repo. It's just put a burden on me. Too much emotional stress, taking up too much of my life with other IRL stuff coming up...

PR #3105, opened 2019-11-08
[Docs] It's 2019. Let's not say it's 2018.

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Change footers to say 2019 instead of 2018.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,7 +58,7 @@ master_doc = "index"
 
 # General information about the project.
 project = "Red - Discord Bot"
-copyright = "2019, Cog Creators"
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -58,7 +58,7 @@
 
 # General information about the project.
 project = "Red - Discord Bot"
-copyright = "2018, Cog Creators"
+copyright = "2019, Cog Creators"
```
Kowlin: This should be a range being ``2018-2019``

Kowlin (changes requested):

Kowlin (approved):

PR #3106, opened 2019-11-09
[Core] Add deprecation warnings about removal of shared libraries.

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This PR:
- adds deprecation note about shared libraries in Downloader Framework docs
- makes bot send deprecation warning when using `[p]load` and `[p]reload` commands if the repos loaded cogs are from have shared libraries (only works with Downloader loaded)
- makes bot send deprecation warning when using install and update commands if the repos installed/updated cogs are from have shared libraries
- makes bot print deprecation warning when some package tries importing from `cog_shared.*` (sadly only works for the first time the import of specific shared lib happens)

Soo, there are 3 deprecation warnings in different places, I think each of them is worth having but if you think some isn't, I can revert it.

### DIFF
--- a/changelog.d/3106.docs.rst
+++ b/changelog.d/3106.docs.rst
@@ -1,0 +1,1 @@
+Add deprecation note about shared libraries in Downloader Framework docs.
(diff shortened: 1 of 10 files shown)


Jackenmen:
This is now ready for review. I updated the PR's description with the made changes.

mikeshardmind (approved):

PR #3047, opened 2019-10-11
Audio PR part 1.5

Drapersniper:
### Type

- [x] Bugfix
- [x] Enhancement
- [x] New feature

### Description of the changes
NO LONGER A DRAFT,  but this one has all the goodies from #2950 minus the new Menu implementation.

1.  Play + Bump Combo Command (#2522)
2.  Don't shuffle bumped tracks (Added an extra argument to [p]shuffle that allows the person with permissions to say whether bumped tracks should be shuffled or kept at the top of the queue, off by default)
3.  Expand local track support to all file formats (Added support for AAC, AU, M3U, M4A, OPUS, RA, TS, WAV, WMA, 3GP, FLV M4V, MKV, MP4, MOV, WEBM, WMV)
4. Cache DJ setting value to reduce api calls  (more significant on Mongo)
5. Rework UX to standardize messages (Summary of what message is about is the title, and a more  descriptive explanation is added to embed description/footer)
6. Add cooldowns to playlist save, eq save and summon, also add resets in error scenarios.
7. Add a global variant of the existing whitelist/blacklist thats thats priority over the server level whitelist/blacklist

@aikaterna  love you please don't kill me :eyes: 

No need to assign to anyone just as of yet but in theory everything here work since #2950  is fully functional, but ... since this hasn't been tested it I may have messed up some things, once I've taken it through i spin i'll flag it as ready for review.

### Update
This PR is ready to be reviewed

PR #3117, opened 2019-11-13
[Audio] Sort local tracks

Nodja:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

The OS doesn't always return local folder/files listings sorted, this ensures everything gets returned sorted.

### DIFF
--- a/changelog.d/audio/3117.bugfix.rst
+++ b/changelog.d/audio/3117.bugfix.rst
@@ -1,0 +1,1 @@
+Sort local tracks/folders
--- a/redbot/cogs/audio/audio_dataclasses.py
+++ b/redbot/cogs/audio/audio_dataclasses.py
@@ -189,7 +189,7 @@ class LocalPath(ChdirClean):
         for track in self.multirglob(*[f"*{ext}" for ext in self._supported_music_ext]):
             if track.exists() and track.is_file() and track.parent != self.localtrack_folder:
                 tracks.append(Query.process_input(LocalPath(str(track.absolute()))))
-        return tracks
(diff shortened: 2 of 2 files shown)


Drapersniper:
This is already handled in #3047 
Unfortunately casefold is not the desired behaviour here and  just casting a Query or LocalTrack object to string won’t give you a user friendly text that is displayed.

Nodja:
The casting is done on the lambda function passed to the sorted function, it still returns a LocalPath object.

But if it's already handled in another PR this isn't needed, apologies for the noise, my search efforts didn't find that PR.

Drapersniper:
No problem, just to explain it though so you/anyone reading this is aware.
> The casting is done on the lambda function passed to the sorted function, it still returns a LocalPath object.
> 
> But if it's already handled in another PR this isn't needed, apologies for the noise, my search efforts didn't find that PR.

Yeah what i meant is that each of these 2 objects have 2 different string stats, 1 to be handled by the code and 1 to be displayed to the user.

When sorting this we would have to sort based on the user shown value not the code value, otherwise when displaying to the users things still may not look sorted.

An example is LocalTrack when cast to string use 
```    def to_string(self):
        try:
            return str(self.path.absolute())
        except OSError:
            return str(self._path)
 ```
 i.e the full path, but when displayed to the user there are formated so hide non relevant paths for privacy and security purposes also so strings that goes into the embeds are not too long

However `.to_string_hidden()` is used to generate the string to shown to the user we we would have to sort this using the output of this method so that the order is actually sorted when displaying it to the user.

PR #3121, opened 2019-11-14
Increased clarity of wording in info command

Vexed01:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

[Draper's variation](https://discordapp.com/channels/133049272517001216/133049878837329920/644274713853427720) of ClearlyElevated's [origianal suggestion](https://discordapp.com/channels/133049272517001216/133049878837329920/644272849758060556) in #feature-ideas, Discord.
Replaces "This is an..." with "Example Bot Username is an...".
Possible variation: use the bot's nick instead of name.

### DIFF
--- a/changelog.d/3121.enhance.rst
+++ b/changelog.d/3121.enhance.rst
@@ -1,3 +1,1 @@
-Use the bot's username in ``[p]info`` instead of "This"
-- Before: "This is an..."
-- Now: "Bot's Username is an..."
+Change ``[p]info`` to say "This bot is an..." instead of "This is an..." for clarity.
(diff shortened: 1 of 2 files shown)


mikeshardmind:
I don't think this message should change in this manner. At most, it should change from

> This is an instance ...

to

> This bot is an instance ...

There's no need to format in a bot's name, and depending on a bot's name could actually make the text *less clear*

If people would like to elaborate about their bot, that is exactly the purpose of the custominfo setting. This section is not about their bot, but about Red.

mikeshardmind (approved):
:+1: Thanks for the quick follow up on this. I do agree that adding that one word actually makes it clearer even with my disagreement on formatting in the bot's name here.

PR #3144, opened 2019-11-19
admin string clarity changes

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This improves the clarity of the strings when a user isn't allowed to do something in the admin cog due to discord hierarchy.

### DIFF
--- a/changelog.d/admin/3144.enhance.rst
+++ b/changelog.d/admin/3144.enhance.rst
@@ -1,0 +1,2 @@
+Improves the clairty of user facing messages in the admin cog when the user is not allowed
+to do something due to Discord hierarchy rules.
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -34,24 +34,21 @@ HIERARCHY_ISSUE_REMOVE = _(
 )
 
 USER_HIERARCHY_ISSUE_ADD = _(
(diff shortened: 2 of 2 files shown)


mikeshardmind:
Additional thanks to [Dav](https://translate.discord.red/profile/Dav_) for reporting this problematic string to us on Crowdin.

image

Dav-Git:
I just started revising the German Translation so more might come up in the near future. If there is a place you want me to collect these at please let me know. If not I will just keep raising issues on Crowdin :)

mikeshardmind:
If one is easier than another, use whichever is easier. Opening issues about strings here may bring it to a wider audience than just those watching the crowdin issues though.

Flame442 (changes requested):
These strings need to also include the fact that if the highest role is the same as the role in question it will fail.
EDIT: The bot hierarchy messages above these also need to be fixed.

Cog-CreatorsBot (changes requested):

mikeshardmind:
Strings in this are still gonna end up being improved, but @Flame442 is working on a better resolution of this (conversation in Discord) which involves cleaning up more of the user interaction than just the raw output, but also the usage and command syntax. As such, I'm closing this in favor of not duplicating work.

PR #3146, opened 2019-11-23
[Core] Dispatch `on_red_api_tokens_update` event on api keys update

### LINKED ISSUE #3134: Emit events when a shared API token is changed
# Feature request

- [ ] Cog
- [ ] Command
- [X] API functionality

Should emit the name of the service which was updated. This will allow cogs to cache the values and only update when if they change.

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Resolves #3134

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -8,6 +8,7 @@ from enum import Enum
 from importlib.machinery import ModuleSpec
 from pathlib import Path
 from typing import Optional, Union, List, Dict, NoReturn
+from types import MappingProxyType
 
 import discord
 from discord.ext.commands import when_mentioned_or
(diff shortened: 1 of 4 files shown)


redbot/core/bot.py
```diff
@@ -582,6 +582,7 @@ def colour(self) -> NoReturn:
 
         async with self._config.custom(SHARED_API_TOKENS, service_name).all() as group:
             group.update(tokens)
+        self.dispatch("red_api_tokens_update", service_name, group)
```
mikeshardmind: This should be changed to either dispatch a mappingproxy or just the service name (indicating to cogs they may need to re-fetch the data if they use that service)

I'd like to avoid issues with dispatching a mutable object here.

Edit: and we would need to prevent it from being a proxy of the object still held by the code calling this too

Edit2: Nope, we take in as kwargs, and config is already getting it's own reference at `__aexit__`

mikeshardmind (changes requested):

mikeshardmind (approved):
:+1:

Issue #3148, opened 2019-11-23
Caching for prefixes/whitelists/blacklists


mikeshardmind:
Caching these should improve performance for DB backends with a minimal memory overhead for doing so.

PR #3150, opened 2019-11-23
Cache prefixes

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

 - This works towards #3148
 - Ends up centralizing some logic
   - Including that prefixes should be a reverse sorted list

Blacklist/whitelist should probably also end up being cached, but as this runs on *every* message to determine if there's a valid command prior to all the other checks, let's start here.


TrustyJAID (approved):
This is working as intended.

PR #3154, opened 2019-11-26
Remove the API from the audio cog

mikeshardmind:

Removes an undocumented API which was added into the audio cog. This API was not in a release version, nor documented as an addition in 3.2. Additionally it is part of a cog. It's removal should not be breaking anyone, as nobody should feasibly be using this.
Removes a couple of dispatches.

 - API as a part of the cog itself is in direct conflict with goals stated in #2804
 - Features this was intended to enable can be enabled in other more appropriate ways later on
  - The majority of dispatches were retained, as they may be useful as informational messages to enable additional functionality in other cogs

### DIFF
--- a/changelog.d/audio/3152.misc.rst
+++ b/changelog.d/audio/3152.misc.rst
@@ -1,0 +1,1 @@
+remove an undocumented API from audio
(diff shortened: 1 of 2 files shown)


Drapersniper (approved):
LGTM

mikeshardmind:
(Will resolve conflicts later, review can wait)

PR #3165, opened 2019-12-03
[Audio] Better error handling after multiple subsequential errors

Drapersniper:
### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

Auto disconnects the player if there are more than 5 errors in the last 10 seconds. (Likely caused by ... YouTube or playing local album with broken tracks etc)

Benefit of this is to reduce message spam caused by the bot when letting them know the playback broke

Also while at it fixed 1 bug where the DJ command was crashing

### DIFF
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -590,38 +590,38 @@ class Audio(commands.Cog):
                     break
             if repeat:
                 player.current = None
-            self._error_counter.setdefault(player.channel.guild.id, 0)
-            if player.channel.guild.id not in self._error_counter:
-                self._error_counter[player.channel.guild.id] = 0
+            with contextlib.suppress(Exception):
+                self._error_counter.setdefault(player.channel.guild.id, 0)
+                if player.channel.guild.id not in self._error_counter:
+                    self._error_counter[player.channel.guild.id] = 0
 
-            early_exit = await self.error_handler(player)
-            if early_exit:
-                self._disconnected_players[player.channel.guild.id] = True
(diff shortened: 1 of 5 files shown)


PredaaA (dismissed):
I tested this PR and works as intended.

mikeshardmind (dismissed):

mikeshardmind:
Needs conflict resolution.

Drapersniper:
Conflicts resolved

redbot/cogs/audio/audio.py
```diff
@@ -122,7 +122,7 @@ def __init__(self, bot):
             status=False,
             use_external_lavalink=False,
             restrict=True,
-            current_version=redbot.core.VersionInfo.from_str("3.0.0a0").to_json(),
+            current_version=redbot.core.VersionInfo.from_str(__version__).to_json(),
```
mikeshardmind: What is this value being used for right now?

If it is or has been used for migrating data, this is not a correct change and will break anyone upgrading from the 3.0.0 -> 3.2.0 as the default here should be the lowest version, with migrations setting the version related to their migration.

Drapersniper: afaik this is not currently used for anything ... but @aikaterna  should be able to confirm it already existed before I touched audio..

Drapersniper:
Do not merge this one ... Merge #3205

redbot/cogs/audio/audio.py
Drapersniper: This is unused anywhere within red ... I've remove this default

Drapersniper:
Will setup a CPR for all of this so itsd easier to follow ... less likely to fuck up)

PR #3183, opened 2019-12-11
[Core] Add `redbot --debuginfo` flag

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

A somewhat cool feature from `redbot-launcher` put in `redbot`

// I'm putting those imports at the top since it looks like `redbot.core.core_commands` 
 would import those anyway.

### DIFF
--- a/changelog.d/3183.enhance.rst
+++ b/changelog.d/3183.enhance.rst
@@ -1,0 +1,1 @@
+Add ``redbot --debuginfo`` flag that shows useful information for debugging.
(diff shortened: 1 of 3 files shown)


mikeshardmind (changes requested):
If you could add `sys.executable` to the debug info output for use in support channels, that would be awesome. Everything else here looks fine.

mikeshardmind (approved):

PR #3203, opened 2019-12-20
[Downloader] Actually use disabled key in updates

### LINKED ISSUE #3173: [Downloader] Actually use disabled key in updates
~~Regression from 3.1.8 -> dev, disabled key doesnt prevent install~~ False Alarm on regression.

Same from 3.1.8 -> dev, disabled key doesnt prevent loading an updated cog which was made disabled.

CC: @jack1142 , discord API issues may make it easier to discuss issue here.

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

WIP - needs testing and I also have alternative solution that detects disabled cogs in `Repo`'s methods, instead of `Downloader`'s methods in case it could work better for this.

Fixes #3173

### DIFF
--- a/changelog.d/downloader/3173.misc.rst
+++ b/changelog.d/downloader/3173.misc.rst
@@ -1,0 +1,1 @@
+Downloader will no longer show update for a cog if the latest version is disabled through ``info.json`` file.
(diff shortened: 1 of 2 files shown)


Jackenmen:
Back when I made this I wanted to test some edge case that wasn't directly related to this PR anyway. Marking this as ready for review as there's no reason to block a rather easy fix.

mikeshardmind (approved):

PR #3209, opened 2019-12-22
[3.2][Audio] Part 1

Drapersniper:
### Type

- [x] Bugfix
- [x] Enhancement
- [x] New feature

### Description of the changes

- Play + Bump Combo Command (#2522)
- Don't shuffle bumped tracks (Added an extra argument to [p]shuffle that allows the person with permissions to say whether bumped tracks should be shuffled or kept at the top of the queue, off by default)
- Expand local track support to all file formats (Added support for AAC, AU, M3U, M4A, OPUS, RA, TS, WAV, WMA, 3GP, FLV M4V, MKV, MP4, MOV, WEBM, WMV)
- Cache DJ setting value to reduce api calls (more significant on Mongo)
- Rework UX to standardize messages (Summary of what message is about is the title, and a more descriptive explanation is added to embed description/footer)
- Add cooldowns to playlist save, eq save and summon, also add resets in error scenarios.
- Add a global variant of the existing whitelist/blacklist thats thats priority over the server level whitelist/blacklist

PR #3210, opened 2019-12-22
[3.2][Audio] Part 2

Drapersniper:
### Type

- [x] Bugfix
- [x] Enhancement
- [x] New feature

## This is the second PR for the CPR, it depends on #3209 and should be merged after it.

Auto disconnects the player if there are more than 5 errors in the last 10 seconds. (Likely caused by ... YouTube or playing local album with broken tracks etc)

Benefit of this is to reduce message spam caused by the bot when letting them know the playback broke

Also while at it fixed 1 bug where the DJ command was crashing


aikaterna:

PR #3211, opened 2019-12-22
[3.2][Audio] Part 3

Drapersniper:
### Type

- [x] Bugfix
- [x] Enhancement
- [x] New feature

## This is the third PR for the CPR, it depends on #3209 and #3210 and should be merged after them.

Fixes a handful of attribute errors that could happen in the event_handle of audio

PR #3212, opened 2019-12-22
[3.2][Audio] Part 4

Drapersniper:
### Type

- [x] Bugfix
- [x] Enhancement
- [x] New feature

## This is the forth PR for the CPR, it depends on #3209, #3210 and #3211 and should be merged after them.

This PR Changed the audio cache database to use APSW instead of Databases, it also changes the Type of the dates stored in the Database


redbot/cogs/audio/apis.py
```diff
@@ -32,129 +17,33 @@
...
-        "update": _UPDATE_LAVALINK_TABLE,
-    },
-}
-
 _TOP_100_GLOBALS = "https://www.youtube.com/playlist?list=PL4fGSI1pDJn6puJdseH2Rt9sMvt9E2M4i"
 _TOP_100_US = "https://www.youtube.com/playlist?list=PL4fGSI1pDJn5rWitrRWFKdm-ulaFiIyoK"
 
+_database: CacheInterface = None
+_bot: Red = None
+_config: Config = None
```
mikeshardmind: These type hints should be `Optional[...]`

Drapersniper: I know the IDE was complaining, but should they are initialised before anything in the module is used and I came across some weirdness with pycharm where when I done optional because it’s set as none here it would consider the value of this always none when it came to code completion so I though I was messing up the typehints

mikeshardmind: Optional is correct here, though ideally during refactoring planned for #3226 we get rid of all module globals anyhow.

aikaterna:

PR #3213, opened 2019-12-22
[3.2][Audio] Part 5 (Last)

Drapersniper:
### Type

- [x] Bugfix
- [x] Enhancement
- [x] New feature

Okay so this should in theory significantly improve audio performance across the border for folks that use the playlist function

@PredaaA using Mart on dev which has #2921 (which significantly reduces JSON size went from 50+MB prior to #2921 to 18.6MB prior to this PR)

With this PR his audio config went from 18.6MB to 287.12 KB... (and say we want to also migrate the EQUALIZER scope to an SQL table ... his config was reduced to 42.97 KB) which should significantly reduce write times and overall make audio commands feel speedier.

(Some more data With this PR @ Cat's config went from over 37MB to only 33kb, speeding up all commands that touches config)


redbot/cogs/audio/apis.py
```diff
@@ -194,12 +85,11 @@ def _make_token_auth(client_id: Optional[str], client_secret: Optional[str]) ->
...
-        if self.client_id is None or self.client_secret is None:
-            tokens = await self.bot.get_shared_api_tokens("spotify")
-            self.client_id = tokens.get("client_id", "")
-            self.client_secret = tokens.get("client_secret", "")
+        tokens = await self.bot.get_shared_api_tokens("spotify")
+        self.client_id = tokens.get("client_id", "")
+        self.client_secret = tokens.get("client_secret", "")
 
-    async def _request_token(self) -> dict:
+    async def _request_token(self) -> Mapping:
```
mikeshardmind: `MutableMapping` should be used here instead of Mapping for the type (Same anywhere else you return a mapping which gets modified later so that dev toolings don't complain when opening the file)

redbot/cogs/audio/audio_dataclasses.py
```diff
@@ -367,18 +448,20 @@ def _parse(track, **kwargs):
...
-                            match = re.search(_re_youtube_timestamp, track)
+                            match = re.search(_RE_YOUTUBE_TIMESTAMP, track)
                             if match:
                                 returning["start_time"] = int(match.group(1))
                         if _has_index:
-                            match = re.search(_re_youtube_index, track)
+                            match = re.search(_RE_YOUTUBE_INDEX, track)
                             if match:
                                 returning["track_index"] = int(match.group(1)) - 1
                         if all(k in track for k in ["&list=", "watch?"]):
```
mikeshardmind, line 467: (Note for later, not necessary to change in this PR, just noticed while reading through the full context)

The logic from here down including all branches attached to this if statement gets convoluted. There are 2 branches with identical implementations, separated by another branch. It's not valid to merge the branches directly either, and it won't be immediately obvious as to why. This should probably be looked at for a clarity improvement later on.

mikeshardmind:
I've not reviewed this or it's components in entirety or tested the full set of changes here.

I saw a few things which can have minor changes to be slightly better with tooling (mostly type hints)

I saw some logic and SQL that looks like it should work but should likely be handled as part of #3226 (The SQL even has a nice comment to some prior discussion regarding this in the code, +1)

I didn't see anything problematic code wise beyond this, but I have not tested the changes for being correct on the behavioral side. As I have been less involved with Audio's overall design and direction, 

I'd prefer to leave judging if the behavioral parts are fine to other people more capable of determining that, especially since some of these changes are going to be more noticeable for people running larger public Red instances (such as cat, bb8, martine)

aikaterna:
Using [p]search list searchterm with dj role on allows people to queue tracks when they shouldn't be allowed to.

Vote mode only, with DJ mode off:
The audioset vote setting should protect the prev command as well. People can use previous without being told no if vote is on.
Summon should also be protected in this mode, which it isn't at the moment.


Jukebox on, user had no currency

```
c:\envs\3.1.8\lib\site-packages\redbot\cogs\audio\audio.py:2770: RuntimeWarning: coroutine 'can_spend' was never awaited
  if not await self._currency_check(ctx, guild_data["jukebox_price"]):
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
c:\envs\3.1.8\lib\site-packages\redbot\cogs\audio\audio.py:2781: RuntimeWarning: coroutine 'can_spend' was never awaited
  if not await self._currency_check(ctx, guild_data["jukebox_price"]):
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
[2019-12-31 10:17:53] [ERROR] red: ValueError
  File "c:\envs\3.1.8\lib\site-packages\discord\ext\commands\core.py", line 79, in wrapped
  File "c:\envs\3.1.8\lib\site-packages\redbot\cogs\audio\audio.py", line 2770, in play
    if not await self._currency_check(ctx, guild_data["jukebox_price"]):
  File "c:\envs\3.1.8\lib\site-packages\redbot\cogs\audio\audio.py", line 7633, in _currency_check
    await bank.withdraw_credits(ctx.author, jukebox_price)
  File "c:\envs\3.1.8\lib\site-packages\redbot\core\bank.py", line 266, in withdraw_credits
    humanize_number(bal, override_locale="en_US"),
ValueError: Insufficient funds 10 > 0
c:\program files\python37\Lib\asyncio\events.py:88: RuntimeWarning: coroutine 'can_spend' was never awaited
  self._context.run(self._callback, *self._args)
RuntimeWarning: Enable tracemalloc to get the object allocation traceback
```

Jackenmen:
Since Draper asked me, I'm mentioning this here.
`[p]playlist download` currently passes `StringIO` object to `discord.File` even though it only supports objects with binary data (i.e. `BytesIO`). While this haven't caused any troubles yet, this isn't the safeest thing to do.
I would advice using `BytesIO` objects instead and also explicitly using utf-8 encoding for saving and loading them to avoid any weird encoding issues.
Using `StringIO` made it save playlist files with ascii encoding, which utf-8 so happens to be compatible with.

aikaterna:
I will be moving further testing and commenting to #3244 via Draper's direction on this.

aikaterna:

PR #3214, opened 2019-12-22
[Core] Use `[p]` for command prefix in help docstring of `[p]removepath`

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature


Flame442 (approved):
Technically this requires #3039 to be fixed, however `[p]` is a better option than `!` if the actual prefix isn't able to be used and this PR prepares this text to be ready for that fix.

Cog-CreatorsBot (approved):
Approved by Flame#2941

PR #3215, opened 2019-12-23
[WIP] [Downloader] Set repo's folder path as cwd when installing requirements

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

I haven't tested this, I didn't think about this very thoroughly either, but this should allow cog creators to install pip package directly from their already downloaded repo.


Jackenmen:
I might do something about this in the future but it will most likely be a new PR as this one is of no much use anyway.

PR #3221, opened 2019-12-25
[Core] Prevent users from locking out themselves or guild owner with localblacklist

### LINKED ISSUE #3207: [Core] Users can lock out themselves with `[p]locablacklist add`
# Command bugs

#### Command name

`[p]localblacklist add`

Core

~~Well, to be honest, I was expecting this to happen.~~ I was expecting to get an error saying that I can't lock out myself from using the bot.

#### What actually happened?

I locked myself out.

1. Have permissions to use `[p]localblacklist` (e.g. have admin role), but don't be bot owner.
2. Use `[p]localblacklist add YOUR_USER_ID`

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #3207
Partial solution for #3222 (guild owners will be able to run commands even when they're not on whitelist but people who aren't guild owners can still lock themselves out if they don't use this `[p]localwhitelist add` command carefully)


Jackenmen:
ehehe, just figured that you can also lock yourself out with localwhitelist 😄

redbot/core/core_commands.py
```diff
@@ -1903,9 +1903,16 @@ class Core(commands.Cog, CoreLogic):
...
-            return
+        if user:
+            if user_or_role.id == ctx.author.id:
+                await ctx.send(_("You cannot blacklist yourself!"))
+                return
+            if user_or_role.id == ctx.guild.owner_id and not await ctx.bot.is_owner(ctx.author):
+                await ctx.send(_("You cannot blacklist the guild owner!"))
+                return
+            if await ctx.bot.is_owner(user_or_role):
+                await ctx.send(_("You cannot blacklist an owner!"))
```
mikeshardmind: I feel like this leaves less room for misunderstanding with end users.
```suggestion
                await ctx.send(_("You cannot blacklist a bot owner!"))
```

mikeshardmind (changes requested):
Also warrants minor changes to the actual logic used in the bot for this to retroactively prevent guild owners from being impacted by either this or local whitelist as well, up to you if you think that should be a separate PR or not.

mikeshardmind (dismissed):

mikeshardmind (approved):

PR #3227, opened 2019-12-27
Allow a small bit of extra async cleanup, without preventing cog unload.

### LINKED ISSUE #2474: Unclosed client session/unclosed connector on 3.0.1 shutdown or restart
# Command bugs

#### Command name

restart or shutdown

I don't believe this is being caused by a cog, unless it's core, as I would get this error with all cogs unloaded

Restart or shutdown shuts down the bot processes cleanly

#### What actually happened?

Received this in the console:
```
_GatheringFuture exception was never retrieved
future: <_GatheringFuture finished exception=CancelledError()>
concurrent.futures._base.CancelledError
Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x101e6e400>
Unclosed connector
connections: ['[(<aiohttp.client_proto.ResponseHandler object at 0x101fac8e8>, 7.72160103)]']
connector: <aiohttp.connector.TCPConnector object at 0x101e6e3c8>
```

Use Red 3.0.1, and use a shutdown or restart command.

While pip installing the bot on one machine I also noticed a version conflict with discord.py and aiohttp and websockets, but I did not save that info and I'm having trouble reproducing that. Also while trying to diagnose this I rolled back aiohttp to 3.4 on two machines but was still getting this error on shutdown.

mikeshardmind:
### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

Adds a new method to the bot for use by cogs: ``submit_for_close``

Passing this an object which implements a coroutine ``close`` which can be called with no arguments, will have the bot handle cleaning up the object asynchronously from synchronous code.

Designed with things such as aiohttp client sessions needing to be closed in `cog_unload` in mind, but allows any object with a similar close method to be passed.

Also has Audio use it, which resolves #2474

### DIFF
--- a/changelog.d/2474.feature.rst
+++ b/changelog.d/2474.feature.rst
@@ -1,0 +1,2 @@
+New method: ``bot.submit_for_close``.
+Can pass an object with a coroutine method ``close`` to be closed by the bot asynchronously from synchronous cleanup code (such as ``cog_unload``)
--- a/changelog.d/audio/2474.bugfix.rst
+++ b/changelog.d/audio/2474.bugfix.rst
@@ -1,0 +1,1 @@
+Prevent some console noise on audio unload
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -6931,9 +6931,9 @@ class Audio(commands.Cog):
                 pass
 
     def cog_unload(self):
+        self.bot.submit_for_close(self.session)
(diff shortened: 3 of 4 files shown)


mikeshardmind:
Rebased over more recent changes.

mikeshardmind:
This isn't even needed anymore with the other startup/shutdown changes handling this more gracefully.

PR #3237, opened 2019-12-30
[Streams] Significantly reduce number the quota usage for YouTube Data api

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

1. Stop using the search endpoint...
2. User the channels RSS feed to check the latest videos for less quota usage ( This can be changed to use the channel endpoing)
3. Check for live streams every 5 minutes instead of 1 minute (`[p]streamset timer` was added so this can be customized down to 1 minute)
4. reduce the number of repeated API calls by caching non-livestreams and skipping them in future calls.

### DIFF
--- a/redbot/cogs/streams/streamtypes.py
+++ b/redbot/cogs/streams/streamtypes.py
@@ -135,15 +135,16 @@ class YoutubeStream(Stream):
         log.critical(f"not_livestreams: {self.not_livestreams}")
         if not self.livestreams:
             raise OfflineStream()
-
-        for livestream_id in self.livestreams:  # This is technically redundant since we have the
-            # info from the RSS ... but incase you dont wanna deal with fully rewritting the
-            # code for this part, as this is only a 2 quota query.
-            params = {"key": self._token["api_key"], "id": livestream_id, "part": "snippet"}
+        # This is technically redundant since we have the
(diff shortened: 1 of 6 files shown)


redbot/cogs/streams/streamtypes.py
```diff
@@ -162,7 +209,7 @@ def __init__(self, **kwargs):
             self.id = await self.fetch_id()
 
         url = TWITCH_STREAMS_ENDPOINT + self.id
-        header = {"Client-ID": str(self._token), "Accept": "application/vnd.twitchtv.v5+json"}
```
palmtree5, line 165: From https://dev.twitch.tv/docs/v5#requests


For client IDs created on or after May 31, 2019, the only available version of the Kraken API is v5.  For client IDs created prior to May 31, 2019, use the `application/vnd.twitchtv.v5+json` header on your requests to access v5 of the Kraken API.

Probably shouldn't remove that header

palmtree5 (dismissed):

redbot/cogs/streams/streamtypes.py
TrustyJAID: This is blocking for #3487 which will remove this issue entirely by moving to the helix API.

TrustyJAID (approved):
Looks good to me.

PR #3240, opened 2019-12-30
[Utils] Privatize internal utils

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature


mikeshardmind (dismissed):
We will still need to do a final pass before release to make sure nothing's been missed, but this is progress towards the more formal version guarantees from 3.2 forward, +1

mikeshardmind (approved):
:+1:

PR #3242, opened 2019-12-31
[3.2.x][Audio]  Persist queues on reboot/reload

Drapersniper:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Queues will now be able to auto restart on bot/cog start up while remembering the queue order and content prior to the shutdown

Note: This is still missing a toggle to allow Server admins to enable this behaviour


Drapersniper:
Blocked by #3459

Drapersniper:
This hasn't been implemented in the feature branch.

Whoever does implemented it, if ever implemented please be careful with the edge cases here.

It should be good to go as it is here, however with the refactor this will not work as expected.

PR #3245, opened 2019-12-31
do better with loop cleanup

mikeshardmind:

It should reduce console noise during cleanup.

Should also fix the issues some people had on windows with a keyboard interrupt hanging rather than quiting.

### DIFF
--- a/.bandit.yml
+++ b/.bandit.yml
@@ -1,0 +1,396 @@
+
+### Bandit config file generated
+
+### This config may optionally select a subset of tests to run or skip by
+### filling out the 'tests' and 'skips' lists given below. If no tests are
+### specified for inclusion then it is assumed all tests are desired. The skips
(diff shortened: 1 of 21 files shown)


mikeshardmind:
This isn't review ready. Some of this needs changing, and it's going to be used to cleanup the logic in main more.

mikeshardmind:
This should be ready to go now.

mikeshardmind:
Handling cleanup properly on windows requires something which was fixed in python3.8 ([bpo-23057](https://bugs.python.org/issue23057))

Since we have to update the deps here for this, we'll be doing it for all platforms to ensure feature availability is uniform across platforms

mikeshardmind:
Current PR state disables our pylint check due to https://github.com/PyCQA/pylint/issues/3247

More internal discussion around this happening.

Kowlin (approved):

PR #3248, opened 2020-01-01
Embed pagination fixing (pt2)

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes buggy first embed of paginated help CC: @jack1142

### DIFF
--- a/changelog.d/3248.misc.rst
+++ b/changelog.d/3248.misc.rst
@@ -1,0 +1,1 @@
+fix a buggy first page of paginated help
--- a/redbot/core/commands/help.py
+++ b/redbot/core/commands/help.py
@@ -278,14 +278,15 @@ class RedHelpFormatter:
 
         for f in fields:
             f_len = len(f.value) + len(f.name)
-            if curr_group and (f_len + current_count > max_chars):
+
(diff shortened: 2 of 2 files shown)


Jackenmen:
This PR doesn't fully address the problem, output of `[p]permissions` still gives empty page:
image

mikeshardmind:
Note with the above, this particular case isn't actually broken, but it might appear so to users. We may want to reconsider the embed generation strategy more later on.

Jackenmen:
In my opinion, we should split "Commands" fields so that some part of it is shown in first page and some is shown in next page with "Commands (continued)", same as we already do when commands in next pages don't fit in one embed.
chrome_2020-01-04_12-38-39

mikeshardmind:
That's actually not how we're splitting fields currently, though I understand why it can appear that way. This needs a larger overhaul than it will get prior to 3.2 with a more generalized embed paginator than what's currently in use.

mikeshardmind:
Long term we should resize fields, short term, we're not ever having an embed with no fields. This can't go over discord limits, but might exceed some page size settings.

Kowlin (approved):
Working as intended.

PR #3249, opened 2020-01-01
Fix ctx.clean_prefix for *new* discord behavior

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Discord made all mentions with the new text boxes use the `<@!` format, instead of following the prior behavior of " `!` is for telling the client to check a nickname for render"

This fixes ctx.clean_prefix which relied on them not changing that without warning.

### DIFF
--- a/changelog.d/3249.bugfix.rst
+++ b/changelog.d/3249.bugfix.rst
@@ -1,0 +1,1 @@
+Fix ``ctx.clean_prefix`` for undocumented changes from discord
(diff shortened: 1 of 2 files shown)


mikeshardmind:
Clarification: While both mention formats have been valid for ages, we made assumptions based on both old behavior and purpose of the varying formats that *at least within the lifespan of a context* this would be "correct enough", this change just makes it more overall correct.

Flame442 (approved):

Cog-CreatorsBot (approved):
Approved by Flame#2941

PR #3261, opened 2020-01-03
Be quieter in expected shutdown cases

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

- shutdown/restart commands are less noisy.
- quitting with keyboard interrupt is warned against, then handled as best as possible.

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -352,22 +352,52 @@ async def shutdown_handler(red, signal_type=None):
     else:
         log.info("Shutting down from unhandled exception")
         exit_code = 1
-    await red.logout()
-    await red.loop.shutdown_asyncgens()
-    pending = [t for t in asyncio.all_tasks() if t is not asyncio.current_task()]
-    [task.cancel() for task in pending]
-    await asyncio.gather(*pending, return_exceptions=True)
-    sys.exit(exit_code)
+
+    try:
+        await red.logout()
+    finally:
+        # Allows transports to close properly, and prevent new ones from being opened.
+        await red.loop.shutdown_asyncgens()
+        # Then cancels all outstanding tasks other than ourselves
+        pending = [t for t in asyncio.all_tasks() if t is not asyncio.current_task()]
+        [task.cancel() for task in pending]
+        await asyncio.gather(*pending, return_exceptions=True)
+        # And we're done!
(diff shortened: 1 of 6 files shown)


Drapersniper (dismissed):
LGTM, also fixed spam in logs i was seeing during a restart

Drapersniper:
During graceful shudowns shudowns are a lot faster on windows.

Forced shutdowns are significantly faster than before too however some times if python is interacting with a task it will throw a trace after shutting the bot down.

image


This is not the case 100% of the time, however we are warning the user with `Please do not use Ctrl+C to Shutdown Red! (attempting to die gracefully...)`  as its an attempt for a graceful shutdown and they are warned to not use `Ctrl+C` so this PR look fine if we are okay with the traces being raise some times on Windows.

Drapersniper (dismissed):

Jackenmen:
I still don't have a graceful shutdown with this PR, getting this when using `[p]shutdown` with non-0 exit code:
```
ExitCodes.SHUTDOWN
CRITICAL:root:Caught unhandled exception: ExitCodes.SHUTDOWN
```

redbot/__main__.py
```diff
@@ -345,31 +345,56 @@ def handle_early_exit_flags(cli_flags: Namespace):
...
+    """
+    Logs unhandled exceptions in other tasks
+    (to standard error, not logs, this applies to all tasks and futures)
+    """
     msg = context.get("exception", context["message"])
-    if isinstance(msg, KeyboardInterrupt):
-        # Windows support is ugly, I'm sorry
-        logging.error("Received KeyboardInterrupt, treating as interrupt")
-        signal_type = signal.SIGINT
+    if isinstance(msg, KeyboardInterrupt, SystemExit):
```
Jackenmen:
```suggestion
    if isinstance(msg, (KeyboardInterrupt, SystemExit)):
```

mikeshardmind: Thanks for catching that one

mikeshardmind:
Proactor event loop has an issue right now not ignoring already being closed. I'll handle this later tonight, and contribute upstream if it doesnt already have someone assigned.

Kowlin (approved):
Working as intended.

PR #3262, opened 2020-01-03
[Docs] Ensure users have setuptools and wheel

mikeshardmind:

Just ensures installing setuptools and wheel in our instructions to make sure users have everything they need.

### DIFF
--- a/changelog.d/3262.docs.rst
+++ b/changelog.d/3262.docs.rst
@@ -1,0 +1,1 @@
+add line about setuptools and wheel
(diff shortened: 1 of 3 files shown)


Jackenmen:
Maybe we should add `pip` there as well?

mikeshardmind:
pip is included with pyenv, but should be installed separately on Debian based systems with the python3-pip package (This is already handled in our docs)

Kowlin (approved):
Looks good

PR #3263, opened 2020-01-04
Fix docs building

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This sidesteps an issue on RTD for now. We won't be able to use `:=` in core until RTD has python3.8 support, but we still benefit from the bugfix which prompted the bump.

3rd party cog creators are free to use such features.


mikeshardmind:
Admin merging this. If it causes issues or RTD doesn't pick it up properly, we'll have it reverted before release.

PR #3269, opened 2020-01-04
[Setup] Stop logging to disk

### LINKED ISSUE #2917: [Setup] Don't use working directory for storing setup logs
Currently `redbot-setup` command creates a folder named "red_setup_logs" in working directory and I think this behaviour is too invasive and unexpected. I am not entirely sure where the setup logs should be stored instead, possibly in appdirs location, but touching working directory without any information about it just isn't a good way to do it.

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Fixes #2917:
> I'm perfectly fine with redbot-setup not logging to disk

### DIFF
--- a/changelog.d/3269.enhance.rst
+++ b/changelog.d/3269.enhance.rst
@@ -1,0 +1,1 @@
+I'm perfectly fine with redbot-setup not logging to disk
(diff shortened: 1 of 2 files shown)


mikeshardmind (dismissed):

mikeshardmind:
Only failing on the linkcheck, rather than cause even more requests, admin merging.
image

mikeshardmind (approved):

PR #3276, opened 2020-01-06
[CI Docs] Allow linkcheck to retry before declaring link broken

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

We all know linkcheck annoyances, this might mitigate our issues with it.

### DIFF
--- a/changelog.d/3276.misc.rst
+++ b/changelog.d/3276.misc.rst
@@ -1,0 +1,1 @@
+Made docs linkcheck try link 3 times before declaring it broken.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -207,6 +207,7 @@ texinfo_documents = [
 # A list of regular expressions that match URIs that should not be
 # checked when doing a linkcheck build.
 linkcheck_ignore = [r"https://java.com*", r"https://chocolatey.org*"]
+linkcheck_retries = 3
 
 
 # -- Options for extensions -----------------------------------------------


mikeshardmind:
I'm actually a little concerned about this one. Most of the time where we've seen a link check fail was when multiple PRs were being updated at once and we were getting timed out by remote hosts.

Tobotimus:
So are you saying that this PR would likely make the problem worse..?

mikeshardmind:
Probably not worse for us, as it still requires it to fail at least once for anything to be different, but I don't like the idea of this potentially spamming out more requests when the main failure mode we've seen are being timed out by the remote host.

It might be more valuable to not treat the link check as an error, at least if we can get a GH action or other clear notification that allows us to check if a link really does need replacing.

Otherwise, we can still just restart builds which are erroring in known good cases, without losing the current benefits.

I definitely think it's worth addressing, I'm just not sure if this is how we should.

Jackenmen:
I'm not sure if you can set up recurring link check, but that might be better than doing link check on each PR. It would however be good to also check if any newly added links in PR work (though I don't really know if this can be done sanely)
I don't think it is that bad to do one or two retries for that one link that sometimes *may* fail if there's high chance that's gonna mitigate the issue but it is true that there are alternatives.

Kowlin (approved):
Looks good.

PR #3289, opened 2020-01-09
minimize the launcher

mikeshardmind:

Strips the launcher of all functionality which is handled by something else.


kennnyshiwa:
Everything seems good here, auto restart and normal start both work and no issues were present

PR #3292, opened 2020-01-09
[3.2.0 Docs] Some clarifications

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

- Ensures we don't advertise the dev version in the normal install guide
- Clarifies that certain chat formatting is pre-escaped

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ notifications:
   email: false
 
 python:
-- 3.8.1
+- 3.8.0
 env:
   global:
     - PIPENV_IGNORE_VIRTUALENVS=1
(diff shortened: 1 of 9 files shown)


Jackenmen:
The issue with missing python 3.8.1 on chocolatey repository isn't the issue, the actual problem is that version flag stopped working in choco commands, see issue: https://github.com/chocolatey/choco/issues/1985

Unless there is some other unstated reason about going back to 3.8.0 as minimum version, we should probably stay with 3.8.1, and possibly remove version flag until the issue gets resolved. Python3.9 isn't gonna get released any time soon, so we would just need to go back to this when the issue gets resolved.

Kowlin (approved):

PR #3327, opened 2020-01-10
[Warnings] Make [p]warnings usable on base of permissions

Dav-Git:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Makes [p]warnings work with permissions and introduces [p]mywarnings with the functionality described in #2900

### DIFF
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -370,14 +370,14 @@ class Warnings(commands.Cog):
                 await ctx.send_interactive(
                     pagify(msg, shorten_by=58), box_lang=_("Warnings for {user}").format(user=user)
                 )
-                
+
     @commands.command()
     @commands.guild_only()
     async def mywarnings(self, ctx: commands.Context):
         """List warnings for yourself."""
-        
+
         user = ctx.author
-        
+
         msg = ""
         member_settings = self.config.member(user)
         async with member_settings.warnings() as user_warnings:
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -345,6 +345,13 @@ class Warnings(commands.Cog):
     async def warnings(self, ctx: commands.Context, user: Union[discord.Member, int]):
         """List the warnings for the specified user."""
 
+        try:
+            userid: int = user.id
+        except AttributeError:
+            userid: int = user
+            user = ctx.guild.get_member(userid)
+            user = user or namedtuple("Member", "id guild")(userid, ctx.guild)
+
         msg = ""
         member_settings = self.config.member(user)
         async with member_settings.warnings() as user_warnings:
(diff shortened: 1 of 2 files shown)


Jackenmen:
Changelog contains wrong command names + I think changelogs should be separated into 2 files - 2900.enhance.1.rst and 2900.enhance.2.rst (though I heard today it might not be necessary so someone please confirm this)

Dav-Git:
Wasn't sure on 2 files. I thought 1 file works

redbot/cogs/warnings/warnings.py
```diff
@@ -342,36 +341,48 @@ def __init__(self, bot: Red):
...
-        appropriate permissions.
-        """
+        """List the warnings for the specified user."""
         if user is None:
-            user = ctx.author
-        else:
-            if not await is_admin_or_superior(self.bot, ctx.author):
-                return await ctx.send(
-                    warning(_("You are not allowed to check warnings for other users!"))
+            await ctx.send(_("Please provide a user you want to check the warnings for."))
```
mikeshardmind: This should either return here, or be removed entirely and remove the optional nature of the user argument (probably the latter, but either would be fine)

redbot/cogs/warnings/warnings.py
```diff
@@ -342,36 +341,48 @@ def __init__(self, bot: Red):
...
+
+        msg = ""
+        member_settings = self.config.member(user)
```
mikeshardmind, line 356: This doesn't properly account for the possibility a user id was passed without it being to a current user (and needs to)

mikeshardmind (changes requested):

redbot/cogs/warnings/warnings.py
Dav-Git: Would removing the "Optional" part of the argument fix this?

redbot/cogs/warnings/warnings.py
Dav-Git: Honestly I have no idea how to go about this.

redbot/cogs/warnings/warnings.py
jack1142: Yes, making the argument required and removing handling for `user is None` would fix this.

redbot/cogs/warnings/warnings.py
Jackenmen: The removed handling here was doing that before:
https://github.com/Cog-Creators/Red-DiscordBot/pull/3327/files#diff-bd66023271557007376afcf648f3b3ffL363-L368

redbot/cogs/warnings/warnings.py
Jackenmen: Yes, making the argument required and removing handling for `user is None` would fix this.
(adding again, cause I accidentally removed the comment)

Dav-Git:
I think that's what you had in mind?

redbot/cogs/warnings/warnings.py
```diff
@@ -342,36 +341,52 @@ def __init__(self, bot: Red):
...
+    async def mywarnings(self, ctx: commands.Context):
+        """List warnings for yourself."""
+
+        user = ctx.author
+
+        msg = ""
+        member_settings = self.config.member(user)
+        async with member_settings.warnings() as user_warnings:
+
             try:
```
mikeshardmind: Not exactly what I meant... 

The git diff here is doing some slight disfavors in discussing review as it's recognizing the code for `mywarnings` now matches the old `warnings` more closely.

The handling for if a user id not a user was passed isn't needed in `mywarnings` but is needed in `warnings` (prior to interacting with config)

mikeshardmind (approved):
Thanks!

PR #3328, opened 2020-01-10
[Audio] Hotfix an edge case where an attribute error can be raised

Drapersniper:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature


mikeshardmind:
For context on this wrt a hotfix changelog, what would a typical user see when this errored?

Drapersniper:
Honestly not quite sure ....I only reproduced while playing Spotify playlists and debug logs aren’t giving me any extra context just after some trial and error this one worked for me and stop reproducing

mikeshardmind:
I see how this would fix that, was just hoping to get something a little clearer for the changelog if possible.

mikeshardmind (dismissed):

Drapersniper:
Hmm turns out this can open in prod 👍

mikeshardmind (approved):
Prior info about this *only* being possible under user error circumstances was wrong.

This does mean this only handles one half of the issue though

PR #3333, opened 2020-01-11
[Modlog API] Add resolution for people inpacted by bad casetypes

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

The worst case right now is just some log spam, but people who migrated from mongo don't have a great resolution for this without it.

The command is hidden, informed about in logging, and does not add a new top-level command to prevent polluting the top level command namespace.

### DIFF
--- a/redbot/cogs/modlog/modlog.py
+++ b/redbot/cogs/modlog/modlog.py
@@ -29,7 +29,8 @@ class ModLog(commands.Cog):
     @checks.is_owner()
     @modlogset.command(hidden=True, name="fixcasetypes")
     async def reapply_audittype_migration(self, ctx: commands.Context):
-        """Command to fix misbehaving casetypes"""
+        """Command to fix misbehaving casetypes."""
+        await ctx.send(_("Fixing up the casetypes, this may take a moment."))
         await modlog.handle_auditype_key()
         await ctx.tick()
 
--- a/redbot/cogs/modlog/modlog.py
+++ b/redbot/cogs/modlog/modlog.py
@@ -26,6 +26,13 @@ class ModLog(commands.Cog):
         """Manage modlog settings."""
         pass
 
+    @checks.is_owner()
+    @modlogset.command(hidden=True, name="fixcasetypes")
+    async def reapply_audittype_migration(self, ctx: commands.Context):
+        """Command to fix misbehaving casetypes"""
+        await modlog.handle_auditype_key()
+        await ctx.tick()
+
     @modlogset.command()
     @commands.guild_only()
     async def modlog(self, ctx: commands.Context, channel: discord.TextChannel = None):
--- a/redbot/cogs/modlog/modlog.py
+++ b/redbot/cogs/modlog/modlog.py
@@ -30,7 +30,6 @@ class ModLog(commands.Cog):
     @modlogset.command(hidden=True, name="fixcasetypes")
     async def reapply_audittype_migration(self, ctx: commands.Context):
         """Command to fix misbehaving casetypes."""
-        await ctx.send(_("Fixing up the casetypes, this may take a moment."))
         await modlog.handle_auditype_key()
         await ctx.tick()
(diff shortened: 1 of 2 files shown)


redbot/cogs/modlog/modlog.py
```diff
@@ -26,6 +26,13 @@ def __init__(self, bot: Red):
         """Manage modlog settings."""
         pass
 
+    @checks.is_owner()
+    @modlogset.command(hidden=True, name="fixcasetypes")
+    async def reapply_audittype_migration(self, ctx: commands.Context):
+        """Command to fix misbehaving casetypes"""
+        await modlog.handle_auditype_key()
```
Drapersniper, line 33: ctx.typing() may be a nice addition here, as this could take a while depending on number of modlogs (FN as an example)

redbot/cogs/modlog/modlog.py
```diff
@@ -26,6 +26,13 @@ def __init__(self, bot: Red):
         """Manage modlog settings."""
         pass
 
+    @checks.is_owner()
+    @modlogset.command(hidden=True, name="fixcasetypes")
+    async def reapply_audittype_migration(self, ctx: commands.Context):
+        """Command to fix misbehaving casetypes"""
```
Drapersniper: > """Command to fix misbehaving casetypes"""

Should be `"""Command to fix misbehaving casetypes."""`

redbot/core/modlog.py
```diff
@@ -508,7 +511,11 @@ def __init__(
         self.case_str = case_str
         self.guild = guild
         if kwargs:
-            log.warning("Got unexpected keys in case %s", ",".join(kwargs.keys()))
+            log.warning(
+                "Fix this using the hidden command: `modlogset fixcasetypes`:  "
```
Drapersniper: We should add the prefix here.
`{prefix}modlogset fixcasetypes` or since we dont have the context ... `[p]modlogset fixcasetypes`, so it's more obvious this command is to be run in Discord and not command line.

Drapersniper (dismissed):
Other than these 3 nitpicks everything is looking good.

redbot/core/modlog.py
mikeshardmind: Not sure I agree there, the logging calls it a command.

Jackenmen: Cli command is a command too, there's nothing indicating it's a bot command.

redbot/cogs/modlog/modlog.py
mikeshardmind: Ironically, the servers which likely would want this the most can't benefit from it as discord disables typing indicators on some larger servers.

redbot/core/modlog.py
Drapersniper: Always assume the user don't know where their head is located :smirk:

redbot/core/modlog.py
```diff
@@ -508,7 +511,11 @@ def __init__(
         self.case_str = case_str
         self.guild = guild
         if kwargs:
-            log.warning("Got unexpected keys in case %s", ",".join(kwargs.keys()))
+            log.warning(
+                "Fix this using the hidden command: `modlogset fixcasetypes` in Discord: "
```
Jackenmen, line 517: This command only fixes the problem if unexpected key is `audit_type`, right?

mikeshardmind: Right now, yes. But that is the only unexpected key possible to exist there currently. I left this generic so that it doesn't need changing in the future.

Jackenmen: I think it would actually be better to change it to just say it on `audit_type`, because this command won't solve the issue about some other unexpected key unless it actually gets updated appropriately and when such change happens, this error can easily be updated for this too.

Jackenmen (approved):
LGTM

PR #3335, opened 2020-01-11
Start making use of typehints for devs

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

A lot of the bot's code is actually covered on type hints. It's not entirely enough for mypy, but lets start having these be seen by type analysis for people using that.

### DIFF
--- a/changelog.d/3335.enhance.rst
+++ b/changelog.d/3335.enhance.rst
@@ -1,0 +1,1 @@
+make typehints accessible to cog developers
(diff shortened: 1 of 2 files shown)


Jackenmen (approved):
// Finally 😄
Might be good to add a comment in `py.typed` file why it's there (i.e. `# Marker file for PEP 516`), it wouldn't hurt and makes it more obvious why we have some random empty file inside package.
But it's just a preference so this PR is fine.

PR #3337, opened 2020-01-11
[3.2.3][Audio] Correct an unsupported LoadType

Drapersniper:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

`V2_COMPACT` was added to the Cache... lets replace any occourance of it with "V2_COMPAT", over time this will get updated (Since future writes to cache will update it to have "V2_COMPAT" automatically.


aikaterna:
Make sure to leave a comment about your change for ytsearch and scsearch query prefix behavior as well because that technically is an issue that is not reported, yet you fix it and don't mention it in this issue, and then this is good to go.

mikeshardmind (approved):

PR #3063, opened 2019-10-16
More help improvements

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

### Description of the changes

 - ``[botname]`` will be dynamically substituted in help
 - help now tries to be smart about newlines in docstrings
 - All user facing help text is i18n safe

resolves #2981

PR #3339, opened 2020-01-11
[Help] formatting additions

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This one was discussed back with #3063 but was delayed.
This includes most of those changes, barring the whitespace ones, which were the hold up.

### DIFF
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -62,7 +62,7 @@ class CogCommandMixin:
         """
         This formats the help string based on values in context
 
-        The steps are (roughly) the following:
+        The steps are (currently, roughly) the following:
 
             - get the localized help
             - substitute ``[p]`` with ``ctx.clean_prefix``
@@ -70,8 +70,8 @@ class CogCommandMixin:
 
         More steps may be added at a later time.
 
-        Cog creators May override this in their own command classes
-        As long as the method signature stays the same
+        Cog creators may override this in their own command classes
+        as long as the method signature stays the same.
 
         Parameters
         ----------
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -4,6 +4,7 @@ This module contains extended classes and functions which are intended to
 replace those from the `discord.ext.commands` module.
 """
 import inspect
+import re
 import weakref
 from typing import Awaitable, Callable, Dict, List, Optional, Tuple, Union, TYPE_CHECKING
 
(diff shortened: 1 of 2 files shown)


Jackenmen:
Would it be bad to have this as async function instead? I.e. for accessing config, unless you would just want cog creators to cache it which probably isn't a bad idea either.

Jackenmen (changes requested):
A very cool addition, saw you're adding i18n here, and noticed 2 places with missing `T_()`, other than that and what I asked in previous comment, looks good.

https://github.com/mikeshardmind/Red-DiscordBot/blob/help-additions/redbot/core/commands/help.py#L554

https://github.com/mikeshardmind/Red-DiscordBot/blob/help-additions/redbot/core/commands/help.py#L564

I'm probably gonna use such Command subclass once this feature goes in:
```py
class MultilineDescription(commands.Command):
    def format_help_for_context(self, ctx: commands.Context) -> str:
        command_help = super().format_help_for_context(ctx)
        pos = command_help.find("\n\n")
        command_help = command_help[:pos].replace("\n", " ") + command_help[pos:]
        return command_help
```

redbot/core/commands/help.py
```diff
@@ -540,25 +548,24 @@ def width_maker(cmds):
...
         if fuzzy_commands:
-            ret = await format_fuzzy_results(ctx, fuzzy_commands, embed=use_embeds)
-            if use_embeds:
-                ret.set_author(name=f"{ctx.me.display_name} Help Menu", icon_url=ctx.me.avatar_url)
-                tagline = (await ctx.bot._config.help.tagline()) or self.get_default_tagline(ctx)
-                ret.set_footer(text=tagline)
-                await ctx.send(embed=ret)
-            else:
-                await ctx.send(ret)
+            msg = await format_fuzzy_results(ctx, fuzzy_commands, embed=use_embeds)
```
Jackenmen: If `use_embeds=True`, `format_fuzzy_results` will return `discord.Embed` so this doesn't work properly with L560 where you create a new embed using `msg` var as embed description.

Jackenmen (changes requested):

redbot/core/commands/help.py
mikeshardmind: Yep.... Will handle consolidating more of the logic another time...

redbot/core/commands/commands.py
```diff
@@ -57,6 +58,49 @@ def __init__(self, *args, **kwargs):
...
+        The steps are (roughly) the following:
+
+            - get the localized help
+            - substitute ``[p]`` with ``ctx.clean_prefix``
+            - substitute ``[botname]`` with ``ctx.me.display_name``
+
+        More steps may be added at a later time.
+
+        Cog creators May override this in their own command classes
+        As long as the method signature stays the same
```
Jackenmen: Letter casing is a little weird here.

redbot/core/commands/commands.py
```diff
@@ -57,6 +58,49 @@ def __init__(self, *args, **kwargs):
             checks=getattr(decorated, "__requires_checks__", []),
         )
 
+    def format_help_for_context(self, ctx: "Context") -> str:
```
Jackenmen, line 61: Preference thing, but we could start using `from __future__ import annotations` for this.

Jackenmen, line 61: It might also be good to mention here that this method can (is intended to) be subclassed.

Jackenmen (changes requested):

redbot/core/commands/commands.py
mikeshardmind: It's mentioned in the docstring already.

mikeshardmind: I've been adding it to files I touch where it doesn't require re-evaluating existing forward references for validity. Planned, but this one can wait.

Jackenmen (approved):
LGTM, thank you for this one, this will allow cog creators to do some great things.

PR #3340, opened 2020-01-11
Add a command to set the bot description

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

Does what it says on the tin.

Follow up from the license info command which was added as a compromise with users on which parts of the bot could be customized while still retaining proper license details.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -76,6 +76,7 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):  # pylint: d
             help__verify_checks=True,
             help__verify_exists=False,
             help__tagline="",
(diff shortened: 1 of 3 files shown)


Jackenmen:
Using too long (over 256 chars long) text here will make help command fail.
While this may need to also be improved in `HelpFormatter`'s handling, it would be good to error out in this command as well so that user immediately knows it's too long when they try to set it.

mikeshardmind:
The length of this is already handled for cogs and commands, it wasn't handled here previously as this wasn't settable before. Restricting it by command is enough in my opinion.

Jackenmen (approved):

PR #3351, opened 2020-01-12
[3.2.3][Docs]Only support venv and virtualenv users

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature


docs/venv_guide.rst
```diff
@@ -7,16 +7,11 @@ Creating a virtual environment is really easy and usually prevents many common i
 problems. Firstly, simply choose how you'd like to create your virtual environment:
 
 * :ref:`using-venv` (quick and easy, involves two commands)
-* :ref:`using-pyenv-virtualenv` (recommended if you installed Python with pyenv)
-
-**Why Should I Use a Virtual Environment?**
-
-90% of the installation and setup issues raised in our support channels are resolved when the user
-creates a virtual environment.
+* :ref:`using-pyenv-virtualenv` (required if you installed Python with pyenv)
```
Jackenmen: It's not required to use `pyenv virtualenv` when you install Python with `pyenv`.

docs/install_linux_mac.rst
```diff
@@ -242,29 +242,19 @@ Installing Red
...
-        python3.8 -m pip install --user -U setuptools wheel
-        python3.8 -m pip install --user -U Red-DiscordBot
-
 To install without additional config backend support:
 
 .. code-block:: none
 
-    python3.8 -m pip install -U setuptools wheel
-    python3.8 -m pip install -U Red-DiscordBot
+    python -m pip install -U setuptools wheel
```
Jackenmen: Can we start updating `pip` here too? If we require venv, we can safely do so and it avoids issues with outdated pip which can be seen in #support.
(remember to update it in all 4 places)

mikeshardmind (approved):

PR #3358, opened 2020-01-13
Make Red shutdown when resetting token

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature


mikeshardmind (approved):

mikeshardmind (changes requested):
This should exit 1 if they don't change the token in this exception. Type hinting this as `NoReturn` and the last line as a `sys.exit(1)` (but not within a finally block!!) should be fine for this, the bot will itself will always shutdown with a system exit or keyboard interrupt

mikeshardmind (approved):
thanks!

PR #3360, opened 2020-01-13
Exit code handling

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This one's my fault, I made a misleading comment when not fully thinking about something yesterday.

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -288,7 +288,16 @@ def handle_edit(cli_flags: Namespace):
         sys.exit(0)
 
 
-async def run_bot(red: Red, cli_flags: Namespace) -> NoReturn:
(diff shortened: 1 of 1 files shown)


Jackenmen (dismissed):
The change itself works, but I still feel like this comment might be unclear.
It's not immediately obvious what is "abnormal shutdown case" on which we need to raise SystemExit manually (from the looks of it, it would be anything that didn't allow us to even make a successful connection to discord?).
And "SystemExit will be triggered as well, though it won't propagate to this function" doesn't say where SystemExit will propagate to (from testing it looks like it propagates to `loop.run_forever()`, this is blowing my mind a little).

Jackenmen (approved):
Awesome, now it's a lot clearer, LGTM 👍

PR #3362, opened 2020-01-14
Some old work and some new

DiscordLiz:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

After talking with @mikeshardmind about it, this is the portion of the prior work I started which is still useful for the way this way the cog is going to be made.

I don't think I can commit the time to finish this on my own, but I'm still willing to offer what I have as a starting point.


redbot/core/utils/discord_helpers.py
```diff
@@ -0,0 +1,154 @@
...
+    This object is considered False for boolean logic when representing no change.
+
+    Attributes
+    ----------
+    allows_added : int
+    allows_removed : int
+    denies_added : int
+    denies_removed : int
+
+    .. versionadded: 3.2.0
```
mikeshardmind: This line should probably go

mikeshardmind (commented):
I'm not sure *exactly* how much of this will end up getting used, but thank you for repackaging what was done and still compatible with the future goals.

mikeshardmind (approved):
Approving this as the starting point to be used for this as even though there's still a lot left to do, it provides a good place for the work to continue from.

PR #3364, opened 2020-01-14
[Downloader] Stop messing with distutils's internals just to copy directory

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Readdressing #2685 (and #2690 that "fixed" it) in a way that doesn't involve messing with distutils's internals.
We require Python 3.8 now, so we can finally just use `shutil.copytree` to copy a directory...
Haven't tested this, basing this solely on discussion from #2685/#2690

### DIFF
--- a/redbot/cogs/downloader/installable.py
+++ b/redbot/cogs/downloader/installable.py
@@ -1,7 +1,7 @@
 from __future__ import annotations
 
 import json
-import distutils.dir_util
+import functools
 import shutil
 from enum import IntEnum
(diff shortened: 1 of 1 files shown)


mikeshardmind (approved):

mikeshardmind:
We probably have a few more places where we can get rid of older behaviors for cleaner options in 3.8, there's at least one similar section to this in `redbot.__main__`

PR #3371, opened 2020-01-15
[Docs] Update docs for new minimum required version of `git`

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

I can reassure that Git 2.11+ is really the minimum version required by Downloader later but after this PR docs should **at least** be fine with git version for all OSes except Debian Stretch and possibly OpenSUSE (waiting for verification on that)

### DIFF
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -19,7 +19,7 @@ Please install the pre-requirements using the commands listed for your operating
 The pre-requirements are:
  - Python 3.8.1 or greater
  - Pip 18.1 or greater
- - Git
+ - Git 2.11+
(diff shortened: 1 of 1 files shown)


mikeshardmind (approved):

Jackenmen:
Since this got merged, I verified all git commands we use and Git 2.11+ is a correct minimum version of `git` that is needed by us. I also verified OpenSUSE's package repository and it looks like Tumbleweed has Git 2.24 and Leap 15.1 has Git 2.16 and even an older Leap 42.1 has Git 2.13, so we're fine in OpenSUSE world too.

PR #3374, opened 2020-01-16
[Audio] Add backticks to commands in docstrings, fix GH-3140

### LINKED ISSUE #3140: [Audio] Help of `[p]audioset emptydisconnect` should mention precedence of `[p]audioset dc`
Help of `[p]audioset dc` already does that, I think it might be good to also say that in help of `[p]audioset emptydisconnect`

Drapersniper:

Closes #3140

### DIFF
--- a/changelog.d/audio/3140.enchance.1.rst
+++ b/changelog.d/audio/3140.enchance.1.rst
@@ -1,0 +1,1 @@
+Update the help strings for ``[p]audioset emptydisconnect``.
--- a/changelog.d/audio/3140.enchance.1.rst
+++ b/changelog.d/audio/3140.enchance.1.rst
@@ -1,1 +1,0 @@
-Update the help strings for ``[p]audioset emptydisconnect``.
--- a/redbot/cogs/audio/audio.py
+++ b/redbot/cogs/audio/audio.py
@@ -1259,7 +1259,10 @@ class Audio(commands.Cog):
     @audioset.command()
     @checks.mod_or_permissions(administrator=True)
     async def emptydisconnect(self, ctx: commands.Context, seconds: int):
-        """Auto-disconnect from channel when bot is alone in it for x seconds, 0 to disable."""
+        """Auto-disconnect from channel when bot is alone in it for x seconds, 0 to disable.
+
+        [p]audioset dc takes precedence over this setting.
+        """
         if seconds < 0:
             return await self._embed_msg(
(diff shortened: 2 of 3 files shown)


redbot/cogs/audio/audio.py
```diff
@@ -1259,7 +1259,10 @@ def _restart_connect(self):
     @audioset.command()
     @checks.mod_or_permissions(administrator=True)
     async def emptydisconnect(self, ctx: commands.Context, seconds: int):
-        """Auto-disconnect from channel when bot is alone in it for x seconds, 0 to disable."""
+        """Auto-disconnect from channel when bot is alone in it for x seconds, 0 to disable.
+
+        [p]audioset dc takes precedence over this setting.
```
Jackenmen: I would put this in backticks.
```suggestion
        `[p]audioset dc` takes precedence over this setting.
```

Drapersniper: This goes agaisn't every use of `[p]command` within the Audio UX. 

I see the benefit of it however since doing this would also have to be applied to all other uses of `[p]command`(for consistency) @aikaterna please confirm if this is a UX behavioural change you are okay with.

(Using backticks for every usage of `[p]commands` within all of audio docstrings)

Yes I know there's 3 occourances in audio where backtics are used but there's a lot more where they aren't used within the docstrings

aikaterna:
I was originally of the mindset that we should add backtick formatting to the rest of the audio commands where applicable, for command examples. However, I took a look at Audio's helpstrings with a bot that was set to codeblocks and not embeds, and the formatting on various commands looks out of place in a codeblock. Slime should be opening an issue regarding stripping formatting from command help for codeblocks and linking back to this PR. If Slime's suggestion is not implemented I lean towards favoring adding formatting towards all audio commands as I would expect most users would be using the embed setting with the bot help.

mikeshardmind (dismissed):

changelog.d/audio/3140.enchance.1.rst
```diff
@@ -0,0 +1 @@
+Update the help strings for ``[p]audioset emptydisconnect``.
```
Jackenmen: This file is called incorrectly:
```
3140.enchance.1.rst
       ^     ^^
```

Drapersniper: The number is fine, but yeah my bad

Jackenmen (approved):
LGTM

PR #3376, opened 2020-01-16
[Audio] Show symbolic link folders

### LINKED ISSUE #3332: Localtracks will not show symlinked folders in Red 3.2.2 on Ubuntu 18.04
# Command bugs

#### Command name

local play

Audio

Listing all the folders and tracks that are in localtracks.

#### What actually happened?

Symlinked folders were not showing up in the bot reply. Only local folders and files and symlinks of files show up.

1. Create a symlink to a folder containing music in localtracks:
`ln -s path/to/local/music/folder path/to/reds/localtracks/`
2. Initiate playing of local music in redbot:
`[p]local play`
3. Red will list folders and tracks for you to choose from. I observed that any local folders and files show up as well as symlinked files but the symlinked folder will not show up in the selection.

#### Additional context
This happens with Red 3.2.2, on Ubuntu 18.04 64 Bit. It has been working as expected prior to updating, which was Red 3.1.8.

Drapersniper:
# Description of the changes

Fixes #3332

### DIFF
--- a/.github/workflows/auto_labeler.yml
+++ b/.github/workflows/auto_labeler.yml
@@ -13,9 +13,14 @@ jobs:
         with:
           github-token: ${{secrets.GITHUB_TOKEN}}
           script: |
+            const is_status_label = (label) => label.name.startsWith('Status: ');
+            if (context.payload.issue.labels.some(is_status_label)) {
+                console.log('Issue already has Status label, skipping...');
+                return;
+            }
             github.issues.addLabels({
               issue_number: context.issue.number,
               owner: context.repo.owner,
               repo: context.repo.repo,
               labels: ['Status: Needs Triage']
-            })
+            });
--- a/changelog.d/3148.misc.1.rst
+++ b/changelog.d/3148.misc.1.rst
@@ -1,0 +1,1 @@
+Cache prefixes rather than lookup from config each time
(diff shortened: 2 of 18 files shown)


Drapersniper:
Okay so @mikeshardmind  Before this is merged we need to disguss the fact that `glob.glob` is a intensive and action and we should look into wrapping it in an async function, as it is it works and on smaller Trees this is not an issue whatsoever but i was messing in a fairly large tree and it proved fustrating as it could take ~5 seconds to process on a tree with a several thousand files/sub directories.

The link i send on #AC should help mitigate this and seems to be a pretty stable wrapper as its used by Django.

Before merging this one and removing the "In progress" label we should decide if we are okay with this and if not what is the best way to work around it.

Drapersniper (commented):
for track in self.multirglob(*[f"*{ext}" for ext in self._all_music_ext]):

Change this to f"{ext}" you complete idiot

Drapersniper:
As discussed with @mikeshardmind  changed to use `glob.iglob` and in turn used an async iterator where we iterate through the result for `iglob`

mikeshardmind (dismissed):
We can probably do some things to improve this in the future... Good enough for now.

Drapersniper:
There no more conflicts

mikeshardmind (approved):

PR #3381, opened 2020-01-16
[Core] Removed redundant event loop check

Redjumpman:
Removed redundant check.

### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Removed a check that forced windows to use ProactorEventLoop. As of Python 3.8, it now uses ProactorEventLoop by default.

### DIFF
--- a/redbot/__init__.py
+++ b/redbot/__init__.py
@@ -181,9 +181,7 @@ class VersionInfo:
 
 
 def _update_event_loop_policy():
-    if _sys.platform == "win32":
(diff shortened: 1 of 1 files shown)


mikeshardmind (approved):
+1, thanks for spotting one of those relics.

Brief note: This should be fine despite the apparent change in behavior, `uvloop` isn't installable on windows, but if it ever were (unlikely as that is), we wouldn't want to exclude windows users from the benefits

PR #3383, opened 2020-01-16
Preemptive patch for discord.py changes

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Patch here is to make it safer to use\* the development version of discord.py without dropping support for 1.2.5 (latest stable) Some of this should be and is marked for further change at 1.3

\* As the changes for 1.3 are not yet finalized, this only handles the known issues in a way which won't break 1.2.5 users

I repeat: This is not a "go ahead and use the dev version of discord.py", this merely allows us to start testing 1.3 internally.

### DIFF
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -323,14 +323,25 @@ class Command(CogCommandMixin, commands.Command):
             if not change_permission_state:
                 ctx.permission_state = original_state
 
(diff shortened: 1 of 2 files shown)


mikeshardmind:
since we're doing a minor update (red 3.3.0) for discord.py  1.3.0, this PR should be updated to fully account for that before merge.

PR #3388, opened 2020-01-17
[Downloader] Pagify any output that might be too long

### LINKED ISSUE #3385: [Downloader] cog update can error if the output of the command is greater than 2000 characters
# Command bugs

Cog update can error when the output is greater than 2000 characters.

#### Command name

`[p]cog update`

Downloader

I was expecting the output of cog update to list the updated cogs.

#### What actually happened?

The command errored stating the request was larger than 2000 characters.

Have a LOT of cogs installed and run `[p]cog update` after migrating from 3.1.9 to 3.2.2.

Jackenmen:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

Fixes #3385, prevents similar issues from happening in whole Downloader cog.

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -418,6 +418,11 @@ class Downloader(commands.Cog):
         elif target.is_file():
             os.remove(str(target))
 
+    @staticmethod
+    async def send_pagified(target: discord.abc.Messageable, content: str) -> None:
+        for page in pagify(content):
+            await target.send(page)
+
     @commands.command()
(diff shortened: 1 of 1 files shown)


mikeshardmind:
While I'm fine with this approach, is it time to reconsider a ctx.send_pagified?

mikeshardmind:
Discord: no ctx.send_pagified

Reasons are mostly the same as before:

- 2 lines + an import
- Promotes thoughtful use of pagination

mikeshardmind (approved):

PR #3363, opened 2020-01-14
Audio Refactor: Part 1  - APIs

Drapersniper:
As the title says this one handles the internal APIs used by Audio ... still WIP.

PR #3398, opened 2020-01-17
Audio Refactor: Part 2 - Cog Class

Drapersniper:
Break up the Cog class into smaller simplied composite classes.


Drapersniper:
*sigh* honestly since everything is so intertwined Imm close this for now and work on it when #3363 is Merged

PR #3410, opened 2020-01-18
[Commands Module] Improve usability of type hints

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This does a few things, but this should be safe enough as is.

1. We don't re-export discord.py modules
2. We did keep the re-export of any names we haven't replaced.
3. Added a few Type Aliases
  - DMContext
  - GuildContext
4. Added a few Converters which look like, and play nice with other type checking
  - UserInputOptional
  - RedOptional
  - Literal
5. Added a setting to ignore discord.py's special casing of typing.Optional
  - While this is usable for many simple cases, you may want to look at added converter types as alternatives to this setting for more complex argument handling.


mikeshardmind:
Lying to mypy while shadowing the commands module like this is proving to be *problematic*

We get a lot of benefits from type analysis if we can get this right, but we may need to not shadow discord.py's commands module anymore.

mikeshardmind:
~~This should no longer be failing, and also not break anyone, but long term we need some more robust handling of this.~~

And now we're looking at a breaking version anyhow...

mikeshardmind:
So, the strain on cog creators for this is something we can hopefully avoid. I'm putting more effort into doing this in a way which may add marginally to our dep update process, but should be otherwise fine and gain in usability and maintainability.

Jackenmen:
I looked to see if there are any names that have disappeared after applying this changes and those would be the module names from discord.ext.commands: `bot`, `cog`, `cooldowns`, `core`, `view`
While I think it's not that good to keep those here forever, it would technically be a breaking change, so we might want to keep them here for now and remove later when doing Red 3.3.

mikeshardmind:
*sigh* okay will add those in later this evening.

mikeshardmind:
Technically breaking again.

Scope of breakage is limited to people importing discord.py submodules (not classes/funcs/etc) from redbot.core and people directly using `PermissionState.TRANSITIONS` and `PermissionState.ALLOWED_STATES`, and people declaring coroutines using the asyncio.coroutine decorator (we're python 3.8+, please don't be doing this) for use as command objects

redbot/core/commands/dpy_reimplements.py
```diff
@@ -0,0 +1,237 @@
...
+
+    if inspect.iscoroutinefunction(predicate):
+        setattr(decorator, "predicate", predicate)
+    else:
+
+        @functools.wraps(predicate)
+        async def wrapper(ctx):
+            return predicate(ctx)
+
+        setattr(decorator, "predicate", wrapper)
```
Jackenmen: Why is this not using dot notation?

redbot/core/commands/dpy_reimplements.py
```diff
@@ -0,0 +1,237 @@
...
+        else:
+            if not hasattr(func, "__commands_checks__"):
+                func.__commands_checks__ = []
+
+            func.__commands_checks__.append(predicate)
+
+        return func
+
+    if inspect.iscoroutinefunction(predicate):
+        setattr(decorator, "predicate", predicate)
```
Jackenmen: Why is this not using dot notation?

redbot/core/commands/dpy_reimplements.py
```diff
@@ -0,0 +1,237 @@
...
+
+        assert isinstance(ctx.me, discord.Member)
+
+        if isinstance(item, int):
+            role = discord.utils.get(ctx.me.roles, id=item)
+        else:
+            role = discord.utils.get(ctx.me.roles, name=item)
+
+        if not role:
+            raise dpy_commands.MissingRole(item)
```
Jackenmen: Implementation in d.py seems to have this instead:
```suggestion
            raise dpy_commands.BotMissingRole(item)
```

redbot/core/commands/dpy_reimplements.py
```diff
@@ -0,0 +1,237 @@
...
+
+        for item in items:
+            if isinstance(item, int):
+                if getter(id=int):
+                    break
+            else:
+                if getter(name=item):
+                    break
+        else:
+            raise dpy_commands.MissingAnyRole(items)
```
Jackenmen: Implementation in d.py seems to have this instead:
```suggestion
            raise dpy_commands.BotMissingAnyRole(items)
```

redbot/core/commands/dpy_reimplements.py
```diff
@@ -0,0 +1,237 @@
...
+            raise dpy_commands.NoPrivateMessage(
+                _("This command cannot be used in private messages.")
+            )
+
+        assert isinstance(ctx.me, discord.Member)
+        getter = functools.partial(discord.utils.get, ctx.me.roles)
+
+        for item in items:
+            if isinstance(item, int):
+                if getter(id=int):
```
Jackenmen:
```suggestion
                if getter(id=item):
```

redbot/core/commands/dpy_reimplements.py
```diff
@@ -0,0 +1,237 @@
...
+        return True
+
+    return check(predicate)
+
+
+PrefixCallable = Callable[[dpy_commands.bot.BotBase, discord.Message], List[str]]
+
+
+def when_mentioned(bot: dpy_commands.bot.BotBase, msg: discord.Message) -> List[str]:
+    return [f"<@{bot.user.id}", f"<@!{bot.user.id}> "]
```
Jackenmen:
```suggestion
    return [f"<@{bot.user.id}> ", f"<@!{bot.user.id}> "]
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +162,175 @@ class APIToken(discord.ext.commands.Converter):
...
+            return ret
 
-    class PartialMeta(type(DictConverter)):
-        __call__ = functools.partialmethod(
-            type(DictConverter).__call__, *expected_keys, delims=delims
-        )
 
-    class ValidatedConverter(DictConverter, metaclass=PartialMeta):
-        pass
+if TYPE_CHECKING:
```
Jackenmen, line 225: Why is this branch needed if type hints are exactly the same?

redbot/core/commands/converter.py
```diff
@@ -162,140 +162,175 @@ class APIToken(discord.ext.commands.Converter):
...
-        If provided, it will additionally try to match integer-only input into
-        a timedelta, using the unit specified. Same units as in `allowed_units`
-        apply.
 
-    Returns
-    -------
-    type
-        The converter class, which will be a subclass of `TimedeltaConverter`
-    """
+if TYPE_CHECKING:
```
Jackenmen, line 298: Why is this branch needed if type hints are exactly the same?

redbot/core/commands/requires.py
```diff
@@ -326,13 +331,13 @@ def __init__(
...
     def get_decorator(
-        privilege_level: Optional[PrivilegeLevel], user_perms: Dict[str, bool]
+        privilege_level: Optional[PrivilegeLevel], user_perms: Optional[Dict[str, bool]]
     ) -> Callable[["_CommandOrCoro"], "_CommandOrCoro"]:
         if not user_perms:
             user_perms = None
 
         def decorator(func: "_CommandOrCoro") -> "_CommandOrCoro":
-            if asyncio.iscoroutinefunction(func):
+            if inspect.iscoroutinefunction(func):
```
Jackenmen, line 344: This will make decorators stop working with generator-based coros (`@asyncio.coroutine`).
I'm completely fine with this, especially that they're deprecated since Python 3.8 but I think this is technically also a breaking change which should be mentioned as part of the warning just in case.

Jackenmen (changes requested):
Found few small problems with d.py reimplements and have few small questions.

Also not included in review comments, but do we want to privatize some parts of `redbot.core.commands.requires` to be able to avoid minor bump when the only changes are in internal stuff?

redbot/core/commands/dpy_reimplements.py
mikeshardmind: Intentional. Mypy doesn't appreciate dynamic access to attributes which aren't declared

redbot/core/commands/converter.py
mikeshardmind: type hints are the same, but there's a dynamic class declaration inside which mypy doesn't allow. For the sake of people using this, making it not appear dynamic since external to the function, the typing is fine.

Jackenmen (dismissed):
Should be ready to go now.

mikeshardmind:
K, this is on hold till the discord.py version bump then, and we'll do a wrap up with anything we have ready around then.

mikeshardmind:
rebased over other newer changes.

Jackenmen:
I didn't look much at `callback` definition, but this breaks:
```py
from typing import Optional

@commands.is_owner()
@commands.command(ignore_optional_for_conversion=True)
async def somecom(ctx: commands.Context, number: Optional[int]) -> None:
    await ctx.send(f"hello {number}")

bot.add_command(somecom)
```

With traceback:
```py
  File "C:\Users\Jakub\OneDrive\Dokumenty\Red-DiscordBot\redbot\core\dev_commands.py", line 171, in _eval
    result = await func()
  File "<string>", line 9, in func
  File "C:\Users\Jakub\OneDrive\Dokumenty\Red-DiscordBot\.venv\lib\site-packages\discord\ext\commands\core.py", line 1241, in decorator
    return cls(func, name=name, **attrs)
  File "C:\Users\Jakub\OneDrive\Dokumenty\Red-DiscordBot\redbot\core\commands\commands.py", line 248, in __init__
  File "C:\Users\Jakub\OneDrive\Dokumenty\Red-DiscordBot\redbot\core\commands\commands.py", line 82, in __init__
  File "C:\Users\Jakub\OneDrive\Dokumenty\Red-DiscordBot\.venv\lib\site-packages\discord\ext\commands\core.py", line 207, in __init__
    self.callback = func
  File "C:\Users\Jakub\OneDrive\Dokumenty\Red-DiscordBot\redbot\core\commands\commands.py", line 314, in callback
    args = vtype.__args__
AttributeError: '_SpecialForm' object has no attribute '__args__'
```

Jackenmen:
I can't get `commands.Literal` to work, example code that doesn't work for me:
```py
from redbot.core.commands import Literal

@commands.is_owner()
@commands.command()
async def somecom(ctx: commands.Context, test: Literal["foo", "bar"]) -> None:
    await ctx.send(test)

bot.add_command(somecom)
```
All I get when using `[p]somecom foo` is `Converting to "Literal" failed for parameter "test".`

Edit: I was able to get this to work after subclassing `Literal` class from `commands.Converter`.
Also, this will have an unexpected behavior when only one argument is passed to `Literal` - you get a string not tuple passed to `__class_getitem__` then which means you're checking if passed arg is a substring.

redbot/cogs/mod/kickban.py
```diff
@@ -7,6 +7,7 @@
 
 import discord
 from redbot.core import commands, i18n, checks, modlog
+from redbot.core.commands import UserInputOptional
```
Jackenmen: You only import the converter but didn't use it.

redbot/core/commands/commands.py
```diff
@@ -209,6 +244,7 @@ def __call__(self, *args, **kwargs):
             return self.callback(*args, **kwargs)
 
     def __init__(self, *args, **kwargs):
+        self.ignore_optional_for_conversion = kwargs.pop("ignore_optional_for_conversion", False)
```
Jackenmen, line 250: This hasn't been documented.

redbot/core/commands/context.py
```diff
@@ -263,3 +265,53 @@ def me(self) -> discord.abc.User:
...
+        def author(self) -> discord.Member:
+            ...
+
+        @property
+        def channel(self) -> discord.TextChannel:
+            ...
+
+        @property
+        def guild(self) -> discord.Guild:
+            ...
```
Jackenmen: These 2 classes aren't documented (not specified in `framework_commands.rst` file and placed under `if TYPE_CHECKING`).

redbot/core/commands/converter.py
```diff
@@ -138,17 +149,17 @@ class APIToken(discord.ext.commands.Converter):
...
-    This will split the argument by either `;` ` `, or `,` and return a dict
+    This will split the argument by a space, colon, or semicolon and return a dict
     to be stored. Since all API's are different and have different naming convention,
     this leaves the onus on the cog creator to clearly define how to setup the correct
     credential names for their cogs.
 
     Note: Core usage of this has been replaced with DictConverter use instead.
 
-    This may be removed at a later date (with warning)
+    This will be removed in 3.4
```
Jackenmen:
```suggestion
    .. warning::
        This will be removed in version 3.4.
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
+if TYPE_CHECKING:
+    RedOptional = Optional
+else:
+
+    class RedOptional:
+        """
+        This can be used instead of typing.Optional
+        To avoid discord.py special casing the conversion behavior
+
+        (see also: the ``ignore_optional_for_conversion`` option of commands)
```
Jackenmen: This suggestion and the one below can only be applied together as I changed the order.
```suggestion
        .. warning::
            This converter class is still provisional.
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
 
-    class PartialMeta(type(TimedeltaConverter)):
-        __call__ = functools.partialmethod(
-            type(DictConverter).__call__,
-            allowed_units=allowed_units,
-            default_unit=default_unit,
-            minimum=minimum,
-            maximum=maximum,
-        )
+        This converter class is still provisional
```
Jackenmen: This suggestion and the one above can only be applied together as I changed the order.
Also, if `ignore_optional_for_conversion` gets documented, this should be a proper reference to that option.
```suggestion
        .. seealso::
            The ``ignore_optional_for_conversion`` option of commands.
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
+        typing.Union[DesiredType, None] for type checking.
+
+        
+        This converter class is still provisional
+        """
+
+        def __class_getitem__(cls, key: _T_OPT) -> _T_OPT:
+            if isinstance(key, tuple):
+                raise TypeError("Must only provide a single type to Optional")
+            return key
```
Jackenmen: This isn't documented because of the place it is in (under `if TYPE_CHECKING:`)

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
-    type
-        The converter class, which will be a subclass of `TimedeltaConverter`
-    """
+if TYPE_CHECKING:
+    RedOptional = Optional
+else:
+
+    class RedOptional:
+        """
+        This can be used instead of typing.Optional
```
Jackenmen:
```suggestion
        This can be used instead of `typing.Optional`
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
+        To avoid discord.py special casing the conversion behavior
```
Jackenmen:
```suggestion
        to avoid discord.py special casing the conversion behavior.
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
 
-    return ValidatedConverter
+    class UserInputOptional:
+        """
+        This can be used when user input should be converted as discord.py
+        treats typing.Optional, but that the type should not be equivalent to
+        typing.Union[DesiredType, None] for type checking.
+
+        
+        This converter class is still provisional
```
Jackenmen:
```suggestion
        .. warning::
            This converter class is still provisional.
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
+
+    class Literal(dpy_commands.Converter):
+        """
+        This can be used as a converter for ``typing.Literal``
+
+        In a type checking context it is ``typing.Literal``
+        In a Runtime context, it's a converter which only matches the literals it was given.
+
+       
+        This converter class is still provisional
```
Jackenmen:
```suggestion
        .. warning::
            This converter class is still provisional.
```

Jackenmen (changes requested):

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
+    UserInputOptional = Optional
+
+
+if TYPE_CHECKING:
+    Literal = _TLiteral
+else:
+
+    class Literal(dpy_commands.Converter):
+        """
+        This can be used as a converter for ``typing.Literal``
```
Jackenmen:
```suggestion
        This can be used as a converter for `typing.Literal`.
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
+
+        In a type checking context it is ``typing.Literal``
```
Jackenmen:
```suggestion
        In a type checking context it is `typing.Literal`.
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
+        In a Runtime context, it's a converter which only matches the literals it was given.
```
Jackenmen:
```suggestion
        In a runtime context, it's a converter which only matches the literals it was given.
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
+
+_T_OPT = TypeVar("_T_OPT", bound=Type)
+
+if TYPE_CHECKING:
 
-    return ValidatedConverter
+    class UserInputOptional:
+        """
+        This can be used when user input should be converted as discord.py
+        treats typing.Optional, but that the type should not be equivalent to
```
Jackenmen:
```suggestion
        treats `typing.Optional`, but the type should not be equivalent to
```

Jackenmen (commented):
Now these are just me being picky, feel free to ignore those, but it's just a matter of merging suggestions.

redbot/core/commands/converter.py
```diff
@@ -138,17 +149,17 @@ class APIToken(discord.ext.commands.Converter):
     This will parse the input argument separating the key value pairs into a 
     format to be used for the core bots API token storage.
     
-    This will split the argument by either `;` ` `, or `,` and return a dict
+    This will split the argument by a space, colon, or semicolon and return a dict
     to be stored. Since all API's are different and have different naming convention,
     this leaves the onus on the cog creator to clearly define how to setup the correct
     credential names for their cogs.
 
     Note: Core usage of this has been replaced with DictConverter use instead.
```
Jackenmen:
```suggestion
    Note: Core usage of this has been replaced with `DictConverter` use instead.
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +173,256 @@ class APIToken(discord.ext.commands.Converter):
...
+_T_OPT = TypeVar("_T_OPT", bound=Type)
+
+if TYPE_CHECKING:
 
-    return ValidatedConverter
+    class UserInputOptional:
+        """
+        This can be used when user input should be converted as discord.py
+        treats typing.Optional, but that the type should not be equivalent to
+        typing.Union[DesiredType, None] for type checking.
```
Jackenmen:
```suggestion
        ``typing.Union[DesiredType, None]`` for type checking.
```

Jackenmen (commented):
Two more nitpicks, might as well add them to the list 😛

Jackenmen:
Soooo, there's a workaround for the issue with `RedOptional` and `Literal` being exported incorrectly...
Re-exporting when importing (`from typing import Optional as RedOptional`) and using `if not TYPE_CHECKING:` to define implementation at runtime would make it work properly.

Jackenmen:
Just FYI, I reverted changes to `kickban.py`, I didn't actually plan to do this for you but my tools betrayed me... Ekhm, anyhow, changes to that file don't really fit into this PR after it no longer is replaced with `UserInputOptional` + #3437 will get rid of the `cast` that I think you wanted to get rid of.

Still planning to do the actual review tomorrow 😄

mikeshardmind:
No worries, I left the PR editable intentionally, sound reversion if it simplifies things.

redbot/core/commands/dpy_reimplements.py
```diff
@@ -0,0 +1,126 @@
...
+    check as check,
+    guild_only as guild_only,
+    dm_only as dm_only,
+    is_nsfw as is_nsfw,
+    has_role as has_role,
+    has_any_role as has_any_role,
+    bot_has_role as bot_has_role,
+    bot_has_any_role as bot_has_any_role,
+    cooldown as cooldown,
+)
```
Jackenmen: mypy screams about the redefinitions of those later in the code ("All conditional function variants must have identical signatures"), one possible solution for this would be putting those imports here under `if not TYPE_CHECKING`

redbot/core/commands/dpy_reimplements.py
```diff
@@ -0,0 +1,126 @@
...
+    return [f"<@{bot.user.id}> ", f"<@!{bot.user.id}> "]
+
+
+def when_mentioned_or(*prefixes) -> PrefixCallable:
+    def inner(bot: dpy_commands.bot.BotBase, msg: discord.Message) -> List[str]:
+        r = list(prefixes)
+        r = when_mentioned(bot, msg) + r
+        return r
+
+    return inner
```
Jackenmen: Do we want to define our own versions of those instead of just putting typed variants under `if TYPE_CHECKING` and re-exporting the ones from d.py?

Jackenmen (changes requested):
If you care about fixing `func.__requires*` stuff in this PR, feel free to but I didn't find it important enough to put review comments on stuff that needs `setattr()` use as it's not important to have it in this PR.

redbot/core/commands/dpy_reimplements.py
mikeshardmind: These ones aren't conditionally defined.

redbot/core/commands/dpy_reimplements.py
mikeshardmind: Conditional import is fine here. It may be ugly, but the point is for this to work for our users and devs, if a little bit more ugly importing in this module is needed for it, that's fine.

redbot/core/bot.py
```diff
@@ -160,6 +163,12 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
         self._red_ready = asyncio.Event()
         self._red_before_invoke_objs: Set[PreInvokeCoroutine] = set()
 
+    def get_command(self, name) -> commands.Command:
```
```suggestion
    def get_command(self, name: str) -> Optional[commands.Command]:
```

redbot/core/bot.py
```diff
@@ -160,6 +163,12 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
...
+        return super().get_command(name)
+
+    def get_cog(self, name) -> commands.Cog:
```
```suggestion
    def get_cog(self, name: str) -> Optional[commands.Cog]:
```

redbot/core/bot.py
```diff
@@ -160,6 +163,12 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
         self._red_ready = asyncio.Event()
         self._red_before_invoke_objs: Set[PreInvokeCoroutine] = set()
 
+    def get_command(self, name) -> commands.Command:
+        return super().get_command(name)
```
Jackenmen: This needs cast or assert.

redbot/core/bot.py
```diff
@@ -160,6 +163,12 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
...
+
+    def get_cog(self, name) -> commands.Cog:
+        return super().get_cog(name)
```
Jackenmen: This needs cast or assert.

redbot/core/commands/context.py
```diff
@@ -263,3 +274,63 @@ def me(self) -> discord.abc.User:
...
+        @property
+        def channel(self) -> discord.DMChannel:
+            ...
+
+        @property
+        def guild(self) -> None:
+            ...
+
+        @property
+        def me(self) -> discord.User:
```
Jackenmen:
```suggestion
        def me(self) -> discord.ClientUser:
```

redbot/core/commands/converter.py
```diff
@@ -138,17 +160,18 @@ class APIToken(discord.ext.commands.Converter):
     This will parse the input argument separating the key value pairs into a 
     format to be used for the core bots API token storage.
     
-    This will split the argument by either `;` ` `, or `,` and return a dict
+    This will split the argument by a space, colon, or semicolon and return a dict
```
Jackenmen:
```suggestion
    This will split the argument by a space, comma, or semicolon and return a dict
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +185,262 @@ class APIToken(discord.ext.commands.Converter):
...
+
+        Parameters
+        ----------
+        maximum : Optional[timedelta]
+            If provided, any parsed value higher than this will raise an exception
+        minimum : Optional[timedelta]
+            If provided, any parsed value lower than this will raise an exception
+        allowed_units : Optional[List[str]]
+            If provided, you can constrain a user to expressing the amount of time
+            in specific units. The units you can chose to provide are the same as the
```
Jackenmen: Not part of PR but since I noticed it, figured I can mention it 🤷‍♂ 
```suggestion
            in specific units. The units you can choose to provide are the same as the
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +185,262 @@ class APIToken(discord.ext.commands.Converter):
...
+
+        Attributes
+        ----------
+        maximum : Optional[timedelta]
+            If provided, any parsed value higher than this will raise an exception
+        minimum : Optional[timedelta]
+            If provided, any parsed value lower than this will raise an exception
+        allowed_units : Optional[List[str]]
+            If provided, you can constrain a user to expressing the amount of time
+            in specific units. The units you can chose to provide are the same as the
```
Jackenmen: Not part of PR but since I noticed it, figured I can mention it 🤷‍♂ 
```suggestion
            in specific units. The units you can choose to provide are the same as the
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +185,262 @@ class APIToken(discord.ext.commands.Converter):
...
+
+        
+        .. warning::
+            This converter class is still provisional.
+
+            This class may not play well with mypy yet
+            and may still require you guard this in a 
+            type checking conditional import vs the desired types
+            
+            We're aware and Looking into improving this
```
Jackenmen:
```suggestion
            We're aware and looking into improving this.
```

redbot/core/commands/converter.py
```diff
@@ -162,140 +185,262 @@ class APIToken(discord.ext.commands.Converter):
...
+    class NoParseOptional:
+        """
+        This can be used instead of `typing.Optional`
+        to avoid discord.py special casing the conversion behavior.
+
+        .. warning::
+            This converter class is still provisional.
+
+        .. seealso::
+            The ``ignore_optional_for_conversion`` option of commands.
```
Jackenmen: Would be convenient to have this reference the option for ease of navigation. I checked and this will get referenced properly by sphinx without adding anything.
```suggestion
            The `ignore_optional_for_conversion` option of commands.
```

redbot/core/commands/requires.py
```diff
@@ -31,7 +32,7 @@
     from .commands import Command
     from .context import Context
 
-    _CommandOrCoro = TypeVar("_CommandOrCoro", Callable[..., Awaitable[Any]], Command)
+    _CommandOrCoro = TypeVar("_CommandOrCoro", bound=Command)  # lie to mypy
```
Jackenmen: This type is supposed to represent both `commands.Command` and `Callable[..., Awaitable[Any]]]`, otherwise developers are gonna run into issues when they use check decorators before `@commands.command()` so I think this shouldn't be left like that unless I'm missing something here.

redbot/core/commands/requires.py
```diff
@@ -211,7 +207,11 @@ def __repr__(self) -> str:
 # result of the default permission checks - the transition from NORMAL
 # to PASSIVE_ALLOW. In this case "next state" is a dict mapping the
 # permission check results to the actual next state.
-PermState.TRANSITIONS = {
+
+TransitionResult = Tuple[Optional[bool], Union[PermState, Dict[bool, PermState]]]
+TransitionDict = Dict[PermState, Dict[PermState, TransitionResult]]
+
+PermStateTransitions: TransitionDict = {
```
Jackenmen, line 218: Since this is supposed to be public, shouldn't it be added to `__all__`?

redbot/core/commands/requires.py
```diff
@@ -248,13 +248,18 @@ def __repr__(self) -> str:
         PermState.ACTIVE_DENY: (False, PermState.ACTIVE_DENY),
     },
 }
-PermState.ALLOWED_STATES = (
+
+PermStateAllowedStates = (
```
Jackenmen, line 256: Since this is supposed to be public, shouldn't it be added to `__all__`?

redbot/core/commands/requires.py
```diff
@@ -248,13 +248,18 @@ def __repr__(self) -> str:
...
     PermState.ACTIVE_ALLOW,
     PermState.PASSIVE_ALLOW,
     PermState.CAUTIOUS_ALLOW,
 )
 
 
+def transition_permstate_to(prev: PermState, next_state: PermState) -> TransitionResult:
```
Jackenmen, line 263: Since this is supposed to be public, shouldn't it be added to `__all__`?

Jackenmen (changes requested):
Aside from stuff I put in review comments, we also need to type hint decorators in `requires` - `bot_has_permissions`, `has_permissions`, `is_owner`, ...
And it would also be good to type-hint `@commands.command()`/`@commands.group()` as even with d.py stubs we still end up with our own deco untyped. But I don't remember full story on why you removed the type hints from those in first place so I am probably missing something here.

There are also some things that I want to ask about but that are gonna be easier to discuss in discord.

mikeshardmind:
Resolved most of the remaining issues, some are small things that we can (and will) continue improving on. The commit has been rebased down to a single commit over V3/develop for sanity purposes with how many revisions this entailed.

redbot/core/commands/context.py
```diff
@@ -254,7 +265,7 @@ def clean_prefix(self) -> str:
         return pattern.sub(f"@{me.display_name}", self.prefix)
 
     @property
-    def me(self) -> discord.abc.User:
+    def me(self) -> Union[discord.User, discord.Member]:
```
Jackenmen:
```suggestion
    def me(self) -> Union[discord.ClientUser, discord.Member]:
```

Jackenmen (dismissed):
Gonna go ahead and approve, there's one suggestion above which I wanted to just commit myself, but I don't have permissions to push to this repository, ping me on discord and I can reapprove if my review gets marked as stale because of committing that.
Would love to get some feedback on this comment too but it's not something that makes it worth to hold off the approval: https://github.com/Cog-Creators/Red-DiscordBot/pull/3410#discussion_r370879326

redbot/core/commands/requires.py
mikeshardmind: I've put this back for now, but I'm not entirely sure we should be encouraging decorator orders which don't stack logically as these end up doing assignments to function objects which gets much more difficult to write good typing for.

Jackenmen (dismissed):

Jackenmen (approved):

Issue #2667, opened 2019-05-15
[Help command] Future plans 


mikeshardmind:
The help redesign in #2628 is not the end of our improvements to help.

There are some default behavior changes in it. While the defaults have changed, there should still a setting added for showing that commands exist or not

Additionally, there are some other planned improvements:

- [x] Exposing a setting for using a menu, rather than paginating output.
- [ ] Adding help attribute use for cog developers, notably: (with more to consider)
  - [ ] Allowing them to set a max page size for their cog/groups lower than the bot
  - [x] (Get cog creator input for other potential useful opts)
- [ ] An API for registering additional 3rd party formatters with the bot and a setting for which formatter to use.  
- [x] Delete delay settings
  - Consider this as a per guild or per channel, not global?
  - Exclude DMs?
- [ ] Guild specific Settings
- [ ] Better whitespace handling
- [ ] Longer Timeout default + customization of timeout

aikaterna:
For the menu style help, could a longer timeout for buttons be considered? In talking with Preda, Kenny, and Trusty it seems that all of us have gotten messages from our bot users at some point asking for longer timeouts on the menu reactions.

PR #3433, opened 2020-01-22
Help delete pages delay feature

DevilXD:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This PR adds a feature that allows you to set a delay after which the help messages / pages will be deleted after using a help command. Disabled by default. Doesn't affect embed-menu-based help since that's handled differently. Doesn't delete when sending to DMs, nor does it delete the issued help command message itself. Handles code blocks and embed-based help. Uses bulk delete. Tested on code-blocks help only, but should work for embed type too.

It addresses one of the future plans in #2667, however the current help config structure doesn't seem to support per-guild settings easily, so this has been implemented as a global bot setting. Should be good enough though.

### DIFF
--- a/changelog.d/3433.feature.rst
+++ b/changelog.d/3433.feature.rst
@@ -1,0 +1,1 @@
+Added a ``[p]helpset deletedelay`` command, that lets you set a delay (in seconds) after which help messages / pages will be deleted.
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -88,6 +88,7 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):  # pylint: d
             custom_info=None,
             help__page_char_limit=1000,
             help__max_pages_in_guild=2,
+            help__delete_delay=0,
             help__use_menus=False,
             help__show_hidden=False,
             help__verify_checks=True,
@@ -186,9 +187,9 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):  # pylint: d
     def before_invoke(self, coro: T_BIC) -> T_BIC:
         """
         Overridden decorator method for Red's ``before_invoke`` behavior.
-        
+
         This can safely be used purely functionally as well.
-        
+
         3rd party cogs should remove any hooks which they register at unload
         using `remove_before_invoke_hook`
 
@@ -199,12 +200,12 @@ class RedBase(commands.GroupMixin, commands.bot.BotBase, RPCMixin):  # pylint: d
             only called if all checks and argument parsing procedures pass
             without error. If any check or argument parsing procedures fail
             then the hooks are not called.
-        
+
         Parameters
         ----------
         coro: Callable[[commands.Context], Awaitable[Any]]
             The coroutine to register as the pre-invoke hook.
-        
+
         Raises
         ------
         TypeError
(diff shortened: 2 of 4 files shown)


mikeshardmind:
I have a few concerns about this implemented as-is, and this was the one thing on that list I was considering removing from it. It mostly comes down to not being able to come up with a good way to pick a lower bound on the timeout which is appropriate for anything from 1 page, to 100 help pages (and we already have at least 1 bot in testing with that many pages) to prevent help from being a useless command. I think the max pages in a guild setting and the menu option both do enough to prevent help spam in channels.

DevilXD:
I mean, I understand this, but it's another thing that this is optional, and it just gives more options - it's disabled by default, and no one is forced to use this. If there's a bot with 100 help pages, and you think there cannot be a lower limit to the amount of time someone might need to read all of that, then they can just not be using this at all? When disabled, the messages stay and nothing changes. Additionally, there's a pretty high chance those 100 pages would end up in DMs due to the pages limit, where this feature is disabled too (no changes).

I feel this setting would be really useful for people who have small amount of pages of help, as well as optional cleaning up after showing help for individual cogs, commands and subcommands (those count as pages too, and they're usually just a single page).

mikeshardmind:
Just because "it's optional" doesn't make it a good option for inclusion in core Red. There are more considerations we have with every addition, and I didn't get into all of the implementation pieces of that because I wasn't even still sure if we should support it as an option.

It's also:

- More calls to config (We want to limit these in anything that should feel responsive)
- More API calls (rate limit concerns)
- A cause for 403s (This does nothing to handle checking the perms required to use that endpoint)
- More behaviors that have to be maintained

As-is, this would be a hard no from me. 

If you were to address confirming having permissions and also only use this setting if max pages is a guild is both set and under 100 (which limits the behavior of delete_messages to a single API call in addition without end-user inconsistent behavior) then it would be possible to consider as an option, but I didn't want to ask you to make these changes if we weren't even still sure it was something we should support.

DevilXD:
I've coded this thing in 3 hours on today's morning (carried by the will of helping and "that'd be useful to have" idea), and was kinda expecting to receive harsh criticism based on this, yes.

I understand the concerns, and have addressed (and tested) them. The current implementation checks the amount of pages / messages to delete and appropriate permissions as requested. It also does one less config call, while reusing the computed permissions based on an assumption, for which I've left an informative (and hopefully sufficient) comment on. 

The only worrying thing left here, would be the additional API call required for the messages to be deleted. I'm guessing this is left up for the discussion, but I think that since this is optional, one could just disable it if it'd cause any problems. That's just my way of thinking about it though.

mikeshardmind:
I'm *okay* accepting this with one other tweak which seems to have been a misunderstanding of intention which was probably my fault.

The 100 messages condition isn't that there are less than 100 messages, it's that there can't ever be more than 100 with the current max pages in guild setting. This prevents an inconsistent behavior on the bot where it won't delete if the full help sends 200 pages, but a specific cog help only sent 2

Drapersniper:
@mikeshardmind 
Seems like @DevilXD addressed most of your points we now actually do less config calls  which is a nice chance (only 1 call)  permissions being checked means 403s shouldn’t happen.

Given that quite a lot of users don’t use menus personally I think this opinions would be really welcomes by the community, given that I happens when there’s less than 100 pages too so it’s just 1 extra api call 

@DevilXD can you add a check to the command itself to only be visible if the bot has manage message perms? Also mention that this will not work if there are more than 100 menu pages, which these to I’d give this feature a +1 from me.

Drapersniper:
Me and sinbad replied at the same time xD, functionally take his review, just the documentation part and check on mine please devil

mikeshardmind (approved):
thanks for the follow-up

PR #3434, opened 2020-01-23
[Docs] Fix `which` cmd for pyenv in systemd docs

Zephyrkul:
### Type

- [X] Docs

Using `which python` when using a pyenv venv returns one of pyenv's shims (e.g. `/home/zephyrkul/.pyenv/shims/python`), which won't work in systemd when using the docs-recommended `pyenv shell` setting.
Rather, `pyenv which python` should be used instead.


mikeshardmind (approved):
Thanks for catching that!

PR #3435, opened 2020-01-23
CI ports from Travis CI

Kowlin:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

The first dedicated port of the Travis tests to GitHub CI.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -9,34 +9,34 @@ jobs:
           python_version:
             - "3.8"
           tox_env:
-          - py
-          - style
-          - docs
-          - postgres
+            - py
+            - style
+            - docs
+            - postgres
           include:
-          - tox_env: py
-            friendly_name: Tests
-          - tox_env: style
-            friendly_name: Style
-          - tox_env: docs
-            friendly_name: Docs
+            - tox_env: py
+              friendly_name: Tests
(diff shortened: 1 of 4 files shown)


.github/workflows/tests.yml
```diff
@@ -0,0 +1,92 @@
+name: Tests
+on: [push, pull_request]
+
+jobs:
+    tox-py:
+      runs-on: ubuntu-latest
+      strategy:
+        matrix:
+          python-version: ['3.8.1']
```
Jackenmen: This is not safe to do like this - when Python gets updated in Actions's virtual environment, previous micro version is no longer gonna be there causing the whole workflow to error. Same applies to all other occurrences of "3.8.1"
```suggestion
          python-version: ['3.8']
```

Kowlin: We can wait a few days until the live release of 3.8.1 for GHA, so I think it would be fine just leaving it like this until its ready.

Kowlin:
As a note to my self, need to port over deploying and Crowdin CI

Jackenmen:
Python 3.8.1 is now available in all images and can be safely used. The comment I made earlier about using "3.8.1" is still valid and needs changing to "3.8"

Jackenmen (approved):

PR #3440, opened 2020-01-24
[Downloader] Do the initialization in task to avoid timeout on bot startup

### LINKED ISSUE #3415: [Downloader] Downloader unloads on bot restart.
# Other bugs

I wanted to update my cogs by running ``[p]cog update``

My cogs to update

#### What actually happened?

Nothing, because downloader was unloaded. I did some troubleshooting and noticed that this happens at every restart.

1. Load downloader
2. Install a 3rd party cog
3. Restart server/machine (Bot as systemd service active)
4. Try to update cog

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #3415

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -54,7 +54,7 @@ class Downloader(commands.Cog):
 
         self._repo_manager = RepoManager()
         self._ready = asyncio.Event()
-        self._init_task = bot.loop.create_task(self.initialize())
+        self._init_task = None
 
     def _create_lib_folder(self, *, remove_first: bool = False) -> None:
         if remove_first:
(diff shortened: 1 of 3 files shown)


redbot/cogs/downloader/downloader.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, bot: Red):
         self._create_lib_folder()
 
         self._repo_manager = RepoManager()
+        self._ready = asyncio.Event()
+        self._init_task = bot.loop.create_task(self.initialize())
```
mikeshardmind: This line probably shouldn't be part of a constructor method as it is possible for the bot to fail to add the cog.

mikeshardmind (changes requested):

mikeshardmind (approved):

PR #3441, opened 2020-01-24
[Warnings]Make it possible to add reason with unwarn

### LINKED ISSUE #3093: `[p]unwarn` has no arg for reason, and modlog says one is addable
# Command bugs

#### Command name

`[p]unwarn`

`warnings`

There to be a third argument where you can put an optional reason for unwarning the user.

#### What actually happened?

There is no argument that allows you to set the reason. The modlog channel says "**Reason:** Use the `reason` command to add it", implying that you should be able to set a reason without needing to do it in two commands.

1. `[p]load warnings`
2. Set up a modlog channel, `[p]modlogset modlog <channel>`
3. `[p]warn` someone
4. `[p]unwarn` them, and note how you cannot add a reason
5. Notice how the modlog says that there is no reason added to the case
6. Use `[p]reason` and notice how you can add a warning to the case

#### Others
I may (attempt!) to PR a fix myself - though I'm a little busy so feel free to do this yourself.
I would like to thank [Skyrider#1337 for reporting this bug](https://discordapp.com/channels/133049272517001216/387398816317440000/639794017596473355). And yes, I waited a whole 11.5 hours for them to open an issue themselves.

Dav-Git:
Might need to be put in Optional[reason: str]

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This closes #3093 .
PR opened since Vexed is kinda on a break iirc.

### DIFF
--- a/changelog.d/warnings/3093.enhance.rst
+++ b/changelog.d/warnings/3093.enhance.rst
@@ -1,0 +1,1 @@
+Added the possibility to directly add a reason when using ``[p]unwarn``.
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -431,7 +431,7 @@ class Warnings(commands.Cog):
             if reason.lower() not in registered_reasons:
                 msg = _("That is not a registered reason!")
                 if custom_allowed:
-                    reason_type = {"description": reason, "points": points}
+                    reason_type = {"description": reason}
                 elif (
                     ctx.guild.owner == ctx.author
                     or ctx.channel.permissions_for(ctx.author).administrator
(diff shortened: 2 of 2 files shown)


Dav-Git:
The check's didn't fail when I tested with `tox` , not sure what Lint is trying to tell me.

Jackenmen:
It's trying to tell you that you used `points` variable that is not defined in that scope.

Jackenmen:
It might be better to not require to use registered reason in `[p]unwarn` command as reasons registered for warnings will most likely not be really useful as reasons for removing a warning.
This is just an opinion though, not a change request so don't change this unless reviewer assigned to this PR tells you to.

Dav-Git:
I was thinking about that but didn't want to rule out the option of having registered "unwarn reasons"

Dav-Git:
This is continued in #3490

PR #3452, opened 2020-01-26
[Permissions] Ordering fix

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

As this is a permissions change, let's start with the TLDR:

No direct security impact, just an ordering issue which could have unintended consequences for guilds. Still important to handle. This does prevent this issue from being able to amplify the effects of using an incorrect decorator.

Further discussion of impact not copied from discord.


Kowlin (approved):
Fix is working as intended, and exploit is no longer achievable.

PR #3460, opened 2020-01-27
[Core] Update licenseinfo command to conform with non-American English.

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Update the licence command to confirm with non-American English and keep it inline with the `colour`/`color` command


mikeshardmind (changes requested):
You swapped en-US to an alias here. That needs to go back, the project default locale is en-US, and the *shown* command names should reflect this.

Aliases being displayed as well can be handled as a separate change to help, but our default view should be consistent across the project here.

Jackenmen (approved):

Issue #3461, opened 2020-01-27
CentOS/RHEL 8


flyingmongoose:
Decided to see if Red would run okay on CentOS/RHEL 8.

It does, however the documentation for CentOS/RHEL 8 would be slightly different from CentOS/RHEL 7

IUS is not planning on supporting RHEL 8, so git2u is not available (at least not in that manner), thus regular Git will need to be installed. Plus other packages that would be in IUS are in epel-release now.
Ref: https://ius.io/faq <-- look at RHEL 8 section.

So, adding a section for CentOS/RHEL 8 would basically just look like this

---
```bash
yum -y install epel-release
yum update -y
yum -y groupinstall development
yum -y install git zlib-devel bzip2 bzip2-devel readline-devel sqlite \
 sqlite-devel openssl-devel xz xz-devel libffi-devel findutils java-11-openjdk
```

Complete the rest of the installation by installing Python 3.8 with pyenv.

---

I have no problem doing a PR for this if you like, this technically works as is 'right now'.

Tobotimus:
Yeah when I updated the Linux docs a few months ago, I figured pyenv was the most reliable way to get Python 3.7/3.8 installed on RHEL 8. I left it as unsupported at the time though, because I wasn't sure if there was some easier way I wasn't aware of. I was kind of in disbelief at the time, since the OS release was so new (May 2019) and released almost a year after Python 3.7.0 (June 2018).

So I'd say go ahead with your PR, if you're not aware of an easier way.

flyingmongoose:
Yeah did some digging with the dnf installer side of things too, looking like still yum packages, and still pyenv, no real better option, but oh well.

After some additional digging, included yum packages for CentOS/RHEL8 only support Python 3.6

Anyway, #3463 PR added specifically for this.

mikeshardmind:
resolved by #3463

PR #3463, opened 2020-01-27
Adding CentOS 8 Documentation

flyingmongoose:
### Type

- [x] Documentation

IUS will not be supporting RHEL 8, so utilizing epel-release and telling the system to use standard git is the best option.

Associated with #3461


mikeshardmind (approved):
Thanks!

Always glad to get docs clarifications in from people who are actually using the relevant portions. We've generally tried to cover what we know or have had time to test and trust to be correct enough.

PR #3472, opened 2020-01-29
[Mod] Move ignored guilds and channels to core

### LINKED ISSUE #3148: Caching for prefixes/whitelists/blacklists

### LINKED ISSUE #3220: [Core] `[p]blacklist add` fails when trying to add an uncached user
# Command bugs

#### Command name

`[p]blacklist add`

Core

Command to run successfully.

#### What actually happened?

Got an error:
```
Exception in command 'blacklist add'
  File "/home/ubuntu/red-venv/lib/python3.7/site-packages/discord/ext/commands/core.py", line 79, in wrapped
  File "/home/ubuntu/red-venv/lib/python3.7/site-packages/redbot/core/core_commands.py", line 1753, in blacklist_add
    if await ctx.bot.is_owner(user):
  File "/home/ubuntu/red-venv/lib/python3.7/site-packages/redbot/core/bot.py", line 481, in is_owner
    if user.id in self._co_owners:
AttributeError: 'int' object has no attribute 'id'
```

Run `[p]blacklist add 123`

TrustyJAID:
### Type

- [x] Bugfix
- [x] Enhancement
- [x] New feature

Add caching for ignored guilds and channels.
Add caching for whitelist and blacklist.
Fix #3220.
Add consume-rest for whitelist and blacklist commands to add multiple users or roles in one command.
Add ability to ignore channel categories ignoring commands in all channels under the category.
Edit: Resolves #3148.


Jackenmen:
Should also resolve #3148, could you add that to description?

redbot/cogs/mod/mod.py
```diff
@@ -85,35 +85,36 @@ def cog_unload(self):
...
+            guild_dict = await self.settings.all_guilds()
+            for guild_id, info in guild_dict.items():
+                delete_repeats = info.get("delete_repeats", False)
+                if delete_repeats:
+                    val = 3
+                else:
+                    val = -1
+                await self.settings.guild(discord.Object(id=guild_id)).delete_repeats.set(val)
+            await self.settings.version.set("1.0.0")  # set version of last update
+        if await self.settings.version() != "1.1.0":
```
mikeshardmind: This isn't forward safe. (that is to say, `<` should be used, not `!=`) Some of the other migrations in the bot use strict integer schema versioning because of this as it's easily comparable and orderable, though turning this into a tuple would work just as well.

mikeshardmind (changes requested):
Mostly good to go, just a small detail on how the versioning here works to prevent this from being a problem in the future.

mikeshardmind (approved):

PR #3474, opened 2020-01-30
[Core] Add a cli flag for setting a max size of message cache

### LINKED ISSUE #3473: [Core] Add a cli flag for setting a max size of message cache
# Feature request

I think it would be nice to have a cli flag for setting max size of message cache (`discord.Client.max_messages`) and maybe also disabling it.

PredaaA:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This adds a cli flag for being able to set a max size for the message cache.
Resolves #3473.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -150,10 +150,8 @@ class RedBase(
             kwargs["command_not_found"] = "Command {} not found.\n{}"
 
         message_cache_size = cli_flags.message_cache_size
-        if message_cache_size == "None":
+        if cli_flags.no_message_cache:
             message_cache_size = None
-        else:
-            message_cache_size = int(message_cache_size)
         kwargs["max_messages"] = message_cache_size
         self._max_messages = message_cache_size
 
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -149,8 +149,13 @@ class RedBase(
         if "command_not_found" not in kwargs:
             kwargs["command_not_found"] = "Command {} not found.\n{}"
 
-        kwargs["max_messages"] = cli_flags.message_cache_size
-        self._max_messages = cli_flags.message_cache_size
+        message_cache_size = cli_flags.message_cache_size
+        if message_cache_size == "None":
+            message_cache_size = None
+        else:
+            message_cache_size = int(message_cache_size)
+        kwargs["max_messages"] = message_cache_size
+        self._max_messages = message_cache_size
 
         self._uptime = None
         self._checked_time_accuracy = None
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -277,6 +277,10 @@ class RedBase(
     def colour(self) -> NoReturn:
         raise AttributeError("Please fetch the embed colour with `get_embed_colour`")
 
+    @property
+    def max_messages(self) -> Optional[int]:
+        return self._max_messages
+
     async def allowed_by_whitelist_blacklist(
         self,
         who: Optional[Union[discord.Member, discord.User]] = None,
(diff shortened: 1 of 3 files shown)


Jackenmen:
IMO it might be better to have a separate `--no-message-cache` bool flag for disabling message cache size instead of having to pass `"None"` to the `--message-cache-size` flag.

PredaaA:
> IMO it might be better to have a separate `--no-message-cache` bool flag for disabling message cache size instead of having to pass `"None"` to the `--message-cache-size` flag.

Good point. I will do that instead.

Jackenmen (dismissed):
LGTM, since this touches core, I'm gonna CC @mikeshardmind before merging but I don't see anything that could go wrong with this.

redbot/core/cli.py
```diff
@@ -74,6 +74,18 @@ def confirm(text: str, default: Optional[bool] = None) -> bool:
     return token
 
 
+def positive_int(arg: str) -> int:
+    try:
+        x = int(arg)
+    except ValueError:
+        raise argparse.ArgumentTypeError("Message cache size has to be a number.")
+    if x <= 0:
```
Jackenmen: After discussion in Discord, we decided that the message cache size shouldn't be allowed to be lower than 1k.
`positive_int` function should have a different name and it should check for `x < 1000` instead here.

redbot/core/bot.py
```diff
@@ -149,6 +149,12 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
         if "command_not_found" not in kwargs:
             kwargs["command_not_found"] = "Command {} not found.\n{}"
 
+        message_cache_size = cli_flags.message_cache_size
+        if cli_flags.no_message_cache:
+            message_cache_size = None
+        kwargs["max_messages"] = message_cache_size
+        self._max_messages = message_cache_size
```
Jackenmen, line 156: While this should remain private, adding a read-only property (in similar way as `bot.uptime`) `bot.max_messages` could be useful for cog creators to know if cache is enabled.
Alternatively, a `True/False` read-only property that just tells cog creator if cache is enabled or not would be enough too.

Jackenmen (changes requested):

redbot/core/cli.py
```diff
@@ -74,6 +74,18 @@ def confirm(text: str, default: Optional[bool] = None) -> bool:
     return token
 
 
+def positive_int(arg: str) -> int:
+    try:
+        x = int(arg)
+    except ValueError:
+        raise argparse.ArgumentTypeError("Message cache size has to be a number.")
+    if x < 1000:
+        raise argparse.ArgumentTypeError("Message cache size has to be greater than 1000.")
```
Jackenmen: Either change "has to be greater than" to "can't be lower than" or to "has to be greater than or equal to"
```suggestion
        raise argparse.ArgumentTypeError("Message cache size can't be lower than 1000.")
```

redbot/core/cli.py
```diff
@@ -74,6 +74,18 @@ def confirm(text: str, default: Optional[bool] = None) -> bool:
...
+    if x > sys.maxsize:
+        raise argparse.ArgumentTypeError(f"Message cache size has to be lower than {sys.maxsize}.")
```
Jackenmen: Either change "has to be lower than" to "can't be greater than" or to "has to be lower than or equal to"
```suggestion
        raise argparse.ArgumentTypeError(f"Message cache size can't be greater than {sys.maxsize}.")
```

redbot/core/bot.py
Rapptz: I would prefer if the attribute would not mirror the one in discord.py's kwargs since it could be potentially confusing.

Jackenmen: For posterity, there was a confusion caused because of my suggestion to use `True/False` as value. If we decided to go that way, it should certainly have a different name than `max_messages` but the approach currently taken in this PR (which is just using the same thing we pass to d.py in kwargs) is fine with `max_messages`.

Jackenmen (approved):
LGTM

PR #3486, opened 2020-02-01
Allow to edit prefixes through `redbot --edit`

### LINKED ISSUE #3481: Allow --prefix to be saved via --edit --no-prompt
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality
- [x] CLI? CLI!

The command line allows for `--edit --no-prompt` to modify and save the instance name, data path, owner, and token. I would like for it to also be able to save the `--prefix` option(s) as well, as this is one of two required configuration values needed when setting up a bot manually (the other being token).

Thank you!

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Allows users to edit prefixes through `redbot --edit`. There's a limitation in interactive use - user can only set prefixes that don't contain spaces. This limitation doesn't exist when using `-p/--prefix` flag(s).

Resolves #3481

### DIFF
--- a/changelog.d/3481.feature.rst
+++ b/changelog.d/3481.feature.rst
@@ -1,0 +1,1 @@
+Allow to edit prefix from command line using ``redbot --edit``.
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -107,6 +107,7 @@ async def edit_instance(red, cli_flags):
     no_prompt = cli_flags.no_prompt
     token = cli_flags.token
     owner = cli_flags.owner
+    prefix = cli_flags.prefix
     old_name = cli_flags.instance_name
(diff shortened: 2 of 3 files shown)


Jackenmen:
This is now ready for review.

Flame442 (approved):
We might want to consider making the list of args a variable somewhere so we don't have to edit multiple strings when we add one, but this change is fine.

mikeshardmind (approved):
It's starting to feel like we have a lot of subtle repeated things here still despite some separation.

Changes are fine, but if we do a sweep later to remove redundancies, this file should be looked at, but with some caution due to some of the startup concerns we need to handle.

PR #3488, opened 2020-02-01
[Mod] Fix shorthelp for `[p]modset dm`

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature


Flame442 (dismissed):
Not an issue with these changes, but the short help is a little awkward, something like "Toggle whether a message should be sent to a user when they are kicked/banned." maybe?

Flame442 (approved):

PR #3490, opened 2020-02-02
[Warnings]Make it possible to add reason with unwarn 

### LINKED ISSUE #3093: `[p]unwarn` has no arg for reason, and modlog says one is addable

Dav-Git:
Type

Bugfix
Enhancement

[X] New feature

Description of the changes

This closes #3093 .
This overwrites #3441 because I'm stupid and deleted the fork.

### DIFF
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -429,7 +429,7 @@ class Warnings(commands.Cog):
         reason_type = None
         async with guild_settings.reasons() as registered_reasons:
             if reason.lower() not in registered_reasons:
-                msg = _("That is not a registered reason!")
+                msg = _("That is not a registered reason.")
                 if custom_allowed:
                     reason_type = {"description": reason}
                 elif (
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -140,9 +140,7 @@ class Warnings(commands.Cog):
                 registered_actions.remove(to_remove)
                 await ctx.tick()
             else:
-                await ctx.send(
-                    _("No action named {name} exists!").format(name=action_name)
-                )
+                await ctx.send(_("No action named {name} exists!").format(name=action_name))
 
     @commands.group()
     @commands.guild_only()
@@ -200,8 +198,7 @@ class Warnings(commands.Cog):
             for r, v in registered_reasons.items():
                 if await ctx.embed_requested():
                     em = discord.Embed(
-                        title=_("Reason: {name}").format(name=r),
-                        description=v["description"],
+                        title=_("Reason: {name}").format(name=r), description=v["description"],
                     )
                     em.add_field(name=_("Points"), value=str(v["points"]))
                     msg_list.append(em)
@@ -227,20 +224,12 @@ class Warnings(commands.Cog):
         async with guild_settings.actions() as registered_actions:
             for r in registered_actions:
                 if await ctx.embed_requested():
-                    em = discord.Embed(
-                        title=_("Action: {name}").format(name=r["action_name"])
-                    )
+                    em = discord.Embed(title=_("Action: {name}").format(name=r["action_name"]))
+                    em.add_field(name=_("Points"), value="{}".format(r["points"]), inline=False)
                     em.add_field(
-                        name=_("Points"), value="{}".format(r["points"]), inline=False
+                        name=_("Exceed command"), value=r["exceed_command"], inline=False,
                     )
-                    em.add_field(
-                        name=_("Exceed command"),
-                        value=r["exceed_command"],
-                        inline=False,
-                    )
-                    em.add_field(
-                        name=_("Drop command"), value=r["drop_command"], inline=False
-                    )
+                    em.add_field(name=_("Drop command"), value=r["drop_command"], inline=False)
                     msg_list.append(em)
                 else:
                     msg_list.append(
@@ -373,9 +362,7 @@ class Warnings(commands.Cog):
             else:
                 for key in user_warnings.keys():
                     mod_id = user_warnings[key]["mod"]
-                    mod = ctx.bot.get_user(mod_id) or _(
-                        "Unknown Moderator ({})"
-                    ).format(mod_id)
+                    mod = ctx.bot.get_user(mod_id) or _("Unknown Moderator ({})").format(mod_id)
                     msg += _(
                         "{num_points} point warning {reason_name} issued by {user} for "
                         "{description}\n"
@@ -405,9 +392,7 @@ class Warnings(commands.Cog):
             else:
                 for key in user_warnings.keys():
                     mod_id = user_warnings[key]["mod"]
-                    mod = ctx.bot.get_user(mod_id) or _(
-                        "Unknown Moderator ({})"
-                    ).format(mod_id)
+                    mod = ctx.bot.get_user(mod_id) or _("Unknown Moderator ({})").format(mod_id)
                     msg += _(
                         "{num_points} point warning {reason_name} issued by {user} for "
                         "{description}\n"
@@ -426,11 +411,7 @@ class Warnings(commands.Cog):
     @commands.guild_only()
     @checks.admin_or_permissions(ban_members=True)
     async def unwarn(
-        self,
-        ctx: commands.Context,
-        user: Union[discord.Member, int],
-        warn_id: str,
-        reason: str,
+        self, ctx: commands.Context, user: Union[discord.Member, int], warn_id: str, reason: str,
     ):
         """Remove a warning from a user."""
 
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -428,29 +428,6 @@ class Warnings(commands.Cog):
         if user_id == ctx.author.id:
             return await ctx.send(_("You cannot remove warnings from yourself."))
 
-        custom_allowed = await self.config.guild(ctx.guild).allow_custom_reasons()
-        guild_settings = self.config.guild(ctx.guild)
-        reason_type = None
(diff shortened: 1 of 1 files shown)


redbot/cogs/warnings/warnings.py
```diff
@@ -433,14 +458,18 @@ def __init__(self, bot: Red):
                 await member_settings.total_points.set(current_point_count)
                 user_warnings.pop(warn_id)
         try:
+            reason_msg = (_("{description}").format(description=reason_type["description"]),)
```
Drapersniper: Not sure if I'm missing anything here but is there a purpose to be running this string through the translator?

redbot/cogs/warnings/warnings.py
```diff
@@ -422,6 +424,29 @@ def __init__(self, bot: Red):
...
+                msg = _("That is not a registered reason!")
+                if custom_allowed:
+                    reason_type = {"description": reason}
+                elif (
+                    ctx.guild.owner == ctx.author
+                    or ctx.channel.permissions_for(ctx.author).administrator
+                    or await ctx.bot.is_owner(ctx.author)
+                ):
+                    msg += " " + _(
+                        "Do `{prefix}warningset allowcustomreasons true` to enable custom "
```
Drapersniper:
```suggestion
                        "Run `{prefix}warningset allowcustomreasons true` to enable custom "
```

redbot/cogs/warnings/warnings.py
```diff
@@ -422,6 +424,29 @@ def __init__(self, bot: Red):
         if user_id == ctx.author.id:
             return await ctx.send(_("You cannot remove warnings from yourself."))
 
+        custom_allowed = await self.config.guild(ctx.guild).allow_custom_reasons()
+        guild_settings = self.config.guild(ctx.guild)
+        reason_type = None
+        async with guild_settings.reasons() as registered_reasons:
+            if reason.lower() not in registered_reasons:
+                msg = _("That is not a registered reason!")
```
Drapersniper:
```suggestion
                msg = _("That is not a registered reason.")
```
Please no shouting is too early in the morning.

redbot/cogs/warnings/warnings.py
```diff
@@ -433,14 +458,18 @@ def __init__(self, bot: Red):
                 await member_settings.total_points.set(current_point_count)
                 user_warnings.pop(warn_id)
         try:
+            reason_msg = (_("{description}").format(description=reason_type["description"]),)
+            prefix = (ctx.prefix,)
```
Drapersniper: What is the purpose of line 462 to 464?

Drapersniper (dismissed):

redbot/cogs/warnings/warnings.py
Dav-Git: You're 100% correct, there isn't. This one slipped through.

Drapersniper (commented):
This is a review for the Code and not functionality:
LGTM (Can't check functionality atm)

redbot/cogs/warnings/warnings.py
```diff
@@ -422,6 +428,29 @@ def __init__(self, bot: Red):
...
 
+        custom_allowed = await self.config.guild(ctx.guild).allow_custom_reasons()
+        guild_settings = self.config.guild(ctx.guild)
+        reason_type = None
+        async with guild_settings.reasons() as registered_reasons:
+            if reason.lower() not in registered_reasons:
+                msg = _("That is not a registered reason.")
+                if custom_allowed:
+                    reason_type = {"description": reason}
+                elif (
```
Drapersniper: Actually stating the valid reasons, when `custom_allowed` is disallowed and `reason.lower() not in registered_reasons`  would be more user friendly (something along the lines of "Please use one of the following: ...")

Right now `if reason.lower() not in registered_reasons and not custom_allowed` this just fails silently

Dav-Git: Will look at this when I return home later today

Dav-Git: This would then require to be edited in the warn command as well. My knowledge in string formatting is limited (everyone tells me my code is to slow) that I'd rather leave that up to someone else as it also doesn't really fit in this PR in my opinion. If you really want to see this included I'll need some guidance via discord.

I'm guessing that ``reason_lower`` is a list of strings that in one way ore another can be concatinated and passed to an embed which is then sent to the user. If we exceed the embed limit with the amount of registered reasons that's going to require some logic I don't know how to code and I'm not really sure if it's worth it at that point. Leaving it with just the instructions on how to enable custom warnings seems reasonable to me.

Dav-Git:
Are they supposed to not report anything? Do I need to do something?

redbot/cogs/warnings/warnings.py
```diff
@@ -422,6 +428,29 @@ def __init__(self, bot: Red):
...
+                ):
+                    msg += " " + _(
+                        "Run `{prefix}warningset allowcustomreasons true` to enable custom "
+                        "reasons."
+                    ).format(prefix=ctx.prefix)
+                    return await ctx.send(msg)
+            else:
+                reason_type = registered_reasons[reason.lower()]
+        if reason_type is None:
+            return
```
Drapersniper: I still think we should return a user friendly message when a valid reason is not present, just silently doing nothing is very confusing imo

Drapersniper: If another QA/Core-dev disagrees let me know and i'll review this as it is.

Jackenmen:
I still think that registered warn reasons shouldn't be used for unwarn - they will simply not make sense as unwarn reasons. *I think* allowing any text as unwarn reason is enough but I guess there's also the option of registering unwarn reasons.

redbot/cogs/warnings/warnings.py
Jackenmen: I do agree we shouldn't fail silently, though I don't think we should list all registered reasons, especially not in scope of this PR.

Dav-Git:
I think just removing the use of registered reasons for unwarn is the way forward with this. You guys let me know if that's what you want me to do.

Drapersniper (commented):
Okay now that we return a message when reason_type is None I’m more okay with this. Although jack made a point And not sure how strongly he feels about this one @jack1142 do you think removing the usage of the built in reason type is important enough here?

Jackenmen:
@Dav-Git after additional discussion, we decided that `[p]uwarn` should just allow any reason.

redbot/cogs/warnings/warnings.py
```diff
@@ -433,14 +462,15 @@ def __init__(self, bot: Red):
                 await member_settings.total_points.set(current_point_count)
                 user_warnings.pop(warn_id)
         try:
+            reason_msg = "{description}".format(description=reason_type["description"])
```
Jackenmen: This can just be (though this part might not even exist after allowing any reason):
```suggestion
            reason_msg = reason_type["description"]
```

Dav-Git:
Sounds good to me. I'll remove it for both commands. Maybe the commands to set the warn reasons are pending removal too though? Especially since the reasons then no longer server a purpose.

Dav-Git:
Correction, I misread that, allowing any reason for just unwarn

redbot/cogs/warnings/warnings.py
```diff
@@ -433,14 +439,15 @@ def __init__(self, bot: Red):
                 await member_settings.total_points.set(current_point_count)
                 user_warnings.pop(warn_id)
         try:
+            reason_msg = "{description}".format(description=reason)
```
Jackenmen: This is unneeded, you can just use the `reason` variable instead of `reason_msg` in `modlog.create_case` call.

redbot/cogs/warnings/warnings.py
```diff
@@ -400,13 +403,16 @@ def __init__(self, bot: Red):
...
+                    pagify(msg, shorten_by=58),
+                    box_lang=_("Warnings for {user}").format(user=user),
                 )
 
     @commands.command()
     @commands.guild_only()
     @checks.admin_or_permissions(ban_members=True)
-    async def unwarn(self, ctx: commands.Context, user: Union[discord.Member, int], warn_id: str):
+    async def unwarn(
+        self, ctx: commands.Context, user: Union[discord.Member, int], warn_id: str, reason: str,
```
Jackenmen: Reason should be optional:
```suggestion
        self,
        ctx: commands.Context,
        user: Union[discord.Member, int],
        warn_id: str,
        reason: str = None,
```

Jackenmen (changes requested):
Looks like I'm gonna review this PR after all, only one change should be needed here before approval :)

redbot/cogs/warnings/warnings.py
Dav-Git: Where is the change?

Dav-Git: oh i'm blind. yeah, agreed

Jackenmen (approved):
LGTM

PR #3493, opened 2020-02-03
Add black config to pyproject.toml

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

A simple change that makes it possible to just run `black .` in repo's folder without bothering with wrong settings. It resembles the flags used in our `Makefile/make.bat`.
Also added the excludes that are used by default in black so that we can simply add new ones if needed in future.
We could technically also modify make files so that we just rely on config in `pyproject.toml` but I left it as is for now.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -79,7 +79,7 @@ If you're not on Windows, you should also have GNU make installed, and you can o
 **Note:** If you're comfortable with setting up virtual environments yourself and would rather do it manually, just run `pip install -Ur tools/dev-requirements.txt` after setting it up.
 
 ### 4.2 Testing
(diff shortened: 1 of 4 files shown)


Jackenmen:
Ready for review I suppose :P

Kowlin (dismissed):

Drapersniper (dismissed):
Can we make the make files and style check use these please so that all configs are in 1 place?

Jackenmen:
Modified make files to only specify files black should run on so that it uses settings from `pyproject.toml` for rest. I left usage of `git ls-files *.py` as it could technically have different behaviour than the settings in `pyproject.toml` but I can change that as well if it's fine.

mikeshardmind (approved):

PR #3495, opened 2020-02-03
[Docs] Restructure venv instructions (put them in install guides)

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This PR moves instructions on creating venv to relevant install guides.

For more tech-savvy people, there's a new section "About Virtual Environments" that contains everything we had there before that wasn't part of creating venv.

Link to built docs from this PR: https://external-builds.readthedocs.io/html/red-discordbot/3495/index.html

### Reasoning behind this change
I decided to restructure the venv docs a little to make it easier for user to install Red just by following install guide without ever having to change the page. Currently, the user has to:
1) Go to linked "Installing Red in a Virtual Environment" page
2) Go through reasons why they should use virtual environment which most users shouldn't be bothered with
3) Choose a method to create venv
4) Use commands in the guide
5) Go back to "Installing Red" section with a link for right install guide.
This makes it more bothersome to go through install process than it needs to to be.

Having instructions right in install guide also make the docs more idiot-proof - users often skip this section maybe because they didn't see the link, maybe because they don't read that it's required, but putting the instructions right in the install guide can make it a lot less likely to happen.

I think one of main reasons behind having a separate page for venv before was because of it being optional and that is not the case anymore so there's simply no need for that. Another reason might have also been trying to avoid repeating same instructions in multiple places but to be fair, there's not much that we repeat here, the venv instructions are actually different (in small details but those details require us to separate most of those instructions in the venv guide already) for Windows and Unix.

For more tech-savvy people, there's a new section "About Virtual Environments" that contains everything we had there before that wasn't part of creating venv.


docs/install_windows.rst
```diff
@@ -68,8 +68,31 @@ Manually installing dependencies
...
+
 We require installing Red into a virtual environment. Don't be scared, it's very
-straightforward. See the section `installing-in-virtual-environment`.
+straightforward.
+
+First, choose a directory where you would like to create your virtual environment. It's a good idea
+to keep it in a location which is easy to type out the path to. From now, we'll call it
+``redenv`` and it will be located in your home directory.
+
+Start with opening a command prompt (open Start, search for "command prompt", then click it)
```
Drapersniper, line 82: Add a warning here and tell them to ensure they close power shell and open command prompt please

mikeshardmind (approved):

Issue #3494, opened 2020-02-03
[RFC] Discussion for Audio fixes to go into 3.3.3


Drapersniper:
https://github.com/Cog-Creators/Red-DiscordBot/issues/3485
https://github.com/Cog-Creators/Red-DiscordBot/issues/3471
https://github.com/Cog-Creators/Red-DiscordBot/issues/3467
https://github.com/Cog-Creators/Red-DiscordBot/issues/3492
https://github.com/Cog-Creators/Red-DiscordBot/issues/3389
https://github.com/Cog-Creators/Red-DiscordBot/issues/3353

The issues above are known issues with Audio 1.1.1 (Red 3.3.0)

There is also an issue with Playlist when you dont provide a scope when creating playlists which doesn't have an open issue for it.

Since Audio is currently in a frozen state I'm not keen on working on these HOWEVER users are important sooo they get to suggest which of these should go in Red 3.3.1

The following i will not consider for 3.3.1 unless you can give me a very strong valid point as to why they should and can't wait for the refactor:
https://github.com/Cog-Creators/Red-DiscordBot/issues/3467
https://github.com/Cog-Creators/Red-DiscordBot/issues/3492

If I missed anything this is your chance to propose it/raise it.

mikeshardmind:
Pushing this to 3.3.2 based on a brief discussion of what is and can be ready in discord.

mikeshardmind:
Note: 3.3.2 should be considered for these, but we're looking to do a 3.3.1 very soon with currently available improvements and fixes.

mikeshardmind:
Pushing to 3.3.3 as there still hasn't been a discussion on these.

PR #3500, opened 2020-02-04
V3/feature/audio/3

Drapersniper:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

Some further tweaks to audio in general

The main change on this one is the edit to `playlist.py`

We now no longer query the playlist twice and also fixed the playlist scope issue mentioned in #3494

This is a merge to `v3/feature/audio` and will not be going into `v3/develop`


aikaterna (dismissed):
This is approved in the idea of functionality only, and not code. 
Two things to address that Draper is aware of already:
* audioset.py: Playlist Are Not Available should be either Playlists Are Not Available or Playlist Is Not Available
* format_time will be modified to add days and maybe also weeks instead of just h/m/s

aikaterna (approved):
The latest commits are approved for resolving what I mentioned in the earlier review.

PR #3508, opened 2020-02-05
[Dev] Allow top-level await in code statements

Zephyrkul:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Allows for top-level await, async for, etc. code in debug and repl using the 3.8+ `ast.PyCF_ALLOW_TOP_LEVEL_AWAIT` flag.

The resulting object may be awaited on twice for backwards compatibility.

### DIFF
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -36,8 +36,8 @@ class Dev(commands.Cog):
         self._last_result = None
         self.sessions = set()
 
-    @classmethod
-    def async_compile(cls, source, filename, mode):
+    @staticmethod
+    def async_compile(source, filename, mode):
         return compile(source, filename, mode, flags=ast.PyCF_ALLOW_TOP_LEVEL_AWAIT, optimize=0)
 
     @staticmethod
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -1,3 +1,4 @@
+import ast
 import asyncio
 import inspect
 import io
@@ -35,6 +36,10 @@ class Dev(commands.Cog):
         self._last_result = None
         self.sessions = set()
 
+    @classmethod
+    def async_compile(cls, source, filename, mode):
+        return compile(source, filename, mode, flags=ast.PyCF_ALLOW_TOP_LEVEL_AWAIT, optimize=0)
+
     @staticmethod
     def cleanup_code(content):
         """Automatically removes code blocks from the code."""
@@ -53,7 +58,7 @@ class Dev(commands.Cog):
         """
         if e.text is None:
             return box("{0.__class__.__name__}: {0}".format(e), lang="py")
-        return box("{0.text}{1:>{0.offset}}\n{2}: {0}".format(e, "^", type(e).__name__), lang="py")
+        return box("{0.text}\n{1:>{0.offset}}\n{2}: {0}".format(e, "^", type(e).__name__), lang="py")
 
     @staticmethod
     def get_pages(msg: str):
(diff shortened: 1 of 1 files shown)


Zephyrkul:
Per Discord conversation, I've added `asyncio` and `aiohttp` to the env for [p]debug and [p]eval.

mikeshardmind:
This set of changes is on hold due to some *exceptional* weird behavior.

Zephyrkul:
It *seems* like adding `__builtins__` in 27b59dd resolved the *exceptionally* weird behavior. Based on the keys in `locals()` for stable's repl (`['ctx', 'bot', 'message', 'guild', 'channel', 'author', '_', '__builtins__']`), compared to the keys in this PR's repl (`['ctx', 'bot', 'message', 'guild', 'channel', 'author', 'asyncio', '_', '__builtins__']`, only `asyncio` is different here), this should theoretically fully resolve the main issue Sinbad found. However, because repl, and especially this version of repl, does some... interesting things with python, waiting for now is probably a good call.

As for the [other, more minor issue](https://discordapp.com/channels/240154543684321280/249967507518128128/674487403557290017) that Sinbad found....
Screenshot

mikeshardmind:
Some of the weirdness here is actually various known issues with python's asyncio repl.
As of the latest changes, this works as well as can be expected reasonably currently.

With that known, should we hold these until python 3.9, accept the weirdness with some warning, or have a separate command to enable to asyncio repl behavior?

(Technically, there are more clever approaches we could use such as some of the things [jishaku](https://github.com/Gorialis/jishaku) does, but this is likely overkill for a cog we don't recommend exist in production)

Jackenmen:
This is IMO great improvement even if it has some small issues so I don't think we should hold this of until Python 3.9. IMO, the best way here would be accepting this as a change to `[p]repl` and `[p]debug`, instead of making separate `[p]arepl` and `[p]adebug` commands. This should still let people run any code that they were able to run before and it's going to pretty much only improve the dev experience.

Zephyrkul:
Given that the weirdness is (apparently) acceptable for the built-in asyncio repl, I don't think that the weirdness is anything that particularly needs to be held off for py 3.9, especially for a cog that isn't for production.

Jackenmen:
Running `[p]debug []` results in error:
```py
TypeError: object list can't be used in 'await' expression
```

Jackenmen:
Value of `__debug__` is inconsistent between eval and debug/repl - eval's `__debug__` value is False when running red with `python -O -m redbot` and debug's and repl's `__debug__` value is True.

zephyrkul:
In my own playing around with this, I've yet to find any hiccups since 4137218. If anyone else wants to play around more with it, feel free.

Zephyrkul:
> Value of `__debug__` is inconsistent between eval and debug/repl - eval's `__debug__` value is False when running red with `python -O -m redbot` and debug's and repl's `__debug__` value is True.

I considered that out of scope of this PR, since this PR is focused more around adding `await` to repl and debug, but I can add this in if you insist.

Jackenmen:
If we would consider this out of scope of this PR, then `[p]debug` and `[p]repl` should remain with the current (pre this PR) behaviour which is using same optimize mode as the one Red runs with.

Zephyrkul:
> If we would consider this out of scope of this PR, then `[p]debug` and `[p]repl` should remain with the current behaviour which is using same optimize mode as the one Red runs with.

Sinbad agreed to keep the optimize mode as this PR has, so I haven't removed it. This PR is, again, mainly about improvements to debug and repl, so I didn't touch eval. But since you insist, I've added this to 62b0861.

Edit: To specify for clarification, the optimize changes were initially unintended, but kept in anyway after a conversation in discord.

Jackenmen:
`aiohttp` is ~~imported~~ added to globals in `[p]eval` and `[p]debug` but not in `[p]repl`.

Zephyrkul:
> `aiohttp` is ~imported~ added to globals in `[p]eval` and `[p]debug` but not in `[p]repl`.

On purpose. repl is meant to have a clean environment, or I assume it was meant to based on the fact that stable has the same "issue" with commands and discord.

mikeshardmind:
Behavior here is largely consistent with prior behavior with the exception of `optimize=0` (which was discussed)

I think the other changes we may consider now that we have people looking at this a bit closer with changes are fine to delay to other PRs.

mikeshardmind (approved):

PR #3514, opened 2020-02-06
Make `[p]announce` send error message only when an actual errors occurs

### LINKED ISSUE #3513: !announce always sends a message to owners
# Command bugs

#### Command name

announce

admin

When I use [p]announce <message>, the bot should announce my message without sending me any error.

#### What actually happened?

When I used [p]announce <message>, the bot announced my message, but also sent me the following error:
`I could not announce to the following servers:`

- Start Red and load admin at the latest version.
- Set up your announcement channel: `[p]announceset channel <channel_id>`
- Run an announcement: `[p]announce <message>`
- The bot will private message you with the error. 

#### Notes

- The bug seems to be in this line and below: [announce.py line 73](https://github.com/Cog-Creators/Red-DiscordBot/blob/1fc4ece14c5a661c748b5aae5773047260fc2007/redbot/cogs/admin/announcer.py#L73). Clearly, the bot should only send the error message to owners if the list failed is not empty.

trundlr:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Announce should only send an error message if an actual error occurs. Issue #3513

### DIFF
--- a/redbot/cogs/admin/announcer.py
+++ b/redbot/cogs/admin/announcer.py
@@ -77,5 +77,5 @@ class Announcer:
         )
         if failed:
             msg += humanize_list(tuple(map(inline, failed)))
-        await self.ctx.bot.send_to_owners(msg)
+            await self.ctx.bot.send_to_owners(msg)
         self.active = False
--- a/redbot/cogs/admin/announcer.py
+++ b/redbot/cogs/admin/announcer.py
@@ -70,12 +70,12 @@ class Announcer:
                 failed.append(str(g.id))
             await asyncio.sleep(0.5)
 
-        msg = (
-            _("I could not announce to the following server: ")
-            if len(failed) == 1
-            else _("I could not announce to the following servers: ")
-        )
         if failed:
+            msg = (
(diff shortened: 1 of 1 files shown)


redbot/cogs/admin/announcer.py
```diff
@@ -77,5 +77,5 @@ def cancel(self):
         )
         if failed:
             msg += humanize_list(tuple(map(inline, failed)))
```
Drapersniper, line 79:
```
		if failed:
			msg = (
        	    _("I could not announce to the following server: ")
        	    if len(failed) == 1
       		    else _("I could not announce to the following servers: ")
       		)
			msg += humanize_list(tuple(map(inline, failed)))
			await self.ctx.bot.send_to_owners(msg)
```
Would be better here There is no purpose to assigning and validating that string here unless there's been failed attempts

Drapersniper (changes requested):
as per the previous comment

Drapersniper (approved):
LGTM

Kowlin (approved):

PR #3517, opened 2020-02-06
Adds traceback logging to task exception handling

Flame442:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

When unhanded exceptions occur in tasks, this change causes the full traceback to be posted instead of just the exception message, which should make it easier to debug these problems.


mikeshardmind (approved):
LGTM

PR #3525, opened 2020-02-08
Use development versions of CI tools

DiscordLiz:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

I noticed `:=` breaks your tests and that it's been fixed already, just not released yet.


mikeshardmind (approved):
Looks fine as a way to allow these into the code base, though I would advise some caution against over using it.

PR #3537, opened 2020-02-10
Make systemd guide clearer on obtaining username and python path

### LINKED ISSUE #3462: [Docs] auto-restart using systemd  can be misleading
Currently the docs say the following:
Paste the following and replace all instances of `username` with the username, and `path` with the location you obtained above:

However in the section mentioned only the `path` is mentioned.

We could have a command in the section above to get the user `echo $USER` or `whoami` and explicitely mention that is the desired output for `username`  and that the outputs of `which python` is the desired output for `path`

Zephyrkul:
### Type

- [X] Docs

Resolves #3462

### DIFF
--- a/docs/autostart_systemd.rst
+++ b/docs/autostart_systemd.rst
@@ -13,6 +13,7 @@ In order to create the service file, you will first need to know two things, you
 First, your Linux :code:`username` can be fetched with the following command:
 
 .. code-block:: bash
+
     whoami
 
 Next, your python :code:`path` can be fetched with the following commands:
--- a/docs/autostart_systemd.rst
+++ b/docs/autostart_systemd.rst
@@ -8,7 +8,14 @@ Setting up auto-restart using systemd on Linux
 Creating the service file
 -------------------------
 
-In order to create the service file, you will first need the location of your :code:`redbot` binary.
+In order to create the service file, you will first need to know two things, your Linux :code:`username` and your Python :code:`path`
+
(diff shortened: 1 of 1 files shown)


docs/autostart_systemd.rst
```diff
@@ -24,7 +24,7 @@ Then create the new service file:
 
 :code:`sudo -e /etc/systemd/system/red@.service`
 
-Paste the following and replace all instances of :code:`username` with the username, and :code:`path` with the location you obtained above:
+Paste the following and replace all instances of :code:`username` with your Linux username (you can use the :code:`whoami` command if you're unsure), and :code:`path` with the location you obtained above:
```
Jackenmen: This is imo a little unclear on *why* would user want to use this command, I think adding something like "if you're unsure what your username is" would be enough here, but anything that makes it more clear when you should use the command and/or what is the reader "unsure" of is gonna work here so if you think something would fit better here, you can of course put that instead ;)
```suggestion
Paste the following and replace all instances of :code:`username` with your Linux username (you can use the :code:`whoami` command if you're unsure what your username is), and :code:`path` with the location you obtained above:
```

Jackenmen (changes requested):
One change on clarity needed

docs/autostart_systemd.rst
Zephyrkul: I decided to rewrite that section entirely since the whole thing was a little unclear at a couple points.

Jackenmen (approved):
Thanks for the PR

Issue #3023, opened 2019-10-02
[Trivia] whosthatpokemon - tinypic urls missing


aikaterna:
# Other bugs

#### What were you trying to do?

Play trivia with the whosthatpokemon list.

#### What were you expecting to happen?

For a silhouette of a pokemon to show.

#### What actually happened?

The image is gone.

#### How can we reproduce this issue?

Play trivia with the whosthatpokemon list and have it show any tinypic url.

mikeshardmind:
If we have the originals for some of these trivia lists, we should try and have them uploaded on a long term host okay with hotlinks, such as imgur, on an account controlled by the organization, while keeping the originals for reupload if needed.

aikaterna:
I have an archive of 151 gen 1 images I can convert to shadows/completely black and provide to whomever in the Red server if needed.

Jackenmen:
Maybe we could put them on GitHub (in separate org repo) and use cdn like jsdelivr.com/raw.githack.com/statically.io/etc to hotlink the images in trivias?

Kevin-Mok:
So, @Redjumpman has a Pokedex cog that has a [CSV file](https://github.com/Redjumpman/Jumper-Plugins/blob/V3/pokedex/data/Pokemon.csv) containing all the Pokemon names along with URL's to their images. The problem is that the URL's contain the Pokemon names in the URL, so you can't directly use them in the YAML file for the question. I extracted the relevant information from the CSV in this [text file](https://github.com/Cog-Creators/Red-DiscordBot/files/3967278/pokemon-name-image.csv.txt) (added `.txt` to end since GitHub doesn't allow attaching CSV files).

I was thinking if we could embed the image in the trivia question like @aikaterna does in their [inspirobot cog](https://github.com/aikaterna/aikaterna-cogs/blob/v3/inspirobot/inspirobot.py#L12-L24), then that'd be great. But, that would require rehauling how the YAML file is parsed, which seems like it could be annoying. Does anyone have any thoughts on this?

Drapersniper:
Aika is in the progress of updating the urls host afaik. 

The trivia cog doesn’t have permissions to explicitly enforce embeds the only cog in core that has such permission is Audio due to its UX. even embedding these images provided does provide the users with the ability to cheat by checking the image url (if I understand what you’re saying I.e use RJM images)

Kevin-Mok:
[...]

PR #3541, opened 2020-02-12
Please update whosthatpokemon.yaml

TBTakoz:
Most of the images (apart from 2) were dead links, so I've created, and updated the links with new silhouettes of the Pokémon.

### Type

- [ ] Enhancement

I went through every link to the images, and replaced them with new images that I have uploaded. I also changed it so each 'Nidoran' has a different name (Nidoran M and Nidoran F)


Flame442:
There's some discussion already happening in #3023, we were intending to have something hosted by the organization for the images.

mikeshardmind:
While I appreciate the effort, at the end of they day, we already have discussion about why we want these under org control.

PR #3542, opened 2020-02-12
[commands module] functools.partial support

mikeshardmind:
### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

As the annotations are derived from the source function, using the source module and globals here is fine for functool partials.


Jackenmen (approved):
LGTM

PR #3553, opened 2020-02-14
new mention behavior, new filter behavior

### LINKED ISSUE #3551: [Core] Filter roles
Filtering roles should be done in ctx.send now that the new discord permissions will allow bots with mention everyone or administrator to ping every role. We do not filter ctx.send currently and user converter does a role mention if you send full role mention as a string.

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

resolves #3551 

It is intentional that this not mix filters and add a new kwarg here.

### DIFF
--- a/redbot/core/commands/context.py
+++ b/redbot/core/commands/context.py
@@ -75,7 +75,9 @@ class Context(DPYContext):
             :func:`~redbot.core.utils.common_filters.filter_mass_mentions`.
             This must take a single `str` as an argument, and return
             the sanitized `str`.
-        \*\*kwargs
+        sanitize_roles : bool
+            Whether or not role mentions should be sanitized for you.
+        **kwargs
             See `discord.ext.commands.Context.send`.
 
         Returns
@@ -86,6 +88,10 @@ class Context(DPYContext):
         """
 
         _filter = kwargs.pop("filter", common_filters.filter_mass_mentions)
+        sanitize_roles = kwargs.pop("sanitize_roles", True)
+
+        if sanitize_roles and content and self.guild:
+            content = common_filters.sanitize_role_mentions(str(content), self.guild.roles)
 
         if _filter and content:
             content = _filter(str(content))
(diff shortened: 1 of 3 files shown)


Jackenmen (dismissed):
Looks good, style check fails (most likely because of `**kwargs` in docstring instead of `\*\*kwargs`) and I think that since `sanitize_role_mentions` is generally useful function, 
it should be documented.

mikeshardmind:
Documented. And it was just me reflexively splitting a line early because of our really long line length compared to other projects I work on. the kwargs bit removed an invalid escape.

Jackenmen:
Oh yeah, I knew that something happened about `\*\*kwargs` in Python 3.8 but got it the other way around 😄

redbot/core/commands/context.py
```diff
@@ -75,7 +75,9 @@ def __init__(self, **attrs):
             :func:`~redbot.core.utils.common_filters.filter_mass_mentions`.
             This must take a single `str` as an argument, and return
             the sanitized `str`.
-        \*\*kwargs
+        sanitize_roles : bool
```
Jackenmen, line 78: We should either include the default for this kwarg in the docstring or put the kwarg in func def so that sphinx puts it there automatically.

redbot/core/utils/common_filters.py
```diff
@@ -173,3 +177,32 @@ def escape_spoilers_and_mass_mentions(content: str) -> str:
         The escaped string.
     """
     return escape_spoilers(filter_mass_mentions(content))
+
+
+def sanitize_role_mentions(content: str, roles: List[discord.Role]) -> str:
```
Jackenmen: Could we support `Iterable[discord.Role]` instead here?

Jackenmen (dismissed):
Tested this already, just some minor stuff.

Jackenmen (approved):

PR #3562, opened 2020-02-17
[Core] Setting the bot's game/listening/watching displays new text in chat message and is capped at 128 chars.

Dav-Git:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

### Description of the changes
This closes #3556

When using ``[p]set game`` it now shows what the game was set to in the "Game set" message and also caps the length of the text to 128 chars to only accept "valid" status messages.

PR #3565, opened 2020-02-17
[Core] i18n fixes for commands that set statuses

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

i18n fixes for #3562


mikeshardmind (approved):
thanks, I missed that with looking over that one.

PR #3606, opened 2020-02-22
[Trivia] Better handling for session errors

### LINKED ISSUE #3578: [Trivia] Errors in game sessions need better handling
Currently any exception inside trivia session's task just stops it without any logging and feedback to session's players.
Example issues that can get hidden because of this issue: #3583, #3584

Flame442:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Properly cleans up a trivia session when it encounters an error and alerts the user that an error had occurred.

Fixes #3578

### DIFF
--- a/redbot/cogs/trivia/session.py
+++ b/redbot/cogs/trivia/session.py
@@ -111,10 +111,12 @@ class TriviaSession:
             LOG.error("A trivia session has encountered an error.\n", exc_info=exc)
             asyncio.create_task(
                 self.ctx.send(
-                    "An unexpected error occurred in the trivia session.\nCheck your console or logs for details."
+                    _(
+                        "An unexpected error occurred in the trivia session.\nCheck your console or logs for details."
+                    )
                 )
             )
-            asyncio.create_task(self.end_game())
+            self.stop()
 
     async def run(self):
         """Run the trivia session.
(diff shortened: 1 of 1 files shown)


redbot/cogs/trivia/session.py
```diff
@@ -98,8 +98,24 @@ def start(cls, ctx, question_list, settings):
...
+        except asyncio.CancelledError:
+            pass
+        except Exception as exc:
+            LOG.error("A trivia session has encountered an error.\n", exc_info=exc)
+            asyncio.create_task(
+                self.ctx.send(
+                    "An unexpected error occurred in the trivia session.\nCheck your console or logs for details."
+                )
+            )
+            asyncio.create_task(self.end_game())
```
Tobotimus: I don't think you should call end_game as it will pay out the winner. We don't pay out the winner on a timeout, we probably shouldn't pay them out on an error. Best just call `self.stop()`, which has the bonus of being a synchronous function.

Moreover, I think the current handler leaves the trivia session object in the trivia cog's cache. `TriviaSession.stop()` takes care of that.

Tobotimus (dismissed):

redbot/cogs/trivia/session.py
Flame442: Although `end_game` does call `stop` at the end, I agree it shouldn't give points regardless.

Tobotimus (approved):

PR #3607, opened 2020-02-23
[Core] Use new PyPI urls

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Source: https://packaging.python.org/guides/migrating-to-pypi-org/

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -310,7 +310,7 @@ class Core(commands.Cog, CoreLogic):
         author_repo = "https://github.com/Twentysix26"
         org_repo = "https://github.com/Cog-Creators"
         red_repo = org_repo + "/Red-DiscordBot"
-        red_pypi = "https://pypi.org/pypi/Red-DiscordBot"
+        red_pypi_data = "https://pypi.org/pypi/Red-DiscordBot/json"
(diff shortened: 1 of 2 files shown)


redbot/core/core_commands.py
```diff
@@ -310,7 +310,7 @@ class Core(commands.Cog, CoreLogic):
         author_repo = "https://github.com/Twentysix26"
         org_repo = "https://github.com/Cog-Creators"
         red_repo = org_repo + "/Red-DiscordBot"
-        red_pypi = "https://pypi.python.org/pypi/Red-DiscordBot"
+        red_pypi = "https://pypi.org/pypi/Red-DiscordBot"
```
Flame442: While these links work, the actual location both the old and new link bring you to is `https://pypi.org/project/Red-DiscordBot` (`pypi` -> `project`). Should we link there instead?

Jackenmen: This isn't actually used to link to the project but rather as an API endpoint. This endpoint only works with url `https://pypi.org/pypi/Red-DiscordBot/json`, not with `https://pypi.org/project/Red-DiscordBot/json`

Kowlin (approved):

PR #3608, opened 2020-02-23
Update deprecation warnings

Jackenmen:

Updates deprecations to be based on earliest date when it can be removed instead of specific minor version number. Removal will still occur in minor versions.
For Downloader's shared libraries, the warning is changed to indefinite "in future" as it is currently unknown when shared libraries will be removed.

This PR also adds a proper DeprecationWarning when APIToken class is imported/used.

### DIFF
--- a/docs/framework_commands.rst
+++ b/docs/framework_commands.rst
@@ -34,3 +34,5 @@ extend functionalities used throughout the bot, as outlined below.
     :members:
     :exclude-members: convert
     :no-undoc-members:
+
+    .. autoclass:: APIToken
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -29,7 +29,7 @@ _ = Translator("Downloader", __file__)
 
 DEPRECATION_NOTICE = _(
     "\n**WARNING:** The following repos are using shared libraries"
-    " which are marked for removal in Red 3.4: {repo_list}.\n"
+    " which are marked for removal in future: {repo_list}.\n"
     " You should inform maintainers of these repos about this message."
 )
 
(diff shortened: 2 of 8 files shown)


Flame442 (changes requested):
The strings are missing a few "the"s.
"...for removal in _**the**_ future..."
"...will be removed in _**the**_ first minor..."

Flame442 (dismissed):
A discussed in discord, these warnings are not showing up, seemingly for no reason, after applying this change. The old warnings do show up on a 3.3.6 bot, however the edited (and new) warnings do not appear. If someone else would like to test this PR to see if it is just me(/windows) being weird, that would be appreciated.

An example eval that produces a warning on 3.3.6 but not this pr is
```py
[p]eval
from redbot.core.utils import bounded_gather_iter
return bounded_gather_iter(ctx.send, loop='aaa')
```

Flame442 (dismissed):
Now that my warnings are not being suppressed, this works fine :P
For consistency's sake, all of the messages could be updated to have a similar format, but they work fine.

Flame442 (dismissed):

Flame442 (dismissed):
whatever...

Jackenmen:
I found an issue with the module name shown by the `AttributeError` when import from `redbot.core.commands`:
image
I fixed it in latest commit.

Flame442 (changes requested):
```py
...\redbot\core\commands\__init__.py:150: DeprecationWarning: `APIToken` is deprecated since Red 3.3.0 and will be removed in the first minor release after 2020-08-05. Use `DictConverter` instead.
  return _converter__getattr__(name)
```
This is the warning I get from putting `from redbot.core.commands import APIToken` in a cog (or eval). It does not explain where the improper code is being used (and has that return line alone for some reason?)

Jackenmen:
Ahh sorry, that's my oversight, I didn't notice that the deprecation warning needs a different stack level when being done from `commands` instead of `commands.converter`.

I pushed the commit that should fix this.

Flame442 (approved):

Jackenmen:
We could probably already merge this PR - this only updates deprecation warnings which means there's no breaking change, and it seems beneficial that people will be able to see them sooner rather than later. It's in 3.4.0 milestone mostly because it needs to be ready *before* 3.4 gets released, but I don't think there's a reason to keep this hanging until we're actually ready with 3.4.

Kowlin (approved):

Issue #3529, opened 2020-02-08
Deprecation warnings


Stonedestroyer:
Deprecation warnings.

- [ ] https://bugs.python.org/issue39015
```
<frozen importlib._bootstrap>:219: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
```

- [ ] Bug in CPython, no tracking issue
```
C:\Python39\lib\asyncio\subprocess.py:196: DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.
  stdin, stdout, stderr = await tasks.gather(stdin, stdout, stderr,
```

- [x] Fixed by https://github.com/Cog-Creators/Red-DiscordBot/pull/3610. 
``` 
..venv/lib/site-packages/redbot/core/config.py:513: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  self, current: collections.Mapping, defaults: Dict[str, Any] = ...
```

- [x] https://github.com/aio-libs/aiohttp/pull/4570 (awaiting bump to aiohttp 3.7+)
```
..venv/lib/site-packages/aiohttp/connector.py:964: DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.
  hosts = await asyncio.shield(self._resolve_host(
```

- [x] https://github.com/aio-libs/aiohttp/pull/4570 (awaiting bump to aiohttp 3.7+)
```
..venv/lib/site-packages/aiohttp/helpers.py:107: DeprecationWarning: "@coroutine" decorator is deprecated since Python 3.8, use "async def" instead
  def noop(*args, **kwargs):  # type: ignore
```

- [x] https://github.com/aio-libs/aiohttp/pull/4570 (awaiting bump to aiohttp 3.7+)
```
..venv/lib/site-packages/aiohttp/locks.py:21: DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.
  self._event = asyncio.Event(loop=loop)
```

- [x] Fixed by #3845
```
..venv/lib/site-packages/websockets/client.py:407: DeprecationWarning: rename klass to create_protocol
  warnings.warn("rename klass to create_protocol", DeprecationWarning)
```

[...]

PR #3610, opened 2020-02-23
Fix few deprecation warnings

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Minimizes some console noise we now have because of enabling deprecation warnings in our logging.

Specifically, it fixes these deprecation warnings:
1. Using ABCs from `collections` instead of `collections.abc`:
```py
redbot/core/config.py:513: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working

  self, current: collections.Mapping, defaults: Dict[str, Any] = ..
```
2. Invalid escape sequence:
```py
redbot/core/events.py:25: DeprecationWarning: invalid escape sequence \

  INTRO = """
```
3. Deprecation of `autodoc_default_flags` in Sphinx (this one is only about docs build, it's not part of bot's output):
```py
RemovedInSphinx30Warning: autodoc_default_flags is now deprecated. Please use autodoc_default_options instead.
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -230,5 +230,5 @@ extlinks = {"dpy_docs": (f"https://discordpy.readthedocs.io/en/v{dpy_version}/%s
 doctest_test_doctest_blocks = ""
 
 # Autodoc options
-autodoc_default_flags = ["show-inheritance"]
+autodoc_default_options = {"show-inheritance": True}
 autodoc_typehints = "none"
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -1,5 +1,5 @@
 import asyncio
-import collections
+import collections.abc
 import logging
 import pickle
 import weakref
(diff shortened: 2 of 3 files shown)


Stonedestroyer:
Fixes #3529

Jackenmen:
It works towards fixing #3529 (doesn't fully fix the issue), but from the warnings mentioned there, it only fixes this one, as rest is related to our dependencies:
```
.redbot\core\config.py:513: DeprecationWarning: Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated since Python 3.3, and in 3.9 it will stop working
  self, current: collections.Mapping, defaults: Dict[str, Any] = ...
```

Kowlin (approved):

PR #3615, opened 2020-02-26
Deprecation warning - alternative

### LINKED ISSUE #3612: Stop showing DeprecationWarnings from all modules by default
This provides no useful info for the end user, and it only causes noise and confusion to be generated

As such the current behaviour should be hidden behind the `--debug` flag
https://github.com/Cog-Creators/Red-DiscordBot/commit/42a4d10ea5ed0d8f337edb73a2c8a3c60e162aaa#diff-3b9bdb4dfee95e89963ffdc7160a5a6fR200 should use 

```py
if "--debug" in _sys.argv:
    _warnings.filterwarnings("default", category=DeprecationWarning)
```

Stonedestroyer:
Alternative to #3613

Hides the known ones for non-debug and shows all for debug. 

I did line number variant but could also do regex but I think line number may be better in this case incase there is two errors with same error but different linenumber. This could also be checked easily if it's fixed by debug.

### DIFF
--- a/redbot/__init__.py
+++ b/redbot/__init__.py
@@ -199,22 +199,23 @@ _warnings.filterwarnings("ignore", module=r"fuzzywuzzy.*")
 # Show DeprecationWarning
 _warnings.filterwarnings("default", category=DeprecationWarning)
 
-# DEP-WARN
-# Individual warnings - tracked in https://github.com/Cog-Creators/Red-DiscordBot/issues/3529
-# DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
-_warnings.filterwarnings("ignore", category=DeprecationWarning, module="importlib", lineno=219)
-# DeprecationWarning: "@coroutine" decorator is deprecated since Python 3.8, use "async def" instead
-#   def noop(*args, **kwargs):  # type: ignore
-_warnings.filterwarnings("ignore", category=DeprecationWarning, module="aiohttp", lineno=107)
-# DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.
-#   hosts = await asyncio.shield(self._resolve_host(..
-_warnings.filterwarnings("ignore", category=DeprecationWarning, module="aiohttp", lineno=964)
-# DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.
-#   self._event = asyncio.Event(loop=loop)
-_warnings.filterwarnings("ignore", category=DeprecationWarning, module="aiohttp", lineno=21)
-# DeprecationWarning: rename klass to create_protocol
-#   warnings.warn("rename klass to create_protocol", DeprecationWarning)
-#
-# discord.py is using deprecated kwarg name when making websockets connection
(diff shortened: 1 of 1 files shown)


Stonedestroyer:
Updated to now also print warnings for both but hides the known ones for non-debug.

redbot/__init__.py
```diff
@@ -198,3 +198,23 @@ def _update_event_loop_policy():
 _warnings.filterwarnings("ignore", module=r"fuzzywuzzy.*")
 # Show DeprecationWarning
 _warnings.filterwarnings("default", category=DeprecationWarning)
+if "--debug" not in _sys.argv:
+    _warnings.filterwarnings("default", category=DeprecationWarning)
```
Jackenmen: This line is unnecessary

redbot/__init__.py
```diff
@@ -198,3 +198,23 @@ def _update_event_loop_policy():
...
+    #   def noop(*args, **kwargs):  # type: ignore
+    _warnings.filterwarnings("ignore", category=DeprecationWarning, module="aiohttp", lineno=107)
+    # DEP-WARN
+    # DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.
+    #   hosts = await asyncio.shield(self._resolve_host(..
+    _warnings.filterwarnings("ignore", category=DeprecationWarning, module="aiohttp", lineno=964)
+    # DEP-WARN
+    # DeprecationWarning: The loop argument is deprecated since Python 3.8, and scheduled for removal in Python 3.10.
+    #   self._event = asyncio.Event(loop=loop)
+    _warnings.filterwarnings("ignore", category=DeprecationWarning, module="aiohttp", lineno=21)
```
Jackenmen: Could you add link to related issue for these 2 warnings?
https://github.com/aio-libs/aiohttp/pull/4570

redbot/__init__.py
```diff
@@ -198,3 +198,23 @@ def _update_event_loop_policy():
...
+    # DEP-WARN
+    # DeprecationWarning: rename klass to create_protocol
+    #   warnings.warn("rename klass to create_protocol", DeprecationWarning)
+    _warnings.filterwarnings(
+        "ignore", category=DeprecationWarning, module="websockets", lineno=407
+    )
```
Jackenmen: Could you also add comment that this warning is caused by d.py? Something like
```
# discord.py is using deprecated kwarg name when making websockets connection
# https://github.com/Rapptz/discord.py/issues/2574
```
should suffice.

Jackenmen (changes requested):
Few things that should be changed **if** we choose to handle #3612 in this way.

Also, this seems to missing one of the warnings mentioned in #3529?
```
<frozen importlib._bootstrap>:219: DeprecationWarning: an integer is required (got type float).  Implicit conversion to integers using __int__ is deprecated, and may be removed in a future version of Python.
```

Stonedestroyer:
I have also updated https://github.com/Cog-Creators/Red-DiscordBot/issues/3529 to make it easier to track these changes.

Stonedestroyer:
@jack1142 I can remove the error messages and put a clearer error message if you want to reduce the clutter.

Jackenmen (approved):
LGTM

PR #3625, opened 2020-02-28
Bring back "new mention behavior, new filter behavior (#3553)"

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Brings back changes from #3553 (reverted by #3619). This needs a discussion (before 3.4.0 release) about whether we want to make this change before Discord's upstream changes (trying to avoid us needing another breakage because of other changes that Discord is planning, supposedly before new role mention behavior will be live).

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -974,6 +974,7 @@ class RedBase(
     async def send_filtered(
         destination: discord.abc.Messageable,
         filter_mass_mentions=True,
+        filter_roles=True,
         filter_invite_links=True,
         filter_all_links=False,
         **kwargs,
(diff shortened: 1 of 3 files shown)


Jackenmen:
Funnily, today we got a more public information on Discord's plans that I have mentioned in the PR description:
https://github.com/discordapp/discord-api-docs/issues/1286#issuecomment-592865198
We still don't know which change is gonna come first, but this should give a more clear picture to people looking at this on why we're considering waiting rather than releasing these changes already.

Stonedestroyer:
I think merging this may be useful unless Discord releases their feature first so we don't get a gap of where bot role mentions go live and we have no patch for it.

PR #3626, opened 2020-02-28
[Mod] Wait until bot ready for _maybe_update_config task.

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

That's to avoid NoneType errors when bot isn't ready on launch.
It also remove initialize from setup() to move it into a task on \_\_init\_\_.

// This PR is not necessary if #3632 gets merged.


Jackenmen:
I don't quite get why this is needed here? Do you have any traceback that this change fixes?

PredaaA:
Sorry I forgot to include traceback of what happen without that PR, there it goes:
```py
Failed to load cog mod
  File "/botsvenv/lib/python3.8/site-packages/redbot/core/bot.py", line 638, in pre_flight
    await asyncio.wait_for(self.load_extension(spec), 30)
  File "/usr/lib/python3.8/asyncio/tasks.py", line 483, in wait_for
    return fut.result()
  File "/botsvenv/lib/python3.8/site-packages/redbot/core/bot.py", line 913, in load_extension
    await lib.setup(self)
  File "/botsdata/1.Martine/cogs/CogManager/cogs/mod/__init__.py", line 7, in setup
    await cog.initialize()
  File "/botsdata/1.Martine/cogs/CogManager/cogs/mod/mod.py", line 85, in initialize
    await self._maybe_update_config()
  File "/botsdata/1.Martine/cogs/CogManager/cogs/mod/mod.py", line 108, in _maybe_update_config
    prefix = re.sub(rf"<@!?{self.bot.user.id}>", f"@{self.bot.user.name}", prefixes[0])
```

redbot/cogs/mod/mod.py
```diff
@@ -77,6 +77,7 @@ def __init__(self, bot: Red):
         self.settings.register_user(**self.default_user_settings)
         self.cache: dict = {}
         self.tban_expiry_task = self.bot.loop.create_task(self.check_tempban_expirations())
+        self.init_task = self.bot.loop.create_task(self.initialize())
```
Jackenmen: This task should be made *after* the cog is added to the bot, otherwise the task might also be created when cog fails to add (e.g. some command name is already taken by other cog)

Just a simple method that you would call from `setup()` after `bot.add_cog(Mod())` should work here:
```py
def create_init_task(self):
    self._init_task = asyncio.create_task(self.initialize())
```

Jackenmen (changes requested):

redbot/cogs/mod/mod.py
PredaaA: Updated.

Jackenmen:
#3632 fixed this, thanks for reporting this issue and taking time to make this PR.

PR #3627, opened 2020-02-28
Handle users leaving the server while we're applying mutes

Kowlin:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Thanks @Sharkytheking. This PR fixes an issue when a member leaves the guild when a mute is being applied.

```
  File "/red-v32/lib/python3.8/site-packages/discord/ext/commands/core.py", line 79, in wrapped
  File "/red-v32/lib/python3.8/site-packages/redbot/cogs/mod/mutes.py", line 257, in guild_mute
    success, issue = await self.mute_user(guild, channel, author, user, audit_reason)
  File "/red-v32/lib/python3.8/site-packages/redbot/cogs/mod/mutes.py", line 422, in mute_user
    await channel.set_permissions(user, overwrite=overwrites, reason=reason)
  File "/red-v32/lib/python3.8/site-packages/discord/abc.py", line 602, in set_permissions
    await http.edit_channel_permissions(self.id, target.id, allow.value, deny.value, perm_type, reason=reason)
  File "/red-v32/lib/python3.8/site-packages/discord/http.py", line 220, in request
    raise NotFound(r, data)
discord.errors.NotFound: 404 NOT FOUND (error code: 10009): Unknown Overwrite

The above exception was the direct cause of the following exception:

  File "/red-v32/lib/python3.8/site-packages/discord/ext/commands/bot.py", line 863, in invoke
    await ctx.command.invoke(ctx)
  File "/red-v32/lib/python3.8/site-packages/redbot/core/commands/commands.py", line 631, in invoke
    await super().invoke(ctx)
  File "/red-v32/lib/python3.8/site-packages/discord/ext/commands/core.py", line 1158, in invoke
    await ctx.invoked_subcommand.invoke(ctx)
  File "/red-v32/lib/python3.8/site-packages/discord/ext/commands/core.py", line 728, in invoke
    await injected(*ctx.args, **ctx.kwargs)
  File "/red-v32/lib/python3.8/site-packages/discord/ext/commands/core.py", line 88, in wrapped
    raise CommandInvokeError(exc) from exc
discord.ext.commands.errors.CommandInvokeError: Command raised an exception: NotFound: 404 NOT FOUND (error code: 10009): Unknown Overwrite
```

### DIFF
--- a/redbot/cogs/mod/mutes.py
+++ b/redbot/cogs/mod/mutes.py
@@ -383,8 +383,8 @@ class MuteMixin(MixinMeta):
 
         unmute_success = []
         for channel in guild.channels:
-            success, message = await self.unmute_user(guild, channel, author, user, audit_reason)
-            unmute_success.append((success, message))
+            success, issue = await self.unmute_user(guild, channel, author, user, audit_reason)
+            unmute_success.append((success, issue))
             await asyncio.sleep(0.1)
 
             if success is False:  # Check the latest entry
--- a/redbot/cogs/mod/mutes.py
+++ b/redbot/cogs/mod/mutes.py
@@ -22,8 +22,7 @@ mute_unmute_issues = {
         "permission and the user I'm muting must be "
         "lower than myself in the role hierarchy."
     ),
-    "left_guild": _("The user has left the server while were applying an overwrite."),
-    "unknown_channel": _("The channel I tried to mute the user in isn't found."),
+    "left_guild": _("The user has left the server while we're applying an overwrite."),
 }
 _ = T_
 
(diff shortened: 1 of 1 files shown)


redbot/cogs/mod/mutes.py
```diff
@@ -422,6 +423,8 @@ class MuteMixin(MixinMeta):
             await channel.set_permissions(user, overwrite=overwrites, reason=reason)
         except discord.Forbidden:
             return False, _(mute_unmute_issues["permissions_issue"])
+        except discord.NotFound:
+            return False, _(mute_unmute_issues["left_guild"])
```
Bakersbakebread: Could this also be if the channel was deleted as well? 🤣

Kowlin: Added error handling for that issue.

redbot/cogs/mod/mutes.py
```diff
@@ -251,27 +253,36 @@ class MuteMixin(MixinMeta):
...
-                until=None,
-                channel=None,
-            )
-        except RuntimeError as e:
-            await ctx.send(e)
-        await ctx.send(_("User has been muted in this server."))
+
+            if success is False:  # Check the latest entry
+                guild_success = False
+                break
```
TrustyJAID: This will break if the bot is explicitly denied permissions to handle overwrites inside any channels when trying to mute a user rendering the command useless in a majority of properly setup permission servers. It would be better to handle this after attempting all channels and checking teh `mute_success` list for displaying where things went wrong. Or you may change the return from `mute_user` to be more than just a boolean return statement and then handle things based on severity of the error that way.

TrustyJAID (changes requested):

redbot/cogs/mod/mutes.py
Jackenmen: Just changing the return value of `mute_user` to `None` when bot gets Forbidden error should suffice, right?

Kowlin: Yes, it would.

TrustyJAID: There is 1 place where the return value is None but not where we want to explicitly stop the funciton from continuing.

TrustyJAID: But the other things to consider is that a server has bot in an admin channel at the top of the server channel list and the bot is explictly denied permission to manage_roles in that channel. In the current way this is setup the entire mute funciton will fail silently without providing the overwrites to subsequent channels which is expected.

Kowlin: You're right in that regard and I suppose I'll redo some stuff later.

TrustyJAID (dismissed):
Looks good to me.

TrustyJAID (approved):

Jackenmen (approved):

PR #3638, opened 2020-03-04
[Mod] Move deletedelay to core

### LINKED ISSUE #3636: [Mod/Core]Move deletedelay to core.
# Feature request

- [ ] Cog
- [X] Command
- [ ] API functionality

I think the ``[p]modset deletedelay`` would better fit to ``[p]set deletedelay`` as it's a setting concerning the Bot's general behaviour. I've found myself and others confused on multiple occasions when looking for the command.

TrustyJAID:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

When mod was refactored some things were put inside a file called movetocore.py suggesting these functions should be moved to core at some point. This completes that and resolves #3636.


redbot/core/events.py
```diff
@@ -192,6 +196,7 @@ def init_events(bot, cli_flags):
                 await ctx.send(embed=await format_fuzzy_results(ctx, fuzzy_commands, embed=True))
             else:
                 await ctx.send(await format_fuzzy_results(ctx, fuzzy_commands, embed=False))
+            await bot._delete_delay(ctx)
```
Jackenmen: `bot._delete_delay` should be called when the error **is not** `CommandNotFound` error - this does the opposite.

redbot/core/core_commands.py
```diff
@@ -903,6 +903,37 @@ class Core(commands.Cog, CoreLogic):
             for page in pagify(settings):
                 await ctx.send(box(page))
 
+    @checks.mod_or_permissions(administrator=True)
```
Jackenmen: This check is too light, it was `@checks.guildowner_or_permissions(administrator=True)` before, in my opinion it should either remain that or at least be `@checks.admin_or_permissions(administrator=True)`.

redbot/cogs/mod/mod.py
```diff
@@ -113,6 +111,16 @@ def cog_unload(self):
...
+        if await self.settings.version() < "1.2.0":
+            prefixes = await self.bot.get_valid_prefixes()
+            prefix = re.sub(rf"<@!?{self.bot.user.id}>", f"@{self.bot.user.name}", prefixes[0])
+            msg = _(
+                "Delete delay settings have been moved. "
+                "Please use `{prefix}movedeletedelay` if "
+                "you were previously using these functions."
+            ).format(prefix=prefix)
+            self.bot.loop.create_task(self.bot.send_to_owners(msg))
+            await self.settings.version.set(__version__)
```
Jackenmen: This code should now use helper function `send_to_owners_with_prefix_replaced` as the part above this fragment (it doesn't show in diff, but if you look at code on `V3/develop`, it will be there).

I also noticed that both this message and the message for `[p]moveignoredchannels` gets sent on fresh Red install, it would be great to avoid it.

Jackenmen (changes requested):

redbot/cogs/mod/mod.py
TrustyJAID: Unfortunately the best solution for removing this message on a fresh install is setting a depracation time on the message say at version 3.x we just remove the message hoping people who want to migrate settings from before this verision they have already.

Jackenmen: We could loop through all guild settings and only send the message if `delete_delay` has been changed from the default in at least one of them.

redbot/cogs/mod/mod.py
```diff
@@ -104,14 +102,20 @@ def cog_unload(self):
...
             msg = _(
                 "Ignored guilds and channels have been moved. "
-                "Please use `{prefix}moveignoredchannels` if "
+                "Please use `[p]moveignoredchannels` if "
                 "you were previously using these functions."
-            ).format(prefix=prefix)
-            self.bot.loop.create_task(self.bot.send_to_owners(msg))
+            )
+            self.bot.loop.create_task(self.bot.send_to_owners_with_prefix_replaced(msg))
+            await self.settings.version.set(__version__)
```
Jackenmen: This should be changed so that we don't preemptively set the version to latest
```suggestion
            await self.settings.version.set("1.1.0")
```

Jackenmen:
Looks like there are some conflicts to solve too.

redbot/cogs/mod/mod.py
```diff
@@ -111,8 +109,15 @@ def cog_unload(self):
                 "Please use `[p]moveignoredchannels` if "
                 "you were previously using these functions."
             )
-            self.bot.loop.create_task(send_to_owners_with_prefix_replaced(self.bot, msg))
-            await self.settings.version.set(__version__)
+            self.bot.loop.create_task(self.bot.send_to_owners_with_prefix_replaced(msg))
```
Jackenmen, line 112:
```suggestion
            self.bot.loop.create_task(self.bot.send_to_owners_with_prefix_replaced(msg))
            await self.settings.version.set("1.1.0")
```

TrustyJAID: Not necessary to write to config twice in this scenario, the save to config was moved to a single operation outside the if statements so it's set once and easier to make changes i nthe future.

Jackenmen: In case there is some error during any migration, we do not want the migrations before it to run another time once you try loading the cog again.

Jackenmen (approved):

fixator10:
Causes `AttributeError: 'Red' object has no attribute 'send_to_owners_with_prefix_replaced'` on load of Mod

Jackenmen:
Made a fix: #3683, thanks

PR #3643, opened 2020-03-08
Fix issues with the black/whitelist cache

### LINKED ISSUE #3642: [Core] Bug in blacklist add
# Command bugs

#### Command name

 ``[p]blacklist add``

Core I think.

 the command to work

#### What actually happened?

It didn't work.

This was brought up by another member in the support discord.(GraveyardEnthusiast#7958) After a bot restart the command worked again. However, this error makes no sense to me. https://pastebin.com/fbgYHHdu

In the code, a dict is accessed with the key "gid" if gid=None. Isn't this guaranteed to cause a KeyError? Maybe i'm overlooking something but this doesn't really make sense to me.

 I don't know.

#### Discord conversation:
Start: https://discordapp.com/channels/133049272517001216/387398816317440000/686123128656625711
End: https://discordapp.com/channels/133049272517001216/387398816317440000/686143352571297808

TrustyJAID:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes some issues with the black/whitelist cache and fixes #3642

### DIFF
--- a/redbot/core/settings_caches.py
+++ b/redbot/core/settings_caches.py
@@ -143,7 +143,7 @@ class WhitelistBlacklistManager:
     async def add_to_whitelist(self, guild: Optional[discord.Guild], role_or_user: List[int]):
         gid: Optional[int] = guild.id if guild else None
         role_or_user = role_or_user or []
-        if not isinstance(role_or_user, list) and not all(
+        if not isinstance(role_or_user, list) or not all(
             isinstance(r_or_u, int) for r_or_u in role_or_user
         ):
             raise TypeError("Whitelisted objects must be a list of ints")
@@ -176,7 +176,7 @@ class WhitelistBlacklistManager:
     async def remove_from_whitelist(self, guild: Optional[discord.Guild], role_or_user: List[int]):
         gid: Optional[int] = guild.id if guild else None
         role_or_user = role_or_user or []
-        if not isinstance(role_or_user, list) and not all(
+        if not isinstance(role_or_user, list) or not all(
             isinstance(r_or_u, int) for r_or_u in role_or_user
         ):
             raise TypeError("Whitelisted objects must be a list of ints")
@@ -220,7 +220,7 @@ class WhitelistBlacklistManager:
     async def add_to_blacklist(self, guild: Optional[discord.Guild], role_or_user: List[int]):
         gid: Optional[int] = guild.id if guild else None
         role_or_user = role_or_user or []
-        if not isinstance(role_or_user, list) and not all(
+        if not isinstance(role_or_user, list) or not all(
             isinstance(r_or_u, int) for r_or_u in role_or_user
         ):
             raise TypeError("Blacklisted objects must be a list of ints")
@@ -252,7 +252,7 @@ class WhitelistBlacklistManager:
     async def remove_from_blacklist(self, guild: Optional[discord.Guild], role_or_user: List[int]):
         gid: Optional[int] = guild.id if guild else None
         role_or_user = role_or_user or []
-        if not isinstance(role_or_user, list) and not all(
+        if not isinstance(role_or_user, list) or not all(
             isinstance(r_or_u, int) for r_or_u in role_or_user
         ):
             raise TypeError("Blacklisted objects must be a list of ints")
--- a/redbot/core/settings_caches.py
+++ b/redbot/core/settings_caches.py
@@ -130,7 +130,7 @@ class WhitelistBlacklistManager:
             ret = self._cached_whitelist[gid].copy()
         else:
             if gid is not None:
-                ret = await self._config.guild_from_id(gid).whitelsit()
+                ret = await self._config.guild_from_id(gid).whitelist()
                 if not ret:
                     ret = []
             else:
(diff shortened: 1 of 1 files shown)


redbot/core/settings_caches.py
```diff
@@ -147,22 +147,24 @@ def __init__(self, config: Config):
             isinstance(r_or_u, str) for r_or_u in role_or_user
```
Jackenmen: This `if` should look like this unless I misunderstood its intent:
```py
if not (
    isinstance(role_or_user, list) and all(
        isinstance(r_or_u, int) for r_or_u in role_or_user
    )
):
```

redbot/core/settings_caches.py
```diff
@@ -175,15 +177,17 @@ def __init__(self, config: Config):
...
```
Jackenmen: This `if` should look like this unless I misunderstood its intent:
```py
if not (
    isinstance(role_or_user, list) and all(
        isinstance(r_or_u, int) for r_or_u in role_or_user
    )
):
```

redbot/core/settings_caches.py
```diff
@@ -214,15 +218,17 @@ def __init__(self, config: Config):
...
```
Jackenmen: This `if` should look like this unless I misunderstood its intent:
```py
if not (
    isinstance(role_or_user, list) and all(
        isinstance(r_or_u, int) for r_or_u in role_or_user
    )
):
```

redbot/core/settings_caches.py
```diff
@@ -242,15 +248,17 @@ def __init__(self, config: Config):
...
```
Jackenmen: This `if` should look like this unless I misunderstood its intent:
```py
if not (
    isinstance(role_or_user, list) and all(
        isinstance(r_or_u, int) for r_or_u in role_or_user
    )
):
```

redbot/core/settings_caches.py
```diff
@@ -242,15 +248,17 @@ def __init__(self, config: Config):
...
         ):
             raise TypeError("Blacklisted objects must be a list of ints")
+        if gid not in self._cached_blacklist:
+            self._cached_blacklist[gid] = []
```
Jackenmen: This `if` may result in adding empty list to the cache when in fact there is something in current blacklist which just wasn't cached just yet.

redbot/core/settings_caches.py
```diff
@@ -214,15 +218,17 @@ def __init__(self, config: Config):
...
```
Jackenmen: This `if` may result in adding empty list to the cache when in fact there is something in current blacklist which just wasn't cached just yet.

redbot/core/settings_caches.py
```diff
@@ -147,22 +147,24 @@ def __init__(self, config: Config):
             isinstance(r_or_u, str) for r_or_u in role_or_user
         ):
             raise TypeError("Whitelisted objects must be a list of ints")
+        if gid not in self._cached_whitelist:
+            self._cached_whitelist[gid] = []
```
Jackenmen: This `if` may result in adding empty list to the cache when in fact there is something in current whitelist which just wasn't cached just yet.

redbot/core/settings_caches.py
```diff
@@ -175,15 +177,17 @@ def __init__(self, config: Config):
...
```
Jackenmen: This `if` may result in adding empty list to the cache when in fact there is something in current whitelist which just wasn't cached just yet.

Jackenmen (changes requested):
Changes that should be made which GitHub didn't let me put comment on:

- [line 238](https://github.com/Cog-Creators/Red-DiscordBot/pull/3643/files#diff-2201ff62582a7905f645d5fbae75cd68R238)
  Suggested change:
  `self._cached_whitelist[gid] = []`

- [line 204](https://github.com/Cog-Creators/Red-DiscordBot/pull/3643/files#diff-2201ff62582a7905f645d5fbae75cd68R204)
  Suggested change:
  `ret = await self._config.guild_from_id(gid).whitelist()`

redbot/core/settings_caches.py
```diff
@@ -210,26 +217,30 @@ def __init__(self, config: Config):
...
-                if obj_id not in self._cached_blacklist:
+                if obj_id not in self._cached_blacklist[gid]:
                     self._cached_blacklist[gid].append(obj_id)
                     async with self._config.guild_from_id(gid).blacklist() as curr_list:
                         curr_list.append(obj_id)
 
     async def clear_blacklist(self, guild: Optional[discord.Guild] = None):
         gid: Optional[int] = guild.id if guild else None
-        self._cached_whitelist[gid]
+        self._cached_blacklist[gid]
```
Jackenmen: I missed this on first review:
```suggestion
        self._cached_blacklist[gid] = []
```

redbot/core/settings_caches.py
```diff
@@ -143,26 +143,31 @@ def __init__(self, config: Config):
     async def add_to_whitelist(self, guild: Optional[discord.Guild], role_or_user: List[int]):
         gid: Optional[int] = guild.id if guild else None
         role_or_user = role_or_user or []
-        if not isinstance(role_or_user, list) and not all(
-            isinstance(r_or_u, str) for r_or_u in role_or_user
+        if not isinstance(role_or_user, list) and all(
+            isinstance(r_or_u, int) for r_or_u in role_or_user
```
Jackenmen: This `if` is still not correct - it only triggers if `role_or_user` is not a list **and** all items in the list (and first part just checked that it's not a list) are integers

It should either be:
```py
if not (
    isinstance(role_or_user, list) and all(
        isinstance(r_or_u, int) for r_or_u in role_or_user
    )
):
```
or if you think it's clearer when using `or` then:
```py
if not isinstance(role_or_user, list) or not all(
        isinstance(r_or_u, int) for r_or_u in role_or_user
):
```
Same thing applies to other 3 `if`s that I previously commented on.

redbot/core/settings_caches.py
Jackenmen: This hasn't been resolved in this function.

redbot/core/settings_caches.py
Jackenmen: This hasn't been resolved in this function.

redbot/core/settings_caches.py
```diff
@@ -238,19 +249,21 @@ def __init__(self, config: Config):
...
             for obj_id in role_or_user:
-                if obj_id in self._cached_blacklist:
+                if obj_id in self._cached_blacklist[gid]:
                     self._cached_blacklist[gid].remove(obj_id)
                     async with self._config.blacklist() as curr_list:
                         curr_list.remove(obj_id)
         else:
             for obj_id in role_or_user:
-                if obj_id not in self._cached_blacklist:
+                if obj_id not in self._cached_blacklist[gid]:
```
Jackenmen: I missed this on first review:
```suggestion
                if obj_id in self._cached_blacklist[gid]:
```

Jackenmen (changes requested):

Jackenmen (approved):
Looks good to me, thanks!

PR #3646, opened 2020-03-10
[Core] Unify commands that set bot's activity status

### LINKED ISSUE #3590: [Core] Set game command not intuitive.
# Feature request

- [X] Command

To set a "playing" status you currently would run the ``[p]set game`` command. While I understand the rationale behind the name, I think changing it to ``[p]set playing`` would be more intuitive especially since listening and watching are set using a command name that matches the result text.

Dav-Git:
### Type

- [X] Enhancement

I aliased ``[p]set game`` with ``[p]set playing`` to resolve #3590 . Not sure if adding an alias or renaming it outright is the way to go. This way users that are used to ``[p]set game`` can keep using it while people who intuitively type ``[p]set playing`` will also get the expected result.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1073,7 +1073,7 @@ class Core(commands.Cog, CoreLogic):
         else:
             await ctx.send(_("Done."))
 
-    @_set.command(name="game", aliases=["playing"])
+    @_set.command(name="playing", aliases=["game"])
     @checks.bot_in_a_guild()
     @checks.is_owner()
     async def _game(self, ctx: commands.Context, *, game: str = None):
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -1073,7 +1073,7 @@ class Core(commands.Cog, CoreLogic):
         else:
             await ctx.send(_("Done."))
 
-    @_set.command(name="game")
+    @_set.command(name="game", aliases=["playing"])
```
Jackenmen: Other commands use `-ing` as their main name, I think it would be good to do it here as well.
```suggestion
    @_set.command(name="playing", aliases=["game"])
```

Jackenmen (changes requested):
I think it would be great to unify the names we use for commands that set activity status, could you also edit `[p]set stream` to `[p]set streaming` and add `[p]set stream` as alias? That way, all command names for setting activity status will be unified and therefore more intuitive.

Jackenmen (approved):

Dav-Git:
Could also rename the func for ``[p]set stream`` went with name= instead for consistency

PR #3667, opened 2020-03-17
[Image] Add `[count]` argument to `[p]imgur` commands

### LINKED ISSUE #3044: imgur search Posts Three Images into Chat

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Adds optional `count` argument that allows user to specify how many images they want to get.
By default, `[p]imgur search` command will send one image now.
Fixes #3044

### DIFF
--- a/redbot/cogs/image/image.py
+++ b/redbot/cogs/image/image.py
@@ -49,7 +49,7 @@ class Image(commands.Cog):
         Use `count` to choose how many images should be returned.
         Command can return up to 5 images.
         """
-        if 1 <= amount <= 5:
+        if 1 <= count <= 5:
             await ctx.send(_("Image count has to be between 1 and 5."))
             return
         url = self.imgur_base_url + "gallery/search/time/all/0"
--- a/redbot/cogs/image/image.py
+++ b/redbot/cogs/image/image.py
@@ -1,4 +1,5 @@
 from random import shuffle
+from typing import Optional
 
 import aiohttp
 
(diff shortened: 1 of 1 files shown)


redbot/cogs/image/image.py
```diff
@@ -42,11 +43,15 @@ def cog_unload(self):
...
     @_imgur.command(name="search")
-    async def imgur_search(self, ctx, *, term: str):
+    async def imgur_search(self, ctx, count: Optional[int] = 1, *, term: str):
         """Search Imgur for the specified term.
 
-        Returns up to 3 results.
+        Use `count` to choose how many images should be returned.
+        Command can return up to 5 images.
         """
+        if 1 <= count <= 5:
```
Flame442: This equality is backwards.

Flame442 (changes requested):

Jackenmen:
Now that I looked at this again... I should probably also do the same change to imgur subreddit

Flame442 (approved):

PR #3674, opened 2020-03-20
[ModLog] Keep last known username in case's data

### LINKED ISSUE #3443: No username in modlog entries
# Command bugs

#### Command name

hackban, reason

Mod, ModLog

ModLog entry should show a nickname and/or username.

#### What actually happened?

For users that aren't in the server, the ModLog entry shows no information indicating what user the entry is for.

Moderate or update a ModLog entry for any user not in the server or not in the bot's cache.

modlog

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Fixes #3443 
This add a `Case.last_known_username` attribute in ModLog so that ModLog doesn't edit the recorded username in case's message with "[Unknown or Deleted User]" after editing the case. ModLog will continue on with not using `bot.fetch_user()`, but if ModLog has recorded last known username then it will be used instead of "[Unknown or Deleted User]"

Better attribute name is welcome :P

### DIFF
--- a/redbot/core/modlog.py
+++ b/redbot/core/modlog.py
@@ -214,12 +214,14 @@ class Case:
         amended_by: Optional[discord.User] = None,
         modified_at: Optional[int] = None,
         message: Optional[discord.Message] = None,
+        last_known_username: Optional[str] = None,
     ):
         self.bot = bot
         self.guild = guild
         self.created_at = created_at
         self.action_type = action_type
         self.user = user
+        self.last_known_username = last_known_username
         self.moderator = moderator
         self.reason = reason
         self.until = until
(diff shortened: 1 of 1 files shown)


Drapersniper:
Wouldn’t it be better using the user ID instead of storing usernames? Due to discord potentially sending data deletion requests, the bot owner would have 7 days to delete PII and User generates data which usernames fall under, iirc bank leaderboard already does this

Jackenmen:
> Wouldn’t it be better using the user ID instead of storing usernames?

We already store user IDs, this PR is precisely for storing last known username - read the issue linked in PR's description.

Drapersniper:
YEah I know, but my point still apply since usernames are User generated data, which fall under what has to be deleted within 7 days if discord sends the notification to delete a users data

Jackenmen:
As said in the linked issue, it has already been discussed internally in Discord and this had been agreed on (you were taking part in that discussion)

And as far as PII goes, what we actually do is store a part of discord message that was sent by the bot to the modlog channel which is required by the bot to function properly.

Kowlin (approved):

PR #3675, opened 2020-03-20
[Docs] Fix Mac instructions in the install guide

### LINKED ISSUE #3436: [Docs] Installing Python on Mac
# Other bugs

Tried to install to install Python 3.8 on Mac.

2 things :
 - The argument `--with-brewed-openssl` could be recognized
 - Homebrew could install Python 3.8

#### What actually happened?

The argument `--with-brewed-openssl` isn't necessary anymore as I saw on [the following issue](https://github.com/Homebrew/homebrew-core/issues/8974)

Using the given command will also install Python 3.7 and not the 3.8 one, and need to be remplaced by the command : `brew install python@3.8`, according to Red 3.2.

Install Python as said on the documentation.

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature
- [x] Docs!

Fixes #3436 

@aikaterna hoping that you could verify this since you actually have Mac :)

Link to rendered version of docs: https://red-discordbot--3675.org.readthedocs.build/en/3675/install_linux_mac.html#mac

### DIFF
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -171,6 +171,8 @@ one-by-one:
 .. code-block:: none
 
     brew install python@3.8
+    echo 'export PATH="/usr/local/opt/python@3.8/bin:$PATH"' >> ~/.profile
+    source ~/.profile
     brew install git
     brew cask install adoptopenjdk/openjdk/adoptopenjdk11
 
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -172,12 +172,7 @@ one-by-one:
 
     brew install python@3.8
     brew install git
-    brew cask install AdoptOpenJDK/openjdk/adoptopenjdk11
-
-.. attention::
-
-    It's possible you will have network issues. If so, go in your Applications folder, inside it, go in
-    the Python 3.8 folder then double click ``Install certificates.command``.
+    brew cask install adoptopenjdk/openjdk/adoptopenjdk11
 
 Continue by `creating-venv-linux`.
 
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -163,7 +163,7 @@ following, then press Enter:
 
 .. code-block:: none
 
-    /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+    /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 
 After the installation, install the required packages by pasting the commands and pressing enter,
 one-by-one:
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -170,13 +170,14 @@ one-by-one:
 
 .. code-block:: none
 
-    brew install python --with-brewed-openssl
+    brew install python@3.8
     brew install git
-    brew tap caskroom/versions
-    brew cask install homebrew/cask-versions/adoptopenjdk11
+    brew cask install AdoptOpenJDK/openjdk/adoptopenjdk11
 
-It's possible you will have network issues. If so, go in your Applications folder, inside it, go in
-the Python 3.8 folder then double click ``Install certificates.command``.
+.. attention::
+
+    It's possible you will have network issues. If so, go in your Applications folder, inside it, go in
+    the Python 3.8 folder then double click ``Install certificates.command``.
 
 Continue by `creating-venv-linux`.
 


docs/install_linux_mac.rst
```diff
@@ -170,13 +170,14 @@ one-by-one:
 
 .. code-block:: none
 
-    brew install python --with-brewed-openssl
+    brew install python@3.8
     brew install git
-    brew tap caskroom/versions
-    brew cask install homebrew/cask-versions/adoptopenjdk11
+    brew cask install AdoptOpenJDK/openjdk/adoptopenjdk11
```
aikaterna: This should be all lowercase.
`brew cask install adoptopenjdk/openjdk/adoptopenjdk11`

docs/install_linux_mac.rst
```diff
@@ -170,13 +170,14 @@ one-by-one:
...
 
-It's possible you will have network issues. If so, go in your Applications folder, inside it, go in
-the Python 3.8 folder then double click ``Install certificates.command``.
+.. attention::
```
aikaterna: If they install through brew, it is installed with —-with-openssl with an openssl package that was built for brew, so there isn’t a need for the Install certificates.command note - they will only have that file if they have installed through the Python installer application.

aikaterna:
I am in the process of testing the brew install command on various Macs. Homebrew has an open issue to migrate formulae to Python 3.8 but it is in progress - the default install on brew currently is Python 3.7.7. Python@3.8 is keg-only which means we will need to tell the system to use it as it is not symlinked into /usr/local. I will provide some suggested install instructions following this, after I confirm some more conditions/install situations.

aikaterna:
The Homebrew install link also should be updated. Using what we have on the guide:
`/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`
Leads to the response:
```
Warning: The Ruby Homebrew installer is now deprecated and has been rewritten in
Bash. Please migrate to the following command:
  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```
This new link provided works as expected.

aikaterna:
These instructions are after installing python@3.8 through brew with `brew install python@3.8`.

ADD TO PATH

* Add Python 3.8 to PATH in the Terminal window:

Check if your terminal is using zsh or bash. This will be at the top of the Terminal window in the title, and will show your Mac user name and then the terminal type.

Mac OS 10.15.x if using zsh:
`echo 'export PATH="/usr/local/opt/python@3.8/bin:$PATH"' >> ~/.zshrc`

Other Mac OS versions or if still using bash on 10.15.x:
`echo 'export PATH="/usr/local/opt/python@3.8/bin:$PATH"' >> ~/.bash_profile`


LINK PYTHON THROUGH BREW

* If you have a previously installed version of Python through brew, or if you are upgrading from an earlier version of Red:

`brew list | grep python` will show which Python versions are installed through brew.

For example, it may reply with a list of versions that are installed:

```
python
python3
python@2
python@3.8
```

Use brew unlink to unlink other Python versions if they are present:

`brew unlink python`
`brew unlink python3`
`brew unlink python@2`

Then link the new Python version:

`brew link python@3.8`


* If this is a brand new OS, or a new Red installation where Red has never been installed before:

Check `brew list | grep python` first to make sure. This command should show no Python installs besides `python@3.8`. If there are other versions listed, use the instructions above in regards to unlinking them.

Link the new Python:

`brew link python@3.8`

If it replies with: `Warning: python@3.8 is keg-only and must be linked with --force`, then use:

`brew link python@3.8 --force`

It should reply that it is creating the symlinks for 3.8 (`Linking /usr/local/Cellar/python@3.8/3.8.2... x symlinks created`).


VERIFY THE PYTHON INSTALL

To test the Python install, type:

`python3.8 --version`

The terminal should respond with Python 3.8.2 or greater.

Drapersniper:
Just an outsiders view:

Instead of editing the bash_profile/zshrc, we may want to edit the .profile, this should simplify the instructions and the .profile should work the same across all MacOS versions.

And doing so reduces the risk of users accidentally messing up their bash_profile/zshrc, as they will be editing only the .profile which is often empty on Mac,

aikaterna:
Also to note, perhaps in those instructions above, if they try to `brew link` python without the bindings being freed though `brew unlink`, the following can occur:
image
We would want to caution users against the --overwrite flag and encourage unlinking if possible (this warning also shows what owns the link, in this case, `python`).

aikaterna (approved):
Turns out all the fun and games with brew link aren't needed.

PR #3679, opened 2020-03-21
Reject package (extension) names that can't be valid Python identifiers

### LINKED ISSUE #3605: Red should only allow valid package names for cog packages
Valid package name should contain only ascii numbers, letters, and underscores and it shouldn't start with a number. And it also shouldn't be a reserved keyword.
To check for this, we could do something like:
```py
package_name = "pkg_name"
if not package_name.isidentifier():
    # can't be identifier, so it's not a valid package name
# ↓ is needed cause `isidentifier()` doesn't check for keywords
import keyword
if keyword.iskeyword(package_name):
    # it's a reserved keyword, so it can't be identifier
    # and therefore isn't a valid package name
```
This check should be performed in CogManager and Downloader in the places where we get available modules (calls to `pkgutil.iter_modules`).
I know for a fact that Downloader already makes assumption that package name is a valid one and one of issues that users may encounter is `[p]cog update` thinking the cog is already up-to-date even though it isn't. 

This is a breaking change (so it's gonna happen in 3.4.0) and while nobody should be using anything else in package name already (not a single cog in approved repositories does for sure), this was naming that Red somewhat supported and it is possible it will break somebody if they don't know about rules for package names in Python.

Jackenmen:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

Resolves #3605

### DIFF
--- a/redbot/cogs/downloader/repo_manager.py
+++ b/redbot/cogs/downloader/repo_manager.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import asyncio
 import functools
+import keyword
 import os
 import pkgutil
 import shlex
@@ -495,6 +496,9 @@ class Repo(RepoJSONMixin):
                     )
         """
         for file_finder, name, is_pkg in pkgutil.iter_modules(path=[str(self.folder_path)]):
+            if not name.isidentifier() or keyword.iskeyword(name):
+                # reject package names that can't be valid python identifiers
+                continue
             if is_pkg:
                 curr_modules.append(
                     Installable(location=self.folder_path / name, repo=self, commit=self.commit)
--- a/redbot/core/cog_manager.py
+++ b/redbot/core/cog_manager.py
@@ -1,4 +1,5 @@
 import contextlib
+import keyword
 import pkgutil
 from importlib import import_module, invalidate_caches
 from importlib.machinery import ModuleSpec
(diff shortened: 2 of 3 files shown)


Jackenmen:
This PR can now be reviewed, it's only labeled as Blocked to prevent an accidental merges before we're actually ready for 3.4 release.

mikeshardmind:
Seems fine, but I'd reword this to us the appropriate d.py term (extensions, not cogs)

Drapersniper (changes requested):
Everything is looking good however
If you had an invalid cog loaded before this change you will see the following, Can you please handle this error and log a `log.error` message saying that we are unable to load the cog as it is not a valid package name (No traceback please as this traceback is super misleading for this case).

```
[2020-07-14 11:02:22] [ERROR] red: Failed to load package zippe$r
  File "C:\Users\guyre\.virtualenvs\red\lib\site-packages\redbot\core\bot.py", line 660, in pre_flight
    await asyncio.wait_for(self.load_extension(spec), 30)
  File "C:\Program Files\Python38\lib\asyncio\tasks.py", line 483, in wait_for
    return fut.result()
  File "C:\Users\guyre\.virtualenvs\red\lib\site-packages\redbot\core\bot.py", line 925, in load_extension
    name = spec.name.split(".")[-1]
AttributeError: 'NoneType' object has no attribute 'name'
```

Drapersniper (approved):
LGTM: #4079 was merged which will now provide a more friendly error ...
However i think we should do another error specific to this PR where the identifier is invalid.

 But won't be not approving this PR due to that nitpick.

PR #3688, opened 2020-03-24
[Cleanup] Add `[p]cleanup spam` command

idle-user:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

An addition to the cleanup cog. Checks the last 50 messages from the current channel, stores the message along with a reference to the author, and compares them. Deletes the duplicate/spam messages, but keeps a single copy.

image

image

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -549,7 +549,7 @@ class Cleanup(commands.Cog):
         """Deletes duplicate messages in the channel from the last X messages. Default 50."""
         msgs = []
         spam = []
-        async for msg in ctx.channel.history(limit=number + 1):
+        async for msg in ctx.channel.history(limit=number, before=ctx.message):
             c = "{0.author.id}-{0.content}".format(msg)
             if c in msgs:
                 spam.append(msg)
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -546,10 +546,10 @@ class Cleanup(commands.Cog):
     @commands.guild_only()
     @commands.bot_has_permissions(manage_messages=True)
     async def cleanup_spam(self, ctx: commands.Context, number: int = 50):
-        """Deletes duplicate messages in the channel from the last X messages, but keeps the original. Default 50."""
+        """Deletes duplicate messages in the channel from the last X messages, but keeps only one. Default 50."""
         msgs = []
         spam = []
-        async for msg in ctx.channel.history(limit=number, before=ctx.message, oldest_first=True):
+        async for msg in ctx.channel.history(limit=number, before=ctx.message):
             if msg.attachments:
                 continue
             c = (msg.author.id, msg.content, [e.to_dict() for e in msg.embeds])
@@ -563,10 +563,13 @@ class Cleanup(commands.Cog):
             if not cont:
                 return
 
+        reason = "{}({}) deleted {} spam messages in channel {}.".format(
+            ctx.author.name,
+            ctx.author.id,
+            humanize_number(len(spam), override_locale="en_US"),
+            ctx.channel.name,
+        )
+        log.info(reason)
+
         spam.append(ctx.message)
         await mass_purge(spam, ctx.channel)
-        log.info(
-            "{0.author} deleted {1} spam messages in channel {0.channel}.".format(
-                ctx, len(spam) - 1
-            )
-        )
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -553,12 +553,14 @@ class Cleanup(commands.Cog):
 
         def check(m):
             if m.attachments:
-                False
+                return False
             c = (m.author.id, m.content, [e.to_dict() for e in m.embeds])
             if c in msgs:
                 spam.append(m)
+                return True
             else:
                 msgs.append(c)
+                return False
 
         to_delete = await self.get_messages_for_deletion(
             channel=ctx.channel, limit=number, check=check, before=ctx.message,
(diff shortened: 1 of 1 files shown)


MeatyChunks:
Can I suggest you keep the default messages at 50, but allow a number to be specified too. I think that'd be handy. Nice work otherwise!

idle-user:
Great idea. Ask, and you shall receive! Thank you.

Edit 1: Also fixed the counts to consider the message that sent the invoking command.
Edit 2: images, oops forgot them.
image
image

redbot/cogs/cleanup/cleanup.py
```diff
@@ -541,3 +541,25 @@ def check(m):
...
             await slow_deletion(to_delete)
+
+    @cleanup.command(name="spam")
+    @commands.guild_only()
+    @commands.bot_has_permissions(manage_messages=True)
+    async def cleanup_spam(self, ctx: commands.Context, number: int = 50):
+        """Deletes duplicate messages in the channel from the last X messages. Default 50."""
+        msgs = []
+        spam = []
+        async for msg in ctx.channel.history(limit=number + 1):
```
Jackenmen: Instead of using `number + 1` for the command to check 50 messages excluding command message, it might be better to use `before` kwarg. This will also make sure that the bot will not include some message that happens to be sent right after the command message but before we the bot actually requests the channel's history.
```suggestion
        async for msg in ctx.channel.history(limit=number, before=ctx.message):
```

redbot/cogs/cleanup/cleanup.py
```diff
@@ -541,3 +541,25 @@ def check(m):
...
+            c = "{0.author.id}-{0.content}".format(msg)
```
Jackenmen: This isn't the safest way to remove spam messages - there can be messages that don't include any content (e.g. attachment-only messages or messages with embeds) and this would result in the bot removing those messages.
There's no good way to compare attachments so unfortunately the best we can do if we want this to be safe is ignore such messages completely. And as for embeds, we could compare them using `embed.to_dict()`:
```suggestion
            if msg.attachments:
                continue
            c = (msg.author.id, msg.content, [e.to_dict() for e in msg.embeds])
```

redbot/cogs/cleanup/cleanup.py
```diff
@@ -541,3 +541,25 @@ def check(m):
             await mass_purge(to_delete, channel)
         else:
             await slow_deletion(to_delete)
+
+    @cleanup.command(name="spam")
+    @commands.guild_only()
+    @commands.bot_has_permissions(manage_messages=True)
+    async def cleanup_spam(self, ctx: commands.Context, number: int = 50):
+        """Deletes duplicate messages in the channel from the last X messages. Default 50."""
```
Jackenmen: It might not be obvious to the user using this command that it deletes duplicates but leaves one instance of the message, could you perhaps make it a bit more clear?

Jackenmen (changes requested):
Thanks for making this command and contributing it to Red, this is a very cool idea 🙂
I have few things (review comments) for you that I think could be improved.

It would also be great if this command (similarly to other commands) used `get_messages_for_deletion()` and `check_100_plus()` methods.
I know that `get_messages_for_deletion()` doesn't support `limit` kwarg which you need here but it shouldn't be hard to add support for that kwarg.
Small note here, while you don't have to use this function, I just think it might be better to reuse what it does rather than duplicate the checks it does in `[p]cleanup spam`'s code. Let me know if I missed something about it though.

If you have any questions, you can find me on [Red's discord](discord.gg/red) (Jackenmen#6607).

idle-user:
Thank you! I really appreciate your comments and made some changes.

For the `check_100_plus()`, I included it after checking the history because the message states it'll delete the `number` of messages when the actual number will be less than that.

~~I also set `oldest_first=True` for the history to keep the original message from the author. I've been thinking about the best way to handle it. *Keep the earliest or the latest?* I went this route. Let me know your opinions on this.~~ Nope. Doesn't work the way I thought.

Also, currently looking into how `get_messages_for_deletion()` would work with this.
Looking forward to more feedback. Thanks again!

redbot/cogs/cleanup/cleanup.py
```diff
@@ -541,3 +541,32 @@ def check(m):
...
+            cont = await self.check_100_plus(ctx, len(spam))
+            if not cont:
+                return
+
+        spam.append(ctx.message)
+        await mass_purge(spam, ctx.channel)
+        log.info(
+            "{0.author} deleted {1} spam messages in channel {0.channel}.".format(
+                ctx, len(spam) - 1
+            )
```
Jackenmen: Putting IDs of involved discord objects can be helpful to bot owners looking at the log so adding them in parentheses should work for that.
Also, I propose using %-formatting as it works better with various logging tooling. While we don't use any, at least currently, some bot users might be.
```suggestion
            "%s (%s) deleted %s spam messages in channel %s (%s).",
            ctx.author,
            ctx.author.id,
            len(spam) - 1,
            channel,
            channel.id,
```

Jackenmen (commented):
Looks good to me, but since you said you'll look into switching it to `get_messages_for_deletion()` let me know when you do, and do ask if there's anything you need to know.

Also I have one more review comment, though it's more of a nitpick.

redbot/cogs/cleanup/cleanup.py
idle-user: I actually just made that change! lol
I wanted to keep it with the same style

idle-user:
Okay, so I made the changes and fixed my all dumb mistakes I hope..

Jackenmen (approved):
Looks good to me, thank you so much for the contribution and I hope it won't be the last ;)

You can expect this feature in next Red release.

PR #3689, opened 2020-03-25
[Streams] Make preview picture for stream alerts bigger

### LINKED ISSUE #3685: [Streams] YouTube Alert Suggestion/Option
Since this alert is very bland and link to the livestream is already used in the preview, why not have the bot just post the link to the stream instead of the bot's own preview. Then you get a large image (like Twitch or just posting a video) which is better visually, plus a lot of info that is added in the description of the stream is also visible in the preview. And you can also press play to start the stream directly in Discord if you want to take a peek without going to YouTube.

This is a lot better than what is used now IMO.
Would be awesome to at least have this as an option to use if it's not changed completely.

Dav-Git:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Closes #3685

### DIFF
--- a/redbot/cogs/streams/streamtypes.py
+++ b/redbot/cogs/streams/streamtypes.py
@@ -153,7 +153,7 @@ class YoutubeStream(Stream):
         vid_data = data["items"][0]
         video_url = "https://youtube.com/watch?v={}".format(vid_data["id"])
         title = vid_data["snippet"]["title"]
-        thumbnail = vid_data["snippet"]["thumbnails"]["default"]["url"]
+        thumbnail = vid_data["snippet"]["thumbnails"]["medium"]["url"]
         channel_title = vid_data["snippet"]["channelTitle"]
         embed = discord.Embed(title=title, url=video_url)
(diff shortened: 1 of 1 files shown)


Drapersniper:
Hmm.. there’s a reason default was used here, iirc not every video will return the medium arg  so this change can leave it susceptible to a crash

Jackenmen:
@Drapersniper hmm, according to https://developers.google.com/youtube/v3/docs/videos#snippet.thumbnails.(key) it should be available for all videos (unlike `standard` and `maxres`) but if you are sure that's not true we can always use it like:
```
thumbnails.get("medium") or thumbnails.get("default")
```

Jackenmen (approved):
Looks good to me.

// According to YouTube Data API Reference, `default`, `medium` and `high` thumbnails are always available and other sources also confirm that so there's no reason to believe that it can be unavailable for any video. But if this will actually raise `KeyError` for some video in future, you can blame *me* for believing in YouTube's docs too much 😄

PR #3707, opened 2020-03-29
Fix bank check (used in Bank, Economy and Trivia cogs)

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This HAS NOT been tested.

Since we didn't notice that something is wrong with the modified check in #3672, please make sure twice that it's done right now.


Kowlin (dismissed):

Kowlin (approved):
Suprise suprise, it works just fine.

PR #3710, opened 2020-03-29
Send error in `[p]alias add` when target command doesn't exist

### LINKED ISSUE #3545: `[p]alias add` should warn (or error) when passed command name doesn't exist
# Command bugs

#### Command name

`[p]alias (global) add aliasname non-existent-command`

Alias

I expected the cog to error out (or just send warning) when non-existent command name is passed.

#### What actually happened?

Alias was created and there was no notice about the target command not existing.

1. `[p]load alias`
2. `[p]alias add aliasname non-existent-command`
3. See that alias gets created successfully without any notice about command not existing.

---

This needs discussion whether we should error out or just send warning when non-existent command name is passed.

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Fixes #3545, this sends an error when target command doesn't exist same as commands in Permissions cog do when target command/cog for the rules doesn't exist.

### DIFF
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -279,6 +279,13 @@ class Alias(commands.Cog):
                 ).format(name=alias_name)
             )
             return
+
+        given_command_exists = self.bot.get_command(command) is not None
+        if not given_command_exists:
+            await ctx.send(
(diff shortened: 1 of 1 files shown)


Kowlin (approved):
LGTM

PR #3717, opened 2020-04-01
Fix `redbot.core.utils.common_filters.INVITE_URL_RE` regex

kennnyshiwa:
Found that the regex used previously could falsely catch strings such as discord member and discord gggggg which aren't invites and shouldn't be caught, testing with this regex string still catches all invite url variations while allowing these other previously caught strings to keep working

### Type

- [X ] Bugfix
- [ ] Enhancement
- [ ] New feature

Found that the regex used previously could falsely catch strings such as discord member and discord gggggg which aren't invites and shouldn't be caught, testing with this regex string still catches all invite url variations while allowing these other previously caught strings to keep working

### DIFF
--- a/redbot/core/utils/common_filters.py
+++ b/redbot/core/utils/common_filters.py
@@ -16,7 +16,7 @@ __all__ = [
 # regexes
 URL_RE = re.compile(r"(https?|s?ftp)://(\S+)", re.I)
 
-INVITE_URL_RE = re.compile(r"(discord.gg\b|discordapp.com\/invite|discord.me\b)(\S+)", re.I)
+INVITE_URL_RE = re.compile(r"(discord\.(?:gg|io|me|li)|discordapp\.com\/invite)\/(\S+)", re.I)
 
 MASS_MENTION_RE = re.compile(r"(@)(?=everyone|here)")  # This only matches the @ for sanitizing
 
--- a/redbot/core/utils/common_filters.py
+++ b/redbot/core/utils/common_filters.py
@@ -16,7 +16,7 @@ __all__ = [
 # regexes
 URL_RE = re.compile(r"(https?|s?ftp)://(\S+)", re.I)
 
-INVITE_URL_RE = re.compile(r"(discord.gg|discordapp.com/invite|discord.me)(\S+)", re.I)
+INVITE_URL_RE = re.compile(r"(discord.gg\b|discordapp.com\/invite|discord.me\b)(\S+)", re.I)
 
 MASS_MENTION_RE = re.compile(r"(@)(?=everyone|here)")  # This only matches the @ for sanitizing
 


redbot/core/utils/common_filters.py
```diff
@@ -16,7 +16,7 @@
 # regexes
 URL_RE = re.compile(r"(https?|s?ftp)://(\S+)", re.I)
 
-INVITE_URL_RE = re.compile(r"(discord.gg|discordapp.com/invite|discord.me)(\S+)", re.I)
+INVITE_URL_RE = re.compile(r"(discord.gg\b|discordapp.com\/invite|discord.me\b)(\S+)", re.I)
```
Jackenmen: Allowing any word boundary is still gonna result in false positives, imo you should just make the regex require the slash:
```suggestion
INVITE_URL_RE = re.compile(r"(discord.gg|discordapp.com\/invite|discord.me)\/(\S+)", re.I)
```

Jackenmen (changes requested):

redbot/core/utils/common_filters.py
kennnyshiwa: I had tried multiple variations and false positives i could try to make it catch something and couldn't get it to trigger on anything else besides valid links, though i did find out there are some other urls we should probably catch 

``(https?:\/\/)?(www\.)?(discord\.(gg|io|me|li)|discordapp\.com\/invite)\/.+[a-z]`` this catches some other url variations that people use though the https and www may not be needed, catches anything i could throw at it with no false positives with the things i saw before, let me know what you think, but i agree, the \b probably isn't the best, so we can do either the end \ or this other regex with the other url endings

Jackenmen: I personally don't think there's a need to add `(https?:\/\/)?(www\.)?` as the regex will trigger without it when using `INVITE_URL_RE.search(text)` anyway and these regexes are supposed to be used as filters (they're in `redbot.core.utils.common_filters` namespace) so that seems like the most likely way this regex will be used in. With that said I'm not opposed to it (so you can keep it), I just don't find it necessary for the use cases they're intended for (filtering).

The additions of discord.io and discord.li are appreciated, though it would be better if you put TLDs into non-capturing group (`(?:gg|io|me|li)`) so that there are always the same amount of captured groups. I would also appreciate if you would send me some working invite link that uses discord.io and discord.li as I wasn't able to get hands on one myself.

As for the last part of the proposed regex (`.+[a-z]`), I think you should leave it as it was before (`(\S+)`), because:
a) it's useful to have the invite code in a captured group
b) it is prone to false-positives like `discord.gg/ abc`
c) it may capture incomplete invite url when it doesn't end with a letter, e.g. `discord.gg/kPdts7`

Alternatively, you could make the regex more strict by acknowledging what characters are allowed for each service:
- `discord.gg`/`discordapp.com/invite` - A-Z, a-z, 0-9
- `discord.me` - A-Z, a-z, 0-9 and `-`
- `discord.io`/`discord.li` - unknown

This is only really possible if you can narrow down possible characters for discord.io/discord.li and I don't think there's much point in being strict when regex's intended usage is filtering so I myself would just prefer `(\S+)`

So in the end, the regex would be:
```suggestion
INVITE_URL_RE = re.compile(r"(discord\.(?:gg|io|me|li)|discordapp\.com\/invite)\/(\S+)", re.I)
```
or if you think we should keep the optional https/www so that the regex could possibly be used in some other situations (doing a match against full string perhaps?) then it would be (using non-capturing groups for `http(s)` and `www`):
```suggestion
INVITE_URL_RE = re.compile(r"(?:https?:\/\/)?(?:www\.)?(discord\.(?:gg|io|me|li)|discordapp\.com\/invite)\/(\S+)", re.I)
```

Jackenmen (changes requested):

kennnyshiwa:
So in looking at this i agreed with your point about not worrying about http/https and or www as if people want a stronger filter based on those they could use retrigger. For discord.io/discord.li they just allow you to create personalized invites, so you give them https://discord.gg/TEeXcDY and you can make discord.io/BB-8 and discord.li redirects to discord.io so they are both the same just different domains

Jackenmen (approved):
Looks okay now, thanks for the fix!

PR #3744, opened 2020-04-11
Add information on how to update to out of date message

Drapersniper:
If the bot is simply out of data and in a supported Python versions users will now see 
image
("Command Prompt" is shown for Windows, any other platform it is called "Terminal")

If the instance is out of date AND in a old Python version the following is the new message:
image

Users are also direct to the correct doct based on their OS.


redbot/core/utils/_internal_utils.py
```diff
@@ -24,6 +26,8 @@
 main_log = logging.getLogger("red")
 
 __all__ = ("safe_delete", "fuzzy_command_search", "format_fuzzy_results", "create_backup")
```
Jackenmen: Could you add the missing `send_to_owners_with_preprocessor`, `send_to_owners_with_prefix_replaced` and newly added `expected_version` to `__all__`?

redbot/core/events.py
```diff
@@ -97,10 +99,54 @@ async def on_ready():
...
+                        "run the following:\n\n"
+                    ).format(console=platform_console)
+                    extra_update += '```"{python}" -m pip install -U Red-DiscordBot```'.format(
+                        python=sys.executable
+                    )
+                    extra_update += _(
+                        "\n\nPlease be aware that the command above is a best guess;"
+                        "\nThis is based on your current OS, enviroment and Python version and "
+                        "there is a small change this could be incorrect."
+                        "\n**Should the above not work for you please check our docs at <{docs}> "
```
Flame442:
```suggestion
                        "\nIt is based on your current OS, environment, and Python version so "
                        "there is a small chance it could be incorrect."
                        "\n**Should the above not work for you, please check our docs at <{docs}> "
```

Flame442 (changes requested):

redbot/core/events.py
```diff
@@ -97,10 +98,58 @@ async def on_ready():
...
+                        else "https://docs.discord.red/en/stable/install_linux_mac.html",
+                    )
+
+                    outdated_red_message += extra_update
+                else:
+                    extra_update = _(
+                        "\n\nYou have Python `{py_version}` and this update "
+                        "requires `{req_py}`; you cannot simply run the update command.\n\n"
+                        "You will need to follow the install instructions in our docs (<{docs}>) "
+                        "to reinstall Python, Red and any other dependency that are out of date "
```
Flame442:
```suggestion
                        "to reinstall Python, Red and any other dependencies that are out of date "
```

redbot/core/events.py
```diff
@@ -97,10 +98,58 @@ async def on_ready():
...
+                        "only skipping the `redbot-setup` command. (This will not delete your data)"
+                        "\nIf you already have a venv you will need to delete it and recreate it."
```
Flame442:
```suggestion
                        "\nIf you already have a venv you will need to delete it and recreate it. "
```

redbot/core/events.py
```diff
@@ -97,10 +98,58 @@ async def on_ready():
...
+                        "there is a small chance it could be incorrect."
+                        "\n**Should the above not work for you, please check our docs at <{docs}> "
+                        "and/or visit our support server at <https://discord.gg/red>.**"
+                    ).format(
+                        docs="https://docs.discord.red/en/stable/install_windows.html"
+                        if platform.system() == "Windows"
+                        else "https://docs.discord.red/en/stable/install_linux_mac.html",
+                    )
+
+                    outdated_red_message += extra_update
```
Flame442: This (and the other line exactly like this below) can both be replaced by one of this at the bottom, unindented once.

Flame442 (changes requested):

Drapersniper:
I've updated it to use the new Update url and resolved conflict let me know if it needs anything else Flame

Flame442 (dismissed):
Seems to work fine.

Jackenmen:
Things that I think should still be addressed here:
- the "personalized" update command doesn't account for extras (e.g. `Red-DiscordBot[postgres]`)
- the disclaimer should be more emphasised, phrases like "there is a small chance" make the disclaimer seem not important, I think it would be better to:
    - put disclaimer **before** the command
    - bold it (not entirely sure about this)
    - word it differently, I'm proposing something with a tone like this:
    > While this command should work in most scenarios, we highly recommend you to read the update docs and make sure there is nothing else that needs to be done during the update.
- it might also be worth to indicate in some way (not sure how exactly) to what command in the docs, the "personalized" command corresponds to
- I don't think it really hurts to be left as is, but imo the string for "unexpected" python version doesn't need to be that specific and could just point to update docs for more information

redbot/core/events.py
```diff
@@ -116,19 +169,16 @@ async def on_ready():
...
+                    extra_update += _(
                         "\n\nYou have Python `{py_version}` and this update "
                         "requires `{req_py}`; you cannot simply run the update command.\n\n"
-                        "You will need to follow the install instructions in our docs (<{docs}>) "
-                        "to reinstall Python, Red and any other dependencies that are out of date "
-                        "only skipping the `redbot-setup` command. (This will not delete your data)"
-                        "\nIf you already have a venv you will need to delete it and recreate it. "
-                        "If you need help updating go to our "
+                        "You will need to follow the update instruction in our docs above, "
+                        "if you still need shelp updating after following the docs go to our "
```
Jackenmen:
```suggestion
                        "if you still need help updating after following the docs go to our "
```

redbot/core/events.py
```diff
@@ -116,19 +169,16 @@ async def on_ready():
...
-                    extra_update = _(
+                    extra_update += _(
                         "\n\nYou have Python `{py_version}` and this update "
                         "requires `{req_py}`; you cannot simply run the update command.\n\n"
-                        "You will need to follow the install instructions in our docs (<{docs}>) "
-                        "to reinstall Python, Red and any other dependencies that are out of date "
-                        "only skipping the `redbot-setup` command. (This will not delete your data)"
-                        "\nIf you already have a venv you will need to delete it and recreate it. "
-                        "If you need help updating go to our "
+                        "You will need to follow the update instruction in our docs above, "
```
Jackenmen:
```suggestion
                        "You will need to follow the update instructions in our docs above, "
```

redbot/core/events.py
```diff
@@ -99,10 +100,96 @@ async def on_ready():
...
+                        import more_itertools
+                    except ImportError:
+                        test = ""
+                    else:
+                        test = "test"
+
+                    if any(i for i in [test, postgres, style, docs]):
+                        package_extras = "[{}]"
+                        valid = [i for i in [test, postgres, style, docs] if i]
+                        package_extras.format(",".join(valid))
```
Jackenmen: I have a better way for you ;)
```suggestion
                    installed_extras = []
                    for extra, reqs in red_pkg._dep_map:
                        if extra is None:
                            continue
                        try:
                            pkg_resources.require(req.name for req in reqs)
                        except pkg_resources.DistributionNotFound:
                            pass
                        else:
                            installed_extras.append(extra)

                    if installed_extras:
                        package_extras = f"[{','.join(installed_extras)}]"
                    else:
                        package_extras = ""
```
Yes, this does use a private attribute of `pkg_resources.Distribution` but it's way cleaner and less prone to omission in future when we add more extras or change some of the dependencies in the extras.
Also, this is used in this function a little below already so there's not much against using here too if it can help.

redbot/core/utils/_internal_utils.py
```diff
@@ -282,3 +295,14 @@ async def preprocessor(bot: Red, destination: discord.abc.Messageable, content:
...
+
+def expected_version(current: str, expected: str) -> bool:
+    re_match = re.match(version_req_re, expected)
+    specifier, expected = re_match.group(1), re_match.group(2)
+    if specifier is None:
+        specifier = ">="
+    try:
+        return operator_lookup[specifier](LooseVersion(current), LooseVersion(expected))
+    except TypeError:  # Invalid Expected version
+        return False
```
Jackenmen: Now that I look at this again... It would be nice if this supported specifier sets, i.e. `">=3.6.2,<3.8"` - we have used such specifier sets in the past on `python_version` and we can't know if we won't use use them in future. We could use `pkg_resources` for that:
```suggestion
def expected_version(current: str, expected: str) -> bool:
    # `pkg_resources` needs a regular requirement string, so "x" serves as requirement's name here
    return current in pkg_resources.Requirement.parse(f"x{expected}")
```

Jackenmen (changes requested):

redbot/core/events.py
```diff
@@ -99,10 +100,96 @@ async def on_ready():
...
+                requires_python = data["info"]["requires_python"]
+                current_python = platform.python_version()
+                extra_update = _(
+                    "\n\nWhile the following should work in most scenarios as it is "
+                    "based on your current OS, environment, and Python version "
+                    "we highly recommend you to read the update docs at <{docs}> and "
+                    "make sure there is nothing else that "
+                    "needs to be done during the update. "
+                    "**Should the above not work for you,"
+                    " ensure you have read the update docs.**"
```
Jackenmen: I think this sentence doesn't fit into this string after it got rewritten. IMO, it could just be removed and rest of the string should be emphasised (or maybe only some part of it? not sure)

Drapersniper: I think just rewording it to say "should the below not work..." I'm not in favour of emboldening a whole paragraph

redbot/core/events.py
```diff
@@ -99,10 +100,67 @@ async def on_ready():
...
                 ).format(data["info"]["version"], red_version)
+                requires_python = data["info"]["requires_python"]
+                current_python = platform.python_version()
+                extra_update = _(
+                    "\n\nWhile the following should work in most scenarios as it is "
+                    "based on your current OS, environment, and Python version "
+                    "we highly recommend you to read the update docs at <{docs}> and "
+                    "make sure there is nothing else that "
+                    "needs to be done during the update. "
+                    "**Should the above not work for you,"
```
Drapersniper:
```suggestion
                    "**Should the information below not work for you,"
```

redbot/core/events.py
Jackenmen: I don't know, it seems kind of out of place, maybe we could just bold the part about docs?
I.e.
```
While the following should work in most scenarios as it is based on your current OS, environment, and Python version, **we highly recommend you to read the update docs at <{docs}> and make sure there is nothing else that needs to be done during the update.**
```

Drapersniper: Can you make a suggestion please on my phone

redbot/core/events.py
```diff
@@ -99,10 +100,67 @@ async def on_ready():
...
+                requires_python = data["info"]["requires_python"]
+                current_python = platform.python_version()
+                extra_update = _(
+                    "\n\nWhile the following should work in most scenarios as it is "
+                    "based on your current OS, environment, and Python version "
+                    "we highly recommend you to read the update docs at <{docs}> and "
+                    "make sure there is nothing else that "
+                    "needs to be done during the update. "
+                    "**Should the information below not work for you,"
+                    " ensure you have read the update docs.**"
```
Jackenmen: Here's the suggestion:
```suggestion
                    "\n\nWhile the following should work in most scenarios as it is "
                    "based on your current OS, environment, and Python version, "
                    "**we highly recommend you to read the update docs at <{docs}> and "
                    "make sure there is nothing else that "
                    "needs to be done during the update.**"
```

redbot/core/utils/_internal_utils.py
```diff
@@ -33,7 +36,17 @@
...
+    "safe_delete",
+    "fuzzy_command_search",
+    "format_fuzzy_results",
+    "create_backup",
+    "send_to_owners_with_preprocessor",
+    "send_to_owners_with_prefix_replaced",
+    "expected_version",
+)
+operator_lookup = {"<": op.lt, "<=": op.le, "==": op.eq, ">=": op.ge, ">": op.gt}
+version_req_re = re.compile(r"^(<=|<|>=|>|==)?(.*)$")
```
Jackenmen:
```suggestion
```

redbot/core/utils/_internal_utils.py
```diff
@@ -18,8 +19,10 @@
     List,
     Optional,
     Union,
+    Set,
     TYPE_CHECKING,
 )
+from distutils.version import LooseVersion
```
Jackenmen:
```suggestion
```

redbot/core/utils/_internal_utils.py
```diff
@@ -18,8 +19,10 @@
     List,
     Optional,
     Union,
+    Set,
```
Jackenmen:
```suggestion
```

redbot/core/utils/_internal_utils.py
```diff
@@ -8,6 +8,7 @@
 import re
 import shutil
 import tarfile
+import operator as op
```
Jackenmen:
```suggestion
```

redbot/core/utils/_internal_utils.py
```diff
@@ -18,8 +19,10 @@
     List,
     Optional,
     Union,
+    Set,
     TYPE_CHECKING,
 )
+from distutils.version import LooseVersion
 
 import discord
```
Jackenmen, line 24:
```suggestion
import discord
import pkg_resources
```

Jackenmen (changes requested):

redbot/core/events.py
```diff
@@ -99,10 +100,65 @@ async def on_ready():
...
+                extra_update = _(
+                    "\n\nWhile the following should work in most scenarios as it is "
+                    "based on your current OS, environment, and Python version, "
+                    "**we highly recommend you to read the update docs at <{docs}> and "
+                    "make sure there is nothing else that "
+                    "needs to be done during the update.**"
+                ).format(docs="https://docs.discord.red/en/stable/update_red.html",)
+                if expected_version(current_python, requires_python):
+                    installed_extras = []
+                    for extra, reqs in red_pkg._dep_map:
```
Jackenmen: My initial suggestion was incorrect.
```suggestion
                    for extra, reqs in red_pkg._dep_map.items():
```

redbot/core/events.py
```diff
@@ -99,10 +100,65 @@ async def on_ready():
...
+                        "You will need to follow the update instructions in our docs above, "
+                        "if you still need help updating after following the docs go to our "
+                        "#support channel in <https://discord.gg/red>"
+                    ).format(
+                        py_version=current_python,
+                        req_py=requires_python,
+                        docs="https://docs.discord.red/en/stable/install_windows.html"
+                        if platform.system() == "Windows"
+                        else "https://docs.discord.red/en/stable/install_linux_mac.html",
+                    )
```
Jackenmen: `{docs}` isn't in the string anymore so this is unneeded
```suggestion
                    ).format(py_version=current_python, req_py=requires_python)
```

redbot/core/events.py
```diff
@@ -99,10 +100,65 @@ async def on_ready():
...
                 )
-                outdated_red_message = (
+                outdated_red_message = _(
                     "Your Red instance is out of date! {} is the current "
                     "version, however you are using {}!"
                 ).format(data["info"]["version"], red_version)
+                requires_python = data["info"]["requires_python"]
+                current_python = platform.python_version()
+                extra_update = _(
+                    "\n\nWhile the following should work in most scenarios as it is "
```
Jackenmen:
```suggestion
                    "\n\nWhile the following command should work in most scenarios as it is "
```

Jackenmen (changes requested):
This is everything left from me, but I'll leave any possibly remaining wording/clarity issues to Flame

Jackenmen (approved):
Looks good to me, leaving this unmerged in case @Flame442 wants to say something.

Flame442 (commented):
Still looks fine to me

PR #3746, opened 2020-04-11
Fix ignored channels list in `[p]ignore`

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Currently ignored channels list is always showing as empty in `[p]ignore`, this fixes it - I decided to only show explicitly ignored channels in ignored channels list so it doesn't include those that are only ignored because of the category they're in. Let me know if that's not how it should be done.


TrustyJAID (approved):
Showing explicitly ignored channels in ignored channels list is a good idea, lets people know which channels are ignored alongside which categories they have ignored. They can easily see which channels are part of the category by looking at the server so listing category channels is not necessary. This looks good to me.

PR #3753, opened 2020-04-14
adds to setup.py

stix26:
Suggested an enhancement to setup.py.


Drapersniper:
Red Follows PEP 8.
https://www.python.org/dev/peps/pep-0008/#imports
So this PR will not get aproved.

Jackenmen:
As Draper said, imports should be in separate lines per PEP 8 recommendations.

PR #3754, opened 2020-04-14
[Drivers] Add Support for a ReJSON driver + a few goodies

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

so lets start with the "goodies"

1. Add `Config.Value.tog()` support for all drivers
```py
# Toggle global value "foo" to the opposite boolean of the currently set value.
await conf.foo.tog()
# Set the global value "foo" True
await conf.foo.tog(True)
```

2. Add `Config.Value.inc()` support for all drivers
```py
# Increment global value "foo" by 2
await conf.foo.inc(2)
```

3. Add the Redis Driver.

NOTE: THIS WILL NOT WORK with a simple Redis driver, You will need the ReJSON plugin enabled in the Redis server.
More instructions can be found in https://rejson.io/
If you want a simple Docker image with it already built and enable use `redislabs/redisjson2:latest` otherwise to do it manually follow https://oss.redislabs.com/redisjson/#building-and-loading-the-module

**WARNING**: __If you are doing this you need to ensure your redis server is persisting data, otherwise all your data will be wiped in a system restart.__

This is still "WIP" as in i need testers ... so far though its been faster than postgres and JSON.

This laos contains #3714 as i needed that fix to test the driver migration to and from JSON/ReJSON

.... PLEASE dont use it on prod (@Stonedestroyer this is for you)
any issues related to this just ping me in the main server.

# Setup
## Hosting the server with Docker

- Saved the following in a file called `docker-compose.yaml`

```yaml
version: "3.3"
services:
  redis:
    image: redislabs/redisjson2:latest
    ports:
      - 6379:6379
    command: ["redis-server", "/usr/local/etc/redis/redis.conf"]
    environment:
      - ALLOW_EMPTY_PASSWORD=yes
    volumes:
      - type: bind
        source:  <full path to redis storage>\data
        target: /data
      - type: bind
        source: <full path to redis storage>\redis.conf
        target: /usr/local/etc/redis/redis.conf
      
```

- Create a file: `<full path to redis storage>\redis.conf` and paste the content of this https://gist.github.com/Drapersniper/b119bf588c0890af75659373579568e2 config inside of it

- Start the reJSON server: `cd` into the folder where you saved the `compose-docker.yaml` and run `compose-docker up -d` 

## Migrate from JSON to Redis
- run : `redbot-setup convert <instance_name> redis`
- Follow the prompts to connecto to your redis server.

- Once migration is completed start the bot


PredaaA:
Currently testing it. I just found two issues:
- Docker compose example needs to be on version 3.3 instead of 3.7
- Conversion from JSON fails with that error `aioredis.errors.AuthError: ERR Client sent AUTH, but no password is set` if you don't set a password, it may needs to be `NONE` by default.

Drapersniper:
The second is by design it follow the same as Postgres connection if no password it required you to explicitly pass "NONE"

PR #3760, opened 2020-04-15
Add `.tog()` and `.inc()` to config value methods

Drapersniper:
Signed-off-by: Drapersniper <27962761+drapersniper@users.noreply.github.com>

### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

1. Add `Config.Value.tog()` support for all drivers
```py
await conf.foo.set(False)
# Toggle global value "foo" to the opposite boolean of the currently set value.
new_value = await conf.foo.tog() # Return True
new_value = await conf.foo.tog() # Return False

await conf.foo.set(True)
# Set the global value "foo" True
new_value = await conf.foo.tog(True) # Return True
new_value = await conf.foo.tog(False) # Return False
```

2. Add `Config.Value.inc()` support for all drivers
```py
await conf.foo.set(10)
# Increment global value "foo" by 2
new_value = await conf.foo.inc(2) # Returns 12

await conf.foo.set(10)
# Decreates global value "foo" by 2
new_value = await conf.foo.inc(-2) # Return 8
```

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -286,17 +286,17 @@ class Value:
         """
         if not isinstance(value, (int, float)):
             raise ValueError("The value is not a Integer or Float")
-
-        if not hasattr(self.driver, "inc"):
-            current = await self._get(default=...)
-            if not isinstance(current, (int, float)):
+        async with self.get_lock():
+            if not hasattr(self.driver, "inc"):
+                current = await self._get(default=...)
+                if not isinstance(current, (int, float)):
+                    raise ValueError("The stored value is not a Integer or Float")
+                await self.set(current + value)
+                return current + value
+            try:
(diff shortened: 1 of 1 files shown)


redbot/core/config.py
```diff
@@ -262,6 +263,86 @@ async def clear(self):
...
+                else:
+                    new_value = value
+                await self.set(new_value)
+                return new_value
+            try:
+                return await self.driver.toggle(
+                    self.identifier_data, value=value, default=self.default
+                )
+            except StoredTypeError:
+                raise ValueError("The stored value is not a Integer or Float")
```
Drapersniper:
```suggestion
                raise ValueError("The stored value is not a boolean")
```

PR #3765, opened 2020-04-18
Normalize names of attributes with Config instances

Drapersniper:
As the title says all this one does ensure the config attribute is always called `config` if public, or `_config` if private

### DIFF
--- a/docs/framework_config.rst
+++ b/docs/framework_config.rst
@@ -202,15 +202,15 @@ the built-in Economy credits::
 
     class Pets:
         def __init__(self):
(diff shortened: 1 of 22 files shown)


Jackenmen:
You missed `admin.conf` in `redbot/cogs/admin/converters.py` and you also didn't update the examples in docstrings in `redbot/core/config.py`

Jackenmen (approved):
LGTM, PR doesn't seem to be missing any place where we get or later use `Config`, I haven't found a single place where `Config` isn't assigned to `config` or `_config` variable, and I didn't find references to old attribute names either.
`pylint` run was successful.

PR #3764, opened 2020-04-18
Reduce calls to config in the `on_command_add` event

Drapersniper:
Signed-off-by: Drapersniper <27962761+drapersniper@users.noreply.github.com>

### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

PR #3766, opened 2020-04-18
Optimize config calls in few places

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This one depends on #3764 and #3765

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -195,8 +195,9 @@ class Downloader(commands.Cog):
             The modules to check off.
 
         """
-        installed_cogs = await self.config.installed_cogs()
-        installed_libraries = await self.config.installed_libraries()
+        core_config = await self.config.all()
+        installed_cogs = core_config["installed_cogs"]
+        installed_libraries = core_config["installed_libraries"]
         for module in modules:
             if module.type == InstallableType.COG:
                 installed = installed_cogs
@@ -207,9 +208,9 @@ class Downloader(commands.Cog):
             module_json = module.to_json()
             repo_json = installed.setdefault(module.repo_name, {})
             repo_json[module.name] = module_json
-
-        await self.config.installed_cogs.set(installed_cogs)
-        await self.config.installed_libraries.set(installed_libraries)
+        async with self.config.all() as config_data:
+            config_data["installed_cogs"] = installed_cogs
+            config_data["installed_libraries"] = installed_libraries
 
     async def _remove_from_installed(self, modules: Iterable[InstalledModule]) -> None:
         """Remove modules from the saved list
@@ -221,8 +222,9 @@ class Downloader(commands.Cog):
             The modules to remove.
 
         """
-        installed_cogs = await self.config.installed_cogs()
-        installed_libraries = await self.config.installed_libraries()
+        core_config = await self.config.all()
+        installed_cogs = core_config["installed_cogs"]
+        installed_libraries = core_config["installed_libraries"]
         for module in modules:
             if module.type == InstallableType.COG:
                 installed = installed_cogs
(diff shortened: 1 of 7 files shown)


redbot/cogs/downloader/downloader.py
```diff
@@ -195,8 +195,9 @@ async def _save_to_installed(self, modules: Iterable[InstalledModule]) -> None:
             The modules to check off.
 
         """
-        installed_cogs = await self.conf.installed_cogs()
-        installed_libraries = await self.conf.installed_libraries()
+        core_config = await self.config.all()
```
Jackenmen: This variable name is confusing - this is Downloader's config, not bot's core config.
You should probably just use `config_data` as you do below for `async with` calls or better yet - `global_data` so that it also indicates scope (in that case, you should also use `global_data` in all other cases where `config_data` is used)

redbot/cogs/downloader/downloader.py
```diff
@@ -221,8 +222,9 @@ async def _remove_from_installed(self, modules: Iterable[InstalledModule]) -> No
             The modules to remove.
 
         """
-        installed_cogs = await self.conf.installed_cogs()
-        installed_libraries = await self.conf.installed_libraries()
+        core_config = await self.config.all()
```
Jackenmen: This variable name is confusing - this is Downloader's config, not bot's core config.
You should probably just use `config_data` as you do below for `async with` calls or better yet - `global_data` so that it also indicates scope (in that case, you should also use `global_data` in all other cases where `config_data` is used)

redbot/cogs/filter/filter.py
```diff
@@ -367,19 +371,21 @@ async def filter_hits(
     async def check_filter(self, message: discord.Message):
         server = message.guild
```
Jackenmen: I know this isn't part of PR, but could you use `guild` here (and instead of `server_data`)?

redbot/cogs/mod/kickban.py
```diff
@@ -409,10 +409,9 @@ async def tempban(
             invite = ""
 
         queue_entry = (guild.id, user.id)
-        await self.settings.member(user).banned_until.set(unban_time.timestamp())
-        cur_tbans = await self.settings.guild(guild).current_tempbans()
-        cur_tbans.append(user.id)
-        await self.settings.guild(guild).current_tempbans.set(cur_tbans)
+        await self.config.member(user).banned_until.set(unban_time.timestamp())
+        async with self.config.guild(guild).all() as guild_data:
+            guild_data["current_tempbans"].append(user.id)
```
Jackenmen:
```suggestion
        async with self.config.guild(guild).current_tempbans() as current_tempbans:
            current_tempbans.append(user.id)

```

redbot/core/commands/help.py
```diff
@@ -123,40 +124,55 @@ async def send_help(self, ctx: Context, help_for: HelpTarget = None):
 
         For most cases, you should use this and only this directly.
         """
+
+        help_config = await HelpSettings.from_context(ctx)
```
Jackenmen: This should be `help_settings` to avoid confusion.

This comment refers to every place where you used `help_config` var name for storing `HelpSettings`.

Jackenmen (changes requested):

Jackenmen (approved):
LGTM

redbot/cogs/filter/filter.py
```diff
@@ -276,8 +278,10 @@ async def filter_names(self, ctx: commands.Context):
         This is disabled by default.
         """
         guild = ctx.guild
-        current_setting = await self.settings.guild(guild).filter_names()
-        await self.settings.guild(guild).filter_names.set(not current_setting)
+
+        async with await self.config.guild(guild).all() as guild_data:
```
Jackenmen:
```suggestion
        async with self.config.guild(guild).all() as guild_data:
```

Jackenmen (changes requested):

Jackenmen (approved):

PR #3767, opened 2020-04-18
Make the largest loops lazier and async to avoid blocking

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This one depends on #3766

### DIFF
--- a/docs/framework_utils.rst
+++ b/docs/framework_utils.rst
@@ -9,8 +9,17 @@ General Utility
 
 .. automodule:: redbot.core.utils
     :members: deduplicate_iterables, bounded_gather, bounded_gather_iter
+
 .. autoclass:: AsyncIter
     :members:
+    :exclude-members: enumerate, filter
+
+    .. automethod:: enumerate
+        :async-for:
+
+    .. automethod:: filter
+        :async-for:
+
 
 Chat Formatting
 ===============
--- a/docs/framework_utils.rst
+++ b/docs/framework_utils.rst
@@ -8,7 +8,7 @@ General Utility
 ===============
 
 .. automodule:: redbot.core.utils
-    :members: deduplicate_iterables, bounded_gather, bounded_gather_iter
+    :members: deduplicate_iterables, bounded_gather, bounded_gather_iter, AsyncIter
 
 Chat Formatting
 ===============
--- a/docs/framework_utils.rst
+++ b/docs/framework_utils.rst
@@ -8,7 +8,9 @@ General Utility
 ===============
 
 .. automodule:: redbot.core.utils
-    :members: deduplicate_iterables, bounded_gather, bounded_gather_iter, AsyncIter
+    :members: deduplicate_iterables, bounded_gather, bounded_gather_iter
+.. autoclass:: AsyncIter
+    :members:
 
 Chat Formatting
 ===============
--- a/redbot/cogs/admin/announcer.py
+++ b/redbot/cogs/admin/announcer.py
@@ -3,7 +3,7 @@ import asyncio
 import discord
 from redbot.core import commands
 from redbot.core.i18n import Translator
-from redbot.core.utils import AsyncGen
+from redbot.core.utils import AsyncIter
 from redbot.core.utils.chat_formatting import humanize_list, inline
 
 _ = Translator("Announcer", __file__)
@@ -56,7 +56,7 @@ class Announcer:
     async def announcer(self):
         guild_list = self.ctx.bot.guilds
         failed = []
-        async for g in AsyncGen(guild_list, delay=0.5):
+        async for g in AsyncIter(guild_list, delay=0.5):
             if not self.active:
                 return
 
--- a/redbot/cogs/admin/announcer.py
+++ b/redbot/cogs/admin/announcer.py
@@ -3,6 +3,7 @@ import asyncio
 import discord
 from redbot.core import commands
 from redbot.core.i18n import Translator
+from redbot.core.utils import AsyncGen
 from redbot.core.utils.chat_formatting import humanize_list, inline
 
 _ = Translator("Announcer", __file__)
@@ -55,7 +56,7 @@ class Announcer:
     async def announcer(self):
         guild_list = self.ctx.bot.guilds
         failed = []
-        for g in guild_list:
+        async for g in AsyncGen(guild_list, delay=0.5):
             if not self.active:
                 return
 
@@ -68,7 +69,6 @@ class Announcer:
                 await channel.send(self.message)
             except discord.Forbidden:
                 failed.append(str(g.id))
-            await asyncio.sleep(0.5)
 
         if failed:
             msg = (
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -71,7 +71,7 @@ class KickBanMixin(MixinMeta):
 
         if author == user:
             return _("I cannot let you do that. Self-harm is bad {}").format("\N{PENSIVE FACE}")
-        elif not await is_allowed_by_hierarchy(self.bot, self.settings, guild, author, user):
+        elif not await is_allowed_by_hierarchy(self.bot, self.config, guild, author, user):
             return _(
                 "I cannot let you do that. You are "
                 "not higher than the user in the role "
@@ -82,7 +82,7 @@ class KickBanMixin(MixinMeta):
         elif not (0 <= days <= 7):
             return _("Invalid days. Must be between 0 and 7.")
 
-        toggle = await self.settings.guild(guild).dm_on_kickban()
+        toggle = await self.config.guild(guild).dm_on_kickban()
         if toggle:
             with contextlib.suppress(discord.HTTPException):
                 em = discord.Embed(
@@ -142,13 +142,13 @@ class KickBanMixin(MixinMeta):
                 except discord.HTTPException:
                     continue
 
-                async with self.settings.guild(guild).current_tempbans() as guild_tempbans:
+                async with self.config.guild(guild).current_tempbans() as guild_tempbans:
                     for uid in guild_tempbans.copy():
                         user = banned_users.get(uid, None)
                         if not user:
                             continue
                         unban_time = datetime.utcfromtimestamp(
-                            await self.settings.member(member(uid, guild)).banned_until()
+                            await self.config.member(member(uid, guild)).banned_until()
                         )
                         if datetime.utcnow() > unban_time:  # Time to unban the user
                             queue_entry = (guild.id, uid)
@@ -186,7 +186,7 @@ class KickBanMixin(MixinMeta):
                 )
             )
             return
-        elif not await is_allowed_by_hierarchy(self.bot, self.settings, guild, author, user):
+        elif not await is_allowed_by_hierarchy(self.bot, self.config, guild, author, user):
             await ctx.send(
                 _(
                     "I cannot let you do that. You are "
@@ -199,7 +199,7 @@ class KickBanMixin(MixinMeta):
             await ctx.send(_("I cannot do that due to discord hierarchy rules"))
             return
         audit_reason = get_audit_reason(author, reason)
-        toggle = await self.settings.guild(guild).dm_on_kickban()
+        toggle = await self.config.guild(guild).dm_on_kickban()
         if toggle:
             with contextlib.suppress(discord.HTTPException):
                 em = discord.Embed(
@@ -255,7 +255,7 @@ class KickBanMixin(MixinMeta):
         author = ctx.author
         guild = ctx.guild
         if days is None:
-            days = await self.settings.guild(guild).default_days()
+            days = await self.config.guild(guild).default_days()
 
         result = await self.ban_user(
             user=user, ctx=ctx, days=days, reason=reason, create_modlog_case=True
@@ -309,7 +309,7 @@ class KickBanMixin(MixinMeta):
             return
 
         if days is None:
-            days = await self.settings.guild(guild).default_days()
+            days = await self.config.guild(guild).default_days()
 
         if not (0 <= days <= 7):
             await ctx.send(_("Invalid days. Must be between 0 and 7."))
@@ -409,10 +409,10 @@ class KickBanMixin(MixinMeta):
             invite = ""
 
         queue_entry = (guild.id, user.id)
-        await self.settings.member(user).banned_until.set(unban_time.timestamp())
-        cur_tbans = await self.settings.guild(guild).current_tempbans()
+        await self.config.member(user).banned_until.set(unban_time.timestamp())
+        cur_tbans = await self.config.guild(guild).current_tempbans()
         cur_tbans.append(user.id)
-        await self.settings.guild(guild).current_tempbans.set(cur_tbans)
+        await self.config.guild(guild).current_tempbans.set(cur_tbans)
 
         with contextlib.suppress(discord.HTTPException):
             # We don't want blocked DMs preventing us from banning
@@ -464,7 +464,7 @@ class KickBanMixin(MixinMeta):
                 )
             )
             return
-        elif not await is_allowed_by_hierarchy(self.bot, self.settings, guild, author, user):
+        elif not await is_allowed_by_hierarchy(self.bot, self.config, guild, author, user):
             await ctx.send(
                 _(
                     "I cannot let you do that. You are "
@@ -540,7 +540,7 @@ class KickBanMixin(MixinMeta):
 
         if await self._voice_perm_check(ctx, user_voice_state, move_members=True) is False:
             return
-        elif not await is_allowed_by_hierarchy(self.bot, self.settings, guild, author, member):
+        elif not await is_allowed_by_hierarchy(self.bot, self.config, guild, author, member):
             await ctx.send(
                 _(
                     "I cannot let you do that. You are "
@@ -620,7 +620,7 @@ class KickBanMixin(MixinMeta):
                 await ctx.send(e)
             await ctx.send(_("Unbanned that user from this server"))
 
-        if await self.settings.guild(guild).reinvite_on_unban():
+        if await self.config.guild(guild).reinvite_on_unban():
             user = ctx.bot.get_user(user_id)
             if not user:
                 await ctx.send(
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -11,6 +11,7 @@ from redbot.core.utils.chat_formatting import pagify, humanize_number, bold
 from redbot.core.utils.mod import is_allowed_by_hierarchy, get_audit_reason
 from .abc import MixinMeta
 from .converters import RawUserIds
+from ...core.utils import AsyncGen
 
 log = logging.getLogger("red.mod")
 _ = i18n.Translator("Mod", __file__)
@@ -134,7 +135,7 @@ class KickBanMixin(MixinMeta):
     async def check_tempban_expirations(self):
         member = namedtuple("Member", "id guild")
         while self == self.bot.get_cog("Mod"):
-            for guild in self.bot.guilds:
+            async for guild in AsyncGen(self.bot.guilds, steps=100):
                 if not guild.me.guild_permissions.ban_members:
                     continue
                 try:
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -7,11 +7,11 @@ from typing import cast, Optional, Union
 
 import discord
 from redbot.core import commands, i18n, checks, modlog
+from redbot.core.utils import AsyncIter
 from redbot.core.utils.chat_formatting import pagify, humanize_number, bold
 from redbot.core.utils.mod import is_allowed_by_hierarchy, get_audit_reason
 from .abc import MixinMeta
 from .converters import RawUserIds
-from ...core.utils import AsyncGen
 
 log = logging.getLogger("red.mod")
 _ = i18n.Translator("Mod", __file__)
@@ -135,7 +135,7 @@ class KickBanMixin(MixinMeta):
     async def check_tempban_expirations(self):
         member = namedtuple("Member", "id guild")
         while self == self.bot.get_cog("Mod"):
-            async for guild in AsyncGen(self.bot.guilds, steps=100):
+            async for guild in AsyncIter(self.bot.guilds, steps=100):
                 if not guild.me.guild_permissions.ban_members:
                     continue
                 try:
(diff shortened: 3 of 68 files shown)


redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+    ) -> None:
+        self.delay = delay
+        self.content = iter(contents)
+        self.i = 0
+        self.steps = steps
+
+    def __aiter__(self):
+        return self
+
+    async def __anext__(self) -> Awaitable[_T]:
```
Jackenmen: Incorrect type hint, it should be:
```suggestion
    async def __anext__(self) -> _T:
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+        try:
+            i = next(self.content)
```
Jackenmen: `item` will be clearer than `i`, especially considering we already use `i` for index.
```suggestion
            item = next(self.content)
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+        except StopIteration:
+            raise StopAsyncIteration
+        self.i += 1
+        if self.i % self.steps == 0:
+            await asyncio.sleep(self.delay)
+        return i
```
Jackenmen: `item` will be clearer than `i`, especially considering we already use `i` for index.
```suggestion
        return item
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+
+    def __await__(self):
```
Jackenmen: Add type hint (remember to import `Generator` from `typing`)
```suggestion
    def __await__(self) -> Generator[Any, None, List[_T]]:
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+
+    def __init__(
+        self, contents: Union[Iterator[_T], Iterable[_T]], delay: float = 0.0, steps: int = 1
+    ) -> None:
+        self.delay = delay
+        self.content = iter(contents)
+        self.i = 0
+        self.steps = steps
+
+    def __aiter__(self):
```
Jackenmen: Add type hint (you need to add `from __future__ import annotations` at the top of the file):
```suggestion
    def __aiter__(self) -> AsyncGen[_T]:
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
     tasks = (_sem_wrapper(semaphore, task) for task in coros_or_futures)
 
     return asyncio.gather(*tasks, return_exceptions=return_exceptions)
+
+
+class AsyncGen(AsyncIterable[_T]):
+    """Yield entry every `delay` seconds per `steps`."""
+
+    def __init__(
+        self, contents: Union[Iterator[_T], Iterable[_T]], delay: float = 0.0, steps: int = 1
```
Jackenmen: Considering we're passing iterable (iterator is an iterable), I think it would be clearer to call it `iterable` instead of `contents`

I also have two suggestions for typings here:
a) we can simplify `Union[Iterator[_T], Iterable[_T]]` to just `Iterable[_T]` as all iterators are iterable
b) since `float` is not an `int`, we need to specify union of both here if we want to accept both (and since it's only used for `asyncio.sleep()`, I think we do want it)
```suggestion
        self, iterable: Iterable[_T], delay: Union[float, int] = 0, steps: int = 1
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+    ) -> None:
+        self.delay = delay
+        self.content = iter(contents)
```
Jackenmen: I think it would be clearer if this attribute were called `iterator` (or `it` if you prefer short version)

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+        function = partial(function, *args, **kwargs)
+        async for i in async_filter(self, function):
+            yield i
+
+    async def enumerate(self, start=0) -> AsyncIterator[Tuple[int, _T]]:
+        async for item in self:
+            yield start, item
+            start += 1
+
+    async def remove_dupes(self):
```
Jackenmen: This needs a better name - this doesn't remove duplicates, it gives you iterator without duplicated entries, maybe it should be called `without_duplicates()`?

Jackenmen: Add type hint:
```suggestion
    async def remove_dupes(self) -> AsyncIterator[_T]:
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+class AsyncGen(AsyncIterable[_T]):
+    """Yield entry every `delay` seconds per `steps`."""
+
+    def __init__(
+        self, contents: Union[Iterator[_T], Iterable[_T]], delay: float = 0.0, steps: int = 1
+    ) -> None:
+        self.delay = delay
+        self.content = iter(contents)
+        self.i = 0
+        self.steps = steps
```
Jackenmen: I think all these attributes should be private.

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+        self, function: Union[AsyncIterable[_T], Iterable[_T]], *args, **kwargs
+    ) -> AsyncIterator[_T]:
+        function = partial(function, *args, **kwargs)
+        async for i in async_filter(self, function):
+            yield i
+
+    async def enumerate(self, start=0) -> AsyncIterator[Tuple[int, _T]]:
+        async for item in self:
+            yield start, item
+            start += 1
```
Jackenmen: - I think using `async_enumerate` would be better here just so that all the logic is in one place.
- `start` is missing type hint
```suggestion
    def enumerate(self, start: int = 0) -> AsyncIterator[Tuple[int, _T]]:
        return async_enumerate(self, start)
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+            i = next(self.content)
+        except StopIteration:
+            raise StopAsyncIteration
+        self.i += 1
+        if self.i % self.steps == 0:
+            await asyncio.sleep(self.delay)
+        return i
+
+    def __await__(self):
+        return self.flatten()
```
Jackenmen: You need to return the result of the `coro.__await__()` not `coro` itself from `__await__` function:
```suggestion
        return self.flatten().__await__()
```

redbot/cogs/admin/announcer.py
```diff
@@ -55,7 +56,7 @@ async def _get_announce_channel(self, guild: discord.Guild) -> discord.TextChann
     async def announcer(self):
         guild_list = self.ctx.bot.guilds
         failed = []
-        for g in guild_list:
+        async for g in AsyncGen(guild_list, delay=0.5):
```
Jackenmen: It would be better to use `asyncio.gather` here but I guess it's kind of out of scope for this PR.

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
     tasks = (_sem_wrapper(semaphore, task) for task in coros_or_futures)
 
     return asyncio.gather(*tasks, return_exceptions=return_exceptions)
+
+
+class AsyncGen(AsyncIterable[_T]):
+    """Yield entry every `delay` seconds per `steps`."""
```
Jackenmen: Better docstring needed (list parameters and their expected types as with any other function in this file)

We should also be a little more descriptive about what this does. Something more like (still needs to be better worded):
```
Asynchronous iterator yielding items from ``iterable`` that sleeps for ``delay`` seconds every ``steps`` items.
```

Last but not least, you need to add `automethod` directive for this class to `docs/framework_utils.rst`

redbot/cogs/mod/kickban.py
```diff
@@ -11,6 +11,7 @@
 from redbot.core.utils.mod import is_allowed_by_hierarchy, get_audit_reason
 from .abc import MixinMeta
 from .converters import RawUserIds
+from ...core.utils import AsyncGen
```
Jackenmen: It's a nitpick but I think it would be clearer to do `from redbot.core.utils import AsyncGen`

redbot/cogs/permissions/converters.py
```diff
@@ -37,26 +38,33 @@ async def convert(
...
+        ]
         objects = itertools.chain(
             bot.get_all_channels(),
             bot.users,
             bot.guilds,
-            *(filter(lambda r: not r.is_default(), guild.roles) for guild in bot.guilds),
+            *(
+                filter(lambda r: not r.is_default(), roles)
+                async for roles in AsyncGen(all_roles, steps=100)
+            ),
```
Jackenmen: This and the `all_roles` list that was added above can be simplified to:
```suggestion
            *(
                filter(lambda r: not r.is_default(), guild.roles)
                async for guild in AsyncGen(bot.guilds, steps=100)
            ),
```
That also means that `all_roles` from above can be removed.

redbot/cogs/permissions/converters.py
```diff
@@ -37,26 +38,33 @@ async def convert(
...
         if ctx.guild is not None:
-            for member in ctx.guild.members:
-                if member.nick == arg and not any(obj.id == member.id for obj in maybe_matches):
+            async for member in AsyncGen(ctx.guild.members, steps=100):
+                if member.nick == arg and not any(
+                    obj.id == member.id async for obj in AsyncGen(maybe_matches, steps=100)
+                ):
                     maybe_matches.append(member)
 
         if not maybe_matches:
```
Jackenmen: This whole thing should be done differently - we should stop when we figure out there's more than one match.

redbot/cogs/permissions/converters.py
```diff
@@ -102,7 +110,7 @@ async def convert(
         )
 
         maybe_matches = []
-        for obj in objects:
+        async for obj in AsyncGen(objects, steps=100):
             if obj.name == arg or str(obj) == arg:
                 maybe_matches.append(obj)
```
Jackenmen, line 108: Same as above, we should stop iterating when figure out there's more than one match.

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
     tasks = (_sem_wrapper(semaphore, task) for task in coros_or_futures)
 
     return asyncio.gather(*tasks, return_exceptions=return_exceptions)
+
+
+class AsyncGen(AsyncIterable[_T]):
```
Jackenmen: This class is not just `AsyncIterable` but also `AsyncIterator` so that generic suits it better. It can also be awaited so I think you should add `Awaitable` generic as well.
```suggestion
class AsyncGen(AsyncIterator[_T], Awaitable[List[_T]]):  # pylint: disable=duplicate-bases
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+        async for i in async_filter(self, function):
+            yield i
+
+    async def enumerate(self, start=0) -> AsyncIterator[Tuple[int, _T]]:
+        async for item in self:
+            yield start, item
+            start += 1
+
+    async def remove_dupes(self):
+        _temp = []
```
Jackenmen: Since we don't care about order in `_temp` here, can you use `set` for better performance?

redbot/core/utils/__init__.py
Jackenmen: This class name (`AsyncGen`) might be confusing - despite of what the name says, this isn't async generator, it's async iterator (and these two terms are different).

redbot/core/utils/__init__.py
```diff
@@ -20,7 +21,7 @@
     TYPE_CHECKING,
 )
 
-__all__ = ("bounded_gather", "deduplicate_iterables")
+__all__ = ("bounded_gather", "deduplicate_iterables", "AsyncGen")
```
Jackenmen: It's not a part of the PR, but could you add `bounded_gather_iter` to `__all__`?
```suggestion
__all__ = ("bounded_gather", "bounded_gather_iter", "deduplicate_iterables", "AsyncGen")
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +256,54 @@ def bounded_gather(
...
+
+    async def flatten(self) -> List[_T]:
+        return [item async for item in self]
+
+    async def filter(
+        self, function: Union[AsyncIterable[_T], Iterable[_T]], *args, **kwargs
+    ) -> AsyncIterator[_T]:
+        function = partial(function, *args, **kwargs)
+        async for i in async_filter(self, function):
+            yield i
```
Jackenmen: - Arguments are passed in wrong order to `async_filter()`
- `function`'s type hint is incorrect
- I think `*args, **kwargs` are unnecessary here, if someone wants to, they can pass partial themselves.
- This whole function can be simplified to regular function returning result of `async_filter()` function:
```suggestion
    def filter(self, function: Callable[[_T], Union[bool, Awaitable[bool]]]) -> AsyncFilter[_T]:
        return async_filter(function, self)
```

redbot/cogs/permissions/converters.py
```diff
@@ -37,26 +38,33 @@ async def convert(
...
 
         if ctx.guild is not None:
-            for member in ctx.guild.members:
-                if member.nick == arg and not any(obj.id == member.id for obj in maybe_matches):
+            async for member in AsyncGen(ctx.guild.members, steps=100):
+                if member.nick == arg and not any(
+                    obj.id == member.id async for obj in AsyncGen(maybe_matches, steps=100)
```
Jackenmen: I didn't test this but I doubt this works at all - `any()` can be used with regular iterables, not with async iterables.

Jackenmen (changes requested):

redbot/cogs/permissions/converters.py
Drapersniper: This is out of the scope of this PR no? I mean im happy to address it but .. just want to be sure

docs/framework_utils.rst
```diff
@@ -9,6 +9,8 @@ General Utility
 
 .. automodule:: redbot.core.utils
     :members: deduplicate_iterables, bounded_gather, bounded_gather_iter
+.. autoclass:: AsyncIter
+    :members:
```
Jackenmen, line 14: For the methods to show as `async for ... in enumerate(start=)` we need to manually indicate it for these two methods:
image

```suggestion

.. autoclass:: AsyncIter
    :members:
    :exclude-members: enumerate, filter

    .. automethod:: enumerate
        :async-for:

    .. automethod:: filter
        :async-for:

```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,128 @@ def bounded_gather(
...
+            >>> async for i in iterator.filter(predicate):
+            1
+            5
+
+           >>> from redbot.core.utils import AsyncIter
+           >>> def predicate(value):
+           >>>    return value <= 5
+           >>> iterator = AsyncIter([1, 10, 5, 100])
+           >>> await iterator.filter(predicate)
+           [1, 5]
```
Jackenmen: bad indentation here
```suggestion
            >>> from redbot.core.utils import AsyncIter
            >>> def predicate(value):
            >>>     return value <= 5
            >>> iterator = AsyncIter([1, 10, 5, 100])
            >>> await iterator.filter(predicate)
            [1, 5]
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,128 @@ def bounded_gather(
...
+        -------
+        AsyncFilter[T]
+            An object which can either be awaited to yield a list of the filtered
+            items, or can also act as an async iterator to yield items one by one.
+
+        Examples
+        --------
+            >>> from redbot.core.utils import AsyncIter
+            >>> def predicate(value):
+            >>>    return value <= 5
```
Jackenmen: Let's use 4 space indent here
```suggestion
            >>>     return value <= 5
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,128 @@ def bounded_gather(
...
+            >>> iterator = AsyncIter([1, 10, 5, 100])
+            >>> async for i in iterator.filter(predicate):
```
Jackenmen, line 325: This looks incomplete, maybe you wanted to add a `print()` here?
```suggestion
            >>> async for i in iterator.filter(predicate):
            >>>     print(i)
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,128 @@ def bounded_gather(
...
+        Returns
+        -------
+        AsyncIterator[Tuple[int, T]]
+            An async iterator of tuples in the form of ``(index, item)``.
+
+        Examples
+        --------
+            >>> from redbot.core.utils import AsyncIter
+            >>> iterator = AsyncIter(['one', 'two', 'three'])
+            >>> async for i in iterator.enumerate(start=10):
```
Jackenmen, line 357: This looks incomplete, maybe you wanted to add a `print()` here?
```suggestion
            >>> async for i in iterator.enumerate(start=10):
            >>>     print(i)
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,128 @@ def bounded_gather(
...
+
+    async def without_duplicates(self) -> AsyncIterator[_T]:
+        """
+        Iterates while omitting duplicated entries.
+
+        Examples
+        --------
+            >>> from redbot.core.utils import AsyncIter
+            >>> iterator = AsyncIter([1,2,3,3,4,4,5])
+            >>> async for i in iterator.without_duplicates():
```
Jackenmen, line 374: This looks incomplete, maybe you wanted to add a `print()` here?
```suggestion
            >>> async for i in iterator.without_duplicates():
            >>>     print(i)
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,128 @@ def bounded_gather(
...
+        The amount of time in seconds to sleep.
+    steps: int
+        The number of iterations between sleeps.
+    """
+
+    def __init__(
+        self, iterable: Iterable[_T], delay: Union[float, int] = 0, steps: int = 1
+    ) -> None:
+        self._delay = delay
+        self._content = iter(iterable)
```
Jackenmen: Wouldn't it make more sense to call this `self._iterator`?

Jackenmen (changes requested):

redbot/core/utils/__init__.py
Jackenmen:
```suggestion
        self._iterator = iter(iterable)
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,128 @@ def bounded_gather(
...
+        self._i = 0
+        self._steps = steps
+
+    def __aiter__(self) -> AsyncIter[_T]:
+        return self
+
+    async def __anext__(self) -> _T:
+        try:
+            item = next(self._content)
```
Jackenmen:
```suggestion
            item = next(self._iterator)
```

Jackenmen (changes requested):

redbot/cogs/permissions/converters.py
```diff
@@ -37,25 +38,27 @@ async def convert(
...
-
         objects = itertools.chain(
             bot.get_all_channels(),
             bot.users,
             bot.guilds,
-            *(filter(lambda r: not r.is_default(), guild.roles) for guild in bot.guilds),
+            *(
+                filter(lambda r: not r.is_default(), guild.roles)
+                async for guild in AsyncIter(bot.guilds, steps=100)
+            ),
```
Jackenmen: This isn't actually a valid code - `*` can't be used with async iterators

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,131 @@ def bounded_gather(
...
+    async def without_duplicates(self) -> AsyncIterator[_T]:
+        """
+        Iterates while omitting duplicated entries.
+
+        Examples
+        --------
+            >>> from redbot.core.utils import AsyncIter
+            >>> iterator = AsyncIter([1,2,3,3,4,4,5])
+            >>> async for i in iterator.without_duplicates():
+            >>>     print(i)
```
Jackenmen:
```suggestion
            ...     print(i)
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,131 @@ def bounded_gather(
...
+        -------
+        AsyncIterator[Tuple[int, T]]
+            An async iterator of tuples in the form of ``(index, item)``.
+
+        Examples
+        --------
+            >>> from redbot.core.utils import AsyncIter
+            >>> iterator = AsyncIter(['one', 'two', 'three'])
+            >>> async for i in iterator.enumerate(start=10):
+            >>>     print(i)
```
Jackenmen:
```suggestion
            ...     print(i)
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,131 @@ def bounded_gather(
...
+            >>>     return value <= 5
+            >>> iterator = AsyncIter([1, 10, 5, 100])
+            >>> async for i in iterator.filter(predicate):
+            >>>     print(i)
+            1
+            5
+
+            >>> from redbot.core.utils import AsyncIter
+            >>> def predicate(value):
+            >>>     return value <= 5
```
Jackenmen:
```suggestion
            ...     return value <= 5
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,131 @@ def bounded_gather(
...
+        -------
+        AsyncFilter[T]
+            An object which can either be awaited to yield a list of the filtered
+            items, or can also act as an async iterator to yield items one by one.
+
+        Examples
+        --------
+            >>> from redbot.core.utils import AsyncIter
+            >>> def predicate(value):
+            >>>     return value <= 5
```
Jackenmen:
```suggestion
            ...     return value <= 5
```

redbot/core/utils/__init__.py
```diff
@@ -255,3 +257,131 @@ def bounded_gather(
...
+            >>> iterator = AsyncIter([1, 10, 5, 100])
+            >>> async for i in iterator.filter(predicate):
+            >>>     print(i)
```
Jackenmen:
```suggestion
            ...     print(i)
```

Jackenmen (changes requested):

Jackenmen (approved):

PR #3771, opened 2020-04-19
One small step for JSON, one giant leap for Redkind

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Stop JSON driver from saving on change, and start saving periodically (1/s if there's been a change in the data)

### DIFF
--- a/redbot/core/drivers/json.py
+++ b/redbot/core/drivers/json.py
@@ -49,6 +49,7 @@ class JsonDriver(BaseDriver):
 
         The path in which to store the file indicated by :py:attr:`file_name`.
     """
+    saving_task = None
 
     def __init__(
         self,
@@ -68,6 +69,9 @@ class JsonDriver(BaseDriver):
             self.data_path = data_manager.cog_data_path(raw_name=cog_name)
         self.data_path.mkdir(parents=True, exist_ok=True)
         self.data_path = self.data_path / self.file_name
+        self.data_changed = False
+        loop = asyncio.get_event_loop()
+        self.saving_task = loop.create_task(self._save_task())
 
         self._lock = asyncio.Lock()
         self._load_data()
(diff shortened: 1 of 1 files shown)


mikeshardmind:
It's been explored before, and it's going to have all the same issues that were shown then.

1. config no longer guarantees a data write when the function returns. 

This particular part can actually be resolved with a pending write queue and asyncio condition waking all of those, thus putting the bottleneck on the write itself while still not returning any unwritten, but this will slow down any methods which write multiple config values

2. data consistency issues

Application crashes can't be reliably caught. There's a [BPO](https://bugs.python.org/issue39232) open for asyncio shutdown on windows specifically still, so we're delaying writes in a way which increases the window for data loss, or inconsistent data state. This would impact the bots that would benefit from this sort of change the most.

3. Inconsistent config behavior between backends

Right now, if a config write returns, as long as the hard drive isn't dead, the data is on disk properly. If this change were to be added, this would have inconsistent meaning across backends. (on Postgres, a response from the server is enough to assume the data is stored resiliently)

If large bots are having performance issues, it may be time to really look into moving more of core to use SQLite (modlog is a good example of where this can be done without much complexity, but with a large gain for big bots), or a full SQLite driver (even if just used as a JSON store, the majority of the time is temp write, then replace + fsync for data resilience, this isn't needed with SQLite)

Exposing parts of this to config and to cog creators could work as well.

For example, adding an async context manager such as `config_group.maybe_delay_writes()` This has other complexities, but at least ensures that developers have control over where it is acceptable to delay or batch writes. Leaving it as a maybe means the actual behavior can be backend dependent, as there isn't always a need to delay, but if batching is a goal, then removing the maybe and having it act more like a transaction might be appropriate.

The large majority of bots aren't impacted by the issues with the JSON driver, only the largest ones which need an actual DB at this point are. 

I think there are solutions that involve improving the backend options, as well as improving control over how writes are handled as developers but I don't think that this solution is better for the majority of bots than the current situation.

Edit: I forgot to mention, but also previously explored was splitting the files using a binning function. This worked to an extent, but could easily lead to an issue with too many file handles open on windows and osx, as well as with some configurations on linux, as well as taking longer for temp files to register as cleaned up by the operating system.

Drapersniper:
Closing in favour of a better implementation.

PR #3773, opened 2020-04-20
Stop using localised display name in stream url

### LINKED ISSUE #3772: Twitch Streamalert links to localized names
#### Command name

[p]twitchalert twitch 

streams

Broadcast message should link to twitch URL not twitch localized name URL.

#### What actually happened?

Stream alert links to Twitch user's localized name instead of their twitch URL resulting in a 404.

[p]streamalert twitch hinamakura

Alert will link to https://www.twitch.tv/%E3%81%B2%E3%81%AA%E3%81%BE%E3%81%8F%E3%82%89

instead of
https://www.twitch.tv/hinamakura

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Should fix #3772 

Bot still display localised display name for channel name but channel URL properly uses login now:
image
image

### DIFF
--- a/redbot/cogs/streams/streamtypes.py
+++ b/redbot/cogs/streams/streamtypes.py
@@ -228,6 +228,7 @@ class TwitchStream(Stream):
             data["followers"] = None
             data["view_count"] = None
             data["profile_image_url"] = None
+            data["login"] = None
 
             game_id = data["game_id"]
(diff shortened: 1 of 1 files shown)


Kowlin (approved):
Logic checks out.

PR #3780, opened 2020-04-22
[Driver] Ensure JSON driver has a singular lock per cog name

Drapersniper:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

JSON driver is supposed to have 1 lock per `self.cog_name` howver currently this is not true, if 2 loaded cogs have the same name but different identifier 2 different locks will be used which could end up writting to file in a incorrect order.

### DIFF
--- a/redbot/core/drivers/json.py
+++ b/redbot/core/drivers/json.py
@@ -70,7 +70,6 @@ class JsonDriver(BaseDriver):
             self.data_path = data_manager.core_data_path()
         else:
             self.data_path = data_manager.cog_data_path(raw_name=cog_name)
-        self._lock = asyncio.Lock()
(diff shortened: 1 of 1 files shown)


redbot/core/drivers/json.py
```diff
@@ -17,6 +17,7 @@
 _shared_datastore = {}
 _driver_counts = {}
 _finalizers = []
+_locks = {}
```
Jackenmen: Wouldn't it be simpler to use `defaultdict(asyncio.Lock)` and get rid of setter (and just use square bracket notation in property instead of `.get()`)?

Jackenmen (approved):
LGTM

PR #3781, opened 2020-04-22
Allow to hide moderator from warnings sent to the users + warn channel fix

Dav-Git:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

On some discords, backlash on the moderator who warned a user is a big concern. Therefore I implemented a setting that would turn off the sending of the username of the moderator if sending DMs is enabled.


Kowlin:
I'd personally be of favor on the moderator's side and making ``show_mod`` be False by default, and I'd like some of the thoughts on this.

Dav-Git:
My thought on that was transparency and have people be able to turn it off once it becomes an issue

Dav-Git:
I went with the kowlin option so it's now default false

Jackenmen:
> I'd personally be of favor on the moderator's side and making `show_mod` be False by default, and I'd like some of the thoughts on this.

I agree with this, even if it's changing current default. Usually it's not important who warned the warned user and it can end up in unwanted backlash so this seems as more reasonable default.

This change might be worth mentioning at the start of changelog section so that people who do want moderators to be shown, can change this after updating.

@Dav-Git this PR is now conflicting, could you handle the conflict resolution?

Dav-Git:
Yep, will do later today/tomorrow.

Dav-Git:
Hold off on merging this, I found an error I will need to test first

Jackenmen:
There's another mention of a moderator at L417

Dav-Git:
I also included a "fix" to the warnchannel... either I broke it or it was broken previosly. i'm not sure at this point.

Dav-Git:
Ah, forgot to include in last comment, but from my testing this works(TM)

Jackenmen (approved):
Looks good to me 👍

PR #3782, opened 2020-04-22
Fix Owner ID failsafe

Kowlin:
# Bugfix request

This PR fixes an issue whereby it will trigger the failsafe and causing the command execution to be horrible delayed by calling application_info on every command.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -703,7 +703,10 @@ class RedBase(
         """
         if user.id in self._co_owners:
             return True
-        if self.owner_id is None:
+        # if bot runs on team application and both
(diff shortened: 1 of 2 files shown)


redbot/core/bot.py
```diff
@@ -703,6 +703,8 @@ async def is_owner(self, user: Union[discord.User, discord.Member]) -> bool:
         """
         if user.id in self._co_owners:
             return True
+        if self.owner_id is None:
```
```suggestion
        # if bot runs on team application and both
        # `--owner` and `--team-members-are-owners` flags aren't set,
        # `bot.owner_id` is `None` (GH-3782)
        if self._use_team_features and self.owner_id is None:
```

Jackenmen (approved):
Looks good to me and Kowlin

PR #3788, opened 2020-04-24
[Alias] Create caching to call config less frequently

TrustyJAID:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Add a caching system to Alias to reduce frequency of calls to config. This also re-arranges some functions inside alias to improve readability and usability.

This also changes aliases to be checked on_message_without_command to avoid the potential of loading a cog with a command with the same name as an alias.

### DIFF
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -1,16 +1,15 @@
 from copy import copy
-from re import findall, search
+from re import search
 from string import Formatter
-from typing import Generator, Tuple, Iterable, Optional
+from typing import Dict
 
 import discord
-from discord.ext.commands.view import StringView
 from redbot.core import Config, commands, checks
 from redbot.core.i18n import Translator, cog_i18n
 from redbot.core.utils.chat_formatting import box
 
 from redbot.core.bot import Red
-from .alias_entry import AliasEntry
+from .alias_entry import AliasEntry, AliasCache, ArgParseError
 
 _ = Translator("Alias", __file__)
 
@@ -26,10 +25,6 @@ class _TrackingFormatter(Formatter):
         return super().get_value(key, args, kwargs)
 
 
-class ArgParseError(Exception):
-    pass
-
-
 @cog_i18n(_)
 class Alias(commands.Cog):
     """Create aliases for commands.
@@ -42,9 +37,9 @@ class Alias(commands.Cog):
     and append them to the stored alias.
     """
 
-    default_global_settings = {"entries": []}
+    default_global_settings: Dict[str, list] = {"entries": []}
 
-    default_guild_settings = {"enabled": False, "entries": []}  # Going to be a list of dicts
+    default_guild_settings: Dict[str, list] = {"entries": []}  # Going to be a list of dicts
 
     def __init__(self, bot: Red):
         super().__init__()
@@ -53,40 +48,14 @@ class Alias(commands.Cog):
 
         self.config.register_global(**self.default_global_settings)
         self.config.register_guild(**self.default_guild_settings)
+        self._aliases: AliasCache = AliasCache(config=self.config, cache_enabled=True)
+        self._load_cache = self.bot.loop.create_task(self.initialize())
 
-    async def unloaded_aliases(self, guild: discord.Guild) -> Generator[AliasEntry, None, None]:
-        return (AliasEntry.from_json(d) for d in (await self.config.guild(guild).entries()))
-
-    async def unloaded_global_aliases(self) -> Generator[AliasEntry, None, None]:
-        return (AliasEntry.from_json(d) for d in (await self.config.entries()))
-
-    async def loaded_aliases(self, guild: discord.Guild) -> Generator[AliasEntry, None, None]:
-        return (
-            AliasEntry.from_json(d, bot=self.bot)
-            for d in (await self.config.guild(guild).entries())
-        )
-
-    async def loaded_global_aliases(self) -> Generator[AliasEntry, None, None]:
-        return (AliasEntry.from_json(d, bot=self.bot) for d in (await self.config.entries()))
-
-    async def is_alias(
-        self,
-        guild: Optional[discord.Guild],
-        alias_name: str,
-        server_aliases: Iterable[AliasEntry] = (),
-    ) -> Tuple[bool, Optional[AliasEntry]]:
-
-        if not server_aliases and guild is not None:
-            server_aliases = await self.unloaded_aliases(guild)
-
-        global_aliases = await self.unloaded_global_aliases()
-
-        for aliases in (server_aliases, global_aliases):
-            for alias in aliases:
-                if alias.name == alias_name:
-                    return True, alias
-
-        return False, None
+    async def initialize(self):
+        await self.bot.wait_until_red_ready()
+        await self._aliases.load_aliases()
+        # When bot level cache disable is added this can be surrounded by an if
+        # Then the AliasCache object just needs its cache_enabled flag set to False
 
     def is_command(self, alias_name: str) -> bool:
         """
@@ -99,56 +68,6 @@ class Alias(commands.Cog):
     @staticmethod
     def is_valid_alias_name(alias_name: str) -> bool:
         return not bool(search(r"\s", alias_name)) and alias_name.isprintable()
-
-    async def add_alias(
-        self, ctx: commands.Context, alias_name: str, command: str, global_: bool = False
-    ) -> AliasEntry:
-        indices = findall(r"{(\d*)}", command)
-        if indices:
-            try:
-                indices = [int(a[0]) for a in indices]
-            except IndexError:
-                raise ArgParseError(_("Arguments must be specified with a number."))
-            low = min(indices)
-            indices = [a - low for a in indices]
-            high = max(indices)
-            gaps = set(indices).symmetric_difference(range(high + 1))
-            if gaps:
-                raise ArgParseError(
-                    _("Arguments must be sequential. Missing arguments: ")
-                    + ", ".join(str(i + low) for i in gaps)
-                )
-            command = command.format(*(f"{{{i}}}" for i in range(-low, high + low + 1)))
-
-        alias = AliasEntry(alias_name, command, ctx.author, global_=global_)
-
-        if global_:
-            settings = self.config
-        else:
-            settings = self.config.guild(ctx.guild)
-            await settings.enabled.set(True)
-
-        async with settings.entries() as curr_aliases:
-            curr_aliases.append(alias.to_json())
-
-        return alias
-
-    async def delete_alias(
-        self, ctx: commands.Context, alias_name: str, global_: bool = False
-    ) -> bool:
-        if global_:
-            settings = self.config
-        else:
-            settings = self.config.guild(ctx.guild)
-
-        async with settings.entries() as aliases:
-            for alias in aliases:
-                alias_obj = AliasEntry.from_json(alias)
-                if alias_obj.name == alias_name:
-                    aliases.remove(alias)
-                    return True
-
-        return False
 
     async def get_prefix(self, message: discord.Message) -> str:
         """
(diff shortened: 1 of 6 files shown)


redbot/cogs/alias/alias_entry.py
```diff
@@ -30,34 +47,195 @@ def inc(self):
...
+            if not self._loaded:
+                await self.load_aliases()
+                self._loaded = True
+            for _, alias in self._aliases[None].items():
+                aliases.append(alias)
+        else:
+            aliases = [AliasEntry.from_json(d) for d in await self.config.entries()]
+        return aliases
+
+    async def is_alias(
```
Jackenmen: Considering this returns an `AliasEntry` object, not a `bool`, we should call this `get_alias` instead of `is_alias`.

redbot/cogs/alias/alias.py
```diff
@@ -424,32 +284,36 @@ async def _del_global_alias(self, ctx: commands.Context, alias_name: str):
     @commands.guild_only()
     async def _list_alias(self, ctx: commands.Context):
         """List the available aliases on this server."""
+        if not await self._aliases.get_guild_aliases(ctx.guild):
```
Jackenmen: This doesn't really matter with cache turned on but assuming it would be possible to disable in future, result of this coro should probably be assigned to a variable to avoid calling Config twice.

redbot/cogs/alias/alias_entry.py
```diff
@@ -30,34 +47,195 @@ def inc(self):
...
+            view.skip_ws()
+        return extra
+
     def to_json(self) -> dict:
-        try:
-            creator = str(self.creator.id)
-            guild = str(self.guild.id)
-        except AttributeError:
-            creator = self.creator
-            guild = self.guild
```
Jackenmen, line 39: There was a reason for this `try: except` here - we want the JSON dict to have user/guild ID, not Member/Guild object.

With that said, you could probably do an `if self.has_real_data` instead of `try: except`

redbot/cogs/alias/alias_entry.py
```diff
@@ -30,34 +47,195 @@ def inc(self):
...
+        self._cache_enabled = cache_enabled
+        self._loaded = False
+        self._aliases: Dict[Optional[int], Dict[str, AliasEntry]] = {None: {}}
+
+    async def load_aliases(self):
+        for alias in await self.config.entries():
+            self._aliases[None][alias["name"]] = AliasEntry.from_json(alias)
+
+        for guild_id in await self.config.all_guilds():
+            aliases = await self.config.guild_from_id(guild_id).entries()
```
Jackenmen: You shouldn't be calling Config on each guild again if you already have all the needed data.

Also, considering all the optimizations for big bots that we've done lately, this should probably be wrapped in `AsyncIter` class from `redbot.core.utils` to avoid blocking event loop on alias loading for bigger bots.

```suggestion
        all_guilds = await self.config.all_guilds()
        async for guild_id, guild_data in AsyncIter(all_guilds.items(), steps=100):
            aliases = guild_data["entries"]
```

redbot/cogs/alias/alias_entry.py
```diff
@@ -30,34 +47,195 @@ def inc(self):
...
+        return global_aliases + server_aliases
+
+    async def get_guild_aliases(self, guild: discord.Guild) -> List[AliasEntry]:
+        """Returns all guild specific aliases"""
+        aliases: List[AliasEntry] = []
+
+        if self._cache_enabled:
+            if not self._loaded:
+                await self.load_aliases()
+                self._loaded = True
```
Jackenmen: Could we do this on cog load instead? It would allow us to skip this check in every method accessing the cache and it would assure that this won't be called several times before the cache is fully loaded (i.e. someone spams an alias). If that's not an option for some reason, the parts that load aliases would need some lock.

redbot/cogs/alias/alias_entry.py
```diff
@@ -30,34 +47,195 @@ def inc(self):
...
+    async def load_aliases(self):
+        for alias in await self.config.entries():
+            self._aliases[None][alias["name"]] = AliasEntry.from_json(alias)
+
+        for guild_id in await self.config.all_guilds():
+            aliases = await self.config.guild_from_id(guild_id).entries()
+            if guild_id not in self._aliases:
+                self._aliases[guild_id] = {}
+            for alias in aliases:
+                self._aliases[guild_id][alias["name"]] = AliasEntry.from_json(alias)
```
Jackenmen: **(this is mostly a nitpick)**
We could use a dict comprehension instead here and (imo) simplify this.
```suggestion
            self._aliases[guild_id] = {
                alias["name"]: AliasEntry.from_json(alias) for alias in aliases
            }
```

redbot/cogs/alias/alias_entry.py
```diff
@@ -30,34 +47,195 @@ def inc(self):
...
+class AliasCache:
+    def __init__(self, config: Config, cache_enabled: bool = True):
+        self.config = config
+        self._cache_enabled = cache_enabled
+        self._loaded = False
+        self._aliases: Dict[Optional[int], Dict[str, AliasEntry]] = {None: {}}
+
+    async def load_aliases(self):
+        for alias in await self.config.entries():
+            self._aliases[None][alias["name"]] = AliasEntry.from_json(alias)
```
Jackenmen, line 98: **(this is mostly a nitpick)**
We could use a dict comprehension instead here and (imo) simplify this.
```suggestion
        aliases = await self.config.entries()
        self._aliases[None] = {alias["name"]: AliasEntry.from_json(alias) for alias in aliases}
```

redbot/cogs/alias/alias_entry.py
```diff
@@ -30,34 +47,195 @@ def inc(self):
...
+
+        if self._cache_enabled:
+            if not self._loaded:
+                await self.load_aliases()
+                self._loaded = True
+            if guild is not None:
+                if guild.id in self._aliases:
+                    server_aliases = list(self._aliases[guild.id].values())
+
+            global_aliases = list(self._aliases[None].values())
```
Jackenmen: It would be faster to search the alias with a key lookup in the cache dict instead of iterating through list of alias objects.

redbot/cogs/alias/alias_entry.py
```diff
@@ -30,34 +47,195 @@ def inc(self):
...
+            if guild:
+                server_aliases = [
+                    AliasEntry.from_json(d) for d in await self.config.guild(guild.id).entries()
+                ]
+            global_aliases = [AliasEntry.from_json(d) for d in await self.config.entries()]
+        all_aliases = global_aliases + server_aliases
+
+        for alias in all_aliases:
+            if alias.name == alias_name:
+                return alias
```
Jackenmen: Having these as iterators as they were before and chaining them would probably be a little better **however you can consider this a nitpick**

redbot/cogs/alias/alias_entry.py
```diff
@@ -30,34 +47,195 @@ def inc(self):
...
+        async with settings.entries() as aliases:
+            for alias in aliases:
+                alias_obj = AliasEntry.from_json(alias)
+                if alias_obj.name == alias_name:
+                    aliases.remove(alias)
+                    if self._cache_enabled:
+                        if global_:
+                            del self._aliases[None][alias_obj.name]
+                        else:
+                            del self._aliases[ctx.guild.id][alias_obj.name]
```
Jackenmen: **I know this isn't a part of this PR so if you'll leave as is, it's fine.**

Could we skip creating alias entry for each alias and just check for `alias["name"]`?
```suggestion
                if alias["name"] == alias_name:
                    aliases.remove(alias)
                    if self._cache_enabled:
                        if global_:
                            del self._aliases[None][alias["name"]
                        else:
                            del self._aliases[ctx.guild.id][alias["name"]]
```

redbot/cogs/alias/alias.py
```diff
@@ -53,40 +48,7 @@ def __init__(self, bot: Red):
...
-    async def loaded_aliases(self, guild: discord.Guild) -> Generator[AliasEntry, None, None]:
-        return (
-            AliasEntry.from_json(d, bot=self.bot)
-            for d in (await self.config.guild(guild).entries())
-        )
-
-    async def loaded_global_aliases(self) -> Generator[AliasEntry, None, None]:
-        return (AliasEntry.from_json(d, bot=self.bot) for d in (await self.config.entries()))
-
-    async def is_alias(
```
Jackenmen, line 72: This method is used in two other places outside Alias cog:
https://github.com/Cog-Creators/Red-DiscordBot/search?q=bot.get_cog%28%22Alias%22%29&unscoped_q=bot.get_cog%28%22Alias%22%29

redbot/cogs/alias/alias.py
```diff
@@ -424,32 +284,36 @@ async def _del_global_alias(self, ctx: commands.Context, alias_name: str):
...
-        if len(names) == 0:
-            await ctx.send(_("There are no aliases on this server."))
-        else:
-            await ctx.send(box("\n".join(names), "diff"))
+        await ctx.send(box("\n".join(names), "diff"))
 
     @global_.command(name="list")
     async def _list_global_alias(self, ctx: commands.Context):
         """List the available global aliases on this bot."""
+        if not await self._aliases.get_global_aliases():
```
Jackenmen: This doesn't really matter with cache turned on but assuming it would be possible to disable in future, result of this coro should probably be assigned to a variable to avoid calling Config twice.

Jackenmen (changes requested):

redbot/cogs/alias/alias_entry.py
TrustyJAID: This was removed because I ensured all creation of AliasEntry objects only accepted the ID not the object itself. I'll further ensure that with typehints to only accept int and not the object themselves. They served no purpose being the object and just a reference ID is good enough.

redbot/cogs/alias/alias_entry.py
TrustyJAID: I was going to but that breaks our tests for the cog.

redbot/cogs/alias/alias_entry.py
Jackenmen: I was basing this comment on `from_json` method which could change it into Member/Guild objects if bot was passed. I know it's never passed so maybe you could just remove this all together or at least add a comment that some changes to `AliasEntry` would be needed if we would start to pass the bot object

redbot/cogs/alias/alias_entry.py
Jackenmen: You can do the loading in the alias fixture here: https://github.com/TrustyJAID/Red-DiscordBot/blob/alias-caching/redbot/pytest/alias.py#L12

redbot/cogs/alias/alias_entry.py
TrustyJAID: I feel like the current method is more readable and easier for others to verify it's working correctly. List and Dict comprehensions while sometimes make the code look cleaner can also make it less readable and this is one where I think it could be misinterpreted.

redbot/cogs/alias/alias_entry.py
```diff
@@ -30,34 +42,182 @@ def inc(self):
...
+        """Returns an AliasEntry object if the provided alias_name is a registered alias"""
+        server_aliases: List[AliasEntry] = []
+
+        if self._cache_enabled:
+            if guild is not None:
+                if guild.id in self._aliases:
+                    if alias_name in self._aliases[guild.id]:
+                        return self._aliases[guild.id][alias_name]
+            if alias_name in self._aliases[None]:
+                return self._aliases[None][alias_name]
```
Jackenmen: These need to be opposite order - global aliases take precedence over guild aliases.
```suggestion
            if alias_name in self._aliases[None]:
                return self._aliases[None][alias_name]
            if guild is not None:
                if guild.id in self._aliases:
                    if alias_name in self._aliases[guild.id]:
                        return self._aliases[guild.id][alias_name]
```

​

If you care for my nitpicks, this code would avoid doing key lookups multiple times, otherwise just ignore this part, I couldn't stop myself from putting this here :P
```suggestion
            if alias := self._aliases[None].get(alias_name):
                return alias
            if guild is not None:
                if guild_aliases := self._aliases.get(guild.id):
                    if alias := guild_aliases.get(guild.id):
                        return alias
```

Jackenmen (changes requested):

Jackenmen (dismissed):

Jackenmen (approved):
Looks good, thanks so much for the PR <3

PR #3793, opened 2020-04-25
Fix behavior of `is_owner` for team applications and put all owner IDs in one attribute

Jackenmen:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

All owner IDs (application owners, co-owners, owner passed by `--owner` flag or set through Config) are now put into one set under `bot.owner_ids` attribute.

Behaviour of bot owners in Red after this PR should be as follows:
a) If bot's application belongs to a team, owners are:
>    1. everyone passed to `--co-owner` flag
>    2. user passed to `--owner` flag or if it's not passed, user that was set through Config (i.e. `redbot --edit` or the no longer existing `[p]set owner`) or **if neither exist**, this part is skipped
>    3. **IF** `--team-members-are-owners` flag is passed, members of application's team are also the owners

b) If bot's application belongs to a user, owners are:
>    1. everyone passed to `--co-owner` flag
>    2. user passed to `--owner` flag or if it's not passed, user that was set through Config (i.e. `redbot --edit` or the no longer existing `[p]set owner`) or **if neither exist**, owner of bot's application is used

Before this PR, `--team-members-are-owners` flag didn't work if `--owner` flag was passed or if there was a user set in Config already.

Two things that I'm not sure about:
1. Should owner saved in Config (not talking about the one explicitly set through `--owner` flag as that one should definitely be included) be included when `--team-members-are-owners` flag is used? Currently it is, but I should be able to easily change it.
**(Mr. Reviewer, I want a response for ↑ one)**
2. Is there any good reason why co-owners were never exposed in public attribute before? IMO, they should be, but maybe I'm missing something.

// I also added a warning if bot is started with no owner set, we might try to exit from there in future but for now I'm leaving this as just a warning.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -77,7 +77,6 @@ class RedBase(
         self._shutdown_mode = ExitCodes.CRITICAL
         self._cli_flags = cli_flags
         self._config = Config.get_core_conf(force_registration=False)
-        self._co_owners = cli_flags.co_owner
         self.rpc_enabled = cli_flags.rpc
         self.rpc_port = cli_flags.rpc_port
         self._last_exception = None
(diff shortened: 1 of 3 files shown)


Kowlin (approved):
Works like a charm

PR #3794, opened 2020-04-26
Show current driver in `[p]debuginfo`

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -21,6 +21,7 @@ import aiohttp
 import discord
 import pkg_resources
 from babel import Locale as BabelLocale, UnknownLocaleError
+from redbot.core.data_manager import storage_type
 
 from . import (
     __version__,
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -1774,6 +1775,7 @@ async def debuginfo(self, ctx: commands.Context):
                 value=escape(sys.executable, formatting=True),
                 inline=False,
             )
+            e.add_field(name="Red Driver", value=driver, inline=False)
```
```suggestion
            e.add_field(name="Storage type", value=driver, inline=False)
```

redbot/core/core_commands.py
```diff
@@ -1786,6 +1788,7 @@ async def debuginfo(self, ctx: commands.Context):
                 + "System arch: {}\n".format(platform.machine())
                 + "User: {}\n".format(user_who_ran)
                 + "OS version: {}\n".format(osver)
+                + "Red Driver: {}\n".format(driver)
```
```suggestion
                + "Storage type: {}\n".format(driver)
```

Jackenmen (commented):
Looks good to me, I think using "Storage type" instead of "Red Driver" would be better but it's fine either way so I'll leave it up to you if you want to change it.

Jackenmen (approved):

PR #3807, opened 2020-04-29
Local Cache but Global

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

Enable audio to make use to the Global Cache service.


Drapersniper:
Will be implemented on the rewrite

PR #3815, opened 2020-04-30
Make server lock work again

### LINKED ISSUE #3814: [Admin] Serverlock doesn't function.
# Command bugs

#### Command name

!serverlock

[Admin] Cog

For people not being able to invite the bot into their server, when the serverlock feature is enabled.

#### What actually happened?

People where able to invite the bot into their server. even when the serverlock feature states it was serverlocked.

1. have the bot in 1 server, and use !serverlock so that it says: "The bot is now serverlocked."
2. Do !invite and the bot will DM you the invite link. follow that link and invite the bot into your server.
3. It will still join said server. Even though it shouldn't.

On my instance which is btw the latest version of Red 3.3.7 - it joins a server from someone that has done that, even though i have the bot serverlocked. i even disabled the Invite command. which prevents most of it, however some people have the invite link saved in their DM's and use that instead the bot. which serverlock normally would prevent, but now doesn't.

Hope this report helps people.

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #3814

* Disclaimer: Untested :P

### DIFF
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -467,6 +467,7 @@ class Admin(commands.Cog):
             await ctx.send(_("The bot is now serverlocked."))
 
     # region Event Handlers
+    @commands.Cog.listener()
     async def on_guild_join(self, guild: discord.Guild):
         if await self.config.serverlocked():
             await guild.leave()


TrustyJAID (approved):
Looks good to me.

PR #3816, opened 2020-04-30
Stop allowing spaces in custom commands names

NeuroAssassin:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

`[p]"testing also"` apparently doesn't work, unfortunately.

@jack1142 ready for another 10 commits of fixing this PR?

### DIFF
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -292,6 +292,10 @@ class CustomCommands(commands.Cog):
 
         Note: This command is interactive.
         """
+        if " " in command:
+            # Haha, nice try
+            await ctx.send(_("Custom command names cannot have spaces in them."))
+            return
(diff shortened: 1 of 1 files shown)


redbot/cogs/customcom/customcom.py
```diff
@@ -292,6 +292,10 @@ async def cc_create_random(self, ctx: commands.Context, command: str.lower):
 
         Note: This command is interactive.
         """
+        if " " in command:
```
Jackenmen: Considering that d.py checks for any whitespace character when getting parts of the command, I wonder if we shouldn't be doing the same thing here too, for example:
```suggestion
        if any(char.isspace() for char in command):
```

d.py source: https://github.com/Rapptz/discord.py/blob/master/discord/ext/commands/view.py#L119

NeuroAssassin: Done

Jackenmen (approved):
LGTM

PR #3819, opened 2020-05-02
Show default avatars in `[p]userinfo` when no avatar is set

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Since default avatars are exposed on d.py, we now can use them on `[p]userinfo`.


Jackenmen (approved):
Looks good to me.

PR #3829, opened 2020-05-07
Fix creation of IdentifierData in config raw methods

### LINKED ISSUE #3796: `[p]permissions acl setguild` is broken for Postgres driver
# Command bugs

#### Command name

`[p]permissions acl setguild`

Permissions

I expected ACL to be set properly

#### What actually happened?

The whole ACL for the guild got borked:
image
- `default` and `176070082584248320` are keys that are result of using `[p]permissions acl setguild`
- `8` represents how it should look - the rule was added with a `[p]permissions addguildrule deny 8 57287406247743488` command right after setting ACL

For the reference, this is how YAML file given to `[p]permissions acl setguild` looks:
```yaml
COG: {}
COMMAND:
  ping:
    default: false
    176070082584248320: true
  roll:
    default: true
    176070082584248320: false
```

Use `[p]permissions acl setguild` on a bot using Postgres driver.

Tobotimus:
### Type

- Bugfix

Resolves #3796.

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -334,7 +334,7 @@ class Group(Value):
         """
         is_group = self.is_group(item)
         is_value = not is_group and self.is_value(item)
-        new_identifiers = self.identifier_data.add_identifier(item)
+        new_identifiers = self.identifier_data.get_child(item)
         if is_group:
             return Group(
                 identifier_data=new_identifiers,
@@ -381,7 +381,7 @@ class Group(Value):
             dict access. These are casted to `str` for you.
         """
         path = tuple(str(p) for p in nested_path)
-        identifier_data = self.identifier_data.add_identifier(*path)
+        identifier_data = self.identifier_data.get_child(*path)
         await self.driver.clear(identifier_data)
 
     def is_group(self, item: Any) -> bool:
@@ -499,7 +499,7 @@ class Group(Value):
             else:
                 default = poss_default
 
-        identifier_data = self.identifier_data.add_identifier(*path)
+        identifier_data = self.identifier_data.get_child(*path)
         try:
             raw = await self.driver.get(identifier_data)
         except KeyError:
@@ -583,7 +583,7 @@ class Group(Value):
             The value to store.
         """
         path = tuple(str(p) for p in nested_path)
-        identifier_data = self.identifier_data.add_identifier(*path)
+        identifier_data = self.identifier_data.get_child(*path)
         if isinstance(value, dict):
             value = _str_key_dict(value)
         await self.driver.set(identifier_data, value=value)
--- a/redbot/core/drivers/base.py
+++ b/redbot/core/drivers/base.py
@@ -109,6 +109,28 @@ class IdentifierData:
     def __hash__(self) -> int:
         return hash((self.uuid, self.category, self.primary_key, self.identifiers))
 
+    def get_child(self, *keys: str) -> "IdentifierData":
+        if not all(isinstance(i, str) for i in keys):
+            raise ValueError("Identifiers must be strings.")
+
+        primary_keys = self.primary_key
+        identifiers = self.identifiers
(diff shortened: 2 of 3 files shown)


Jackenmen (dismissed):
Looks good to me, if you want to, we could use some tests for clear/clear_raw on partial pkeys as well but it isn't important enough to block this.

Thanks so much for this PR :)

mikeshardmind:
I haven't had a chance to check this myself yet. While at a glance, and based on my memory, it looks like it should, does this handle the case where people used this intentionally to bulk set values with config?

Should there also be a test for this case so that this driver (and future ones, see #3832 #3831 ) continue supporting the same overall usage without extra edge cases that aren't either disallowed or broken?

Tobotimus:
@mikeshardmind, I think those test cases already exist in the PR, unless there are some other ones I'm missing?

mikeshardmind:
Unless I'm missing something, all of the test cases in the PR reduce to a singular nested config value. There isn't a test to ensure dicts and subdicts get split properly should the set be caused by a the set which occurs when exiting the context manager over a group. This is something which has been used in cogs, and does work on JSON, and isn't restricted by config (and should work in postgres too IIRC the logic for multiple documents) so it would be good to ensure that this is tested for consistency in the future

Edit: Sorry, I wasn't fully clear on why/what's missing from this, continuing below...

While the test cases here cover all the ways in which the IdentifierData class might exist, the case where this was problematic isn't actually currently covered by a test. none of the tests in config actually check the behavior setting above a group with primary keys not all in the group itself.

Jackenmen:
> Edit: Sorry, I wasn't fully clear on why/what's missing from this, continuing below...
> 
> While the test cases here cover all the ways in which the IdentifierData class might exist, the case where this was problematic isn't actually currently covered by a test. none of the tests in config actually check the behavior setting above a group with primary keys not all in the group itself.

So you're thinking of a test for wrong input? e.g.
```py
config.init_custom("TEST", 3)
await config.custom("TEST").set({"1": {"foo": "bar"}}) # incomplete primary keys - only 1
```

mikeshardmind:
No, sorry, I mean that because the JSON driver and other drivers are going to deal with groups fundamentally differently, the fact that none of the config tests make any attempt at setting above a "normal" group leaves room for surprises in the differences between the drivers.

Here's an example of something which should either be guaranteed to work across all drivers, or be disallowed, but that there is no test for.

```py
config.init_custom("TEST", 1)
await config.custom("TEST").set(
    {"identifier_one": {"some_valname" : 1}, "identifier_two": {"some_valname": 1}}
)
```
~~Notice that in this case, there should actually be 2 instances of identifer data and values~~
It would be easy to assume this should require 2 seperate set operations with 2 different instances of identifier data, however, this can happen relatively naturally since config.custom is documented as returning a config group, and config groups are documented as being able to be used as a context manager to modify all values within. So this somewhat unnatural looking set would be the resulting call from `__aexit__` in the below. 

```py
config.init_custom("TEST", 1)

group = config.custom("TEST")
async with group() as test_data:
    for identifier in ("identifier_one", "identifier_two"):
        test_data[identifier] = {"some_valname": 1}
```

All of this is technically within what config supports as documented, and the JSON driver handles this with zero surprises, though this is more a consequence of the JSON driver not making a distinction between config groups and values which are dicts, as was already brought up before.

If there's going to be more drivers added, the behavior of this should probably be ensured to be uniform across all of them.

Tobotimus:
@mikeshardmind thanks for bringing that up, I'll add a test for that today.

This usage was in fact always supported, we just never had a test for it and thus the bug in config never surfaced until later. The postgres driver supports it still as a single set() operation and a single INSERT query which inserts many rows

Tobotimus:
Tests added

Jackenmen (approved):
Looks good to me, we will have to look into our tests for Config more before pursuing all the new drivers but this PR contains the needed tests for the stuff it's supposed takes care of.

Issue #3112, opened 2019-11-11
[Informational/RFC] The removal of MongoDB as a backend choice and the future of Config


mikeshardmind:
## Why remove support for MongoDB?

Recently, we realized a larger issue with Red's config existed. It was an awkward issue in which something we assumed to be true but did not strictly ensure, created several options nobody deciding was happy about.

### The options available to us were:

 - Modify the mongo driver to support the edge case.
 - Modify config to be stricter and prevent the edge case.
 - Drop Mongo and let the fact that the other drivers handled the edge case gracefully be enough.

The first option here was very quickly ruled out. We did not have anyone actively developing for Red who wanted to write a large amount of logic for this. Mongo had already been proven to be less suited to our use than originally thought when added. The various issues we've had to work around for it over time made people hesitant to put keep putting more work into it with the limited time available to the people working on Red.

The second option was considered, however, the number of users it could break was unquantifiable. 
The large majority of our users use the JSON driver. There are also a large number of private or at least not publicized cogs in existence. Changing config here, especially so close to our next release could have broken any number of these cogs and users, without easy handling available as it would require cog creators changing (and migrating) their data structures.

The third option was also unpopular among people deciding on this, though ultimately is what we have gone with. (#3099) At least people who are using MongoDB have alternatives, and those alternatives will not require any knowledge to swap to beyond the ability to follow basic instructions.
(#3108)

## So what does this mean for Config going forward?

Red's config module should largely be in a stable place from here forward. There are still things we would like to improve based on what we've learned from it and from how users have given feedback on how they find it to be useful and what they would like to see in it.

However, I don't think Config should be changed any further going forward.

### What is Config2?

Just a separate module allowing us to keep Config as-is for people to keep using without any additional work for existing cogs. While Config2 is intended to be easy to switch to, we don't need to tear down Config to make that a reality.

### The goals of Config2

 - Make certain restrictions on data stricter.
    - This actually would include the case which led to MongoDB's removal as a backend option for config.
    - A clear set of allowed data types which we enforce consistently will allow more backends, as some backends were previously excluded from consideration because they needed to deal with some levels of ambiguity.
    - This might require some additional work on the part of cog creators, and is an additional reason to leave in the existing config for being easier to use initially.
  - Make it easier to extend to other uses.
    - Removal of ``__getattr__`` use
    - Add a clear way to extend support for specific types
  - Have a clear path to move from config to config2 if a creator wants to.
[...]

PR #3835, opened 2020-05-07
[A Config Alternative]

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

- This provides the bare necessities of storing data in Red. It
  doesn't do anything fancy.
- it utilizes an extremely small standalone library I've personally
  written as a means of storing non-relational
  data quickly.
- This is directly related to ideas originally expressed in #3112
- This is a draft form of the technical side of things, I won't put in
  the time on documenting this if the ideas are no longer wanted. If the
  slimmer alternative is still wanted, I'll get to documenting, but I
  wanted to be able to show the direction I was willing to provide
  before putting in more on this to ensure it was still something people
  more active in the project were still interested in

### Limitations and considerations

1. The provided wrapper here in red isn't technically necessary, and cog devs could easily use the library without its inclusion in Red. However, there is a lot of code which would be repeated due to how red handles cog imports, and that shared libs have been deprecated.

2. This allows for less arbitrary identifiers. I consider this a necessary concession to get the most out of it while still not requiring cog devs be committing to a full DB schema

3. Nested access isn't handled for cog devs. What they put in is what they get back. This makes it less magical, but it improves performance significantly.

4. Stored values must be msgpack serializable with this wrapper. This actually allows storing things config can't store, such as dicts utilizing tuple keys. (This can be changed, the serialization can be swapped out)

5. There are no docs yet. I'll be fixing this on the library side first, then if this is still something which this project is interested in, I'll update this PR with additional documentation.

6. This isn't a replacement for config. It's another tool with different considerations. It's slimmer, should perform faster for most use cases, but it also does less for the user automatically.

### DIFF
--- a/redbot/core/rapid_storage.py
+++ b/redbot/core/rapid_storage.py
@@ -1,0 +1,55 @@
+from rapid_dev_storage import SQLiteBackend, Storage, StoredValue, StorageGroup, NoValue
+import discord
+
+from .data_manager import cog_data_path
+
+__all__ = ["RapidStorage", "StorageGroup", "StoredValue", "NoValue"]
+
+
+class RapidStorage(Storage):
+
+    @classmethod
+    async def get_cog_storage(cls, cog_name: str, unique_identifier: int):
+
+        path = cog_data_path(raw_name=cog_name) / "rapid_storage.db"
+        backend = await SQLiteBackend.create_backend_instance(path, cog_name, unique_identifier)
+        return cls(backend)
+
+    @property
+    def botwide(self) -> StorageGroup:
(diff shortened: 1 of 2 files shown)


mikeshardmind:
Someone else can add a wrapper for red if people want it, it works without such a wrapper's inclusion, and such a wrapper is trivial to write correctly. I've not gotten so much as a "still interested in the idea" from people in the discord when bringing it up, and I'm not going to be more invested in this happening for red than people who still use red are, I was just going to offer something back to the project since this project was one of the original things which pushed me to consider how the average (non-DBA) developer interacts with storing data.

PR #3848, opened 2020-05-13
Add a note about how to manage selfroles to help text for selfrole

RheingoldRiver:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Add discoverability of selfroleset to selfrole doc/help text

### DIFF
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -367,7 +367,7 @@ class Admin(commands.Cog):
     @commands.guild_only()
     @commands.group()
     async def selfrole(self, ctx: commands.Context):
-        """Apply selfroles."""
+        """Apply selfroles. Admins: see selfroleset to edit this list."""
         pass
 
     @selfrole.command(name="add")


Jackenmen:
Hi, sorry this has gone unnoticed for so long... I think that we're not currently interested in this - this tip shows for everyone trying to use the command which can confuse regular users, and the commands are already discoverable through the help command *and* now also the Admin cog guide.
A possible alternative here would be to have this tip shown dynamically *only* if the command's caller is an Admin but I believe that this would add unwanted complexity that isn't present anywhere else in the Core Red currently.
Perhaps if a more convincing use case for the same complexity presents itself in the future, we'll consider doing the same thing here but for now, I'll be closing this PR to not have this issue be in limbo forever.

RheingoldRiver:
Hi, I understand the reasoning, but I just want to note that just today I was unable to remember the command for changing selfroles and I couldn't find it after about 5 minutes of looking. I thought it was in `set`, read through all those commands, couldn't find `adminset`, finally I remembered that I had aliased `asar` and I was able to use it. My `help` command is currently 10 messages on one of the bots I admin and 18 on the other; that's effectively useless and may as well not exist, I'm never going to use it for anything. While it's cool that there's an admin cog guide, a progressive disclosure pattern does require an entry point for the secondary information, which is currently missing: currently the progressive part is missing (I had no idea that there was an admin cog guide).

I've been using Red for over 3 years now and I'm still stumbling through the admin cogs and I have no idea what anything is called or how to find it, I feel like additional help texts is a worthy goal and not unnecessary complexity in the code. I think this would actually be an awesome feature to add throughout all of the core commands to add extra documentation helping admins use & configure core features.

PR #3857, opened 2020-05-18
Fix local blacklist and whitelist commands

NeuroAssassin:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

#3643 was a complete lie

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1945,8 +1945,8 @@ class Core(commands.Cog, CoreLogic):
             await ctx.send_help()
             return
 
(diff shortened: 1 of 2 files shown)


Jackenmen (changes requested):
The fixes in here work, however there's one more thing that's not working - `[p]localblacklist clear`

This line:
https://github.com/NeuroAssassin/Red-DiscordBot/blob/f778c600954cfefad50cf2489324a26d1f85deaf/redbot/core/core_commands.py#L2080

Should be:
```py
        await self.bot._whiteblacklist_cache.clear_blacklist(ctx.guild)
```

Jackenmen (approved):

PR #3862, opened 2020-05-22
[Docs] Add information about provisional status of RPC

NeuroAssassin:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature
- [X] Other

Document the current state of support for RPC functionality

See message in #v3-burndown on Red's main server for the announcement:
https://discord.com/channels/133049272517001216/411381123101491200/714560168465137694

### DIFF
--- a/docs/framework_rpc.rst
+++ b/docs/framework_rpc.rst
@@ -4,6 +4,10 @@
 RPC
 ===
 
+.. important::
+
+  RPC support is included in Red on a provisional basis. Backwards incompatible changes (up to and including removal of the RPC) may occur if deemed necessary.
+
 V3 comes default with an internal RPC server that may be used to remotely control the bot in various ways.
 Cogs must register functions to be exposed to RPC clients.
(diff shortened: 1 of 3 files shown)


Jackenmen:
This PR seems fine to me, though I wonder if it would make sense to also add some notice to docstring of `(un)register_rpc_handler` methods.

NeuroAssassin:
I personally think users will be more likely to look at the docs than the docstrings, as I do not usually look at docstrings for methods that are not often used, but I can see the reason behind putting it in the docstring.

Jackenmen:
> I personally think users will be more likely to look at the docs than the docstrings

docstrings are part of the docs - here: https://docs.discord.red/en/latest/framework_bot.html#redbot.core.bot.RedBase.register_rpc_handler

I think we should add the notice to them, just so this information won't get missed if someone would only look at this part of docs.

Jackenmen (approved):

PR #3865, opened 2020-05-22
Case-Insensitive SelfRole

### LINKED ISSUE #3853: Make selfrole case insensitive
# Feature request

- [x] Cog
- [ ] Command
- [ ] API functionality

In Admin cog, could selfrole be made case insensitive? In the event that a server has two separate roles that are both allowed in selfrole, it could print a warning about the situation every time a role is added or removed via selfrole.

chasehult:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This allows input into selfrole to be case insensitive in accordance with https://github.com/Cog-Creators/Red-DiscordBot/issues/3853


redbot/cogs/admin/converters.py
```diff
@@ -7,15 +7,33 @@
 
 class SelfRole(commands.Converter):
     async def convert(self, ctx: commands.Context, arg: str) -> discord.Role:
-        admin = ctx.command.cog
+        admin = ctx.command.cog.bot.get_cog("Admin")
```
Drapersniper: This line is not necessary since the SelfRole converter is a converter specifically for the admin cog.

And a FYI, this also incorrectly assumes that every cog instance has a bot attribute, which is not true.

redbot/cogs/admin/converters.py
```diff
@@ -7,15 +7,33 @@
...
             raise commands.BadArgument(_("The Admin cog is not loaded."))
 
+        selfroles = await admin.config.guild(ctx.guild).selfroles()
         role_converter = commands.RoleConverter()
-        role = await role_converter.convert(ctx, arg)
 
-        selfroles = await admin.config.guild(ctx.guild).selfroles()
+        pool = []
+        for role_id in selfroles:
+            role = ctx.guild.get_role(role_id)
```
Drapersniper, line 20: Since there can be quite a few role I'd personally suggest using AsyncIter here, yes won't make much of a diff but it's good to get in the habit of using it for medium/large loops that does a lot of processing

redbot/cogs/admin/converters.py
```diff
@@ -7,15 +7,33 @@
...
+            if role is None:
+                continue
+            if role.name.lower() == arg.lower():
```
Drapersniper: What if the argument is a role ID, you didn't account for that here

redbot/cogs/admin/converters.py
```diff
@@ -7,15 +7,33 @@
...
+                pool.append(role)
+
+        if len(pool) == 0:
```
Drapersniper:
```suggestion
        if not pool :
```

redbot/cogs/admin/converters.py
```diff
@@ -7,15 +7,33 @@
...
+        admin = ctx.command.cog.bot.get_cog("Admin")
         if admin is None:
             raise commands.BadArgument(_("The Admin cog is not loaded."))
 
+        selfroles = await admin.config.guild(ctx.guild).selfroles()
         role_converter = commands.RoleConverter()
-        role = await role_converter.convert(ctx, arg)
 
-        selfroles = await admin.config.guild(ctx.guild).selfroles()
+        pool = []
```
Drapersniper:
```suggestion
        pool = set()
```

redbot/cogs/admin/converters.py
```diff
@@ -7,15 +7,33 @@
...
+        for role_id in selfroles:
+            role = ctx.guild.get_role(role_id)
+            if role is None:
+                continue
+            if role.name.lower() == arg.lower():
+                pool.append(role)
```
Drapersniper:
```suggestion
                pool.add(role)
```

redbot/cogs/admin/converters.py
```diff
@@ -7,15 +7,33 @@
...
+                raise commands.BadArgument(_("The provided role is not a valid selfrole."))
+        elif len(pool) > 1:
+            raise commands.BadArgument(
+                _(
+                    "This selfrole has more than one capitalization"
+                    " possibilities.  Please inform a moderator."
+                )
+            )
+        else:
+            role = pool[0]
```
Drapersniper:
```suggestion
            role = pool.pop()
```

Drapersniper (changes requested):

redbot/cogs/admin/converters.py
chasehult: Oh yeah whoops, I repurposed that line from old code that I was using it for.

redbot/cogs/admin/converters.py
chasehult: I did.  It would catch in the if `len(pool) == 0` suite.  Unless there's a role that happens to be named something like `355573989923487765`, which I would think should have priority over a role with that as its ID.  That's a VERY edge case though.

redbot/cogs/admin/converters.py
```diff
@@ -15,7 +16,7 @@ async def convert(self, ctx: commands.Context, arg: str) -> discord.Role:
         role_converter = commands.RoleConverter()
 
         pool = set()
-        for role_id in selfroles:
+        for role_id in AsyncIter(selfroles):
```
Drapersniper:
```suggestion
        async for role_id in AsyncIter(selfroles, steps=100):
```

Drapersniper (changes requested):
AsyncIter returns an async iterator, so this needs to be changed here.

Flame442 (approved):
Thanks!

PR #3868, opened 2020-05-26
Special case new RLL message (RLL 0.5.1 bump inside)

Drapersniper:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/cogs/audio/core/utilities/player.py
+++ b/redbot/cogs/audio/core/utilities/player.py
@@ -386,7 +386,10 @@ class PlayerUtilities(MixinMeta, metaclass=CompositeMetaClass):
                 title = _("Nothing found.")
                 embed = discord.Embed(title=title)
                 if result.exception_message:
-                    embed.set_footer(text=result.exception_message[:2000].replace("\n", ""))
+                    if "Status Code" in result.exception_message:
+                        embed.set_footer(text=result.exception_message[:2000])
+                    else:
+                        embed.set_footer(text=result.exception_message[:2000].replace("\n", ""))
                 if await self.config.use_external_lavalink() and query.is_local:
                     embed.description = _(
                         "Local tracks will not work "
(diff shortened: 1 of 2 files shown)


Jackenmen (dismissed):
Works great with https://github.com/Cog-Creators/Red-Lavalink/pull/78
I wonder if it wouldn't be good to set an embed description to something like "Something went wrong when fetching track" or whatever as all (or most) other errors in Audio have a concise friendly error and I wouldn't really say this is a concise or friendly error:
image
Example of what it could look like with additional one-liner:
image

I don't think this the most important thing so I'm approving this PR anyway, but I think the addition I mentioned is worth considering.

Drapersniper:
it wouldnt follow the format set by audio for all of its error messages.

this is something to consider for the rewrite though...  other things can be in the description field

Jackenmen:
I guess what I mean is that the error in footer is quite verbose and "Nothing found." text while simple, doesn't really explain *that* much about it and isn't exactly correct either. This isn't a most typical error so I just went with something generic but I guess even the error "Service Unavailable" sent after "Message: " could work *somewhat* better than "Nothing found."
I'm not convinced about either of these so approving status quo is better than leaving the error like this 😄:
image

Drapersniper:
To expand on it the other exceptions, are exceptions provided by the jar, this is the one of new that arent, so the handling of it is iffy at best... (all other exceptions provided by the jar are handled in the footer, so further special casing this imo would:
1. Overly complicate the code
2. Add something that will be removed shortly.
3. Break from current pattern

Although like you mentioned ... adding it does make it clearer to the user so i'll leave aika to make the final call on this.

aikaterna:
After a discussion in Discord I agree about adding the clarification that Jack asked for, but my opinion is that it should be done in a subsequent PR as it's not something simple to tack onto this one in a quick manner. I agree with this PR and will be approving it but we should refer back to this later on the audio todo list.

aikaterna (dismissed):

Jackenmen (approved):
Bumped RLL version to 0.5.1, reapproving.

Will need to rerun checks when RLL 0.5.1 is released.

PR #3845, opened 2020-05-12
Preparations for d.py 1.4 (includes breaking changes related to mass mentions)

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

### Description of the changes
This PR contains changes needed due to d.py 1.4 release as well as this (breaking) change:
> 1. By default, none of the `.send()` methods will mention roles or @​everyone/@​here.
> 1.1. d.py 1.4 introduces allowed_mentions kwarg to `.send()` methods that can be used to make the message mention roles or @​everyone/@​here (NOTICE: this may still change before d.py 1.4 release)
> 1.2. `sanitize_roles` kwarg mentioned in the previous breaking change notice from Flame will NOT be implemented.
> 1.3. The default value of the `filter` kwarg in `Context.send()` will be changed to `None` which means that enabling @​everyone/@​here mentions will now require using the `allowed_mentions` kwarg instead.

Fixes #4069 (indirectly, through dependency update)

PR #3870, opened 2020-05-27
Add non-generic message when loading a cog with command name that is already registered

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

@Drapersniper this is what's needed to have a friendly message:
image

~~We might want to go for a regex instead though, as that would allow us to improve this message easily now or in future, instead of being limited by whatever d.py already gives us in exception message.~~
[Update 1](https://github.com/Cog-Creators/Red-DiscordBot/pull/3870#issuecomment-634992977): ~~I already did that, see comment below.~~
[Update 2](https://github.com/Cog-Creators/Red-DiscordBot/pull/3870#issuecomment-646903416): Fuck all these 
hacky ideas, Danny agreed on adding `CommandRegistrationError` that solves all our problems <3

~~In any case, this isn't the most future-proof implementation, but I think it's still worth the enhancement we get nonetheless.~~ see [update 2](https://github.com/Cog-Creators/Red-DiscordBot/pull/3870#issuecomment-646903416)

Blocked by #3845

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -60,6 +60,14 @@ _ = i18n.Translator("Core", __file__)
 
 TokenConverter = commands.get_dict_converter(delims=[" ", ",", ";"])
 
+# DEP-WARN
+# we're basing this handling on d.py's exception message which might change
+_ALREADY_REGISTERED_PATTERN = re.compile(
+    r"The alias (?P<alias>.+) is already an existing command or alias\."
+    r"|"
+    r"Command (?P<command>.+) is already registered\."
+)
+
 
 class CoreLogic:
     def __init__(self, bot: "Red"):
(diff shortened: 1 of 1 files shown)


Jackenmen:
Yeah... Regex definitely seems like a better idea here, especially if we want this to support i18n properly.
I'm not entirely happy that we have to match it like that, but imo it's worth it so I'll live.

Jackenmen:
This is now depending on d.py 1.4 ([this change](https://github.com/Rapptz/discord.py/pull/4217) to be exact)

The implementation can now actually be considered future-proof and not hacky so that's good 😄

TrustyJAID (approved):
This is a really cool addition and great quality of life for end users to know why a cog fails to load.

PR #3878, opened 2020-05-29
[Audio] Redefine max length with livestream exception

### LINKED ISSUE #3877: [Audio] Max length should ignore livestreams
# Command bugs

#### Command name

[p]play with [p]audioset maxlength set to a reasonable value like 1h

Audio

For audio to deny playback of an item like "10 hours of x" but still allow livestreams/icecast/shoutcast/http streams

#### What actually happened?

"Track exceeds maximum length." when attempting to play a http stream

1. [p]audioset maxlength 3600
2. [p]audioset restrict       (toggled to false)
3. [p]play http://ice6.somafm.com/gsclassic-128-mp3

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Resolves #3877

### DIFF
--- a/redbot/cogs/audio/apis/interface.py
+++ b/redbot/cogs/audio/apis/interface.py
@@ -566,7 +566,7 @@ class AudioAPIInterface:
                     if len(player.queue) >= 10000:
                         continue
                     if guild_data["maxlength"] > 0:
-                        if self.cog.is_track_too_long(single_track, guild_data["maxlength"]):
+                        if self.cog.is_track_length_allowed(single_track, guild_data["maxlength"]):
                             enqueued_tracks += 1
                             player.add(ctx.author, single_track)
                             self.bot.dispatch(
--- a/redbot/cogs/audio/core/abc.py
+++ b/redbot/cogs/audio/core/abc.py
@@ -165,7 +165,7 @@ class MixinMeta(ABC):
         raise NotImplementedError()
 
     @abstractmethod
-    def is_track_too_long(self, track: lavalink.Track, maxlength: int) -> bool:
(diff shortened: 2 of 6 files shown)


Drapersniper:
line 168 in redbot/cogs/audio/core/utilities/formatting.py
line 1492 in redbot/cogs/audio/core/commands/playlists.py

In both of these we pass Track.length instead of Track (These are the only places we do)

can we change the 
```py
    def is_track_too_long(self, track: Union[lavalink.Track, int], maxlength: int) -> bool:
        try:
            length = round(track.length / 1000)
        except AttributeError:
            length = round(track / 1000)
        if length > 900000000000000:  # livestreams return 9223372036854775807ms
            return True
        elif length >= maxlength:
            return False
        else:
            return True
```

to

```py
    def is_track_too_long(self, track: lavalink.Track, maxlength: int) -> bool:
        if track.is_stream:
            return False
        length = track.length / 1000
        if length > maxlength:
            return False
        else:
            return True
```

This way we dont depend on an assumed number in case they change it down the line
and all that will need is to update
line 168 in redbot/cogs/audio/core/utilities/formatting.py
line 1492 in redbot/cogs/audio/core/commands/playlists.py
to pass the full track object


Note if we do this, line 168 in redbot/cogs/audio/core/abc.py will also need to be updated to 

```py
    @abstractmethod
    def is_track_too_long(self, track: lavalink.Track, maxlength: int) -> bool:
        raise NotImplementedError()
```

Drapersniper (approved):
Everything here is looking solid.

Jackenmen (approved):

PR #3893, opened 2020-05-31
[Docs] Autostart and restart on Windows

Limmek:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Added how to create a batch file that allows the bot to be restarted and it also checks if a windows scheduled task exist to start the bot on windows startup, if not it will create it.

### DIFF
--- a/docs/autostart_windows.rst
+++ b/docs/autostart_windows.rst
@@ -8,9 +8,9 @@ Setting up auto-restart on Windows
 
 This script will setup windows schedule task to start the bot when the computer starts, it also checks for updates when the bot starts or is restarted by command ``[p]restart``.
 
--------------------------
-Creating the batch file
--------------------------
+--------------------------------------
+Creating the auto restart batch file
+--------------------------------------
 
 .. important:: If you followed the `install_windows` your virtual environment should be ``"%userprofile%\redenv"``.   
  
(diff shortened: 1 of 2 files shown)


Drapersniper:
I'm personally not in favour of auto updates, updates should be something a owner takes action to ensure they have up to date requirements.

Yes it's more convenient, but convenience is not always better than explicitly

Drapersniper:
Also not to mention, you shouldn't run red as an admin. Ever.

Limmek:
I will change the update to be false  from the beginning and if the user want to have auto update he can just enable it.

That is true you should never do it but the bot will only run as admin if you right click and run it as admin.

Drapersniper:
I mean it shouldn't be an option to start with it available in the docs I mean.

By having it there it means it's supported, supporting auto updates is problematic


Same applies to telling them to run as admin.

Limmek:
Ohh i see,  i will make some changes!

Limmek:
I removed auto update and task creation from the script and instead added how to sett up the task by command.

Jackenmen:
Considering there is already one PR adding auto-restart (#2726) and that there's been no prior discussion on the approach taken here, I would like to mention that this PR might not get accepted if our team decides that we can't support this approach if for example there are some flaws in it.

Also for the next time, please file an issue or start a conversation on [Red server](https://discord.gg/red) for any new additions that have not been proposed already or, if you have alternative proposals to already requested additions, please propose them in already existing issue. It may save you from losing time on an addition that we may not be able to accept. Communication is key :)

Kowlin:
Superseded by #5996.

PR #3895, opened 2020-06-01
[Mod] Consistency periods & proper logging

Flame442:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Adds some periods to some strings that were missing them, replaces `print` statements with `log` statements.

### DIFF
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -78,7 +78,7 @@ class KickBanMixin(MixinMeta):
                 "hierarchy."
             )
         elif guild.me.top_role <= user.top_role or user == guild.owner:
-            return _("I cannot do that due to discord hierarchy rules")
+            return _("I cannot do that due to discord hierarchy rules.")
         elif not (0 <= days <= 7):
             return _("Invalid days. Must be between 0 and 7.")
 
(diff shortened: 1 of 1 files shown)


redbot/cogs/mod/kickban.py
```diff
@@ -101,14 +101,19 @@ async def ban_user(
...
         except discord.Forbidden:
             return _("I'm not allowed to do that.")
         except Exception as e:
-            return e  # TODO: improper return type? Is this intended to be re-raised?
+            log.exception(
+                "{}({}) attempted to kick {}({}), but an error occurred.".format(
+                    author.name, author.id, user.name, user.id
+                )
+            )
+            return _("An unexpected error occurred. ({})".format(e))
```
Jackenmen: String formatting needs to be done *after* calling `_()`
```suggestion
            return _("An unexpected error occurred. ({})").format(e)
```

redbot/cogs/mod/kickban.py
```diff
@@ -101,14 +101,19 @@ async def ban_user(
...
         except discord.Forbidden:
             return _("I'm not allowed to do that.")
         except Exception as e:
-            return e  # TODO: improper return type? Is this intended to be re-raised?
+            log.exception(
+                "{}({}) attempted to kick {}({}), but an error occurred.".format(
+                    author.name, author.id, user.name, user.id
+                )
+            )
+            return _("An unexpected error occurred. ({})").format(e)
```
Jackenmen: I'm not sure if it's good idea to send the exception message to a random (from bot owner's perspective) person here.

Jackenmen:
Other than that, looks good, would be nice to use %-formatting instead of `.format()` but it was like this before so that's up to you.

Jackenmen (changes requested):
Only the change from the review comment above is needed. I'm making this review so that it properly shows as "changes requested".

Jackenmen (approved):
LGTM

PR #3634, opened 2020-03-02
Move mutes to new cog, add role-based and temporary mutes

TrustyJAID:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

### Description of the changes
This moves mutes outside of the mod cog into its own separate cog. 
- Allow mutes to be role based
- `[p]mute` defaults to serverwide muting and `[p]mutechannel` mutes in the current channel
- Mutes all have an optional time in the reason field to automatically remove the mute
- Mutes allow for more than 1 user to be muted at a time utilizing commands.Greedy

PR #3896, opened 2020-06-01
[Core] Guild scoped I18n

### LINKED ISSUE #1970: [V3 i18n] Guild-overides for localisations
It would be nice if I could change bot localisations per-server and per-user

*Edit by Tobotimus:* And per-channel? ;)

### LINKED ISSUE #3801: Per-server localising
# Feature request

- [ ] Cog
- [ ] Command
- [X] API functionality

So, in a bit of ye olde necroposting, I thought I'd bring back up the issue of Per-server languages. Currently we have bot-wide language setting and regional format setting, but a per-server option might be useful for larger bots that may not strictly be in English only servers.

Found it referenced in PR https://github.com/Cog-Creators/Red-DiscordBot/pull/2246, and also in issue https://github.com/Cog-Creators/Red-DiscordBot/issues/1970

Kowlin:
Never again!

### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

Add Guild i18n support.

~~TODO:
Clean up code
Add Guild scoped Babel Numeric formatter support.~~

Fixes #1970


redbot/core/core_commands.py
```diff
@@ -1302,7 +1302,7 @@ async def serverprefix(self, ctx: commands.Context, *prefixes: str):
 
     @_set.command()
     @checks.is_owner()
-    async def locale(self, ctx: commands.Context, language_code: str):
+    async def globallocale(self, ctx: commands.Context, language_code: str):
         """
         Changes bot's locale.
```
Flame442: Should be changed to something like "Changes the default locale".

Flame442 (dismissed):

redbot/core/core_commands.py
Kowlin: Solved in https://github.com/Cog-Creators/Red-DiscordBot/pull/3896/commits/2210270f0bacd2556a566c0cd9b871eb24aef3ec

Kowlin:
Guild scoped i18n currently has an issue whereby modlog cases **cannot** be translated, this is not the fault of i18n, Modlog doesn't account for i18n existing at all, this PR should be good to merge and modlog handling can be done at a latter date. If it proofs that people are interested in,

mikeshardmind:
This appears to be working as intended, but I'm going to take a closer look a little later at the memory use and how it scales based on concerns the last time this came up (toby's old PR)

mikeshardmind:
We may want a command for the bot owner to designate which locales are available, defaulting to all and release notes about this. Resource use grows as needed but will be higher.

Kowlin:
Its a fair suggestion, but based on the calculations I did, I came around a total memory usage of 1.4 MB from French.

accounting for the fact that its ~460KB at 70% completion, adding another 30% on-top of that estimate leads it to being ~600KB of memory, doubling that to account for the overhead from Python.

Meaning that its very roughly ~1.5MB per language loaded.

mikeshardmind:
I was seeing about 15mb increase from french alone when accounting for everything from contextvars as well, i wonder if there may be more to that growth than the i18n itself though, give me a few more to look into this.

mikeshardmind:
Prior concern inflated by a tooling issue and a tool incorrectly assesing the lifecycle of contextvar held references.

mikeshardmind (changes requested):
There needs to be an exposed method for cog creators to set the contextvar to the appropriate locale from an event. Right now, `set_contextual_locale` works, but there's no supported way for cog creators to get the appropriate locale to set. (such as by asking by guild or guild id) This isn't necessary for command invokes, but would be needed for any messages sent by events to be handlable by cog creators.

aikaterna:
While people are directed to Red’s Crowdin page to see what translations are available in [p]set globallocale or [p]set locale, we don’t seem to have a list of the language-country code pairs listed anywhere. For ease of use I would suggest having them listed somewhere, whether that is a command (might be a bit overboard imo), more help text in those commands (might be too crowded/long), or a link to somewhere where we can list those. If something like that is already available in the bot elsewhere or listed on Crowdin I am not aware of it.

Edit: [p]listlocales was a thing at one point... I don't know where it went, checking the bot now.
2nd Edit: Please add listlocales back to the bot. Was removed in #3676 

It might also be helpful to have a display of the current guild and global locales in the informational display under [p]set, as the regional format override is displayed there already. Ideally I would like to see both global and local information for locale and regional format exposed in that info display, but of course this is just a suggestion for ux and I’d like to see what other people think, or if there is another way to display that for users.

Jackenmen:
This is a little bit problematic - locale can be set to *any* valid language code, not just those that we offer translations for. For one, 3rd-party cog creators might provide translations for languages we don't provide translations for, but also the regional formatting functionality allows you to set the formatting of any region. That's why we started to link to Crowdin to see available translations **for Red**. But you are right that it could not be immediately obvious what to set it to.
It would be great if Crowdin could just show it right under the language in the grid but sadly it doesn't... I think the link would probably still be the best, but if there is some website that lists all language code in `en-US`-like format, we could possibly link to that.

mikeshardmind:
The locales which we support are technically limited to validity under CLDR and recognized by babel

`babel.localedata.locale_identifiers()` [currently] generates a list of 758 of these, which would obviously be unruly for any realistic use by end-users. 

We could create an interactive command which is hidden but pointed to which helps pick a locale based on an interactive prompt and use of `babel.negotiate_locale` (which compares requested locales with a set of "available" ones and tries to find the best fit, sacrificing region if no exact match) though this might create misleading circumstances when it results in recommending any of the languages we technically have Crowdin support for but under 10% translation.

I don't really have a good solution for this other than *perhaps* a webpage which we control or trust which checks the browser info to tell the user what their locale code actually is and how much if any of that locale is translated.

Kowlin:
Delaying this until I find time to implement the requested features.

aikaterna:
I tested this for close to a week with a few thousand users/few servers using it and no one was able to find a way to cause a traceback or unintended behaviour. Made sure to try different combos of the global regional format override and various languages as well, but only tested languages with a good amount of coverage (mostly FR, RU, and DE).

TurnrDev:
I'm going to give this a try on a live bot and report back if I get any errors

redbot/core/core_commands.py
```diff
@@ -1327,11 +1328,47 @@ async def locale(self, ctx: commands.Context, language_code: str):
         standardized_locale_name = f"{locale.language}-{locale.territory}"
         i18n.set_locale(standardized_locale_name)
         await ctx.bot._config.locale.set(standardized_locale_name)
+        await ctx.send(_("Global locale has been set."))
+
+    @_set.command()
+    @checks.guildowner_or_permissions(manage_guild=True)
+    async def locale(self, ctx: commands.Context, language_code: str):
```
TurnrDev, line 2067: This needs `@commands.guild_only()`
```suggestion
    @commands.guild_only()
    async def locale(self, ctx: commands.Context, language_code: str):
```

redbot/core/core_commands.py
```diff
@@ -1359,6 +1396,42 @@ async def regionalformat(self, ctx: commands.Context, language_code: str = None)
...
         await ctx.bot._config.regional_format.set(standardized_locale_name)
+        await ctx.send(
+            _("Global regional formatting will now be based on `{language_code}` locale.").format(
+                language_code=standardized_locale_name
+            )
+        )
+
+    @_set.command(aliases=["region"])
+    @checks.guildowner_or_permissions(manage_guild=True)
+    async def regionalformat(self, ctx: commands.Context, language_code: str = None):
```
TurnrDev, line 2140: This needs `@commands.guild_only()`
```suggestion
    @commands.guild_only()
    async def regionalformat(self, ctx: commands.Context, language_code: str = None):
```

TurnrDev (commented):
You forgot to restrict the command to guilds only

redbot/core/core_commands.py
Drapersniper: Not really necessary 

```py
@checks.guildowner_or_permissions(manage_guild=True)
```
Will account for that

Edit: That said .... If the check is added before `@checks.guildowner_or_permissions(manage_guild=True)` it could be benefitial here, for both clarity and an early exit.

Drapersniper: AAAA As Jack points out ..... Although this is the case forr everyone thats not a bot owner ... this would still fail for bot owners 🤦 

So yes `@commands.guild_only()` would be needed here to avoid bot owners running this in DMs

TurnrDev: > AAAA As Jack points out ..... Although this is the case forr everyone thats not a bot owner ... this would still fail for bot owners facepalm
> 
> So yes `@commands.guild_only()` would be needed here to avoid bot owners running this in DMs

That would be why it was in my DMs

Kowlin: Added in a pending commit.

redbot/core/core_commands.py
```diff
@@ -1557,8 +1557,22 @@ async def set_showsettings(self, ctx: commands.Context):
             mod_role_ids = guild_data["mod_role"]
             mod_role_names = [r.name for r in guild.roles if r.id in mod_role_ids]
             mod_roles_str = humanize_list(mod_role_names) if mod_role_names else "Not Set."
-            guild_settings = _("Admin roles: {admin}\nMod roles: {mod}\n").format(
-                admin=admin_roles_str, mod=mod_roles_str
+
+            guild_locale = await self.bot._i18n_cache.get_locale(ctx.guild)
+            guild_regional_format = await self.bot._i18n_cache.get_regional_format(ctx.guild) or _(
+                "Same as bot's locale"
```
TurnrDev: Might I suggest putting the bot's locale in brackets so guild admins know what the setting is?

Vexed01: ~~_pst_ and maybe add a lil dot on the end of the sentence as well.~~
Yeah let's read the context next time.

Drapersniper: a dot doesn't make sense here.

Kowlin:
At this stage this is probably ready for an re-review.

redbot/core/core_commands.py
Kowlin: Resolved in https://github.com/Cog-Creators/Red-DiscordBot/pull/3896/commits/9f07ae0b7e17f5e2dfaacb9d0ad8c58454d2d7bf

TurnrDev:
I am going to put this live on my bot for a week to see if there's any issues

TurnrDev (dismissed):

TurnrDev:
It's been 8 days since I've had this running on a live instance now and I've had no issues related to this PR. :+1: Would recommend.

redbot/core/core_commands.py
```diff
@@ -1557,24 +1557,38 @@ async def set_showsettings(self, ctx: commands.Context):
...
+                guild_locale=guild_locale,
+                guild_regional_format=guild_regional_format,
             )
         else:
             guild_settings = ""
 
         prefixes = await ctx.bot._prefix_cache.get_prefixes(ctx.guild)
         global_data = await ctx.bot._config.all()
         locale = global_data["locale"]
-        regional_format = global_data["regional_format"] or _("Same as bot's locale")
```
Jackenmen, line 1581: This was intentional. It makes it obvious that the regional format is currently set to use bot's locale.
Same suggestion applies to guild regional format (which can be "Same as server's locale") as well as to guild's locale (which can be "Same as bot's locale").

redbot/core/core_commands.py
```diff
@@ -2032,6 +2085,42 @@ async def regionalformat(self, ctx: commands.Context, language_code: str = None)
...
+    @_set.command(aliases=["region"])
+    @checks.guildowner_or_permissions(manage_guild=True)
+    async def regionalformat(self, ctx: commands.Context, language_code: str = None):
+        """
+        Changes bot's regional format in this server. This is used for formatting date, time and numbers.
+
+        `<language_code>` can be any language code with country code included,
+        e.g. `en-US`, `de-DE`, `fr-FR`, `pl-PL`, etc.
+
+        Leave `<language_code>` empty to base regional formatting on bot's locale.
```
Jackenmen: This might be a bit clearer.
```suggestion
        Leave `<language_code>` empty to base regional formatting on bot's locale in this server.
```

redbot/core/core_commands.py
```diff
@@ -2032,6 +2085,42 @@ async def regionalformat(self, ctx: commands.Context, language_code: str = None)
...
+        """
+        if language_code is None:
+            i18n.set_contextual_regional_format(None)
+            await self.bot._i18n_cache.set_regional_format(ctx.guild, None)
+            await ctx.send(_("Regional formatting will now be based on bot's locale."))
```
Jackenmen: This might be a bit clearer.
```suggestion
            await ctx.send(_("Regional formatting will now be based on bot's locale in this server."))
```

redbot/core/events.py
```diff
@@ -308,6 +308,11 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
 
     @bot.event
     async def on_message(message):
+        locale = await bot._i18n_cache.get_locale(message.guild)
+        regional_format = await bot._i18n_cache.get_regional_format(message.guild)
+        set_contextual_locale(locale)
+        set_contextual_regional_format(regional_format)
```
Jackenmen: Let's use our own APIs here:
```suggestion
        await set_contextual_locales_from_guild(bot, message.guild)
```

redbot/core/i18n.py
```diff
@@ -159,16 +245,32 @@ def __call__(self, untranslated: str) -> str:
...
+            return self.translations[locale][untranslated]
         except KeyError:
             return untranslated
 
+    def check_cache_for_locale(self, locale: str) -> bool:
+        """Check is an locale exists with in the cache"""
+        if locale in self.translations:
+            return True
+        return False
+
```
Jackenmen: This method can be removed as it isn't used anywhere.

redbot/core/settings_caches.py
```diff
@@ -56,6 +56,54 @@ async def set_prefixes(
             await self._config.guild_from_id(gid).prefix.set(prefixes)
 
 
+class I18nManager:
+    def __init__(self, config: Config):
+        self._config: Config = config
+        self._guild_locale: Dict[int, str] = {}
+        self._guild_regional_format: Dict[int, str] = {}
+
+    async def get_locale(self, guild: Union[discord.Guild, None]) -> str:
```
Jackenmen, line 65: We should really be caching everything. Currently, in most cases we still end up calling config:
- in non-guild contexts we always get global locale from config
- in guild contexts, we get guild and global locale from config in cases where guild locale is `None` (which will most likely be the case for most guilds we check it for)

redbot/core/settings_caches.py
```diff
@@ -56,6 +56,54 @@ async def set_prefixes(
...
+            else:
+                self._guild_locale[guild.id] = out
+        return out
+
+    async def set_guild_locale(self, guild: discord.Guild, locale: Union[str, None]) -> None:
+        """Set the guild locale in the config and cache"""
+        self._guild_locale[guild.id] = locale
+        await self._config.guild(guild).locale.set(locale)
+
+    async def get_regional_format(self, guild: Union[discord.Guild, None]) -> Optional[str]:
```
Jackenmen, line 111: Same comment as for `get_locale()` method.

But I also have one more thought. I think it would be more logical to use this resolution order for guild regional format:
- use guild regional format, if not None
- use guild locale, if not None
- use global regional format, if not None
- use global locale

It does seem to make some sense to me, but I am not really sure if it's actually better or worse than current resolution order which is:
- use guild regional format, if not None
- use global regional format, if not None
- use guild locale, if not None
- use global locale

// Edit: Hmm, it might require more rewriting, so I'm not sure whether it's really worth pursuing.

redbot/core/i18n.py
```diff
@@ -159,16 +245,32 @@ def __call__(self, untranslated: str) -> str:
...
         """
-        self.translations = {}
+        locale = get_locale()
+
+        if locale.lower() == "en-us":
+            # Red is written in en-US, no point in loading it
+            return
+        if locale in self.translations:
+            # Locales cannot be loaded twice as they have an entry in
+            # self.translations, which is created in L#92.
```
Jackenmen: This line number reference doesn't make sense anymore and even if you update it, it will probably not make sense after some time again. Can you just refer to the relevant method name?

Jackenmen (changes requested):
Review comments contain everything that needs to be changed.

Random thoughts that don't need addressing, but I thought might be worth saying:
- does it only feel weird to me to call it context***ual*** locale?
- there's a bit of code repetition happening in core commands, which we may want to address at some point.

redbot/core/settings_caches.py
```diff
@@ -59,23 +59,31 @@ async def set_prefixes(
 class I18nManager:
     def __init__(self, config: Config):
         self._config: Config = config
-        self._guild_locale: Dict[int, str] = {}
-        self._guild_regional_format: Dict[int, str] = {}
+        self._guild_locale: Dict[int, Union[str, None]] = {}
+        self._guild_regional_format: Dict[int, Union[str, None]] = {}
```
Jackenmen:
```suggestion
        self._guild_locale: Dict[Union[int, None], Union[str, None]] = {}
        self._guild_regional_format: Dict[Union[int, None], Union[str, None]] = {}
```

redbot/core/settings_caches.py
```diff
@@ -59,23 +59,31 @@ async def set_prefixes(
...
+        else:  # Uncached guild
+            out = await self._config.guild(guild).locale()  # No locale set
             if out is None:
-                out = await self._config.locale()
+                self._guild_locale[guild.id] = None
+                return self._guild_locale[None]
             else:
-                self._guild_locale[guild.id] = out
-        return out
+                return self._guild_locale[guild.id]
```
Jackenmen: Looks like you didn't update this method and only updated `get_regional_formatting()` method:
```suggestion
                self._guild_locale[guild.id] = out
                return out
```

redbot/core/settings_caches.py
```diff
@@ -84,17 +92,26 @@ async def set_guild_locale(self, guild: discord.Guild, locale: Union[str, None])
...
+            global_regional_format = await self._config.regional_format()
+            self._guild_regional_format[None] = global_regional_format
+
+        if guild is None:  # Not a guild so cannot support guild locale
+            return self._guild_regional_format[None]
+        elif guild.id in self._guild_regional_format:  # Cached guild
+            if self._guild_regional_format[guild.id] is None:
+                return self._guild_regional_format[None]
             else:
+                return self._guild_regional_format[None]
```
Jackenmen:
```suggestion
                return self._guild_regional_format[guild.id]
```

Jackenmen (changes requested):
I know you haven't done all the changes yet, but I just wanted to let you know about the issues I've seen.

redbot/core/i18n.py
Kowlin: Removed

redbot/core/settings_caches.py
Kowlin: Should be done.

redbot/core/settings_caches.py
Kowlin: Theoretically should be done?

Jackenmen (changes requested):
`[p]set locale` and `[p]set regionalformat` don't update the i18n cache.

Update: I meant `[p]set globallocale` and `[p]set globalregionalformat`, sorry about that 😬

Jackenmen (changes requested):
To add, I've tested everything else and the aforementioned cache is the only issue BUT there are some core cogs that need to set contextual locale manually in events/tasks:

- [x] Audio's `lavalink_event_handler()` (I am about 76% certain that this is the only place in Audio that needs this, we can ask Draper to get this closer to 99% territory)
- [x] Filter's `maybe_filter_name()` and `on_message()`
- [x] Mod's `on_message()`
- [x] Report's `on_message()` and `on_raw_reaction_add()`
- [ ] Streams's `check_streams()`

I can help out with this if you don't feel like doing it all, just hit me up ;)

Jackenmen (dismissed):
Looks good to me, but I'll probably hold this unmerged until #3634 (unless that doesn't happen in 3.4.1) and then add i18n support for Mutes here and merge after testing.

redbot/core/core_commands.py
```diff
@@ -1569,25 +1569,39 @@ async def set_showsettings(self, ctx: commands.Context):
...
 
         prefix_string = " ".join(prefixes)
         settings = _(
             "{bot_name} Settings:\n\n"
             "Prefixes: {prefixes}\n"
             "{guild_settings}"
-            "Locale: {locale}\n"
-            "Regional format: {regional_format}\n"
+            "Global locale: {locale}\n"
+            "Global regional format: {regional_format}"
```
```suggestion
            "Global regional format: {regional_format}\n"
```

Jackenmen:
The changes I just pushed around Mutes cog need testing which I'll be doing later today.

Jackenmen (approved):
LGTM

TurnrDev:
:heart_eyes:

PR #3897, opened 2020-06-02
Fix setting guild prefixes when the prefixes aren't in cache

NeuroAssassin:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

I really love these one line PRs.

Fix for if you're a person who likes to repeatedly spam bot internals.  Before, would raise KeyError when trying to delete it from cache, since it wasn't in cache.  Now, it ignores it if it isn't in the cache.


Jackenmen (approved):
Looks good to me 👍

PR #3901, opened 2020-06-03
[Streams] Only send message about missing client secret once

### LINKED ISSUE #3587: Bot sends messages to owners on each load of streams cog
# Other bugs

Load streams cog

Nothing

#### What actually happened?

I got a message that I need to set up a client secret key

1. Set twitch client_id (can be invalid) in shared api tokens:
`[p]set api twitch client_id anything`
2. Reload cog
3. Bot sends message to owners about missing client secret
4. Repeat step 2
5. Bot does sends the message again.

I wouldn't even mind this error if there were option to clear api tokens for a service but since there isn't, this is getting annoying.

Drapersniper:
… 
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

fixes #3587

### DIFF
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -116,7 +116,7 @@ class Streams(commands.Cog):
                     await self.config.notified_owner.set(False)
             except KeyError:
                 message = _(
-                    "You need a client secret key to correctly use Twitch API on this cog.\n"
+                    "You need a client secret key to correctly use the Twitch API on this cog.\n"
                     "Follow these steps:\n"
                     "1. Go to this page: https://dev.twitch.tv/console/apps.\n"
                     '2. Click "Manage" on your application.\n'
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -116,7 +116,7 @@ class Streams(commands.Cog):
                     await self.config.notified_owner.set(False)
             except KeyError:
                 message = _(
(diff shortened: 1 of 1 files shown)


redbot/cogs/streams/streams.py
```diff
@@ -109,22 +109,28 @@ async def move_api_keys(self) -> None:
...
         if tokens.get("client_id"):
+            notify_owner = await self.config.notified_owner()
             try:
                 tokens["client_secret"]
+                if notify_owner is True:
+                    await self.config.notified_owner.set(False)
             except KeyError:
                 message = _(
-                    "You need a client secret key to use correctly Twitch API on this cog.\n"
+                    "You need a client secret key to correctly use the Twitch API on this cog.\n"
```
Jackenmen: "to correctly use" seems weird here, maybe we could just use "if you want to" or "to be able to"?
```suggestion
                    "You need a client secret key if you want to use the Twitch API on this cog.\n"
```

Jackenmen (dismissed):
Works great, I only wonder if `notified_owner` is the best name here - there could be more situations where we would want to notify owner in future and it could be confusing if there were `notified_owner` but also `notified_owner_about_something`.
I'm approving this anyway, but if you want to change that name, feel free to.
There's also one string suggestion but it's not related to the actual change so it can be ignored too.

Drapersniper:
wait no, dont merge i agree with you, and to avoid a migration later lets do it now

Jackenmen (approved):
Looks good to me 👍

PR #3903, opened 2020-06-03
Make `localwhitelist` check if caller will still be able to use bot after changes

### LINKED ISSUE #3222: [Core] Users may accidentally lock guild owner and themselves with `[p]localwhitelist add`
# Command bugs

#### Command name

`[p]localwhitelist add`

Core

I was expecting to get an error saying that I can't lock out myself (and guild owner) from using the bot.

#### What actually happened?

I locked myself and guild owner out.

1. Have permissions to use `[p]localwhitelist` (e.g. have admin role), but don't be bot owner.
2. Use `[p]localwhitelist clear`
3. Use `[p]localwhitelist add SOMEONE_ELSES_USER_ID`

---

This doesn't seem like a trivial task but it would surely be great to have it improved.

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

fixes #3222

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1948,7 +1948,7 @@ class Core(commands.Cog, CoreLogic):
         names = [getattr(u_or_r, "name", u_or_r) for u_or_r in users_or_roles]
         uids = {getattr(u_or_r, "id", u_or_r) for u_or_r in users_or_roles}
         if not (ctx.guild.owner == ctx.author or await self.bot.is_owner(ctx.author)):
-            current_whitelist = await self.bot._whiteblacklist_cache.get_whitelist(ctx.guild)
+            current_whitelist = set(await self.bot._whiteblacklist_cache.get_whitelist(ctx.guild))
             theoretical_whitelist = current_whitelist + uids
             ids = {i for i in (ctx.author.id, *(getattr(ctx.author, "_roles", [])))}
             if ids.isdisjoint(theoretical_whitelist):
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1954,7 +1954,8 @@ class Core(commands.Cog, CoreLogic):
             return await ctx.send(
                 _(
                     "I cannot allow you to do this, as it would "
-                    "remove your ability to run commands"
+                    "remove your ability to run commands, "
+                    "please ensure to add yourself to the whitelist first."
                 )
             )
         await self.bot._whiteblacklist_cache.add_to_whitelist(ctx.guild, uids)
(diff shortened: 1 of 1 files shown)


Jackenmen:
Adding guild owner is actually unnecessary - the linked issue description is slightly outdated and we always return `True` for guild owner already:
https://github.com/Cog-Creators/Red-DiscordBot/blob/2c313594dacf298e47e3e5ec339f2af2c545b9b2/redbot/core/bot.py#L383-L385

*(I just happened to look at issue tracker :P)*

Drapersniper:
> Adding guild owner is actually unnecessary - the linked issue description is slightly outdated and we always return `True` for guild owner already:
> https://github.com/Cog-Creators/Red-DiscordBot/blob/2c313594dacf298e47e3e5ec339f2af2c545b9b2/redbot/core/bot.py#L383-L385
> 
> _(I just happened to look at issue tracker :P)_

yeah i saw this after i added but forgot to remove it, thanks for spotting it

Jackenmen (changes requested):
Similar check is needed `[p]localwhitelist remove` as well - I can add myself, then someone else and later remove myself and then I will lose my access to the bot.

Also, I was thinking that the error message in `[p]localwhitelist add` could mention that to prevent access loss, you should add yourself to the whitelist first.

Jackenmen (changes requested):
The comment I made in Discord that would be good to address in this PR:
These checks are not needed for guild owners and bot owners as they're not affected by the local whitelist and it may make it unnecessarily harder to add/remove some role on the whitelist.

redbot/core/core_commands.py
```diff
@@ -1946,8 +1946,20 @@ async def localwhitelist_add(
             return
 
         names = [getattr(u_or_r, "name", u_or_r) for u_or_r in users_or_roles]
-        uids = [getattr(u_or_r, "id", u_or_r) for u_or_r in users_or_roles]
-        await self.bot._whiteblacklist_cache.add_to_whitelist(ctx.guild, uids)
+        uids = {getattr(u_or_r, "id", u_or_r) for u_or_r in users_or_roles}
+        if not (ctx.guild.owner == ctx.author or await self.bot.is_owner(ctx.author)):
+            current_whitelist = await self.bot._whiteblacklist_cache.get_whitelist(ctx.guild)
+            theoretical_whitelist = current_whitelist + uids
```
Jackenmen: You're concatenating list and set here which raises an error.

Jackenmen (changes requested):

Jackenmen (approved):
Looks good to me.

PR #3911, opened 2020-06-04
[Mod] Only send the DM about migration if at least 1 scope entry has a non-default value

### LINKED ISSUE #3909: Mod cog sends owner notifications on fresh install.
# Other bugs

I got reminded about it when I saw a fix for #3587. Mod cog sends owner notifications about `[p]moveignoredchannels` and `[p]movedeletedelay` on fresh Red installs. Only viable solution seems to be looping through all guild settings and only send the message if `delete_delay` has been changed from the default in at least one of them though I'm basing that on my comment [here](https://github.com/Cog-Creators/Red-DiscordBot/pull/3638#discussion_r392119234).

Drapersniper:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

fixes #3909

### DIFF
--- a/redbot/cogs/mod/mod.py
+++ b/redbot/cogs/mod/mod.py
@@ -6,6 +6,8 @@ from collections import defaultdict
 from typing import List, Tuple
 
 import discord
+from redbot.core.utils import AsyncIter
+
 from redbot.core import Config, modlog, commands
 from redbot.core.bot import Red
 from redbot.core.i18n import Translator, cog_i18n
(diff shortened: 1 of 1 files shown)


redbot/cogs/mod/mod.py
```diff
@@ -104,20 +106,24 @@ async def _maybe_update_config(self):
...
-            )
-            self.bot.loop.create_task(send_to_owners_with_prefix_replaced(self.bot, msg))
+            async for e in AsyncIter((await self.config.all_channels()).values(), steps=25):
+                if e["ignored"] is not False:
+                    msg = _(
+                        "Ignored guilds and channels have been moved. "
+                        "Please use `[p]moveignoredchannels` to migrate the old settings."
+                    )
+                    self.bot.loop.create_task(send_to_owners_with_prefix_replaced(self.bot, msg))
+                    break
```
Jackenmen: Both guilds and channels can be ignored - we need code that checks guilds as well.

Jackenmen (changes requested):

Jackenmen (approved):
Looks good to me, thanks :)

PR #3914, opened 2020-06-04
[Core] Add `[p]set mentionhelp` command to the bot

### LINKED ISSUE #3912: Posting general info about the bot on ping
Command- So for instance someone pings the bot to figure out what it does, it should reply with some general info like;

Prefix,
help command
Some general info about the bot, maybe uptime, how many servers its in, how many commands, maybe the support server (requires a setup obviously), some general information like that.

it would really help out new people figuring out what the bot is, what it does, and how to use it.

Drapersniper:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

fixes #3912
- Bot will now respond to a blank mention of itself with the available prefixes.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -916,13 +916,25 @@ class RedBase(
             and message.clean_content.strip() == f"@{message.mentions[0].display_name}"
             and message.mentions[0].id == self.user.id
         ):
-            prefixes = await self._prefix_cache.get_prefixes(message.guild)
+            prefixes = await self.get_valid_prefixes(guild=message.guild)
+            prefixes = sorted(prefixes, key=len)
+            counter = 0
+            prefixes_string = humanize_list(
+                [
+                    pf
+                    for p in prefixes
+                    if (pf := f"`{p}`")
+                    and len(pf) < 1800
+                    and ((counter + len(pf)) < 1800)
+                    and (counter := counter + len(pf))
+                ]
+            )
             await message.channel.send(
                 _(
-                    "Hello there, my prefix here are the following:\n{}"
+                    "Hello there, my prefix here are the following:\n{}\n"
                     "Why don't you try `{}help` to see everything I can do."
                     # There's a chance help is disabled ... but that's a gray area anyways so dont think supporting it is necessary.
-                ).format(humanize_list([f"`{i}`\n" for i in prefixes]), prefixes[0])
+                ).format(prefixes_string, prefixes[0])
             )
             return
         if ctx is None or ctx.valid is False:
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -47,7 +47,7 @@ from .settings_caches import (
     PrefixManager,
     IgnoreManager,
     WhitelistBlacklistManager,
-    MentionPredixManager,
+    MentionPrefixManager,
 )
 
 from .rpc import RPCMixin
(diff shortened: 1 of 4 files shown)


Jackenmen:
I don't think we should be adding such event to the bot. IMO, it would be better to just do all the necessary checks in newly added `on_message_without_command` event listener.
And they should definitely happen *after* checking the `mentionhelp` setting is enabled.

Drapersniper:
> I don't think we should be adding such event to the bot. IMO, it would be better to just do all the necessary checks in newly added `on_message_without_command` event listener.
> And they should definitely happen _after_ checking the `mentionhelp` setting is enabled.

yeah the checks after is a good point.

The event, the reason for an early exit here, is i dont want the `message_without_command` dispatched if this mention this setting is enabled, and the checks all pass.

This is to reduce the risk of cogs also doing things on this signal.
For example, cleverbot cog, it replies if you just use @botMention the bot will also reply, and with this on it would send 2 messages. but using a separate event that is dispatched earlier, and stop `on_message_without_command` being dispatched this wont be the case.

Jackenmen:
> using a separate event that is dispatched earlier, and stop on_message_without_command being dispatched this wont be the case.

This is exactly something that IMO we should NOT be doing.

Drapersniper:
> > using a separate event that is dispatched earlier, and stop on_message_without_command being dispatched this wont be the case.
> 
> This is exactly something that IMO we should NOT be doing.

Personally i feel that if this is enable it should take priority over other cogs.

As a bot owner if i enabled this behaviour for a single mention i wouldnt other cogs to mess up with it.

But input from the community here could be useful.

and example is a user goes and run @botMention

With my proposal if the setting is enable it would show the prefixes and not send a `on_message_without_command` event.

With what jack is suggesting we could send the dispatch for that event which means cogs that listen to that event and single mention would also respond.

So for example, cleverbot would also respond to the mention sending 2 messages to the user, the prefixes, and the clever bot message.

mikeshardmind:
not dispatching `message_without_command` would definitely break people using that event. 

Personally, I find any bot that responds to just a plain mention to be a malfunction at best, mentions *aren't* obviously commands and could very easily just be an easy way to reference a bot in conversation knowing that bots don't get annoyed at pings. 

To that end, I'm not sure why this isn't just a 3rd party cog rather than another setting that has to be checked on every non-command message even for people not using this.

Even if this is to be included though, changing that event behavior would be problematic.

redbot/core/bot.py
```diff
@@ -898,7 +905,19 @@ async def process_commands(self, message: discord.Message):
             await self.invoke(ctx)
         else:
             ctx = None
-
+        if (
+            ctx
+            and ctx.valid
```
Drapersniper, line 910: Just a note so that i dont forget .. this is wrong.

Kowlin:
I feel like this strongly impacts performance and should not be added to core at all sadly enough.

This ``if`` block seems like an extreme drain in performance considering how much stuff it has to check. The ``on_message`` event is already a hotly contested. Before the commands are processed i18n will have to run, context is being generated, and finally this if block will have to go through its steps. This all adds time before command execution for something which has a very low chance of occurring in a live situation.

https://github.com/Cog-Creators/Red-DiscordBot/pull/3914/files#diff-b4e8893536462f5cd5d7e291288c4d5eR908-R920

A good way to look at this is to ask your self, how many times am I gonna mention the bot for its prefix. And from what I've seen, the use case of where the bot acts as the primary moderation bot for an extremely scaled server. This is nearly never gonna occur.

Simply put, this adds too much time between execution to be beneficial at all when the bot is under constant stress already.

mikeshardmind:
I've gone ahead and created this [as an additional cog](https://github.com/mikeshardmind/SinbadCogs/blob/v3/mentionhelp/mentionhelp.py) to demonstrate a correct implementation of this which does not require making a listener which cannot be removed by people who do not want this behavior heavier for it. Given the simplicity of it, I have no issue keeping this in the repo long term, and anyone should be able to trivially modify it.

Given how easy this is to add on without being in core, I hope this is enough reason to not make core's `on_message` handling heavier.

Drapersniper:
The aforementioned cog looks good to me.

Closing this in favour of it.

PR #3920, opened 2020-06-05
[redbot.setup] Ask for confirmation when passed instance name contains hyphens + allow to use dots

### LINKED ISSUE #3749: Instance names can start with `-`
Instance names that start with `-` should not be allowed. Names such as `--list-instances` or `-h` can currently be created however a bot with that name cannot be launched. If we ever add a new arg to `redbot`, we could block someone out of their bot.

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Users can create instance names can contain `-`, which causes issues with systemd and if it's at the start, the bot can't run.
Changes:
- Periods are now allowed
- If there is a `-`, users are warned about potential issues and asked to confirm if they want to continue

_This has been edited to reflect changes discussed below._


redbot/setup.py
```diff
@@ -132,10 +132,11 @@ def get_name() -> str:
             " A-z, numbers, underscores, and hyphens."
         )
         name = input("> ")
-        if re.fullmatch(r"[a-zA-Z0-9_\-]*", name) is None:
+        if re.fullmatch(r"[a-zA-Z0-9_][a-zA-Z0-9_\-]*", name) is None:
```
```suggestion
        if re.fullmatch(r"[a-zA-Z0-9]*", name) is None:
```

This is not final .... but like ... in all honesty ... this is an identifier ... with all the problem we seen due to symbols i would be leaning toward just making this identifier alpha numerical.

Vexed01: I'd lean towards agreeing with this -- this would also need an owner DM/log warning for users who already have a `_ ` or `-` asking them to use `redbot --edit-instance` if we want users to transition away from this.
Edit: A single sentence log warning might be best, as it would just be a small push to users to update

Jackenmen: I'm against it, I like my hyphens in instance names and also being able to use number at the beginning of the name, I would even want some other characters (like dots for example) allowed.

On the other hand, hyphens specifically (though there might be other characters that have this issue) need to be escaped when used in argument to systemd service. So while I would want my freedom, there are definitely some good reasons to forbid this.

What would you say about... Adding a prompt which will warn the user if the characters used in instance name *may* cause problems (i.e. hyphen in name)? On the bad side, this would need us to define 2 patterns (or give complete freedom which I'm not so sure is a good idea either, though arguably not terrible if it explicitly says it can cause problems) but on the bright side, it gives more freedom when choosing the instance name.

Jackenmen: Oh and I forgot, I don't think there's really a need to *transition* users to new restrictions. Until 3.2, Red allowed any character except spaces and we did not do it - it's not like old instance names don't work and if there were some issues with them, I think the user would already notice it quite early.

Drapersniper: I can get behind the prompt.

I mean if we allow alphanumetic `string.isalnum` could do the trick, then regex is used for prompt less strict. what do you think?

Vexed01: I'd be happy with that, though I don't use hyphens in instance names, as well or we don't have hyphens and continue to allow underscores `_`.

That's two nos for a transition then! Fine by me.

So draper you mean that if it's not alphanumeric there's a prompt warning about possible issues?

Drapersniper: Yeah thats the way i'd go about it.

If not alpha-numeric prompt for warning about potential issues, then if they say yes validate their entry agaisnt the more lenient regex.

Jackenmen: > I mean if we allow alphanumetic `string.isalnum` could do the trick

`.isalnum()` is too lenient to be used here - it allows characters like "Ą". I think going with `[A-Za-z0-9_]+` regex (so just removing the hyphen from current one) would be the best to go about it for the strict one. As far as I know, there's nothing in Python that would make it easier than that, the closest is `str.isidentifier()` though I don't think there's a need to be *that* strict in the strict check :)

> If not alpha-numeric prompt for warning about potential issues, then if they say yes validate their entry agaisnt the more lenient regex.

Is there any reason not to validate the name against the more lenient regex *before* prompting? (only if it fails the strict one and we would actually need to prompt of course)

Now that I think about it... We could simply do a regex `[A-Za-z0-9_\.\-]+` and warn `if "-" in string` 🤔 I don't know, this doesn't really give much more freedom *and only adds what I want* while making it so regular user doesn't use the problematic hyphens... But I honestly don't know what character should be allowed in lenient check... Hyphen specifically proven to be problematic, so warning about it makes sense, but other than that I'm not sure if splitting into strict and lenient check makes as much sense as I initially thought it would. Maybe I should give up on hyphens, I don't know at this point :/

Jackenmen: Maybe I should add a disclaimer that I'm a windows user and that hyphens work great here 😄 (well, other than the case where you start instance name with hyphen)

Vexed01:
Ok, so I've tried to take what we talked about above into consideration. This is the logic if it doesn't pass the RegEx `[a-zA-Z0-9_\.]*`
- if on Linux and contains a "-" anywhere, not allowed
- if it contains a space or starts with a "-", not allowed
- if the top two don't trigger, then `confirm()` with the user.

How does this sound? The code doesn't look great, but it works. Note I haven't tested this on Linux but I did substitute `linux` for `win32` and it worked fine.

Jackenmen:
I don't think* there's a need to special-case Linux here, imo it should just ask to confirm whenever `-` is used, no matter where and no matter in what OS, I just mentioned Windows to explain *why* I didn't really want to completely prevent people from using it.

*but we can wait for someone else to say something about it

Jackenmen:
It's been 11 days, could someone look at my comment above and send their opinion on this here? I would be grateful :)

aikaterna:
I tend to agree with Jack in thinking there’s not a need to special case Linux here and to just ask for confirmation, but I’m also ok with it either way.

Only other comment I have is I think Vexed meant "aesthetics" instead of "ascetics", re: the comment on the print() line.

Jackenmen (changes requested):
Okay, so to sum up:
First we should check the name with regex `[A-Za-z0-9_\.\-]*`:
a) if the regex doesn't match, we should just send an error:
> ERROR: Instance names can only include characters A-z, numbers, underscores (_) and periods (.).

b) if the regex matches, we should check if `-` is in `name` and if it is, we need to ask for confirmation, i.e. something like:
> Hyphens (-) in instance names may cause issues. Are you sure you want to use this instance name?

Vexed01:
Sure, ^ commit does this.

redbot/setup.py
```diff
@@ -128,16 +129,25 @@ def get_name() -> str:
...
+        elif "-" in name and confirm(
+            "Hyphen (-) in instance names may cause issues. Are you sure you want to continue with this instance name?",
+            default=False,
+        ):
+            return name
+        else:
+            return name
+
+        print()  # new line for aesthetics
+        name = ""
```
Jackenmen: This needs the change so that Red still doesn't allow to set an empty name, let's just do it as we did before:
```suggestion
            name = ""
        elif "-" in name and not confirm(
            "Hyphen (-) in instance names may cause issues. Are you sure you want to continue with this instance name?",
            default=False,
        ):
            name = ""

        print()  # new line for aesthetics
```

Jackenmen (changes requested):
One change needed for the cases where empty name is passed.

Vexed01:
Wait no don't merge that
Taking that out means that correct names aren't accepted 
http://i.vex.my.to/IhyxSM

redbot/setup.py
```diff
@@ -128,16 +129,23 @@ def get_name() -> str:
...
+            " A-z, numbers, underscores (_) and periods (.)."
         )
         name = input("> ")
-        if re.fullmatch(r"[a-zA-Z0-9_\-]*", name) is None:
+        if re.fullmatch(r"[A-Za-z0-9_\.\-]*", name) is None:
             print(
-                "ERROR: Instance name can only include"
-                " characters A-z, numbers, underscores, and hyphens!"
+                "ERROR: Instance names can only include characters A-z, numbers, "
+                "underscores (_) and periods(.)."
```
Jackenmen: Missing space
```suggestion
                "underscores (_) and periods (.)."
```

redbot/setup.py
```diff
@@ -128,16 +129,23 @@ def get_name() -> str:
...
             )
             name = ""
+        elif "-" in name and not confirm(
+            "Hyphen (-) in instance names may cause issues. Are you sure you want to continue with this instance name?",
```
Jackenmen: And plural
```suggestion
            "Hyphens (-) in instance names may cause issues. Are you sure you want to continue with this instance name?",
```

Jackenmen (approved):
Looks good to me, thanks for the PR ❤️

PR #3923, opened 2020-06-05
[Core Commands] Fix `[p]set custominfo` error

Flame442:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes an `AttributeError` trying to use `ctx.bot.send` instead of `ctx.send`.
Also makes the message a little clearer while I'm editing the code anyway.


Jackenmen (approved):
LGTM 👍

PR #3925, opened 2020-06-05
[Filter] Fix confusing behavior detecting quotes

Flame442:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Certain strings passed to filter command would give unexpected responses.
`"word"` / `"word` / `"many words` would get interpreted as the start of a quoted phrase, but because the phrase never ended, they would be ignored. This would output the standard response when no words are added to the filter list, "Those words were already in the filter.".
~This PR solves that issue by using `shlex.split()` instead of the funky custom logic that was there before.~
This PR solves that issue by using `*arg` instead of `*, arg` and funky custom logic.

~@jack1142~
~1. Should it use `posix=False`? I don't think so, but Draper doesn't know so I'll ask you.~
~2. For some reason, `shlex.split("here are a\" few \" words")` returns `['here', 'are', 'a few', 'words']`, which I didn't expect. What's the deal with that?~

### DIFF
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -1,5 +1,6 @@
 import discord
 import re
+import shlex
 from typing import Union, Set
 
 from redbot.core import checks, Config, modlog, commands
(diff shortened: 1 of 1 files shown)


Jackenmen:
I have a question, why can't we just use d.py's native way of handling this kind of situations by using `*args`?

Jackenmen (approved):
LGTM 👍

PR #3933, opened 2020-06-07
Let Downloader use the plural

madebylydia:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Downloader will now use the plural when needed, like when there's only one repository, only one cog loaded, unloaded, etc...
It may need fix about grammar.

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -548,7 +548,9 @@ class Downloader(commands.Cog):
         if len(repos) == 0:
             joined = _("There is no repo installed.")
         else:
-            joined = _("Installed Repo{plural}:\n\n").format(plural="s" if len(repos) > 1 else "")
+            joined = _("Installed {plural}:\n\n").format(
+                plural=_("Repos") if len(repos) > 1 else _("Repo")
+            )
         for repo in sorted_repos:
             joined += "+ {}: {}\n".format(repo.name, repo.short or "")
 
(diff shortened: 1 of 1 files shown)


Drapersniper:
Hey just a quick headsup.
This pural here is not i18n friendly.

which means it cannot be used with the current implementation, just wanted to give you a early warning here.

The way for you to implement this which would be more friendly would be 


```py
				message += _("Successfully uninstalled {plural}: ").format(plural=_("cogs") if len(uninstalled_cogs) > 1 else _("cog")
                ) + humanize_list(uninstalled_cogs)
``` 

as an example

Jackenmen:
> Hey just a quick headsup.
> This pural here is not i18n friendly.

I guess I'm too slow
image

madebylydia:
Alright, I change that and push later.

Jackenmen:
> The way for you to implement this which would be more friendly would be
> 
> ```python
> 				message += _("Successfully uninstalled {plural}: ").format(plural=_("cogs") if len(uninstalled_cogs) > 1 else _("cog")
>                 ) + humanize_list(uninstalled_cogs)
> ```
> 
> as an example

Please don't do this either, it's not much better with i18n as other parts of a sentence can also change between singular and plural, i.e in Polish (more or less, I don't know what would be used for `cog`):
`<Cog> został pomyślnie odinstalowany.` for singular 
`<Cogs> zostały pomyślnie odinstalowane.` for plural

So please use two completely separate sentences here.

Drapersniper:
> > The way for you to implement this which would be more friendly would be
> > ```python
> > 				message += _("Successfully uninstalled {plural}: ").format(plural=_("cogs") if len(uninstalled_cogs) > 1 else _("cog")
> >                 ) + humanize_list(uninstalled_cogs)
> > ```
> > 
> > 
> > as an example
> 
> Please don't do this either, it's not much better with i18n as other parts of a sentence can also change between singular and plural, i.e in Polish (more or less, I don't know what would be used for `cog`):
> `<Cog> został pomyślnie odinstalowany.` for singular
> `<Cogs> zostały pomyślnie odinstalowane.` for plural
> 
> So please use two completely separate sentences here.

Damned Polski... but yeah good point

Jackenmen:
@Predeactor So yeah, what Draper said to do isn't good enough as I said above, sorry for confusion 😛

Jackenmen:
@Predeactor you removed your fork.

madebylydia:
0_0
What a mistake, sorry

madebylydia:
Replaced by #4004.

PR #3934, opened 2020-06-07
[Drivers] Add SQLite3 local driver support

### LINKED ISSUE #3832: Deploy SQLite (apsw - JSON)  driver
because it is "simple" fast and relatively easy to implement.

### LINKED ISSUE #3922: [Drivers] Add a SQL driver using JSON1
as the title says.

Drapersniper:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

fixed #3922
fixes #3832

Look... I know it aint pretty... yes it is monkey patched... but it works.

There are some improvements i need to do here but time for me to put my legs up and go watch some anime.

### Key Points
1. This driver will always be avaliable for Red instances just the JSON driver, as it requires no extra dependencies.
2. This driver will not support external SQL servers, and will store files in the local host.
3. Each cog identifier get its own database.
4. Each Scope gets its own table.
5. this driver makes use of `json_set` and `json_type` JSON1 functions to only write the necessary data, instead of writting the scope file (unlike the JSON driver)


redbot/core/drivers/sql/sql.py
```diff
@@ -0,0 +1,246 @@
+import asyncio
+import concurrent
+import ujson as json
```
Drapersniper: so this needs to be looked at im really tempted to make ujson a core dependency, but if not then this can just be changed to normal json.

redbot/core/drivers/sql/sql.py
```diff
@@ -0,0 +1,246 @@
...
+                except ValueError:
+                    print("Please type a number corresponding to one of the options.")
+                else:
+                    break
+        if drop_db is True:
+            for cog_name in data_manager.cog_data_path().iterdir():
+                for cog_id in cog_name.iterdir():
+                    if cog_id.suffix in {".db"}:
+                        cog_id.unlink()
+        # else:
```
Drapersniper: Welp may or may not have forgoten to do this...

Drapersniper:
Note to self: I also need to handle malicious injections.

PR #3938, opened 2020-06-07
[i18n/Streams] Change strings mentioning commands to use {command} variable

### LINKED ISSUE #3919: [Crowdin/i18n] Make every mention of a command in a string a variable so it can't be translated
This is an enhancement.
Wherever a command is mentioned within a string, it should follow the following format
(_("I am a string with a {command}").format(command=f"{ctx.clean_prefix}somecommand"))

This will prevent translators from translating the command itself.

I shall PR this in a time not far in the future. (But I'm not so smart so I'll have to make this issue to remember)

Note to self: Copy the translation for the twitch api help first... That was a pain to translate.

Dav-Git:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Changing all strings that mention a command to use the {command} variable and a .format method to put the command in place. This prevents commands from being translated by crowdin volunteers.

This does things mentioned in #3919 (but doesn't close)


Drapersniper:
Can you please ... do cogs individually if its not asking too much, makes conflict and review easier as it breaks it up.

Dav-Git:
Good point

Dav-Git:
@jack1142 @Kowlin , @Drapersniper and me were wondering if I can just format the docstrings like that or if that would break something...

Jackenmen:
No, you can't. Docstring must be a literal string. If you *really need to*, you *could* use `format_text_for_context` but ehh I don't think the added complexity is worth it.

Edit: just to be clear, this is surely not the only solution, but I can't think of any that would be very simple while keeping i18n support properly.

Dav-Git:
Copy, i'll revert that in a bit then

Dav-Git:
I am dumb

redbot/cogs/streams/streams.py
```diff
@@ -301,12 +301,14 @@ async def picarto_alert(self, ctx: commands.Context, channel_name: str):
...
 
-        `[p]streamalert stop` will disable this channel's stream
+        {command} will disable this channel's stream
         alerts.
 
-        Do `[p]streamalert stop yes` to disable all stream alerts in
+        Do {command2} to disable all stream alerts in
         this server.
-        """
+        """.format(
```
Drapersniper: @Dav-Git  This one still editing the docstring

redbot/cogs/streams/streams.py
```diff
@@ -380,17 +382,16 @@ async def stream_alert(self, ctx: commands.Context, _class, channel_name):
             except InvalidTwitchCredentials:
                 await ctx.send(
                     _(
-                        "The Twitch token is either invalid or has not been set. See "
-                        "`{prefix}streamset twitchtoken`."
-                    ).format(prefix=ctx.clean_prefix)
+                        "The Twitch token is either invalid or has not been set. See " "{command}."
```
Drapersniper:
```suggestion
                        "The Twitch token is either invalid or has not been set. See {command}."
```

Drapersniper (dismissed):
LGTM as long as the 2 comments below are addressed

Jackenmen:
I wonder, wouldn't it actually make sense to translate `<your_client_id_here>`? It's probably not worth the additional trouble but I want to at least mention this 😛

Drapersniper:
> I wonder, wouldn't it actually make sense to translate `<your_client_id_here>`? It's probably not worth the additional trouble but I want to at least mention this 😛

I mean could just simplify by saying `<client_id>`  but i can see people running the command like this and not replacing their client id... but yeah thats a good point thanslating that does make sense

Dav-Git:
Would you put that as a seperate string then? Otherwise the start message will be pretty fucked up in crowdin

Kowlin:
No, keep it all as one string as that would make more sense, Crowdin marks stuff in ``{ }`` and alikes to not be translated by the translators. And will alert them of such.

Jackenmen:
> Would you put that as a seperate string then? Otherwise the start message will be pretty fucked up in crowdin

Something like this would probably make most sense if we were to do so
```py
(
     "blah, blah, {command}"
).format(
    command="`[p]set api twitch client_id {} client_secret {}`".format(
        _("<your_client_id_here>"), _("<your_client_secret_here>")
    )
)

**Comment by Dav-Git** [comment]
Yeah, so as I suggested. Not sure what you mean kowlin

**Comment by Jackenmen** [comment]
> No, keep it all as one string as that would make more sense, Crowdin marks stuff in `{ }` and alikes to not be translated by the translators. And will alert them of such.

That's not quite what I meant @Kowlin - I mean that `<your_client_id_here>` part of the shown example **would make some sense** to actually be translated, see the example above.

File: redbot/cogs/streams/streams.py [hunk 3]
```diff
@@ -226,16 +230,14 @@ async def check_online(
         except InvalidTwitchCredentials:
             await ctx.send(
                 _(
-                    "The Twitch token is either invalid or has not been set. See "
-                    "`{prefix}streamset twitchtoken`."
-                ).format(prefix=ctx.clean_prefix)
+                    "The Twitch token is either invalid or has not been set. See " "{command}."
```
Jackenmen: Could you join these strings? Black doesn't do it automagically right now.
```suggestion
                    "The Twitch token is either invalid or has not been set. See {command}."
```

redbot/cogs/streams/streams.py
```diff
@@ -226,16 +230,14 @@ async def check_online(
...
+                ).format(command=f"`{ctx.clean_prefix}streamset twitchtoken`")
             )
         except InvalidYoutubeCredentials:
             await ctx.send(
                 _(
-                    "The YouTube API key is either invalid or has not been set. See "
-                    "`{prefix}streamset youtubekey`."
-                ).format(prefix=ctx.clean_prefix)
+                    "The YouTube API key is either invalid or has not been set. See " "{command}."
```
Jackenmen: Could you join these strings? Black doesn't do it automagically right now.
```suggestion
                    "The YouTube API key is either invalid or has not been set. See {command}."
```

redbot/cogs/streams/streams.py
```diff
@@ -458,10 +462,10 @@ async def youtubekey(self, ctx: commands.Context):
             "3. Set up your API key \n"
             "(see https://support.google.com/googleapi/answer/6158862 for instructions)\n"
             "4. Copy your API key and run the command "
-            "`{prefix}set api youtube api_key <your_api_key_here>`\n\n"
+            "{command}\n\n"
             "Note: These tokens are sensitive and should only be used in a private channel\n"
             "or in DM with the bot.\n"
-        ).format(prefix=ctx.clean_prefix)
+        ).format(command=f"`{ctx.clean_prefix}set api youtube api_key <your_api_key_here>`")
```
Jackenmen: You missed this one
```suggestion
        ).format(
            command="`{}set api youtube api_key {}`".format(
                ctx.clean_prefix, _("<your_api_key_here>")
            )
        )
```

Jackenmen:
Also needs conflict resolution

Dav-Git:
Sorry, I only just saw this. Will take care of it tomorrow or on Sunday.

Dav-Git:
I think I solved this correctly (TM)

Jackenmen (dismissed):
Looks good to me 👍

Jackenmen:
@Dav-Git Well, excluding the style check that is 😛

Dav-Git:
... whenever I next get to a PC...
Aaaaaaaaaaaaaaaaaaaaaaaaaaaa

Dav-Git:
Blaaaaaaaaaaaaack fixed?

Jackenmen (approved):
Yup 👍

PR #3941, opened 2020-06-08
[Alias] Remove guild_only deco from main alias command group

### LINKED ISSUE #3940: [Alias] `alias global` does not work in DMs
# Command bugs

#### Command name

`[p]alias global` (in DMs)

Alias

To be able to use `[p]alias global` and the subcommands in DMs

#### What actually happened?

That command is not available in DMs.

1 Open a DM with a Red bot
2 Run `[p]alias global` or any of the subcommands

#### Other details
As global aliases can be executed in DMs, it would make sense to be able to create/edit them in DMs.

I'll take a look at this later today.

Dav-Git:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Makes alias global work in DMs.
Closes #3940

### DIFF
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -252,20 +252,24 @@ class Alias(commands.Cog):
         )
 
     @alias.command(name="help")
-    @commands.guild_only()
     async def _help_alias(self, ctx: commands.Context, alias_name: str):
         """Try to execute help for the base command of the alias."""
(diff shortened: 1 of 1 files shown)


Flame442 (changes requested):
`[p]alias show` and `[p]alias help` should also be usable in the global scope with this change. I believe the code will work by just removing the decos from my light testing.

Dav-Git:
I think this should work now, however I haven't actually found the time to test this myself yet. Handle with care

Flame442 (approved):

PR #3942, opened 2020-06-08
Only trigger cooldown when message is parsed in `[p]contact`

madebylydia:
### Type

- [x] Bugfix
- [X] Enhancement
- [ ] New feature

As long as we receive the help message if no message is parsed, it's better to activate the cooldown only if the message is sent.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1570,7 +1570,7 @@ class Core(commands.Cog, CoreLogic):
         await ctx.bot._config.help.tagline.set(tagline)
         await ctx.send(_("The tagline has been set."))
 
-    @commands.command()
+    @commands.command(cooldown_after_parsing=True)
     @commands.cooldown(1, 60, commands.BucketType.user)
     async def contact(self, ctx: commands.Context, *, message: str):
         """Sends a message to the owner"""


Drapersniper:
How does this affect you in the real world, as in when is the cooldown triggered before it is parsed.

Flame442:
`[p]contact` (no message passed)

madebylydia:
> How does this affect you in the real world, as in when is the cooldown triggered before it is parsed.

If I made this PR for this, then I'm too affected to let that here.

Drapersniper:
> > How does this affect you in the real world, as in when is the cooldown triggered before it is parsed.
> 
> If I made this PR for this, then I'm too affected to let that here.


Sure I get that, my question was to try to understand under what cases you were noticing the impact in the current code.

As in what scenarios did it affect you.

When considering changes we need to take all users into consideration, not saying that for this specific feature as its small enough and Flame pointed a pretty good case when this PR would make sense.

But Im more wondering is there others you noticed.

madebylydia:
> what cases you were noticing the impact in the current code.

Well, I don't mind the actual code, it' just I found useless to trigger a cooldown just because the help showed, you did not really run the command at this point, triggering the cooldown on help depend if there's code running if we don't pass arguments, actually, there's none.

> When considering changes we need to take all users into consideration

I don't think users would complain about that.

> But Im more wondering is there others you noticed.

Actually, no, I saw no one complaining about, maybe because it is not that used.

Drapersniper:
Question. When you say it was triggered by help do you mean when running [p]help and [p]help contact ? Or the missing argument context embed that is sent ?

madebylydia:
When you just run contact without message argument. (Try running it 2 times)

Jackenmen (approved):
LGTM 👍

PR #3954, opened 2020-06-11
[Bank] Display in settings if bank is global

aikaterna:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Display the bank's scope in bankset's settings display so that a user will not have to use bankset toggleglobal to check the current scope.

### DIFF
--- a/redbot/cogs/bank/bank.py
+++ b/redbot/cogs/bank/bank.py
@@ -46,8 +46,8 @@ class Bank(commands.Cog):
     @commands.group(autohelp=True)
     async def bankset(self, ctx: commands.Context):
         """Base command for bank settings."""
-        cur_setting = await bank.is_global()
         if ctx.invoked_subcommand is None:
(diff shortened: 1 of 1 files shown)


redbot/cogs/bank/bank.py
```diff
@@ -55,15 +55,18 @@ async def bankset(self, ctx: commands.Context):
                 group = bank._config.guild(ctx.guild)
             group_data = await group.all()
             bank_name = group_data["bank_name"]
+            bank_scope = _("Global") if await bank.is_global() else _("Server")
```
Jackenmen: Could you reuse the output we get above (Line 50)?

Jackenmen (commented):
Just one nitpick since @Drapersniper tries to avoid these :)

Jackenmen (approved):
Looks good 👍

PR #3955, opened 2020-06-11
[Permissions] Prevent guild owner lockouts

### LINKED ISSUE #3107: Permissions cog can be denied losing access to the cog entierly
# Command bugs

#### Command name

permissions setdefaultserverrule

permissions

when trying to do ``[p]permissions setdefaultserverrule deny Permissions`` for it to say it can't disbale access to the permissions command/cog similar to how ``[p]command`` can't be disabled

#### What actually happened?

The cog was denied and unable to be allowed back

1. ``[p]permissions setdefaultserverrule deny Permissions``
2. try and use any permissions commands

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

  - Guild owners will always be able to access the guild configuration
  commands in permissions
  - This includes `permissions explain` and `permissions canrun` as
  informational tools useful for ensuing a correct configuration

I've also added a comment explaining the purpose of the hook itself, as well as what should be contained in the potential allowances such that future modifications can be made with clarity

This change requires some testing, I have not tested it yet (will update this comment should that change)

  resolves #3107

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -137,12 +137,12 @@ class Permissions(commands.Cog):
                     self.permissions_acl,  # acl group
                     self.permissions_acl_setguild,
                     self.permissions_acl_getguild,
-                    self.permissions_canrun,
                     self.permissions_acl_updateguild,
                     self.permissions_addguildrule,
+                    self.permissions_clearguildrules,
                     self.permissions_removeguildrule,
                     self.permissions_setdefaultguildrule,
-                    self.permissions_clearguildrules,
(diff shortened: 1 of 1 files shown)


mikeshardmind:
This comes after some discussion in discord, There is one drawback to this approach though. It will ignore gloabal rules denying these specific commands to the guild owner, however, I think that's probably acceptable since bot owners can outright disable the guild specific commands if they want, without any negative impact on their own ability to configure using global rules.

Jackenmen (approved):
Looks good to me, thanks so much for the PR 😄

PR #3956, opened 2020-06-11
Fix state transition for subcommands modified by permission hooks

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Discovered during testing of #3955 This fixes an issue introduced by #2375 in which subcommands of command groups who were allowed or denied by permission hooks could not transition state properly.

Due to limited use of permission hooks, and that the current hook users had not directly run into this prior, this is a very limited scope issue.

### DIFF
--- a/redbot/core/commands/requires.py
+++ b/redbot/core/commands/requires.py
@@ -183,11 +183,15 @@ class PermState(enum.Enum):
 
     ALLOWED_BY_HOOK = enum.auto()
     """This command has been actively allowed by a permission hook.
-    check validation doesn't need this, but is useful to developers"""
+    check validation swaps this out, but the information may be useful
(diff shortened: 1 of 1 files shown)


Jackenmen:
For anyone interested in specific example:
```py
async def __permissions_hook(self, ctx: commands.Context) -> Optional[bool]:
    if ctx.command.name == "permissions":
        return True

    return None
```
Trying to run `[p]permissions` or `[p]help permissions` with the hook above and without this PR applied should make the bot error.

mikeshardmind:
To add to this, until this is resolved, hook authors can avoid this issue by either not impacting groups, or by manually declaring an allow or deny on all subcommands of groups they do decide to allow.

Jackenmen (approved):
LGTM.

PR #3964, opened 2020-06-12
Fix unnecessary typing before running commands in Downloader

### LINKED ISSUE #3948: [Downloader] `repo incorrectsubcommand` causes bot to type
# Command bugs

#### Command name

`[p]repo/cog/pipinstall*` (without any subcommands)

Downloader

The bot to not start typing.

#### What actually happened?

The bot started typing.

1. Run `[p]cog` or `[p]repo` without any subcommands

#### Aditional info

*`[p]pipinstall` only types for under a second, whilst the others type for a few seconds.

I haven't tested this on downloader's hidden commands.

I had a quick look for any "stray" `ctx.typing`s and it appears there aren't any unless I cannot use Ctrl + F (which is likely).

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #3948
Technically it doesn't directly fix whatever causes this but it makes it impossible for the bot to start typing when cog is ready therefore fixing the issue 😛 
I had that done before but I thought I could figure out exactly what causes it. It doesn't really matter in the end so I'm just creating this PR now instead of who knows when...

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -77,8 +77,9 @@ class Downloader(commands.Cog):
                 pass
 
     async def cog_before_invoke(self, ctx: commands.Context) -> None:
-        async with ctx.typing():
-            await self._ready.wait()
+        if not self._ready.is_set():
+            async with ctx.typing():
+                await self._ready.wait()
         if self._ready_raised:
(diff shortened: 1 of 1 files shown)


Drapersniper (approved):
LGTM

PR #3970, opened 2020-06-12
[Trivia] Minor correction in greekmyth trivia

owocado:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

In greekmyth trivia, there is a question: `Who is the God of Medicine and the son of Apollo?` and the only answer written in the yaml file is `Asclepius` , however, `Aesculapius` or `Hepius` can also be considered as a right answer. 

Source: https://en.wikipedia.org/wiki/Asclepius


Drapersniper (approved):
LGTM

PR #3976, opened 2020-06-16
[Docs] Trigger update on `sudo add-apt-repository ...` v2

yamikaitou:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Adds the `-u` flag back to `add-apt-repository` to force the cache update (for 18.04 and 20.04). Someone in #support had wasn't able to get Python from deadsnakes as he didn't run `apt update` after adding it (they claim they had 18.04)

### DIFF
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -299,13 +299,13 @@ We recommend adding the ``git-core`` ppa to install Git 2.11 or greater:
 
     sudo apt update
     sudo apt -y install software-properties-common
-    sudo add-apt-repository -y ppa:git-core/ppa
+    sudo add-apt-repository -yu ppa:git-core/ppa
 
 We recommend adding the ``deadsnakes`` ppa to install Python 3.8.1 or greater:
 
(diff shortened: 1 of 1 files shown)


Jackenmen:
@yamikaitou Are you 100% positive that they actually had Ubuntu 18.04? Cause that flag is in 18.04 and 20.04 for compatibility reasons and it should not actually do anything on either of these systems:
https://git.launchpad.net/ubuntu/+source/software-properties/tree/add-apt-repository?h=ubuntu/bionic#n71

yamikaitou:
Turns out they are actually using Mint 19.3

Jackenmen:
Just to add some info (I know we don't directly support Mint right now but it's based Ubuntu so it is a little uncommon that it doesn't *just work*), it looks like `add-apt-repository` in Mint does not support all the options that Ubuntu's `add-apt-repository` does and it also doesn't behave in exactly the same way:
https://github.com/linuxmint/mintsources/issues/171

PR #3982, opened 2020-06-18
Bump dependencies (this switches to HTML 5 writer in Sphinx)

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature
- [x] Dependency Update

This PR bumps all our dependencies to the latest versions.
Notably:
- Sphinx 2.4.4 -> 3.1.1
  - sphinx_rtd_theme 0.4.3 -> 0.5.0 - Sphinx 2.x and 3.x is now fully supported
    - this includes full support for HTML 5 writer which this PR switches our docs to use now
  - there are few goodies in Sphinx 3.x we could take use of, but this PR focuses on dependency update itself

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -108,9 +108,6 @@ rst_prolog += f"\n.. |DPY_VERSION| replace:: {dpy_version}"
 #
 html_theme = "sphinx_rtd_theme"
 
-# This will be needed until sphinx_rtd_theme supports the html5 writer
-html4_writer = True
-
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
 # documentation.
--- a/redbot/core/data_manager.py
+++ b/redbot/core/data_manager.py
@@ -44,13 +44,6 @@ if sys.platform == "linux":
 if not config_dir:
     config_dir = Path(appdir.user_config_dir)
 config_file = config_dir / "config.json"
-if not config_file.exists() and sys.platform == "darwin":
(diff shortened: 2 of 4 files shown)


Jackenmen:
Yeah... I think I broke something with this judging by the postgres tests...

Jackenmen:
We need to wait for `pytest-asyncio 0.13.0` release or back down on the `pytest-asyncio` bump. This is not a time-critical dependency update so I'm gonna wait, especially that it most likely will be released very soon™

Jackenmen:
This is now ready for review.
Testing is mostly needed for Mac (more detailed information in PR description) and changes in how docs are rendered (some changes, like bigger font size in class/method documentation are expected, but anything that seems wrong is worth mentioning).

aikaterna:
I'll help with getting the Mac stuff tested on a couple devices.

aikaterna:
On a Mac that's never had Red installed:
I confirmed that it installed appdirs==1.4.4, but it placed a config.json in `~/Library/Application Support/Red-DiscordBot/config.json` when we were expecting `~/Library/Preferences/Red-DiscordBot/config.json` instead.

Jackenmen:
After some additional debugging (thanks aika), I figured that I'm just dumb and I looked at wrong diff when checking the changes in appdirs 1.4.4 and it turns out that the change that would break us doesn't exist in that version (probably due to the reason that I needed to do this compatibility fix in the first place).

Thanks for the testing @aikaterna and helping me figure out that I'm just dumb 👍

Kowlin (approved):

Issue #3961, opened 2020-06-11
`[p]permissions acl setguild/setglobal` doesn't accept rule with empty dictionary


Jackenmen:
# Command bugs

#### Command name

`[p]permissions acl setguild/setglobal`

#### What cog is this command from?

Permissions

#### What were you expecting to happen?

The command to accept the file

#### What actually happened?

It said "Your YAML file did not match the schema: Top-level keys must be either COG or COMMAND." (the error message is unrelated, see #3185)

#### How can we reproduce this issue?

1. Make a yaml file with these contents:
```yaml
COMMAND:
  permissions: {}
```
2. Use the file with `[p]permissions acl setguild/setglobal`
3. See error

Jackenmen:
Fixed by #3987

PR #3987, opened 2020-06-18
[Permissions] Fix integer commands and empty dict rules

### LINKED ISSUE #3185: [Permissions] Schema checking gives unrelated errors
# Command bugs

#### Command name

`[p]permissions acl setglobal/setserver/updateglobal/updateserver`

Permissions

Either get error about incorrect type of key or have Permissions automatically convert int to str.

#### What actually happened?

I got error "Top-level keys must be either `COG` or `COMMAND`"

1. Create such yaml permissions file:
```yaml
COMMAND:
  8:
    176070082584248320: true
```
2. Upload it using one of the commands: `[p]permissions acl setglobal/setserver/updateglobal/updateserver`
3. See error.

---

The above is reproduction for issue that was already fixed, but the unrelated errors can still happen for yaml file like this:
```yaml
COMMAND:
  ping:
    default_but_spelt_incorrectly: false
```

NeuroAssassin:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #3185 and #3961

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -40,12 +40,13 @@ YAML_SCHEMA = Schema(
         {
             UseOptional(COMMAND): Or(
                 {
-                    str: And(
+                    Or(str, int): Or(
                         {
                             Or(int, "default"): And(
                                 bool, error=_("Rules must be either `true` or `false`.")
                             )
                         },
+                        {},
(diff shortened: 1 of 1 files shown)


Jackenmen (approved):
Looks good to me 👍 

We could be more strict here and not allow people to use integers without wrapping them in quotes but this way, it's a little bit more convenient.
I still would want us to use strictyaml one day but that's a whole other discussion...

PR #3990, opened 2020-06-18
[Streams] Invalidate old bearer token when api key is updated.

### LINKED ISSUE #3917: [Streams] ttv_bearer_cache attribute needs to be invalidate on a token change
This is due to it containing outdated info that causes users to see 
`The Twitch token is either invalid or has not been set. See !streamset twitchtoken.` after a valid token is set.

This can be temporarely "fixed" by reload the streams cog.

flaree:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Calls the `get_twitch_bearer_token` when a user updates the twitch tokens.
Fixes #3917

### DIFF
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -95,6 +95,11 @@ class Streams(commands.Cog):
 
         self._ready_event.set()
 
+    @commands.Cog.listener()
+    async def on_red_api_tokens_update(self, service_name, api_tokens):
+        if service_name == "twitch":
+            await self.get_twitch_bearer_token()
+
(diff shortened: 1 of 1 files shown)


Jackenmen:
This could be further optimized if you would make `get_twitch_bearer_token()` accept the tokens dict and only get shared api tokens when the dict is not passed.

Drapersniper (approved):
LGTM

Drapersniper:
I only added this comment to appease a formal code review requirement in the hopes you will reciprocate some day. I didn't actually look at the diff.

PR #3991, opened 2020-06-18
Stop putting text about invite when invite isn't sent in tempban message

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature


Drapersniper (approved):
lgtm

PR #3993, opened 2020-06-19
Make `[p]tempban` respect `default_days` setting

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This might need an additional note in changelog as this can change the expected default behavior of `[p]tempban` - up to 3.3.9 `[p]tempban` did not delete any user's messages on ban but if we now make it respect `default_days` setting, someone might get surprised by it.

If anyone has any concerns about this, please comment.
And I'm guessing that some feedback from people that actually use `[p]tempban` could be helpful too.

↑ is also something reviewer should probably consider when reviewing, as I don't think anyone looked at this.


Drapersniper (approved):
lgtm

PR #3994, opened 2020-06-19
[Image] Update instructions for setting the GIPHY API

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

GIPHY has changed how to register API keys, so the instructions under `[p]giphycreds` should be updated as well.
This is how it looks as an embed:
http://i.vex.my.to/ZOclyj

### DIFF
--- a/redbot/cogs/image/image.py
+++ b/redbot/cogs/image/image.py
@@ -250,7 +250,11 @@ class Image(commands.Cog):
             "6. Add an app description, for example *Used for Red's image cog*.\n"
             "7. Click *Create App*. You'll need to agree to the GIPHY API Terms.\n"
             "8. Copy the API Key.\n"
-            "9. In Discord, run the command `{prefix}set api GIPHY api_key <your_api_key_here>`.\n"
-        ).format(prefix=ctx.clean_prefix)
+            "9. In Discord, run the command {command}.\n"
+        ).format(
+            command="`{prefix}set api GIPHY api_key <your_api_key_here>`".format(
+                prefix=ctx.clean_prefix
+            )
+        )
 
         await ctx.maybe_send_embed(message)
--- a/redbot/cogs/image/image.py
+++ b/redbot/cogs/image/image.py
@@ -242,7 +242,7 @@ class Image(commands.Cog):
 
         message = _(
             "To get a GIPHY API Key:\n"
-            "1. Login to, or create, a GIPHY account.\n"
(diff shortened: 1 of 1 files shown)


Drapersniper:
While you are editing this string ... could you also apply https://github.com/Cog-Creators/Red-DiscordBot/pull/3938 to this? specifically the 9th point. so that the translation of it is cleaner.

redbot/cogs/image/image.py
```diff
@@ -238,17 +238,23 @@ async def gifr(self, ctx, *keywords):
...
+            "4. Click *Select API*, then *Next Step*.\n"
+            "5. Add an app name, for example *Red*.\n"
+            "6. Add an app description, for example *Used for Red's image cog*.\n"
+            "7. Click *Create App*. You'll need to agree to the GIPHY API Terms.\n"
+            "8. Copy the API Key.\n"
+            "9. In Discord, run the command {command}.\n"
+        ).format(
+            command="`{prefix}set api GIPHY api_key <your_api_key_here>`".format(
+                prefix=ctx.clean_prefix
+            )
```
Jackenmen: As said in #3938, "<your_api_key_here>" makes some sense to be translated:
```suggestion
            command="`{prefix}set api GIPHY api_key {placeholder}`".format(
                prefix=ctx.clean_prefix, placeholder=_("<your_api_key_here>")
            )
```

Vexed01:
I got this email from GIPHY a few days ago regarding the development key I created whilst updating the instructions. This may be intimidating to new users. The production keys require more information so they are a bit iffy with users.

I haven't got any email afaik about my key that's about half a year old now, so I assume they don't continue follow-ups.

http://i.vex.my.to/CArrdl

I think we should still go ahead with merging this but open an issue regarding possibly moving to a production key or search endpoint as they are what GIPHY prefer (or just discussion in Discord which results in a direct PR).

Jackenmen (approved):
Looks good to me, thanks for updating the instructions.

Jackenmen:
> I think we should still go ahead with merging this but open an issue regarding possibly moving to a production key or search endpoint as they are what GIPHY prefer (or just discussion in Discord which results in a direct PR).

I don't know about this search endpoint you mentioned, but as for the production key, we shouldn't bother users with them as it requires verification that includes demonstrating the features of the GIPHY integration and other stuff like this.

PR #3995, opened 2020-06-19
[Audio] Allow users to set their java path.

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature


aikaterna:
This can be considered at a later date when audio isn't technically in a bugfixes only mode.

PR #3997, opened 2020-06-19
[Audio] Auto deafen the bot when it joins VC

### LINKED ISSUE #3996: [Audio] Should Audio auto deafen the bot when it joins a VC?
Making this to hear from prople.

Currently d.py and in turn Red does not have any way to receive audio from VCs but some users don't care about that and as skeptical about their privacy.

So my question is should we set the bot to auto deafen itself the moment it joins a VC, it would be a rather trivial change to make this happen, but will only do it if theres enough agreement and demand for it.

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Fixes #3996

### DIFF
--- a/redbot/cogs/audio/core/abc.py
+++ b/redbot/cogs/audio/core/abc.py
@@ -502,3 +502,7 @@ class MixinMeta(ABC):
     @abstractmethod
     async def command_prev(self, ctx: commands.Context):
         raise NotImplementedError()
+
+    @abstractmethod
+    async def self_deafen(self, player: lavalink.Player) -> None:
+        raise NotImplementedError()
--- a/redbot/cogs/audio/core/cog_utils.py
+++ b/redbot/cogs/audio/core/cog_utils.py
@@ -8,7 +8,7 @@ from redbot.core.i18n import Translator
 
 from ..converters import get_lazy_converter, get_playlist_converter
 
-__version__ = VersionInfo.from_json({"major": 2, "minor": 0, "micro": 0, "releaselevel": "final"})
+__version__ = VersionInfo.from_json({"major": 2, "minor": 0, "micro": 1, "releaselevel": "final"})
 
 __author__ = ["aikaterna", "Draper"]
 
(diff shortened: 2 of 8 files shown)


Vexed01:
Now whilst I welcome these changes, I would _like_ (again, don't need this it would just be nice) the ability to not have the bot deafen itself, eg `[p]audioset deafen` enabled by default. 

Why? Well, would prefer to not have the deafen symbol taking up space in Discord. In my opinion, this increases clutter. I also feel this looks off, as I expect that deafening also comes with muteing (based of non-API actions) though I am not often in music channels so I'm not accustomed to deafened bots. 

However, this is just me wanting total customisation of my bot and not something most users would actually benefit from as they have no reason to change the new default.

Drapersniper:
> Now whilst I welcome these changes, I would _like_ (again, don't need this it would just be nice) the ability to not have the bot deafen itself, eg `[p]audioset deafen` enabled by default.
> 
> Why? Well, would prefer to not have the deafen symbol taking up space in Discord. In my opinion, this increases clutter. I also feel this looks off, as I expect that deafening also comes with muteing (based of non-API actions) though I am not often in music channels so I'm not accustomed to muted bots.
> 
> However, this is just me wanting total customisation of my bot and not something most users would actually benefit from as they have no reason to change the new default.

I +1 this, however.... I'll only add a guild/global option on rewrite because it will be cleaner to do it there.

Unless I get people shouting at me ... I would rather just have deafen as the default and non customisable. But if you ya'll think customisation here is important... than so be it.

aikaterna:
Deafen as a default is ok with me, in the idea that discord.py doesn't have voice receive right now anyway. If that's added in the future that can be considered then. This comment is also not a final say on this behavior, willing to listen to other people's opinions on this.

In regards to this being marked as a draft, just wanted to mention to anyone looking here that this will be considered after another rewrite for audio... right now audio is technically in a bugfixes only mode.

Drapersniper:
To be handled by rewrite

Mattz-P:
For people who have trouble using `[p]audioset deafen`. It's called `autodeafen` now

aikaterna:
We're aware of that, thanks.

PR #3999, opened 2020-06-21
[Core] Add `su` and `unsu`  and `sudo <command>` commands to bot

### LINKED ISSUE #3598: [Core] Allow dropping and elevating permissions
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality
- [x] Core bot functionality

`[p]sudo`/`[p]unsudo` commands (actual names might be different when this gets added) that would allow owners to elevate their permissions to be able to use owner-only commands and drop their permissions to be treated as if they weren't bot owner. Functionality would be optional.

From original issue (#3457):
> Now that teams are here, people using their features may want to be able to have the bot treat them like a normal user most of the time but elevate to owner permissions if they are allowed to do so as needed.

> This one may need more discussion before any implementation as it would likely need to be an always available command with inlined checks to prevent lockouts.

Drapersniper:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This PR does a couple of things.

1. It adds `[p]su` which allows a bot owner to elevate themselves to currently bot owner level.
2. It adds `[p]unsu` which allows a bot owner to delavate their permissions back to a standard user
3. Makes `[p]su` and `[p]unsu` always avaliable to bot owners.
4. By default at bot start up the no bot owners have elevated permissions until `[p]su` is run

So theres a couple of things that can be done/changed here.

1. we could add a CLI flag to toggle this behaviour. without the CLI flag the bot would behave exactly as it does now with bot owners always in an elevated state
2. Add tasks to auto delevate a bot owner after N minutes 

The reasoning behind this PR is that bot owners should not have elevated permissions by default and should only elevate themselves when needed (for example a user will need to run `sudo` on terminal to elevate their shell,  and even here after a period of time the user will need to reelevate themselves).

I would appreciate any feedback on behaviour and discussion on default behaviour introduced by this PR.

resolves #3598


flaree:
I feel as if the owners should have the elevated privileges by default. A lot of users may only host for their needs and this would hinder their setup, maybe a flag to have it off by default?

yamikaitou:
Personally, if I think the default should be unelevated for Team Tokens. If the bot is being run with just a single owner, it should default to elevated. If Co-Owners are used, maybe elevate the Token owner and unelevate the co-owners by default?

The idea of being able to "unsudo" is nice though, save us from having to always test stuff with an alt

Jackenmen:
We could put a recommendation to *enable* this functionality somewhere in our docs, but I really don't think we should make this a default. I also think that status-quo wins by default so IMO we would need very good reasons to change the current default behavior and they imo would need to outweight the reasons for the current behavior.

For one, we have a lot of less-knowledgeable users and this is yet another obstacle thrown at them.

I personally also don't think that `[p]sudo` actually necessarily has to make *that* much sense if your bot is in one (or few) servers. In these cases, I would be quite surprised if the person owning the bot isn't at least a very trusted mod (since mods don't have perms to add bots) and I would assume that usually they might be even admins or guild owners. Adding `[p]sudo` as default makes the management of the bot harder by default and considering that currently this is discussed as a cli flag, it would require the user to always put additional flags when running the bot if they want to disable this. Truth to be told, they should be running auto-restart service, but there might be some less or more valid reasons why they don't.

Last but not least, while the point of what should be the default is debatable, there **definitely needs to** be a way that allows you to (depending on what will be chosen by default) enable/disable this functionality.

Jackenmen:
From other things:
- names of the commands can still be discussed - `sudo` definitely makes sense for us, but it might not be the most user-friendly for people that aren't familiar with Unix
    - also, there is at least one approved 3rd-party cog that already takes use of this command name
- the implementation here is very important, so if this change affects someone's cogs, it's important for us to know it now

Dav-Git:
I agree with what jack said. I believe this is a cool feature but the default shouldn't be changed. 

As for names of the commands... 
I think a "ownermode" setting would make more sense to users that are unfamiliar with unix.

This could either be made part of ``[p]set`` though this time contrary to the proposed audio changes I'd agree to having them as standalone command. Maybe ``[p]ownermode enable`` and ``[p]ownermode disable``? Potentially a book converter instead of seperate commands.

Jackenmen:
I don't have anything specific against the `ownermode` name in itself, but one thing to note here is that it might be a little bit long considering that owner will need to run it whenever they want to use some owner-only command (though only once within the set timeout). It is not hard to add an alias though so I guess it wouldn't necessarily be that problematic.

Drapersniper:
Added `[p]sudotimer` which will allow bot owners to set a time so that the sudo priviledges auto expire.

Added the `--enable-sudo` cli.

Jackenmen:
I think `[p]set sudotimer` (or `[p]set sudotimeout`) would make more sense here

redbot/core/bot.py
```diff
@@ -180,11 +181,17 @@ async def prefix_manager(bot, message) -> List[str]:
...
+
         # to prevent multiple calls to app info in `is_owner()`
         self._app_owners_fetched = False
         super().__init__(*args, help_command=None, **kwargs)
         # Do not manually use the help formatter attribute here, see `send_help_for`,
         # for a documented API. The internals of this object are still subject to change.
+        if self._sudo_enabled is False:
+            self._true_owner_ids = self.owner_ids
         self._help_formatter = commands.help.RedHelpFormatter()
         self.add_command(commands.help.red_help)
```
Jackenmen: By putting this here, you moved the help formatter line a little more below the comment about it
```suggestion
        self._help_formatter = commands.help.RedHelpFormatter()
        self.add_command(commands.help.red_help)
        if self._sudo_enabled is False:
            self._true_owner_ids = self.owner_ids
```

mikeshardmind:
This can be enabled more elegantly with a kwarg to is_owner check based on my own use of similar in the past.

redbot/core/commands/commands.py
```diff
@@ -1159,3 +1159,22 @@ async def can_run(self, ctx, *args, **kwargs) -> bool:
         return await ctx.bot._config.datarequests.allow_user_requests()
 
     can_see = can_run
+
+
+# This is intentionally left out of `__all__` as it is not intended for general use
+class _IsTrueBotOwner(_AlwaysAvailableMixin, _RuleDropper, Command):
```
Jackenmen: nitpick, `_AlwaysAvailableMixin` is an unneeded subclass here.

redbot/core/core_commands.py
```diff
@@ -3393,6 +3393,54 @@ async def count_ignored(self, ctx: commands.Context):
...
+    async def sudo(self, ctx: commands.Context):
+        """Enable your bot owner privileges.
+
+        Sudo permission is auto removed after interval set with `[p]sudotimer` (Default to 15 minutes).
+        """
+        if ctx.author.id not in self.bot.owner_ids:
+            self.bot.owner_ids.add(ctx.author.id)
+            await ctx.send(_("Your bot owner privileges have been enabled."))
+            await asyncio.sleep(delay=await self.bot._config.sudotime())
+            self.bot.owner_ids.discard(ctx.author.id)
```
Jackenmen: This might make more sense as a cancellable task - currently using `[p]sudo`, `[p]unsudo` after 5 minutes, and `[p]sudo` right after will lead to "sudo time" being shorter by 5 minutes.

redbot/core/core_commands.py
```diff
@@ -3401,9 +3449,9 @@ async def count_ignored(self, ctx: commands.Context):
         aliases=["licenceinfo"],
         i18n=_,
     )
-    async def license_info_command(self, ctx):
+    async def license_info_command(self, ctx: commands.Context):
         """
-        Get info about Red's licenses.
+        Get info about Red's licenses
```
Jackenmen: Why remove dot here?

Jackenmen (changes requested):
I don't know what's the view of org members on this functionality, so I wouldn't be able to approve this either way, but here's a code review.

From things not included in review comments:
- are we sure we want to call these `sudo`/`unsudo`? I do like these names myself, but as mentioned earlier, there could be better choices we (mostly you and I, I guess) didn't think of
- can it possibly be important for 3rd-party cogs to know "true owners"? If it could be (and I'm quite worried it might), then we need this to be exposed in the API
    - one could also say that modifying what is put in owner ids is changing behavior and therefore breaking. The importance of this needs to be determined.

Drapersniper:
So in theory this is ready to merge, although @TrustyJAID  asked for the sudo command to accept commands, this would make it so it runs a singular command as sudo, and then it would auto remove sudo from you after, this needs to be looked into further before it is properly implemented.

redbot/core/bot.py
```diff
@@ -197,20 +200,38 @@ async def prefix_manager(bot, message) -> List[str]:
...
+    def true_owner_ids(self):
+        return copy(self._true_owner_ids)
+
+    @true_owner_ids.setter
+    def true_owner_ids(self, value: Any) -> NoReturn:
+        raise RuntimeError("Don't try to assign to this attribute")
+
+    @true_owner_ids.deleter
+    def true_owner_ids(self) -> NoReturn:
+        raise RuntimeError("Don't try to delete to this attribute")
```
Drapersniper:
```suggestion
        raise RuntimeError("Don't try to delete this attribute")
```

Drapersniper:
So adding the optional arg is gonna be far more painful  than antecipated, since this PR is not dependent on it i would suggest we leave that for a follow up PR and keep this PR as it is.

Drapersniper:
Keeping this one open  as jack spent a lot of time on it already.

PR #3910, opened 2020-06-04
Reduce config calls when changing white/blacklist and use sets for their cache

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

PR #4002, opened 2020-06-21
Make local blacklist and local whitelist mutually exclusive

### LINKED ISSUE #4001: `localwhitelist` and `localblacklist` can both be non-empty at the same time
# Other bugs

Something I noticed when testing #3903 (but out of scope for it), removing last user from localwhitelist or clearing it makes the bot use the previously configured localblacklist as adding people to `localwhitelist` does not automatically clear `localblacklist`.

IMO, adding people to one of these lists when other isn't empty should prompt the user to clear the other before allowing the user to add people.

Edit: I forgot to add reproduction steps, here they are (Replace `You` with your username):

Pre-requirements: You can't be bot or guild owner
1. Run both `[p]localblacklist clear` and `[p]localwhitelist clear` if they're not already empty
2. Run `[p]localwhitelist add You`
3. Run `[p]localblacklist add You`
4. Run `[p]localwhitelist remove You` or `[p]localwhitelist clear`
5. You can't run commands anymore because there was a blacklist you didn't know about configured

NeuroAssassin:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Prevent setting local blacklist/whitelist when the opposite is not empty.  Fixes #4001.


Jackenmen (changes requested):
I haven't tested these changes yet (although it seems fine from the code), but there are two things that I noticed:
- the error messages should be worded differently - they should say something more along these lines (but not necessarily *exactly* like this - this is just a rough idea on what it should say):
    ```
    Users can't be added to local whitelist while local blacklist is in use. You can use `{prefix}localblacklist clear` if you want to clear the local blacklist and use the local whitelist instead.
    ```
    The important part that needs to be communicated is that **the other list is currently in use/non-empty**; the information about how to clear it is less important and is more of a "tip".
- this should *probably* also have same notice for bot-wide whitelist/blacklist - while it's a lot less important because bot owners are the only people that can change it (and they're not affected by it), it would still be good if they knew that their action might not work as intended

Jackenmen:
There are also some conflicts here, but you shouldn't worry about those until #3910 is merged (unless you want to deal with the conflict resolution twice :P)

Jackenmen:
For your information, #3910 is now merged

PR #4008, opened 2020-06-23
Another tiny PR :awesome:

Drapersniper:
This one is larger than I'd like ...

Howver it does the following:

1. It adds the backbone for the node management
2. It adds the base event handler.
3. It adds some contants
4. It adds the base for the bo managed node.
5. It adds the base for the setting cache
6. It ads the sql queries

This still not in a functional stage ... but didnt want to carry on adding stuff...


aikaterna (approved):
Approved in concept and not in function.

PR #4009, opened 2020-06-23
Audio API and Audio Rewrite

Drapersniper:
This PR, will track the work gone into the Audio Feature branch. and should only be merged after a full QA pass is completed agaisn't that branch.


Drapersniper:
I'm officially closing this.

I've made the decision that we will no longer be moving to Wavelink and therefore a rewrite will not be needed, 

With that being said I do want to explore LL.py and potentially move us to a fork of it, (Allowing for easy updates from upstream while also giving us full freefrom for customization)

PR #4012, opened 2020-06-24
Add `discord.com` to supported domains in `INVITE_URL_RE`

bobloy:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Discord.com is now a discord domain, so added that as an option to an invite URL regex search.
Passed tests and formatted (though unsure if this is even in a test)

### DIFF
--- a/redbot/core/utils/common_filters.py
+++ b/redbot/core/utils/common_filters.py
@@ -16,7 +16,7 @@ __all__ = [
 # regexes
 URL_RE = re.compile(r"(https?|s?ftp)://(\S+)", re.I)
 
-INVITE_URL_RE = re.compile(r"(discord\.(?:gg|io|me|li)|discordapp\.com\/invite)\/(\S+)", re.I)
+INVITE_URL_RE = re.compile(r"(discord\.(?:gg|io|me|li|com)|discordapp\.com\/invite)\/(\S+)", re.I)
 
 MASS_MENTION_RE = re.compile(r"(@)(?=everyone|here)")  # This only matches the @ for sanitizing
 
--- a/redbot/core/utils/common_filters.py
+++ b/redbot/core/utils/common_filters.py
@@ -16,7 +16,7 @@ __all__ = [
 # regexes
 URL_RE = re.compile(r"(https?|s?ftp)://(\S+)", re.I)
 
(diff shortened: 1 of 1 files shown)


redbot/core/utils/common_filters.py
```diff
@@ -16,7 +16,7 @@
 # regexes
 URL_RE = re.compile(r"(https?|s?ftp)://(\S+)", re.I)
 
-INVITE_URL_RE = re.compile(r"(discord\.(?:gg|io|me|li)|discordapp\.com\/invite)\/(\S+)", re.I)
+INVITE_URL_RE = re.compile(r"(discord\.(?:gg|io|me|li|com)|discordapp\.com\/invite)\/(\S+)", re.I)
```
Jackenmen: The invites at discord.com need the `/invite` part same as discordapp.com
You should instead make the `app` part of the discordapp.com optional

Jackenmen (changes requested):

redbot/core/utils/common_filters.py
bobloy: This edit has been added

redbot/core/utils/common_filters.py
```diff
@@ -16,7 +16,7 @@
 # regexes
 URL_RE = re.compile(r"(https?|s?ftp)://(\S+)", re.I)
 
-INVITE_URL_RE = re.compile(r"(discord\.(?:gg|io|me|li)|discordapp\.com\/invite)\/(\S+)", re.I)
+INVITE_URL_RE = re.compile(r"(discord\.(?:gg|io|me|li)|discord(app)?\.com\/invite)\/(\S+)", re.I)
```
Jackenmen: The `(app)` group needs to be non-capturing (`(?:app)`), so that developers can use `match.group(1)` and `match.group(2)` to get first and second part of the invite link.

Jackenmen (changes requested):
One more thing and this can be merged.

redbot/core/utils/common_filters.py
bobloy: Ah, I wasn't familiar with capturing vs non-capturing. Taken care of now

Jackenmen (approved):
Looks good to me, thanks for the PR 👍

PR #4017, opened 2020-06-25
Use aware objects when storing and reading UTC timestamps

DevilXD:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

`.utcfromtimestamp()` takes local timezone into account when converting a timestamp back into a `datetime` object: https://cdn.discordapp.com/attachments/160386989819035648/725707305320185876/unknown.png
Considering that the timestamp that ends up being stored is in UTC, this is probably not what was intended.

Ref of where the value is stored: https://github.com/Cog-Creators/Red-DiscordBot/blob/632840384b0cc3d67ac31a74e20a3dfd1c5cff03/redbot/cogs/mod/kickban.py#L445

There appears to be one other place where `.utcfromtimestamp()` is used - I modified that one accordingly too: https://github.com/Cog-Creators/Red-DiscordBot/blob/ad979180e57e1f770c7ca013f0d46ca043bcb3b2/redbot/core/bank.py#L136


Dav-Git (dismissed):

Jackenmen:
Hmm, the actual issue here is with how we create the timestamp - `datetime.timestamp()` assumes that naive `datetime` instances (such as the one generated by `datetime.utcnow()`) represents local time and therefore generates incorrect POSIX timestamp that is relative to local timezone.

I don't think we can really solve that though, when we already have existing data, so I guess this might be needed middle ground 😐

DevilXD:
> `datetime.timestamp()` assumes that naive datetime instances (such as the one generated by `datetime.utcnow()`) represents local time (...)

This is not really the case though? `datetime.utcnow()` always generates a UTC timestamp object, no matter the timezone. The `.timestamp()` method converts that to a POSIX timestamp, and `datetime.fromtimestamp()` converts it back - also no matter the timezone. You should always operate on UTC timestamps internally, only **optionally** using timezone conversions when displaying the information to the user - otherwise you're exposing yourself to lots of potential bugs, such as the stored time intervals changing when you change the system timezone, among many others.

My recommendation would be to stick to UTC stamps, and as explained, only deal with timezones when it's time (duh) to display it to the user.

Jackenmen:
> This is not really the case though? `datetime.utcnow()` always generates a UTC timestamp object, no matter the timezone. The `.timestamp()` method converts that to a POSIX timestamp, and `datetime.fromtimestamp()` converts it back - also no matter the timezone.

Well, not really. `.timestamp()` **assumes** the time is local when the datetime object is naive. And `datetime.utcnow()` returns a naive datetime object that is not in local timezone.

DevilXD:
This is technically ready to go. Aware datetime objects are now used in places where they are (or can end up) being stored in the config, along with UTC conversion back.

One thing I'd like to point out tho - the audio cog has lots of `datetime.utcnow()` usage, and probably lots of timestamp conversions too. Because the internals of that cog appear to be too complicated for me, I haven't touched anything there. I'm not sure if its in the scope of this PR, but if someone knowledgeable about that would / could point me to the right places, I'll add the changes as needed.

redbot/core/modlog.py
```diff
@@ -374,7 +376,7 @@ async def message_content(self, embed: bool = True):
         last_modified = None
         if self.modified_at:
             last_modified = "{}".format(
-                datetime.fromtimestamp(self.modified_at).strftime("%Y-%m-%d %H:%M:%S")
+                datetime.utcfromtimestamp(self.modified_at).strftime("%Y-%m-%d %H:%M:%S")
```
Jackenmen: Would it be possible to include " UTC" suffix or something like that so that it's clear to users in what timezone this is in?

redbot/core/modlog.py
```diff
@@ -351,8 +353,8 @@ async def message_content(self, embed: bool = True):
         until = None
         duration = None
         if self.until:
-            start = datetime.fromtimestamp(self.created_at)
-            end = datetime.fromtimestamp(self.until)
+            start = datetime.utcfromtimestamp(self.created_at)
+            end = datetime.utcfromtimestamp(self.until)
             end_fmt = end.strftime("%Y-%m-%d %H:%M:%S")
```
Jackenmen: Would it be possible to include " UTC" suffix or something like that so that it's clear to users in what timezone this is in?

redbot/core/events.py
```diff
@@ -55,7 +54,7 @@ async def on_ready():
         if bot._uptime is not None:
             return
 
-        bot._uptime = datetime.datetime.utcnow()
+        bot._uptime = datetime.now(timezone.utc)
```
Jackenmen: Bot's uptime from this var is exposed in `bot.uptime` property, so this change could break someone (+ it does break `[p]uptime` command). I don't think the change here is really needed.

Jackenmen (changes requested):
A very late review, sorry for that.

Aside from review comments:
- The `message.created_at` in `redbot/cogs/filter/filter.py:L376` was missed
- Per your comment, I checked Audio's usage of `utcnow()` and it shouldn't need any changes, so this PR doesn't need to touch it.

Jackenmen (approved):
Looks good to me, thanks for contributing these fixes for the timezone madness 😄

PR #4018, opened 2020-06-25
[Docs] Update framework_i18n.rst

Dav-Git:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Updates the Internationalization framework docs to explain how to get the ``messages.pot`` file using redgettext which works reliably across operating systems.

Rendered page here: https://red-discordbot--4018.org.readthedocs.build/en/4018/framework_i18n.html

### DIFF
--- a/docs/framework_i18n.rst
+++ b/docs/framework_i18n.rst
@@ -33,21 +33,16 @@ Tutorial
 
 After making your cog, generate a :code:`messages.pot` file
 
-The process of generating this will depend on the operating system
-you are using
+We recommend using redgettext a modified version of pygettext for Red.
+You can install redgettext by running :code:`pip install redgettext` in a command prompt.
 
-In a command prompt in your cog's package (where yourcog.py is),
-create a directory called "locales".
-Then do one of the following:
-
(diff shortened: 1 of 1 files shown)


docs/framework_i18n.rst
```diff
@@ -33,22 +33,17 @@ Tutorial
...
-
-Windows: :code:`python <your python install path>\Tools\i18n\pygettext.py -D -n -p locales`
-
-Mac: ?
-
-Linux: :code:`pygettext3 -D -n -p locales`
-
-This will generate a messages.pot file with strings to be translated, including
+To generate the :code:`messages.pot` file, you will now need to run
+:code:`python -m redgettext -D [path_to_cog]`
```
Jackenmen: I'm not sure if most people needs **all** docstrings rather than just command/cog docstrings to be translated, perhaps using `-c` flag instead of `-D` would make more sense here? 
We could also use `-r` flag so that the messages.pot files are generated recursively, but I'm not sure on that.

Dav-Git: I agree on -c, however with the current explanation -r shouldn't be used. I'll add a note to check -h

Jackenmen (approved):
Looks good to me.

PR #4020, opened 2020-06-26
[Core] Add two missing full stops to `licenseinfo`

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Add missing full stops in the `[p]licenseinfo` command.
Note that I'm not a lawyer so hopefully this doesn't cause any legal issues.


redbot/core/core_commands.py
```diff
@@ -2658,10 +2658,10 @@ async def license_info_command(ctx):
     """
 
     message = (
-        "This bot is an instance of Red-DiscordBot (hereafter referred to as Red)\n"
+        "This bot is an instance of Red-DiscordBot (hereafter referred to as Red).\n"
```
Jackenmen: I know it's not part of PR, but shouldn't `hereafter` be `hereinafter`?

Vexed01: Yes, looking at that I think it should be.

Jackenmen (dismissed):
Thanks!

Jackenmen (approved):
Thanks! (again 😄)

PR #4021, opened 2020-06-26
Vendor `discord.ext.menus` until they are released to PYPI

Drapersniper:
Sooner or later we will be using `discord.ext.menus` this PR vendors it until its officially released as there is no current release date for it.

Audio UI will be making use of these to improve performance and user interaction


Jackenmen:
- vendor folder needs to be excluded from black check
- there should (imo) be some information about the commit at which the vendor is for our convenience
    - just having that info in commit message works too
- there's been a new commit; it's not all that important to include but I figured it's worth mentioning
- this was mentioned earlier but I'm repeating it here so that it's not forgotten:
    there should be info about the exclusion from guarantees in docs

Jackenmen:
One more thing that came to my mind is licensing, I know `vendored` folder includes `LICENSE` file already but we might need to have another mention about it in `LICENSE` file in root repo folder.

I think something like this would work:
```
This project vendors discord.ext.menus package (https://github.com/Rapptz/discord-ext-menus) made by Danny Y. (Rapptz) which is distributed under MIT License.
Copy of this license can be found in LICENSE file in redbot/vendored folder of this repository.
```

Mention of it at the end of `README` wouldn't hurt either, although I'm guessing that's more of a courtesy for the work Danny have done rather than actual requirement.

Jackenmen:
I think there's an unwritten convention to put vendored packages in `_vendor`/`vendor` folders (some examples: 
 [pip](https://github.com/pypa/pip/tree/master/src/pip/_vendor), [setuptools](https://github.com/pypa/setuptools/tree/master/setuptools/_vendor), [pipenv](https://github.com/pypa/pipenv/tree/master/pipenv/vendor), [poetry](https://github.com/python-poetry/poetry/tree/master/poetry/_vendor)), or at least I have never seen other names used. We could also use `_vendor` rather than `vendor` since we don't guarantee it anyway, but I guess there's nothing wrong with people using it as long as the docs say it's excluded from guarantees.
Another thing is that it seems a little unnecessary to include *everything* from the `discord-ext-menus` repo instead of just the package itself (+ LICENSE).
Also, is there any reason for us not to use namespace packages? I see that you also made the `discord.ext` namespace package into a regular one as well.

Drapersniper:
Closing in favour of  #4039 so that it goes into V3/develop

PR #4011, opened 2020-06-23
Add a provisional API for replacing the help formatter

mikeshardmind:
### Type

- [ ] Bugfix
- [X] Enhancement
- [X] New feature

### Description of the changes

- Adds an API for replacing the help formatter
- Defines which interfaces are required
- Adds some information for 3rd party developers
- Adds commands for resetting help behavior (to allow end users to fix issues with misbehaving cogs)

This is follow up from things which were planned originally in #2667 but tracked in #3601 and resolves a portion of that issue.

PR #4022, opened 2020-06-26
[Core Commands] Add Help settings view

### LINKED ISSUE #4013: [Help] Add current settings to `[p]helpset` cmd
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality

Add current settings display (as in `[p]modset`/`[p]set`) for `[p]helpset` command

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

Adds a command for viewing the current help settings.

resolves #4013 depends on #4011

### DIFF
--- a/redbot/core/commands/help.py
+++ b/redbot/core/commands/help.py
@@ -30,7 +30,7 @@
 import abc
 import asyncio
 from collections import namedtuple
-from dataclasses import dataclass
+from dataclasses import dataclass, asdict as dc_asdict
 from typing import Union, List, AsyncIterator, Iterable, cast
 
 import discord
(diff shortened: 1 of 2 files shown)


mikeshardmind:
Just for a quick reference of how this looks with the default formatter, though there's a small bit about prepending a warning that these settings may not be used if a non-default formatter is in use.
image

Jackenmen:
Maybe this would look better in code block?

mikeshardmind:
I'm not convinced that it looks better [in a code block], especially when a custom tagline is in use (which I realize isn't shown in the quick example)

Drapersniper:
Since the tagline is only added as `"{tagline_info}"` couldnt we add the first art into a code block followed by the tagline.
IMO it will kee in line with the existing Setting dislays in core.

Jackenmen:
Needs conflict resolution

redbot/core/commands/help.py
```diff
@@ -94,6 +94,40 @@ async def from_context(cls, context: Context):
...
+                return _("Yes")
+            return val
+
+        data = {k: bool_transformer(v) for k, v in dc_asdict(self).items()}
+
+        if not self.delete_delay:
+            data["delete_delay"] = _("Disabled")
+
+        if tag := data.pop("tagline", ""):
+            tagline_info = _("\nCustom Tagline: {tag}").format(tag)
```
Jackenmen:
```suggestion
            tagline_info = _("\nCustom Tagline: {tag}").format(tag=tag)
```

Jackenmen (changes requested):
It would be nice if `Delete delay` were suffixed with " seconds" when it's set.

If you won't be able to do this, I'll just do a follow up PR for it later after merging

Jackenmen (approved):
Looks good to me, thanks for the follow up!

PR #4025, opened 2020-06-28
[Mod] Make tempbans permanent when using `[p]hackban`

### LINKED ISSUE #2898: Can't make tempbans permanent
# Command bugs

#### Command name

tempban

Mod

To be able to make the tempban permanent later on

#### What actually happened?

Unbanning, hackbanning, or any combination did not remove the user from the tempban list

Tempban a user
Unban them
Hackban them
After tempban expires, the user is unbanned despite being unbanned and hackbanned.

Dav-Git:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Closes #2898 by removing the user entry in the list of current tempbans which prevents them from bein automatically unbanned.

### DIFF
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -311,6 +311,7 @@ class KickBanMixin(MixinMeta):
 
         author = ctx.author
         guild = ctx.guild
+        tempbans = await self.config.guild(guild).current_tempbans()
 
         if not user_ids:
             await ctx.send_help()
@@ -330,7 +331,7 @@ class KickBanMixin(MixinMeta):
         for entry in ban_list:
             for user_id in user_ids:
                 if entry.user.id == user_id:
-                    if user_id in await self.config.guild(guild).current_tempbans():
+                    if user_id in tempbans:
                         # We need to check if a user is tempbanned here because otherwise they won't be processed later on.
                         continue
                     else:
@@ -347,7 +348,7 @@ class KickBanMixin(MixinMeta):
         for user_id in user_ids:
             user = guild.get_member(user_id)
             if user is not None:
-                if user_id in await self.config.guild(guild).current_tempbans():
+                if user_id in tempbans:
                     # We need to check if a user is tempbanned here because otherwise they won't be processed later on.
                     continue
                 else:
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -369,19 +369,24 @@ class KickBanMixin(MixinMeta):
             user = discord.Object(id=user_id)
             audit_reason = get_audit_reason(author, reason)
             queue_entry = (guild.id, user_id)
-            try:
-                await guild.ban(user, reason=audit_reason, delete_message_days=days)
-                log.info("{}({}) hackbanned {}".format(author.name, author.id, user_id))
-            except discord.NotFound:
-                errors[user_id] = _("User {user_id} does not exist.").format(user_id=user_id)
-                continue
-            except discord.Forbidden:
-                errors[user_id] = _("Could not ban {user_id}: missing permissions.").format(
-                    user_id=user_id
-                )
-                continue
-            else:
-                banned.append(user_id)
+            async with self.config.guild(guild).current_tempbans() as tempbans:
+                if user_id in tempbans:
+                    tempbans.remove(user_id)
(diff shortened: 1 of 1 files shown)


Dav-Git:
This is now fully tested and definitely functional. I added the info that a tempban was upgraded to a permaban as an "error" so it would show in the result of the command because it seemed like the best and easiest option. I don't think that this info should be "left out" and isn't necessarilly worth the effort of adding additional handling in the output function.

redbot/cogs/mod/kickban.py
```diff
@@ -330,9 +330,13 @@ def remove_processed(ids):
         for entry in ban_list:
             for user_id in user_ids:
                 if entry.user.id == user_id:
-                    errors[user_id] = _("User {user_id} is already banned.").format(
-                        user_id=user_id
-                    )
+                    if user_id in await self.config.guild(guild).current_tempbans():
```
Jackenmen: You should get the list of tempbans *before* looping so that we can just reuse it instead of calling for it on each iteration.

redbot/cogs/mod/kickban.py
```diff
@@ -343,21 +347,25 @@ def remove_processed(ids):
...
             if user is not None:
-                # Instead of replicating all that handling... gets attr from decorator
-                try:
-                    result = await self.ban_user(
-                        user=user, ctx=ctx, days=days, reason=reason, create_modlog_case=True
-                    )
-                    if result is True:
-                        banned.append(user_id)
-                    else:
+                if user_id in await self.config.guild(guild).current_tempbans():
```
Jackenmen: You can use the tempbans list you get above already

Jackenmen (dismissed):

redbot/cogs/mod/kickban.py
Dav-Git: You're absolutely right... I thought I needed those for the user IDs specifically (don't ask me how that got into my brain). I put them up top to where the other local variables are defined.

redbot/cogs/mod/kickban.py
```diff
@@ -369,19 +379,34 @@ def remove_processed(ids):
...
-                )
-                continue
-            else:
-                banned.append(user_id)
+            async with self.config.guild(guild).current_tempbans() as tempbans:
+                if user_id in tempbans:
+                    tempbans.remove(user_id)
+                    errors[user_id] = _("Upgraded tempban for {user_id} to a permaban.").format(
+                        user_id=user_id
+                    )
```
Jackenmen: This shouldn't show under errors - this is just an additional information about the ban, it's not an error. I don't have any specific way of showing it in mind though (if we *do* want to show this information).

Jackenmen (changes requested):
Looks good, there's just one issue with the message shown on successful "upgrade" to a permaban.

redbot/cogs/mod/kickban.py
Dav-Git: I think it's important to show that an upgrade happen and made some changes. Let me know what you think of them.

redbot/cogs/mod/kickban.py
```diff
@@ -292,6 +292,7 @@ async def hackban(
         using this command"""
         banned = []
         errors = {}
+        upgrades = {}
```
Jackenmen: You can just use a list and append IDs to it instead of using a dict with two same values.

redbot/cogs/mod/kickban.py
```diff
@@ -300,6 +301,11 @@ async def show_results():
             if errors:
                 text += _("\nErrors:\n")
                 text += "\n".join(errors.values())
+            if upgrades:
+                text += _(
+                    "\nFollowing user IDs have been upgraded from a temporary to a permanent ban:\n"
+                )
+                text += "\n".join(upgrades.values())
```
Jackenmen: It would probably be better to use a humanised list (`humanize_list()` util) here, rather than put each user ID in separate line.

Jackenmen (changes requested):
Just two small things that should be changed here and it should be good to go.

redbot/cogs/mod/kickban.py
Dav-Git: yes... that would've been smart

Jackenmen (approved):
Looks good to me, thanks for the PR.

PR #4030, opened 2020-06-30
[Utils] Fix regex for role mentions in `MessagePredicate`

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

The MessagePredicate.valid_role() predicate was not matching a role mention, specifically in the _find_role function via the regex match.

This condition can be tested by loading Audio with a blank/new settings file, or on a bot that has not had a DJ role specified for a guild, by using `[p]audioset dj` with no role specified, the bot prompts for a role name, and the user uses a role mention in the response to the bot.


Jackenmen (approved):
Looks good to me. It's first time I'm seeing that `MessagePredicate.has_role()` and `MessagePredicate.valid_role()` exist though 😄

PR #4038, opened 2020-07-05
[Mod] Add kick and warn to mention spam

### LINKED ISSUE #3786: Banmention options
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality

`modset banmentionspam` Instead of having this purely set to ban only, it would be nice to have options to have the bot warn / kick / softban / ban. Having different options rather than having just a pure straight ban for mentions.

Yes, this can be achieved by making a cog or using a third party, though if mentions is already a core function, why not add options and customization to it.

SharkyTheKing:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Moved banmentionspam to its own subcommand. `mentionspam ban`
Added warn mention spam. `mentionspam warn`
Added kick mention spam `mentionspam kick`
Adjusted config to account for new warning and kicking mention spam

Copied Warning case type from `Warning` cog into `Mod`'s casetype.py file.

Fixes #3786


redbot/cogs/mod/events.py
```diff
@@ -72,6 +71,60 @@ async def check_mention_spam(self, message):
...
+                            message.created_at,
+                            "kick",
+                            author,
+                            guild.me,
+                            _("Mention spam (Autokick)"),
+                            until=None,
+                            channel=None,
+                        )
+                    except RuntimeError as e:
+                        print(e)
```
Flame442: This should be actual error logging, not a print.

redbot/cogs/mod/events.py
```diff
@@ -72,6 +71,60 @@ async def check_mention_spam(self, message):
                         print(e)
                         return False
                     return True
+
+        if mention_spam["kick"]:
+            if len(mentions) >= mention_spam["kick"]:
+                try:
+                    await guild.kick(author, reason=_("Mention Spam (Autokick)"))
+                except discord.HTTPException:
+                    log.info(
```
Flame442: I'm not sure if this log level should be `info`, this needs some discussion.

I understand this and the two suggestions below reference code modeled after existing code, however the existing code should be changed as well.

redbot/cogs/mod/events.py
```diff
@@ -72,6 +71,60 @@ async def check_mention_spam(self, message):
...
+                        "Failed to kick member for mention spam in server {}".format(guild.id)
```
Flame442: Should be "Failed to kick **a** member for...". Also maybe also include the member's ID in this log?

redbot/cogs/mod/events.py
```diff
@@ -72,6 +71,60 @@ async def check_mention_spam(self, message):
...
+                    except RuntimeError as e:
+                        print(e)
+                        return False
+                    return True
+
+        if mention_spam["warn"]:
+            if len(mentions) >= mention_spam["warn"]:
+                try:
+                    await author.send(
+                        _("Please do not mention {} or more people!".format(mention_spam["warn"]))
```
Flame442: I'm not sure stating the warn number is the *best* idea, it's not always a good idea to let users know exactly what trips the bots in a server (beyond knowing something is against the rules in general).

redbot/cogs/mod/events.py
```diff
@@ -72,6 +71,60 @@ async def check_mention_spam(self, message):
...
+                    )
+                except (discord.HTTPException, discord.Forbidden):
```
Flame442, line 101: A fallback option for this could be mentioning the user in the channel the mention spam occurred in, since users could just have "allow direct messages from server members" disabled.

redbot/cogs/mod/events.py
```diff
@@ -72,6 +71,60 @@ async def check_mention_spam(self, message):
...
+                    log.info(
+                        "Failed to warn member for mention spam in server {}".format(guild.id)
```
Flame442: Same comments as above.

redbot/cogs/mod/events.py
```diff
@@ -72,6 +71,60 @@ async def check_mention_spam(self, message):
...
+                            message.created_at,
+                            "warning",
+                            author,
+                            guild.me,
+                            _("Mention spam (Autowarn)"),
+                            until=None,
+                            channel=None,
+                        )
+                    except RuntimeError as e:
+                        print(e)
```
Flame442: Same comments as above.

redbot/cogs/mod/settings.py
```diff
@@ -85,35 +79,155 @@ async def hierarchy(self, ctx: commands.Context):
                 _("Role hierarchy will be ignored when moderation commands are issued.")
             )
 
-    @modset.command()
+    @modset.group()
+    @commands.guild_only()
+    async def mentionspam(self, ctx: commands.Context):
+        """
+        Manage the automoderation for mentionspam
```
Flame442:
```suggestion
        Manage the automoderation settings for mentionspam.
        
        Using this command with no subcommands will send the current settings.
```

redbot/cogs/mod/settings.py
```diff
@@ -85,35 +79,155 @@ async def hierarchy(self, ctx: commands.Context):
...
+
+            await ctx.send(box(msg))
+
+    @mentionspam.command()
+    @commands.guild_only()
+    async def warn(self, ctx: commands.Context, max_mentions: int = 0):
+        """
+        Sets the autowarn conditions for mention spam.
+
+        Users will be warned if they send any messages with contain more than
```
Flame442:
```suggestion
        Users will be warned if they send any messages which contain more than
```

redbot/cogs/mod/settings.py
```diff
@@ -85,35 +79,155 @@ async def hierarchy(self, ctx: commands.Context):
...
+            return await ctx.send(_("Autowarn for mention spam disabled."))
+
+        if max_mentions < 2:
+            return await ctx.send(_("`<max_mentions>` must be 2 or higher to autowarn."))
+
+        mismatch_message = ""
+
+        if mention_spam["kick"]:
+            if max_mentions >= mention_spam["kick"]:
+                mismatch_message += _("\nAutowarn is equal or higher than autokick.")
```
Flame442:
```suggestion
                mismatch_message += _("\nAutowarn is equal to or higher than autokick.")
```

redbot/cogs/mod/settings.py
```diff
@@ -85,35 +79,155 @@ async def hierarchy(self, ctx: commands.Context):
...
+        )
+
+    @mentionspam.command()
     @commands.guild_only()
-    async def banmentionspam(self, ctx: commands.Context, max_mentions: int = 0):
+    async def kick(self, ctx: commands.Context, max_mentions: int = 0):
+        """
+        Sets the autokick conditions for mention spam.
+
+        Users will be kicked if they send any messages with contain more than
```
Flame442:
```suggestion
        Users will be kicked if they send any messages which contain more than
```

redbot/cogs/mod/settings.py
```diff
@@ -85,35 +79,155 @@ async def hierarchy(self, ctx: commands.Context):
...
+        `<max_mentions>` mentions.
+
+        `<max_mentions>` must be at least 3. Set to 0 to disable.
```
Flame442: Why does max mentions have to be at least three, besides to allow for autowarn to function at two? Couldn't autowarn be disabled and autokick be set to two?

redbot/cogs/mod/settings.py
```diff
@@ -85,35 +79,155 @@ async def hierarchy(self, ctx: commands.Context):
...
+
+    @mentionspam.command()
+    @commands.guild_only()
+    async def ban(self, ctx: commands.Context, max_mentions: int = 0):
         """Set the autoban conditions for mention spam.
 
         Users will be banned if they send any message which contains more than
         `<max_mentions>` mentions.
 
         `<max_mentions>` must be at least 5. Set to 0 to disable.
```
Flame442: Again, this kinda just seems like an arbitrary number (even though it is existing code).

Flame442 (changes requested):
*Disclaimer: I have not actually run the code yet, this review is just based on things I saw reading the code changes*

redbot/cogs/mod/settings.py
SharkyTheKing: I forgot why I set it to 3, I just know the ban part (as you commented as well on) has an arbitrary number to lock it down from the previous coding. We could remove the number lock, or at least require all of them to be 2 (or even 1) or more.

redbot/cogs/mod/events.py
SharkyTheKing: I fixed the message, added in the user.id but not resolving this as I'd like to get feedback per `not sure if log level should be info` comment you posted above. Not sure if you were wanting to possibly change this debug level.

redbot/cogs/mod/events.py
SharkyTheKing: Accounted, removed the number telling. Changed to `Please do not mass mention people!`, if the DM fails. It'll send in the channel `@member, Please do not mass mention people!`

redbot/cogs/mod/settings.py
SharkyTheKing: Fixed, changed ban/kick/warn to at least be 1. So people don't attempt to action on...every message...

redbot/cogs/mod/events.py
SharkyTheKing: Changed to `log.warning` not entirely sure what else to go for here. Willing to change this type of logging if anyone has something to suggest.

redbot/cogs/mod/mod.py
```diff
@@ -50,7 +50,7 @@ class Mod(
     default_global_settings = {"version": ""}
 
     default_guild_settings = {
-        "ban_mention_spam": False,
```
Flame442, line 52: It might be a good idea to have some kind of config migration.

redbot/cogs/mod/settings.py
```diff
@@ -87,35 +99,131 @@ async def hierarchy(self, ctx: commands.Context):
                 _("Role hierarchy will be ignored when moderation commands are issued.")
             )
 
-    @modset.command()
+    @modset.group()
+    @commands.guild_only()
+    async def mentionspam(self, ctx: commands.Context):
+        """
+        Manage the automoderation settings for mentionspam
```
Flame442:
```suggestion
        Manage the automoderation settings for mentionspam.
```

redbot/cogs/mod/settings.py
```diff
@@ -87,35 +99,131 @@ async def hierarchy(self, ctx: commands.Context):
...
+        """
+        mention_spam = await self.config.guild(ctx.guild).mention_spam.all()
+        if not max_mentions:
+            if not mention_spam["warn"]:
+                return await ctx.send(_("Autowarn for mention spam is already disabled."))
+            await self.config.guild(ctx.guild).mention_spam.warn.set(False)
+            return await ctx.send(_("Autowarn for mention spam disabled."))
+
+        if max_mentions < 1:
+            return await ctx.send(_("`<max_mentions>` must be 1 or higher to autowarn."))
```
Flame442: ~~This entire section looks confusing to me. The docstring says "**Set** to 0 to disable" (never stating that is the default value of the command), the `if not max_mentions` part looks like it is accepting `0` to view the current settings (consider using `None` and testing `if max_mentions is None`), and any value below `1` (including 0) is blocked.~~

~~Look through this part of the code again, and consider what it should actually be *doing* for every possible value.~~

The docstring should state what will happen if no value is passed.

It is also *slightly* contradictory to say "must be at least one" then "set to zero to disable", but I don't know how you could word that better.

(This comment applies to all 3 subcommands of `[p]mentionspam`)

Flame442 (changes requested):
Still haven't run the code yet, but there's a pretty glaring issue that needs fixing regardless.

redbot/cogs/mod/mod.py
SharkyTheKing: Added migration.

redbot/cogs/mod/settings.py
SharkyTheKing: Don't know a better way to word it either, though it still gets the point across regardless. If someone has a better wording for this, I'm happy to change. 

Added information that it defaults if no value is passed.

Zephyrkul: You could take a page from [p]set deletedelay. Still not the best, but at least it'd be consistent.
> Must be 0 or greater. Set to 0 to disable this feature.

SharkyTheKing: Possibly, I'll look into adding this change when I can later today and see. Thanks for this! :).

SharkyTheKing: Done, thanks @zephyrkul. Changes looked a lot cleaner so decided to go with that for the mention spam.

Jackenmen:
Needs conflict resolution +  the `try` blocks on `modlog.create_case()` are unnecessary (and have already been removed from all places that had them in V3/develop)

redbot/cogs/mod/settings.py
```diff
@@ -87,35 +99,131 @@ async def hierarchy(self, ctx: commands.Context):
...
+
+        mismatch_message = ""
+
+        if mention_spam["kick"]:
+            if max_mentions >= mention_spam["kick"]:
+                mismatch_message += _("\nAutowarn is equal to or higher than autokick.")
```
Flame442, line 134: The "mismatch" messages should explain better what they mean. On the one hand, a user should know that their `warn` will usually not trigger since `ban` is the same value, however there are cases where it is good to have them at the same value as backups (if a ban fails, it will still warn).

redbot/cogs/mod/events.py
```diff
@@ -42,18 +42,19 @@ async def check_duplicates(self, message):
...
 
-        max_mentions = await self.config.guild(guild).ban_mention_spam()
-        if max_mentions:
-            mentions = set(message.mentions)
-            if len(mentions) >= max_mentions:
+        mentions = set(message.mentions)
+        if mention_spam["ban"]:
+            if len(mentions) >= mention_spam["ban"]:
                 try:
                     await guild.ban(author, reason=_("Mention spam (Autoban)"))
```
Flame442, line 52: `ban` defaults to deleting a day's worth of messages from a user (including the mention message). I know this was the previous behavior, but doing so creates ghost pings, so I don't know if we want to override that and not delete any messages.

Flame442 (changes requested):
Code works wonderfully, just a few minor things I think need a little discussion.

redbot/cogs/mod/settings.py
SharkyTheKing: Couldn't think of a good message to explain, so kept it simple in hopes someone could help write a better explanatory.

redbot/cogs/mod/events.py
SharkyTheKing: I'm good with whichever, I do think maybe keeping the message content there is better, so disallowing it from removing messages is good.

Flame442 (dismissed):
After some discussion, I will be making a PR for the requested change after this PR is merged as it is out of the scope of this PR.

redbot/cogs/mod/events.py
```diff
@@ -68,6 +69,61 @@ async def check_mention_spam(self, message):
...
+            if len(mentions) >= mention_spam["warn"]:
+                try:
+                    await author.send(_("Please do not mass mention people!"))
+                except (discord.HTTPException, discord.Forbidden):
+                    try:
+                        await message.channel.send(
+                            _("{member}, Please do not mass mention people!").format(
+                                member=author.mention
+                            )
+                        )
```
Jackenmen, line 107: Even if this runs successfully, the case will not be created

redbot/cogs/mod/casetypes.py
```diff
@@ -103,4 +103,10 @@
         "image": "\N{SPEAKER WITH CANCELLATION STROKE}",
         "case_str": "Voice Kick",
     },
+    {
+        "name": "warning",
+        "default_setting": True,
+        "image": "\N{WARNING SIGN}",
+        "case_str": "Warning",
+    },
```
Jackenmen: This needs to be put in `redbot/core/generic_casetypes.py` + it needs a variation selector for the emoji (i.e. `"\N{WARNING SIGN}\N{VARIATION SELECTOR-16}"`).

redbot/cogs/mod/events.py
SharkyTheKing: Hopefully solved.

redbot/cogs/mod/casetypes.py
SharkyTheKing: Hopefully got this fixed and working. I assume what I put is what you were recommending. Never used the Variation selector before so just copied from here.

Jackenmen (commented):
The changes I asked for look good 👍 
Leaving the approval to @Flame442

Flame442 (approved):

PR #4045, opened 2020-07-07
Begin work on a data request API

### LINKED ISSUE #3239: Privacy API
# Feature request

- [ ] Cog
- [ ] Command
- [X] API functionality

Would like to see Red implement a Privacy API so that cog creators can hook their cogs up to it and be more GDPR and Discord ToS compliant.

Relevant Discord API ToS sections: Sections 2.2.v, 2.4, 2.5.c, 2.5.d, 2.5.f, and Section 13
https://discordapp.com/developers/docs/legal

Functionality should include:

- Getting and storing user consent so other cogs can check for that consent
- Providing an interface for cogs to support delete or download data requests from end users
~~- Checking stored cog data periodically to see if it's associated with a user account that has been deleted, which according to Discord ToS end user data for deleted accounts needs to be deleted within 7 days (unless the data is required for the base functionality of the cog)~~ No longer necessary because Discord eliminated this oddly specific requirement in their newly updated API TOS. There are new requirements that should instead be considerated, see https://github.com/Cog-Creators/Red-DiscordBot/pull/4045

Obviously Discord has not enforced these terms in the API ToS because I've never in my life seen a bot with the ability to delete End User Data upon request, or check for data that is associated with accounts that are deleted and remove that data within 7 days, both of which are requirements according to the Discord API Terms of Service:

>You shall delete all End User Data upon Discord’s or the End User’s request and within seven (7) days following the End User’s termination of the End User’s account. You may only retain chat logs as necessary for the operation of your Applications. Discord’s offering of an API or SDK that allows you to reproduce or display End User message content is not an express or implied license to such content from Discord.

However I believe we as a community have an opportunity here to provide tools for people to "do the right thing", even if there's no fear of violating the ToS.

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

This adds an API for showing users what type of data red stores (including statements from cog creators) as well as providing the framework for and commands for users to request data deletion.

This framework provides enough information for cog creators to determine whether or not they need to comply with various requests. (ie a user should not be able to request deleting a record of themself using the bot to spam)

This is a breaking change (technically speaking, blocked by breaking change that this PR depends on in #4085) as it changes the used method names of the cog base class and could conflict with an existing cog.

this resolves #3239

### Things worked on in this PR (check indicating ready for review)

 - [X] Working deletion of core data
 - [X] Working calling of methods provided for 3-rd party deletion of data
 - [X] Working presentation of 3rd-party End User Data Statements
 - [x]  Cog Creator documentation
 - [x] A User facing docs page which explains what data red stores in core.
 - [x] User feedback for crashed calls to extension methods
 - [x] Tuned default cooldowns
 - [x] Owner setting for user self deletion availability.

### Things which will need doing in other PRs

 - Downloader handling of a new info.json key
 - Handling requests for getting what data is stored, not just deleting what data is stored.

### DIFF
--- a/docs/guide_cog_creation.rst
+++ b/docs/guide_cog_creation.rst
@@ -98,6 +98,7 @@ Open :code:`__init__.py`. In that file, place the following:
 
     from .mycog import Mycog
 
+
     def setup(bot):
         bot.add_cog(Mycog())
 
@@ -238,3 +239,20 @@ Not all of these are strict requirements (some are) but are all generally advisa
     but a cog which takes actions based on messages should not.
 
 15. Respect settings when treating non command messages as commands.
+
+16. Handle user data responsibly
+
+  - Don't do unexpected things with user data.
+  - Don't expose user data to additional audiences without permission.
+  - Don't collect data your cogs don't need.
+  - Don't store data in unexpected locations.
+    Utilize the cog data path, Config, or if you need something more
+    prompt the owner to provide it.
+
+17. Utilize the data deletion and statement APIs
+
+  - See `redbot.core.commands.Cog.red_delete_data_for_user`
+  - Make a statement about what data your cogs use with the module level
+    variable ``__red_end_user_data_statement__``.
+    This should be a string containing a user friendly explanation of what data
+    your cog stores and why.
--- a/docs/red_core_data_statement.rst
+++ b/docs/red_core_data_statement.rst
@@ -36,7 +36,7 @@ Discord account was the target of a moderation action.
 
 3rd party extensions to Red are able to delete data when this command
 is used as well, but this is something each extension must implement.
-If a loaded extension does not implmenent this, the user will be informed.
+If a loaded extension does not implemenent this, the user will be informed.
 
 Additional Notes for Bot Owners and Hosts
 *****************************************
--- a/docs/red_core_data_statement.rst
+++ b/docs/red_core_data_statement.rst
@@ -1,0 +1,86 @@
+.. Red Core Data Statement
+
+=====================
+Red and End User Data
+=====================
+
+Notes for everyone
+******************
+
+What data Red collects
+----------------------
+
+Red and the cogs included with it collect some amount of data
+about users for it's normal operations. 
+
+In particular the bot will keep track of a short history of usernames/nicknames
+which actions refer to your discord account (such as creating a playlist)
+as well as the content of specific messages used directly as commands for the bot
+(such as reports sent to servers).
+
+By default, Red will not collect any more data than it needs, and will not use it
+for anything other than the portion of the Red's functionality that necessitated it.
+
+3rd party extensions may store additional data beyond what Red does by default.
+You can use the command ``[p]mydata 3rdparty``
+to view statements about how extensions use your data made by the authors of 
+the specific 3rd party extensions an instance of Red has installed.
+
+How can I delete data Red has about me?
+---------------------------------------
+
+The command ``[p]mydata forgetme`` provides a way for users to remove
+large portions of their own data from the bot. This command will not
+remove operational data, such as a record that your
+discord account was the target of a moderation action.
+
+3rd party extensions to Red are able to delete data when this command
+is used as well, but this is something each extension must implement.
+
+Additional Notes for Bot Owners and Hosts
+*****************************************
+
+How to comply with a request from Discord Trust & Safety
+--------------------------------------------------------
+
+There are a handful of these available to bot owners in the command group
+``[p]mydata ownermanagement``.
+
+The most pertinent one if asked to delete data by a member of Trust & Safety
+is
+
+``[p]mydata ownermanagement processdiscordrequest`` 
+
+This will cause the bot to get rid of or disassociate all data
+from the specified user ID. 
+
+.. warning::
+
+    You should not use this unless
+    Discord has specifically requested this with regard to a deleted user.
+    This will remove the user from various anti-abuse measures.
+    If you are processing a manual request from a user, read the next section
+
+
+How to process deletion requests from users
+-------------------------------------------
+
+You can point users to the command ``[p]mydata forgetme`` as a first step.
+
+If users cannot use that for some reason, the command
+
+``[p]mydata ownermanagement deleteforuser``
+
+exists as a way to handle this as if the user had done it themselves.
+
+Be careful about using the other owner level deletion options on behalf of users,
+as this may also result in losing operational data such as data used to prevent spam.
+
+What owners and hosts are responsible for
+-----------------------------------------
+
+Owners and hosts must comply both with Discord's terms of service and any applicable laws.
+Owners and hosts are responsible for all actions their bot takes.
+
+We cannot give specific guidance on this, but recommend that if there are any issues
+you be forthright with users, own up to any mistakes, and do your best to handle it.
--- a/docs/red_core_data_statement.rst
+++ b/docs/red_core_data_statement.rst
@@ -14,7 +14,7 @@ Red and the cogs included with it collect some amount of data
 about users for the bot's normal operations. 
 
 In particular the bot will keep track of a short history of usernames/nicknames
-which actions refer to your discord account (such as creating a playlist)
+which actions refer to your Discord account (such as creating a playlist)
 as well as the content of specific messages used directly as commands for the bot
 (such as reports sent to servers).
 
@@ -32,10 +32,11 @@ How can I delete data Red has about me?
 The command ``[p]mydata forgetme`` provides a way for users to remove
 large portions of their own data from the bot. This command will not
 remove operational data, such as a record that your
-discord account was the target of a moderation action.
+Discord account was the target of a moderation action.
 
 3rd party extensions to Red are able to delete data when this command
 is used as well, but this is something each extension must implement.
+If a loaded extension does not implmenent this, the user will be informed.
 
 Additional Notes for Bot Owners and Hosts
 *****************************************
--- a/docs/red_core_data_statement.rst
+++ b/docs/red_core_data_statement.rst
@@ -11,7 +11,7 @@ What data Red collects
 ----------------------
 
 Red and the cogs included with it collect some amount of data
-about users for it's normal operations. 
+about users for the bot's normal operations. 
 
 In particular the bot will keep track of a short history of usernames/nicknames
 which actions refer to your discord account (such as creating a playlist)
--- a/docs/red_core_data_statement.rst
+++ b/docs/red_core_data_statement.rst
@@ -36,7 +36,7 @@ Discord account was the target of a moderation action.
 
 3rd party extensions to Red are able to delete data when this command
 is used as well, but this is something each extension must implement.
-If a loaded extension does not implemenent this, the user will be informed.
+If a loaded extension does not implement this, the user will be informed.
 
 Additional Notes for Bot Owners and Hosts
 *****************************************
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -113,7 +113,7 @@ class Alias(commands.Cog):
 
             if modified:
                 await self.config.guild_from_id(guild_id).entries.set(to_set)
-            
+
             await asyncio.sleep(0)
             # control yielded per loop since this is most likely to happen
             # at bot startup, where this is most likely to have a performance
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -1,7 +1,7 @@
 from copy import copy
 from re import search
 from string import Formatter
-from typing import Dict, List
+from typing import Dict, List, Literal
 
 import discord
 from redbot.core import Config, commands, checks
@@ -50,6 +50,17 @@ class Alias(commands.Cog):
         self.config.register_global(**self.default_global_settings)
         self.config.register_guild(**self.default_guild_settings)
         self._aliases: AliasCache = AliasCache(config=self.config, cache_enabled=True)
+
+    async def red_delete_data_for_user(
+        self,
+        *,
+        requester: Literal["discord_deleted_user", "owner", "user", "user_strict"],
+        user_id: int,
+    ):
+        if requester != "discord_deleted_user":
+            return
+
+        await self._aliases.anonymize_aliases(user_id)
 
     async def initialize(self):
         # This can be where we set the cache_enabled attribute later
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -1,3 +1,5 @@
+import asyncio
+import logging
 from copy import copy
 from re import search
 from string import Formatter
@@ -13,6 +15,8 @@ from redbot.core.bot import Red
 from .alias_entry import AliasEntry, AliasCache, ArgParseError
 
 _ = Translator("Alias", __file__)
+
+log = logging.getLogger("red.cogs.alias")
 
 
 class _TrackingFormatter(Formatter):
@@ -38,18 +42,15 @@ class Alias(commands.Cog):
     and append them to the stored alias.
     """
 
-    default_global_settings: Dict[str, list] = {"entries": []}
-
-    default_guild_settings: Dict[str, list] = {"entries": []}  # Going to be a list of dicts
-
     def __init__(self, bot: Red):
         super().__init__()
         self.bot = bot
         self.config = Config.get_conf(self, 8927348724)
 
-        self.config.register_global(**self.default_global_settings)
-        self.config.register_guild(**self.default_guild_settings)
+        self.config.register_global(entries=[], handled_string_creator=False)
+        self.config.register_guild(entries=[])
         self._aliases: AliasCache = AliasCache(config=self.config, cache_enabled=True)
+        self._ready_event = asyncio.Event()
 
     async def red_delete_data_for_user(
         self,
@@ -60,12 +61,87 @@ class Alias(commands.Cog):
         if requester != "discord_deleted_user":
             return
 
+        await self._ready_event.wait()
         await self._aliases.anonymize_aliases(user_id)
 
-    async def initialize(self):
-        # This can be where we set the cache_enabled attribute later
+    async def cog_before_invoke(self, ctx):
+        await self._ready_event.wait()
+
+    async def _maybe_handle_string_keys(self):
+        # This isn't a normal schema migration because it's being added
+        # after the fact for GH-3788
+        if await self.config.handled_string_creator():
+            return
+
+        async with self.config.entries() as alias_list:
+            bad_aliases = []
+            for a in alias_list:
+                for keyname in ("creator", "guild"):
+                    if isinstance((val := a.get(keyname)), str):
+                        try:
+                            a[keyname] = int(val)
+                        except ValueError:
+                            # Because migrations weren't created as changes were made,
+                            # and the prior form was a string of an ID,
+                            # if this fails, there's nothing to go back to
+                            bad_aliases.append(a)
+                            break
+
+            for a in bad_aliases:
+                alias_list.remove(a)
+
+        # if this was using a custom group of (guild_id, aliasname) it would be better but...
+        all_guild_aliases = await self.config.all_guilds()
+
+        for guild_id, guild_data in all_guild_aliases.items():
+
+            to_set = []
+            modified = False
+
+            for a in guild_data.get("entries", []):
+
+                for keyname in ("creator", "guild"):
+                    if isinstance((val := a.get(keyname)), str):
+                        try:
+                            a[keyname] = int(val)
+                        except ValueError:
+                            break
+                        finally:
+                            modified = True
+                else:
+                    to_set.append(a)
+
+            if modified:
+                await self.config.guild_from_id(guild_id).entries.set(to_set)
+            
+            await asyncio.sleep(0)
+            # control yielded per loop since this is most likely to happen
+            # at bot startup, where this is most likely to have a performance
+            # hit.
+
+        await self.config.handled_string_creator.set(True)
+
+    def sync_init(self):
+        t = asyncio.create_task(self._initialize)
+
+        def done_callback(fut: asyncio.Future):
+            try:
+                t.result()
+            except Exception as exc:
+                log.exception("Failed to load alias cog", exc_info=exc)
+                self.bot.remove_cog(self)
+
+        t.add_done_callback(done_callback)
+
+    async def _initialize(self):
+        """ Should only ever be a task """
+
+        await self._maybe_handle_string_keys()
+
         if not self._aliases._loaded:
             await self._aliases.load_aliases()
+
+        self._ready_event.set()
 
     def is_command(self, alias_name: str) -> bool:
         """
@@ -337,6 +413,8 @@ class Alias(commands.Cog):
 
     @commands.Cog.listener()
     async def on_message_without_command(self, message: discord.Message):
+
+        await self._ready_event.wait()
 
         if message.guild is not None:
             if await self.bot.cog_disabled_in_guild(self, message.guild):
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -122,14 +122,14 @@ class Alias(commands.Cog):
         await self.config.handled_string_creator.set(True)
 
     def sync_init(self):
-        t = asyncio.create_task(self._initialize)
+        t = asyncio.create_task(self._initialize())
 
         def done_callback(fut: asyncio.Future):
             try:
                 t.result()
             except Exception as exc:
                 log.exception("Failed to load alias cog", exc_info=exc)
-                self.bot.remove_cog(self)
+                # Maybe schedule extension unloading with message to owner in future
 
         t.add_done_callback(done_callback)
 
--- a/redbot/cogs/alias/alias_entry.py
+++ b/redbot/cogs/alias/alias_entry.py
@@ -107,7 +107,7 @@ class AliasCache:
             else:
                 continue
             # basically, don't build a context manager wihout a need.
-            async with self.config.guild_from_id(guild_id).entries as entry_list:
+            async with self.config.guild_from_id(guild_id).entries() as entry_list:
                 for a in entry_list:
                     if a.get("creator", 0) == user_id:
                         a["creator"] = 0xDE1
--- a/redbot/cogs/alias/alias_entry.py
+++ b/redbot/cogs/alias/alias_entry.py
@@ -90,6 +90,30 @@ class AliasCache:
         self._loaded = False
         self._aliases: Dict[Optional[int], Dict[str, AliasEntry]] = {None: {}}
 
+    async def anonymize_aliases(self, user_id: int):
+
+        async with self.config.entries() as global_aliases:
+            for a in global_aliases:
+                if a.get("creator", 0) == user_id:
+                    a["creator"] = 0xDE1
+                    if self._cache_enabled:
+                        self._aliases[None][a["name"]] = AliasEntry.from_json(a)
+
+        all_guilds = await self.config.all_guilds()
+        async for guild_id, guild_data in AsyncIter(all_guilds.items(), steps=100):
+            for a in guild_data["entries"]:
+                if a.get("creator", 0) == user_id:
+                    break
+            else:
+                continue
+            # basically, don't build a context manager wihout a need.
+            async with self.config.guild_from_id(guild_id).entries as entry_list:
+                for a in entry_list:
+                    if a.get("creator", 0) == user_id:
+                        a["creator"] = 0xDE1
+                        if self._cache_enabled:
+                            self._aliases[guild_id][a["name"]] = AliasEntry.from_json(a)
+
     async def load_aliases(self):
         if not self._cache_enabled:
             self._loaded = True
--- a/redbot/cogs/audio/core/events/red.py
+++ b/redbot/cogs/audio/core/events/red.py
@@ -1,5 +1,6 @@
+import asyncio
 import logging
-from typing import Mapping
+from typing import Literal, Mapping
 
 from redbot.core import commands
 from ..abc import MixinMeta
@@ -19,3 +20,37 @@ class RedEvents(MixinMeta, metaclass=CompositeMetaClass):
             self.api_interface.spotify_api.update_token(api_tokens)
         elif service_name == "audiodb":
             self.api_interface.global_cache_api.update_token(api_tokens)
+
+    async def red_delete_data_for_user(
+        self,
+        *,
+        requester: Literal["discord_deleted_user", "owner", "user", "user_strict"],
+        user_id: int,
+    ):
+
+        await self.cog_ready_event.wait()
+
+        if requester in ("discord_deleted_user", "owner"):
+            await self.playlist_api.handle_playlist_user_id_deletion(user_id)
+
+            all_equalizers = await self.config.custom("EQUALIZER").all()
+
+            collected_for_removal = []
+
+            c = 0
+            for guild_id, guild_equalizers in all_equalizers.items():
+                c += 1
+                if not c % 100:
+                    await asyncio.sleep(0)
+
+                for preset_name, preset in guild_equalizers.get("eq_presets", {}).items():
+                    c += 1
+                    if not c % 100:
+                        await asyncio.sleep(0)
+
+                    if preset.get("author", 0) == user_id:
+                        collected_for_removal.append((guild_id, preset_name))
+
+            async with self.config.custom("EQUALIZER").all() as all_eqs:
+                for guild_id, preset_name in collected_for_removal:
+                    all_eqs[str(guild_id)]["eq_presets"][preset_name]["author"] = 0xDE1
--- a/redbot/cogs/audio/sql_statements.py
+++ b/redbot/cogs/audio/sql_statements.py
@@ -10,6 +10,8 @@ __all__ = [
     "PRAGMA_SET_read_uncommitted",
     "PRAGMA_FETCH_user_version",
     "PRAGMA_SET_user_version",
+    # Data Deletion statement
+    "HANDLE_DISCORD_DATA_DELETION_QUERY",
     # Playlist table statements
     "PLAYLIST_CREATE_TABLE",
     "PLAYLIST_DELETE",
@@ -80,6 +82,33 @@ PRAGMA_SET_user_version: Final[
     str
 ] = """
 pragma user_version=3;
+"""
+
+# Data Deletion
+# This is intentionally 2 seperate transactions due to concerns
+# Draper had. This should prevent it from being a large issue,
+# as this is no different than triggering a bulk deletion now.
+HANDLE_DISCORD_DATA_DELETION_QUERY: Final[
+    str
+] = """
+BEGIN TRANSACTION;
+
+UPDATE playlists
+SET deleted = true
+WHERE scope_id = :user_id ;
+
+UPDATE playlists
+SET author_id = 0xde1
+WHERE author_id = :user_id ;
+
+COMMIT TRANSACTION;
+
+BEGIN TRANSACTION;
+
+DELETE FROM PLAYLISTS
+WHERE deleted=true;
+
+COMMIT TRANSACTION;
 """
 
 # Playlist table statements
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -1,9 +1,10 @@
+import asyncio
 import re
 import random
 from datetime import datetime, timedelta
 from inspect import Parameter
 from collections import OrderedDict
-from typing import Iterable, List, Mapping, Tuple, Dict, Set
+from typing import Iterable, List, Mapping, Tuple, Dict, Set, Literal
 from urllib.parse import quote_plus
 
 import discord
@@ -11,7 +12,7 @@ from fuzzywuzzy import process
 
 from redbot.core import Config, checks, commands
 from redbot.core.i18n import Translator, cog_i18n
-from redbot.core.utils import menus
+from redbot.core.utils import menus, AsyncIter
 from redbot.core.utils.chat_formatting import box, pagify, escape, humanize_list
 from redbot.core.utils.predicates import MessagePredicate
 
@@ -40,14 +41,34 @@ class OnCooldown(CCError):
 
 class CommandObj:
     def __init__(self, **kwargs):
-        config = kwargs.get("config")
+        self.config = kwargs.get("config")
         self.bot = kwargs.get("bot")
-        self.db = config.guild
+        self.db = self.config.guild
 
     @staticmethod
     async def get_commands(config) -> dict:
         _commands = await config.commands()
         return {k: v for k, v in _commands.items() if _commands[k]}
+
+    async def redact_author_ids(self, user_id: int):
+
+        all_guilds = await self.config.all_guilds()
+
+        for guild_id in all_guilds.keys():
+            await asyncio.sleep(0)
+            async with self.config.guild_from_id(guild_id).commands() as all_commands:
+                async for com_name, com_info in AsyncIter(all_commands.items(), steps=100):
+                    if not com_info:
+                        continue
+
+                    if com_info.get("author", {}).get("id", 0) == user_id:
+                        com_info["author"]["id"] = 0xDE1
+                        com_info["author"]["name"] = "Deleted User"
+
+                    if editors := com_info.get("editors", None):
+                        for index, editor_id in enumerate(editors):
+                            if editor_id == user_id:
+                                editors[index] = 0xDE1
 
     async def get_responses(self, ctx):
         intro = _(
@@ -200,6 +221,17 @@ class CustomCommands(commands.Cog):
         self.commandobj = CommandObj(config=self.config, bot=self.bot)
         self.cooldowns = {}
 
+    async def red_delete_data_for_user(
+        self,
+        *,
+        requester: Literal["discord_deleted_user", "owner", "user", "user_strict"],
+        user_id: int,
+    ):
+        if requester != "discord_deleted_user":
+            return
+
+        await self.commandobj.redact_author_ids(user_id)
+
     @commands.group(aliases=["cc"])
     @commands.guild_only()
     async def customcom(self, ctx: commands.Context):
@@ -209,7 +241,7 @@ class CustomCommands(commands.Cog):
     @customcom.command(name="raw")
     async def cc_raw(self, ctx: commands.Context, command: str.lower):
         """Get the raw response of a custom command, to get the proper markdown.
-        
+
         This is helpful for copy and pasting."""
         commands = await self.config.guild(ctx.guild).commands()
         if command not in commands:
@@ -472,12 +504,14 @@ class CustomCommands(commands.Cog):
         if isinstance(responses, str):
             responses = [responses]
 
-        author = ctx.guild.get_member(cmd["author"]["id"])
-        # If the author is still in the server, show their current name
-        if author:
-            author = "{} ({})".format(author, cmd["author"]["id"])
+        _aid = cmd["author"]["id"]
+
+        if _aid == 0xDE1:
+            author = _("Deleted User")
+        elif member := ctx.guild.get_member(_aid):
+            author = f"{member} ({_aid})"
         else:
-            author = "{} ({})".format(cmd["author"]["name"], cmd["author"]["id"])
+            author = f"{cmd['author']['name']} ({_aid})"
 
         _type = _("Random") if len(responses) > 1 else _("Normal")
 
--- a/redbot/cogs/permissions/converters.py
+++ b/redbot/cogs/permissions/converters.py
@@ -138,12 +138,19 @@ class CogOrCommand(NamedTuple):
     # noinspection PyArgumentList
     @classmethod
     async def convert(cls, ctx: commands.Context, arg: str) -> "CogOrCommand":
-        cog = ctx.bot.get_cog(arg)
-        if cog:
-            return cls(type="COG", name=cog.__class__.__name__, obj=cog)
-        cmd = ctx.bot.get_command(arg)
-        if cmd:
-            return cls(type="COMMAND", name=cmd.qualified_name, obj=cmd)
+        ret = None
+        if cog := ctx.bot.get_cog(arg):
+            ret = cls(type="COG", name=cog.qualified_name, obj=cog)
+
+        elif cmd := ctx.bot.get_command(arg):
+            ret = cls(type="COMMAND", name=cmd.qualified_name, obj=cmd)
+
+        if ret:
+            if isinstance(ret.obj, commands.commands._RuleDropper):
+                raise commands.BadArgument(
+                    "You cannot apply permission rules to this cog or command."
+                )
+            return ret
 
         raise commands.BadArgument(
             _(
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -2,7 +2,7 @@ import asyncio
 import io
 import textwrap
 from copy import copy
-from typing import Union, Optional, Dict, List, Tuple, Any, Iterator, ItemsView, cast
+from typing import Union, Optional, Dict, List, Tuple, Any, Iterator, ItemsView, Literal, cast
 
 import discord
 import yaml
@@ -10,6 +10,7 @@ from schema import And, Or, Schema, SchemaError, Optional as UseOptional
 from redbot.core import checks, commands, config
 from redbot.core.bot import Red
 from redbot.core.i18n import Translator, cog_i18n
+from redbot.core.utils import AsyncIter
 from redbot.core.utils.chat_formatting import box
 from redbot.core.utils.menus import start_adding_reactions
 from redbot.core.utils.predicates import ReactionPredicate, MessagePredicate
@@ -113,6 +114,52 @@ class Permissions(commands.Cog):
         self.config.register_custom(COG)
         self.config.init_custom(COMMAND, 1)
         self.config.register_custom(COMMAND)
+
+    async def red_delete_data_for_user(
+        self,
+        *,
+        requester: Literal["discord_deleted_user", "owner", "user", "user_strict"],
+        user_id: int,
+    ):
+        if requester != "discord_deleted_user":
+            return
+
+        count = 0
+
+        _uid = str(user_id)
+
+        # The dict as returned here as string keys. Above is for comparison,
+        # there's a below recast to int where needed for guild ids
+
+        for typename, getter in ((COG, self.bot.get_cog), (COMMAND, self.bot.get_command)):
+
+            obj_type_rules = await self.config.custom(typename).all()
+
+            count += 1
+            if not count % 100:
+                await asyncio.sleep(0)
+
+            for obj_name, rules_dict in obj_type_rules.items():
+
+                count += 1
+                if not count % 100:
+                    await asyncio.sleep(0)
+
+                obj = getter(obj_name)
+
+                for guild_id, guild_rules in rules_dict.items():
+
+                    count += 1
+                    if not count % 100:
+                        await asyncio.sleep(0)
+
+                    if _uid in guild_rules:
+                        if obj:
+                            await self._remove_rule(
+                                CogOrCommand(typename, obj.qualified_name, obj),
+                                user_id,
+                                int(guild_id),
+                            )
 
     async def __permissions_hook(self, ctx: commands.Context) -> Optional[bool]:
         """
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -392,14 +392,6 @@ class Permissions(commands.Cog):
         if not who_or_what:
             await ctx.send_help()
             return
-        if isinstance(cog_or_command.obj, commands.commands._AlwaysAvailableCommand):
-            await ctx.send(
-                _(
-                    "This command is designated as being always available and "
-                    "cannot be modified by permission rules."
-                )
-            )
-            return
         for w in who_or_what:
             await self._add_rule(
                 rule=cast(bool, allow_or_deny),
@@ -434,14 +426,6 @@ class Permissions(commands.Cog):
         """
         if not who_or_what:
             await ctx.send_help()
-            return
-        if isinstance(cog_or_command.obj, commands.commands._AlwaysAvailableCommand):
-            await ctx.send(
-                _(
-                    "This command is designated as being always available and "
-                    "cannot be modified by permission rules."
-                )
-            )
             return
         for w in who_or_what:
             await self._add_rule(
@@ -520,14 +504,6 @@ class Permissions(commands.Cog):
         `<cog_or_command>` is the cog or command to set the default
         rule for. This is case sensitive.
         """
-        if isinstance(cog_or_command.obj, commands.commands._AlwaysAvailableCommand):
-            await ctx.send(
-                _(
-                    "This command is designated as being always available and "
-                    "cannot be modified by permission rules."
-                )
-            )
-            return
         await self._set_default_rule(
             rule=cast(Optional[bool], allow_or_deny),
             cog_or_cmd=cog_or_command,
@@ -551,14 +527,6 @@ class Permissions(commands.Cog):
         `<cog_or_command>` is the cog or command to set the default
         rule for. This is case sensitive.
         """
-        if isinstance(cog_or_command.obj, commands.commands._AlwaysAvailableCommand):
-            await ctx.send(
-                _(
-                    "This command is designated as being always available and "
-                    "cannot be modified by permission rules."
-                )
-            )
-            return
         await self._set_default_rule(
             rule=cast(Optional[bool], allow_or_deny), cog_or_cmd=cog_or_command, guild_id=GLOBAL
         )
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -162,7 +162,8 @@ class Permissions(commands.Cog):
                                 int(guild_id),
                             )
                         else:
-                            await self.config.custom(typename,).clear_raw(guild_id, user_id)
+                            grp = self.config.custom(typename, obj_name)
+                            await grp.clear_raw(guild_id, user_id)
 
     async def __permissions_hook(self, ctx: commands.Context) -> Optional[bool]:
         """
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -155,11 +155,14 @@ class Permissions(commands.Cog):
 
                     if _uid in guild_rules:
                         if obj:
+                            # delegate to remove rule here
                             await self._remove_rule(
                                 CogOrCommand(typename, obj.qualified_name, obj),
                                 user_id,
                                 int(guild_id),
                             )
+                        else:
+                            await self.config.custom(typename,).clear_raw(guild_id, user_id)
 
     async def __permissions_hook(self, ctx: commands.Context) -> Optional[bool]:
         """
--- a/redbot/cogs/reports/reports.py
+++ b/redbot/cogs/reports/reports.py
@@ -81,7 +81,7 @@ class Reports(commands.Cog):
                 if not steps % 100:
                     await asyncio.sleep(0)  # yield context
 
-            if ticket.get("user_id", 0) == user_id:
+            if ticket.get("report", {}).get("user_id", 0) == user_id:
                 paths.append((guild_id_str, ticket_number))
 
         async with self.config.custom("REPORT").all() as all_reports:
--- a/redbot/cogs/reports/reports.py
+++ b/redbot/cogs/reports/reports.py
@@ -86,7 +86,7 @@ class Reports(commands.Cog):
 
         async with self.config.custom("REPORT").all() as all_reports:
             async for guild_id_str, ticket_number in AsyncIter(paths, steps=100):
-                r = all_reports[guild_id_str][ticket_number]
+                r = all_reports[guild_id_str][ticket_number]["report"]
                 r["user_id"] = 0xDE1
                 # this might include EUD, and a report of a deleted user
                 # that's been unhandled for long enough for the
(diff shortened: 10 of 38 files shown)


Dav-Git:
Would handling the delete/request command be the responsibility of the CC or do you plan to have something like ``[p]data request <User>`` and ``[p]data delete <User>`` somewhere in core?
I think other than those two types of operations not much more is needed in terms of the API

Drapersniper:
I think is infesable to leave this to core and core alone. 

Core can't know all the data stored (Custom scopes) so it should be left to CC unless there is a way to make core away of EUD stored in custom scopes.

In adition to the above, I feel this should be made a requirement for Approved CCs

Dav-Git:
Couldn't the invocation of the command do an event thingie which CCs would then be able to handle?

Dav-Git:
I just think it would be pretty complicated if the deletion request would have to be done individually per COG

Drapersniper:
Oh I see what you meant, when a request is done it is dispatched as an event that cogs have to handle.

Yeah that doesnt sound bad and leave it to CC To implement their data deletion on a per cog basis while the user would only have to send the request once. via a global command that dispatch the event to core and cogs to do the right thing.

Dav-Git:
Exactly what I meant with better words

mikeshardmind:
The intent is that CCs *only* have to handle those 2 methods, core handles every other facet of this process. I'll have more to add to this later and just wanted to ensure since this is the part cog creators will need to interact with to support, that there was visibility on this part first.

mikeshardmind:
I need to take the time to write up a few additional thoughts that came up for this (which probably need discussion before continuing) and have not yet. This PR is not being abandoned but may end up slightly delayed.

mikeshardmind:
So one of the goals this is intended to work towards is allowing Red to be used in use cases that need to consider various laws that allow users to request their data. I'm struggling to come up with a universal API here, as there are various things cogs can store that won't fit into a simple solution of "just shove it all in a JSON file" and am currently thinking that each cog on the bot will have to handle sending the user a message with data the cog has.

Clearly this is not ideal, but I'm not sure how else this can remain compliant with requirements that include the data be human-understandable, and clear as to what the data represents.

The request API might need to shift in this direction, but I'm open to suggestions which allow something more streamlined.

mikeshardmind:
Proposed APIs have been updated, these may need more review before commands, events, and UX are designed around this.

mikeshardmind:
I've linked a PR from my cog repo to this to serve as a demonstration for the intended way of using the deletion portion of this, as well as how some of the handlings of this may be subject to some discretion.

None of my cogs make great candidates for needing to share data to end-users, except the one which was already documented as being a potential liability for bot owners and included a giant warning on install that it did not and would not respect this sort of API and to only use if you could manually accommodate requests as legally required.

I'm hoping for feedback on the design from both a compliance standpoint and from cog creators to ensure the ease of use before attaching too many more mechanisms to this.

mikeshardmind:
Cog creators need enough time to get their cogs in line with what we need prior to discord's deadline in mid August, and I need enough agreement on how we're handling this to have the remainder of this finished in time for then as well

For technical handling, the above APIs cover the cogs actually interacting with the bot on removal requests.

The following are also things I'd like approval on being the way to handle

- A new info.json key named `end_user_data_statement` which should hold the same info as the similarly named field in the cog class, but be accessible prior to loading the cog. This one is for use by downloader to ensure that the bot owner is aware of if a cog is going to be a liability for them prior to loading it.

- Unloading all cogs forcibly on the next bot update so that owners get an opportunity to see these statements.

- Downloader warning users if these statements change on a cog update

- Adding a core command which ejects a cog from the bot and which wipes its data folder (for use with non-compliant cogs that the owner does not want anymore)

As a matter of policy, I'm suggesting other things in an additional comment, but as long as I can get approval on the technical portions above, we can present this to cog creators to give them enough notice, and I can get to finishing work on this.

Dav-Git:
One thing about this I would like to amend is the forced unloading of cogs. When we do this (which I agree is a good way to raise awareness) I think we should show which cogs were loaded prior to forcefully unloading all.

mikeshardmind:
Policy:

The following are merely suggestions for how this should interact with the cog ecosystem

### Core cogs
 
Core cogs should support as much of it as possible, including user sourced deletion requests which don't have conflicts of interest (like removing self from a blocked user list) as well as explicitly saying what data each cog + core itself stores and for what purposes. This should additionally be documented in our user-facing docs somewhere

### Approved CCs

Cogs in an approved repo should at least mark what data they store and why + how much they support the Data API

This avoids current approved Cog Creators needing to be on this immediately if they are not capable, while still giving end-users enough information to make an informed decision about the cogs (However, note on this below)

Approved CCs which do not comply with this should be removed.

### Sr CCs

(Note: this references an original term which was intended for use and designed as part of Cog Creator applications being opened, but the Sr Apps have never been available. Should this remain so, all of these requirements need to shift to Approved CCs for that distinction to do its job of protecting users and not just being a label handed out)

Sr. Cog Creators should mark what data they store and why + how much they support the Data API
They should also remove data if it's a request from discord or the bot owner.
They may keep some EUD as needed for the bot.
If they receive a user request, they can keep data needed for the cog
If they receive a user_strict request, they can keep data (such as snowflakes and timestamps of interactions) needed for anti-abuse measures, but not EUD

Drapersniper:
> Cog creators need enough time to get their cogs in line with what we need prior to discord's deadline in mid August, and I need enough agreement on how we're handling this to have the remainder of this finished in time for then as well
> 
> For technical handling, the above APIs cover the cogs actually interacting with the bot on removal requests.
> 
> The following are also things I'd like approval on being the way to handle
> 
> * A new info.json key named `end_user_data_statement` which should hold the same info as the similarly named field in the cog class, but be accessible prior to loading the cog. This one is for use by downloader to ensure that the bot owner is aware of if a cog is going to be a liability for them prior to loading it.
> * Unloading all cogs forcibly on the next bot update so that owners get an opportunity to see these statements.
> * Downloader warning users if these statements change on a cog update
> * Adding a core command which ejects a cog from the bot and which wipes its data folder (for use with non-compliant cogs that the owner does not want anymore)
> 
> As a matter of policy, I'm suggesting other things in an additional comment, but as long as I can get approval on the technical portions above, we can present this to cog creators to give them enough notice, and I can get to finishing work on this.

I have nothing wrong with this section.

Sure it can be percieved as a one time annoyance ... but its for the best, ensure bot owners see the EUD statement, so +1 from me.

Flame442:
Two things that cannot be easily handled when a data deletion request occurs are logs and data from unloaded cogs. There should be some additional documentations / warnings to inform users that these may still be storing data after a deletion.

Kowlin:
@mikeshardmind You have my go ahead.

mikeshardmind:
the "discord" arg will instead be "discord_deleted_user" to ensure future proofness on this, as this is currently the only request like this discord has informed devs of, but may not be the only one they ever have.

mikeshardmind:
```py
    #: This is a string which may be presented to users about what data you store and why
    `__red_end_user_data_statement__ = None`
```
This is being removed from the cog and moved to the extension level (parallel with the setup function in an extension) to better support varied cog types.

mikeshardmind:
Due to a valid concern that was brought up, the API for getting user data has been changed to not make each cog responsible for sending to the user. Sorry for any disruption this causes to people already planning.

LGACode:
This PR looks closely related to an issue that I posted a while back for a Privacy API. I can't speak much to how it should be implemented technically, but that issue discusses what I would expect as a business owner to be able to provide to my end users. My concerns are mostly rooted in the new Colorado privacy law which was recently passed and came into effect and can be viewed below, however it's not necessary to consider it specifically per say because it follows the standard of many other laws already in place such as California's and GPDR which I suspect to be a major factor in this change.

https://github.com/Cog-Creators/Red-DiscordBot/issues/3239

https://leg.colorado.gov/bills/hb18-1128

mikeshardmind:
oh right, this does close that issue too (updates the PR description)

mikeshardmind:
Python doesn't have built-in support for multipart archives (in either `zipfile` or `tarfile`), which brings up an unfortunate issue with getting data back to users in standard formats when combined with the prior concerns brought up about rate-limits and with the potential number of cogs loaded.

The cog creator api for this particular function should probably not need to change at all, but it may not be used as part of this PR.

mikeshardmind:
A few editorial decisions about cogs that are in so far:

1. If a user wants to wipe their own bank acct, there's no reason not to let them.
2. Modlog doesn't allow user requests, anonymizes the user id and drops the nickname if asked by discord.
3. Alias anonymizes the user (we can probably do a migration to kill this key, it doesn't ever get used, however, I left it in case of future plans I'm unaware of)

Waiting on a response to a question in discord to handle audio, and I have a few other places in core that need looking at.

Oh, and anonymized user IDs are consistently: `0xde1` because I found it amusing and it can't conflict with a valid snowflake

redbot/cogs/audio/sql_statements.py
```diff
@@ -82,6 +84,23 @@
...
 """
 
+# Data Deletion
+HANDLE_DISCORD_DATA_DELETION_QUERY: Final[
+    str
+] = """
+BEGIN TRANSACTION;
+
+UPDATE playlists
+SET deleted = true
```
DiscordLiz, line 97: Is this guaranteed to cause a deletion?

redbot/cogs/permissions/permissions.py
```diff
@@ -114,6 +115,49 @@ def __init__(self, bot: Red):
...
+        user_id: int,
+    ):
+        if requester != "discord_deleted_user":
+            return
+
+        count = 0
+
+        _uid = str(user_id)
+
+        # The dict as returned here as tring keys. Above is for comparison,
```
DiscordLiz: "tring"

redbot/cogs/warnings/warnings.py
```diff
@@ -45,6 +46,21 @@ def __init__(self, bot: Red):
...
+        user_id: int,
+    ):
+        if requester != "discord_deleted_user":
+            return
+
+        all_members = await self.config.all_members()
+
+        async for guild_id, guild_data in AsyncIter(all_members.items(), steps=100):
+            if user_id in guild_data:
+                await self.config.member_from_ids(guild_id, user_id).clear()
```
DiscordLiz: I see a lot of blocks like this. Isn't this horrible for config?

redbot/core/modlog.py
```diff
@@ -47,10 +47,35 @@
...
+        async for guild_id_str, guild_cases in AsyncIter(all_cases.items(), steps=100):
+            async for case_num_str, case in AsyncIter(guild_cases.items(), steps=100):
+                if case.get("user", 0) == user_id:
+                    key_paths.append((guild_id_str, case_num_str))
+
+        async with _config.custom(_CASES).all() as all_cases:
+            for guild_id_str, case_num_str in key_paths:
+                case = all_cases[guild_id_str][case_num_str]
+                case["user"] = 0xDE1
+                case.pop("last_nown_username", None)
```
DiscordLiz: Why can't the guild scope do something like this to reduce writes?

redbot/cogs/audio/sql_statements.py
mikeshardmind: Technically... No.

redbot/cogs/warnings/warnings.py
mikeshardmind: Yes, yes it is.

redbot/core/modlog.py
mikeshardmind: "It's complicated"

I'd either need to extend config as part of this PR too or use things we can't support being done by 3rd party creators. As many creators use core cogs as a reference, I'll leave this to be improved in future work.

mikeshardmind:
This PR is *dangerous* to test right now as it has changes to prevent core commands from being locked out with permissions and to keep certain commands available. (0ddb7ea) While these changes should be fine, there's a non-zero chance (needs further testing) that there is an issue with the implementation. 

Please exercise caution when testing this.

mikeshardmind:
Live testing shows that the changes which had potential to be dangerous are safe enough, but the caution was still warranted in testing this.

redbot/cogs/audio/sql_statements.py
mikeshardmind: handled now

mikeshardmind:
Ready for review.

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,76 @@
...
+If users cannot use that for some reason, the command
+
+``[p]mydata ownermanagement delete for user``
+
+exists as a way to handle this as if the user had done it themselves
+
+Be careful about using the other owner level deletion options on behalf of users,
+as this may also result in losing operational data such as data used to prevent spam.
+
+What owners and hosts responsible for
```
Dav-Git: Should be "What owners and hosts are responsible for"

redbot/core/commands/commands.py
```diff
@@ -809,6 +820,100 @@ def help(self):
...
+        This method may also be implemented for an extension.
+
+        Parameters
+        ----------
+        user_id: int
+
+        Returns
+        -------
+        io.BytesIO
+            A BytesIO object suitable to send as a faile or as part
```
Dav-Git: There's an a in "file" that shouldn't be there

Dav-Git (changes requested):
Found these while reading the docs. Otherwise documentation looks good so far. I did not review the actual code.

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,76 @@
...
+
+There are a handful of these available to bot owners in the command group
+``[p]mydata ownermanagement``
+
+The most pertinent one if asked to delete data by a member of Trust & Safety
+is
+
+``[p]mydata ownermanagement processdiscordrequest``
+
+which will more completely remove a user.
```
Stonedestroyer, line 53: Sounds grammatically wrong.

Stonedestroyer: Sounds wrong, not sure what to change it to. 

which will completely purge a user may be one.

redbot/core/commands/commands.py
```diff
@@ -809,6 +820,100 @@ def help(self):
...
+        user_id: int,
+    ):
+        """
+        This should be overridden by all cogs that store data.
+
+        This method may also be implemented for an extension.
+
+        Parameters
+        ----------
+        requester: Literal["discord", "owner", "user"]
```
Jackenmen:
```suggestion
        requester: Literal["discord_deleted_user", "owner", "user", "user_strict"]
```

docs/guide_cog_creation.rst
```diff
@@ -229,3 +232,20 @@ Not all of these are strict requirements (some are) but are all generally advisa
 
   - We announce this in advance.
   - If you need help, ask.
+
+14. Handle user data responsibly
+
+  - Don't do unexpected things with user data.
+  - Don't expose user data to additional audiences without permission.
+  - Don't collect data your cogs don't need.
+  - Don't store data in unepected locations.
```
Flame442:
```suggestion
  - Don't store data in unexpected locations.
```

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,76 @@
...
+=====================
+
+Notes for everyone
+******************
+
+What data Red collects
+----------------------
+
+Red, and the cogs included with it collect some amount of data
+about users for it's normal operations.
```
Flame442: A bit pedantic, but the commas here are awkward

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,76 @@
...
+which actions refer to your discord account (such as creating a playlist)
+as well as the content of specific messages used directly as commands for the bot
+(such as reports sent to servers).
+
+By default, Red will not collect any more data than it needs, and will not use it
+for anything other than the portion of the Red's functionality that necessitated it.
+
+3rd party extensions may store additional things beyond this. You can use the command
+``[p]mydata 3rdparty`` to view statements about this made by the authors of 
+the specific 3rd party extensions an instance of Red has installed.
```
Flame442: A few usages of "things" and "this" that make this block sound somewhat lacking in direction.

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,76 @@
...
+which will more thoroughly remove data connected to a Discord user ID.
+
+How to process deletion requests from users
+-------------------------------------------
+
+You can point users to the command ``[p]mydata forgetme`` as a first step.
+
+If users cannot use that for some reason, the command
+
+``[p]mydata ownermanagement delete for user``
```
Flame442: Is this supposed to have spaces in `delete for user`?

redbot/core/commands/commands.py
```diff
@@ -809,6 +820,100 @@ def help(self):
         if doc:
             return inspect.cleandoc(translator(doc))
 
+    async def red_get_data_for_user(self, *, user_id: int) -> MutableMapping[str, io.BytesIO]:
+        """
+        This should be overridden by cogs
+        that support users requesting their data
+        as well as cogs that don't store data.
+
+        Overriden implementations should send a io.BytesIO containing a
```
Flame442:
```suggestion
        Overridden implementations should send a io.BytesIO containing a
```

Flame442: Should this say "send" or "return"?
Also, it just says it needs to return the `io.BytesIO`, but the function states it returns a `MutableMapping[str, io.BytesIO]`. This should probably be explained in this more human-readable section.

redbot/core/commands/commands.py
```diff
@@ -809,6 +820,100 @@ def help(self):
...
+        user_id: int
+
+        Returns
+        -------
+        MutableMapping[str, io.BytesIO]
+            A mapping of filenames to BytesIO objects
+            suitable to send as a files or as part of an archive to a user.
+
+            .. note::
+                Cogs which explicitly do not store EUD should return
```
Flame442: Is this return *just* returning `commands.RED_NO_UED` (NOT in a MutableMapping)? Maybe add this to the return hints? `typing.Union[MutableMapping[str, io.BytesIO], commands.RED_NO_UED]`

redbot/core/commands/commands.py
```diff
@@ -809,6 +820,100 @@ def help(self):
...
+                The request should be processed as if
+                Discord has asked for the data removal
+                This then additionally must treat the
+                user ID itself as something to be deleted.
+                The user ID is no longer operational data
+                as the ID no longer refers to a valid user.
+            owner:
+                The request was made by the bot owner.
+                If removing the data requested by the owner
+                would be an operational hazzard
```
Flame442:
```suggestion
                would be an operational hazard
```

redbot/core/commands/commands.py
```diff
@@ -809,6 +820,100 @@ def help(self):
...
+                but there must remain a way for them to process this request.
+            user_strict:
+                The request was made by a user,
+                the bot settings allow a user to request their own data
+                be deleted, and the bot is configured to respect this
+                at the cost of functionality.
+                Cogs may retain data needed for anti abuse measures
+                such as IDs and timestamps of interactions,
+                but should not keep EUD such
+                as user nicknames if recieving a request of this nature.
```
Flame442:
```suggestion
                as user nicknames if receiving a request of this nature.
```

redbot/core/core_commands.py
```diff
@@ -443,6 +468,407 @@ async def uptime(self, ctx: commands.Context):
...
+    @commands.cooldown(1, 600, commands.BucketType.user)
+    @mydata.command(cls=commands.commands._AlwaysAvailableCommand, name="whatdata")
+    async def mydata_whatdata(self, ctx: commands.Context):
+        """ Find out what type of data [botname] stores and why """
+
+        ver = "latest" if VersionInfo.from_str(__version__).dev_release else "stable"
+        link = f"https://docs.discord.red/en/{ver}/red_core_data_statement.html"
+        await ctx.send(
+            _(
+                "This bot stores some data about users as neccessary to function. "
```
```suggestion
                "This bot stores some data about users as necessary to function. "
```

redbot/core/core_commands.py
```diff
@@ -443,6 +468,407 @@ async def uptime(self, ctx: commands.Context):
...
+                check=lambda m: m.channel.id == ctx.channel.id and m.author.id == ctx.author.id,
+                timeout=30,
+            )
+        except asyncio.TimeoutError:
+            await ctx.send(_("Did not get confirmation, cancelling."))
+        else:
+            if message.content.strip() == confirm_token:
+                return True
+            else:
+                await ctx.send(_("Did not get a matching confimration, cancelling."))
```
```suggestion
                await ctx.send(_("Did not get a matching confirmation, cancelling."))
```

redbot/core/core_commands.py
```diff
@@ -443,6 +468,407 @@ async def uptime(self, ctx: commands.Context):
...
+
+        closing = "</body></html>"
+
+        html = "\n".join((pretty_things, generated, closing))
+
+        fp = io.BytesIO(html.encode())
+
+        await ctx.send(file=discord.File(fp, filename="3rd-party.html"))
+
+    async def get_serious_confirmation(self, ctx: commands.Context, prompt: str) -> bool:
```
Flame442, line 575: Does this return `None` on fail? I don't see any explicit `return False`s even though the return type is `bool`.

redbot/core/core_commands.py
```diff
@@ -443,6 +468,407 @@ async def uptime(self, ctx: commands.Context):
...
+            )
+        except asyncio.TimeoutError:
+            await ctx.send(_("Did not get confirmation, cancelling."))
+        else:
+            if message.content.strip() == confirm_token:
+                return True
+            else:
+                await ctx.send(_("Did not get a matching confimration, cancelling."))
+
+    # 1 per day, not stored to config to avid this being more stored data.
```
```suggestion
    # 1 per day, not stored to config to avoid this being more stored data.
```

redbot/core/core_commands.py
```diff
@@ -443,6 +468,407 @@ async def uptime(self, ctx: commands.Context):
...
+    # and small bots that do restart often don't have enough
+    # users for this to be an issue.
+    @commands.cooldown(1, 86400, commands.BucketType.user)
+    @mydata.command(cls=commands.commands._AlwaysAvailableCommand, name="forgetme")
+    async def mydata_forgetme(self, ctx: commands.Context):
+        """
+        Have [botname] forget what it knows about you.
+
+        This may not remove all data about you, data needed for operation,
+        such as command cooldowns will be kept until no longer neccessary.
```
```suggestion
        such as command cooldowns will be kept until no longer necessary.
```

redbot/core/core_commands.py
```diff
@@ -443,6 +468,407 @@ async def uptime(self, ctx: commands.Context):
...
+
+        Further interactions with [botname] may cause it to learn about you again.
+        """
+        if ctx.assume_yes:
+            # lol, no, we're not letting users schedule deletions every day to thrash the bot.
+            ctx.command.reset_cooldown(ctx)  # We will however not let that lock them out either.
+            return await ctx.send(_("This command does not support non-interactive usage"))
```
Flame442: You *might* want to include the command name in this message.

redbot/core/core_commands.py
```diff
@@ -443,6 +468,407 @@ async def uptime(self, ctx: commands.Context):
...
+
+        if not await self.get_serious_confirmation(
+            ctx,
+            _(
+                "This will cause the bot to get rid of and/or dissasociate "
```
```suggestion
                "This will cause the bot to get rid of and/or disassociate "
```

redbot/core/core_commands.py
```diff
@@ -443,6 +468,407 @@ async def uptime(self, ctx: commands.Context):
...
+                ).format(mention=ctx.author.mention, cogs=humanize_list(results.failed_cogs))
+            )
+        elif results.failed_modules:
+            await ctx.send(
+                _(
+                    "I tried to delete all non-operational data about you "
+                    "(that I know how to delete) "
+                    "{mention}, however the following modules errored: {modules}.\n"
+                    "Please contact the owner of this bot to address this."
+                ).format(mention=ctx.author.mention, modules=humanize_list(results.failed_modules))
```
Flame442: You might want to say that the modules / cogs not included *did* work.

redbot/core/core_commands.py
```diff
@@ -443,6 +468,407 @@ async def uptime(self, ctx: commands.Context):
...
+                    "I tried to delete all data about that user, "
+                    "(that I know how to delete) "
+                    "however the following modules errored: {modules}.\n"
+                    "Please check your logs and contact the creators of "
+                    "these cogs and modules."
+                ).format(modules=humanize_list(results.failed_modules))
+            )
+        else:
+            await ctx.send(
+                _("I've deleted all data about that user " "that I know how to delete.")
```
Flame442, line 971: Why is this string weird?

Flame442 (changes requested):
Didn't look at *too* much code, this is mostly spelling fixes

redbot/core/core_commands.py
mikeshardmind: I preemptively split it for screen width since I'm used to a smaller allowed width, black reassembled it.

redbot/core/commands/commands.py
mikeshardmind: Can't put that in the type hint, type would be

`typing.Union[MutableMapping[str, io.BytesIO], object]` at that point.

docs/red_core_data_statement.rst
mikeshardmind: Nope, fixing that...

redbot/core/core_commands.py
mikeshardmind: I figured the "however" bit implies what follows is an exception, but I added clarity in each place where this happens anyhow.

docs/red_core_data_statement.rst
mikeshardmind: Agreed, I've removed that one.

redbot/core/core_commands.py
mikeshardmind: I missed this in the original batching but fixing this as well.

redbot/core/core_commands.py
mikeshardmind: Same on missing this.

redbot/core/core_commands.py
mikeshardmind: I was being lazy with the translation handling here, but I've gone and added that for clarity in case other commands decide to vocally complain about being scheduled.

docs/red_core_data_statement.rst
mikeshardmind: I think the updated handling should resolve this, but let me know if more should be done.

redbot/core/commands/commands.py
mikeshardmind: So, I've updated this a bit more. `commands.RED_NO_EUD` is gone. If `typing.Literal` were a bit more powerful, I'd be happier about special sentinel values.

mikeshardmind:
This won't break anyone who doesn't implement the methods, but per discussion in discord, the methods must be overridden now so that we can properly signal if a cog handled it (even if that handling involves nothing due to no data)

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,77 @@
...
+In particular the bot will keep track of a short history of usernames/nicknames
+which actions refer to your discord account (such as creating a playlist)
+as well as the content of specific messages used directly as commands for the bot
+(such as reports sent to servers).
+
+By default, Red will not collect any more data than it needs, and will not use it
+for anything other than the portion of the Red's functionality that necessitated it.
+
+3rd party extensions may store additional data beyond what Red does by default.
+You can use the comman ``[p]mydata 3rdparty``
```
Flame442:
```suggestion
You can use the command ``[p]mydata 3rdparty``
```

Flame442 (dismissed):

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,77 @@
...
+
+There are a handful of these available to bot owners in the command group
+``[p]mydata ownermanagement``.
+
+The most pertinent one if asked to delete data by a member of Trust & Safety
+is
+
+``[p]mydata ownermanagement processdiscordrequest``
+
+which will more thoroughly remove data connected to a Discord user ID.
```
Jackenmen: Could this be made a bit less vague and include some information that the confirmation from the command itself provides?

mikeshardmind:
This is missing the settings for user/user_strict/off I'm placing this back as a draft so that this doesn't get missed because with the scope of this PR, it was something which could have fallen through the cracks had questions not been asked.

mikeshardmind:
This has the settings commands which were intended and utilizes them now.

docs/guide_cog_creation.rst
```diff
@@ -238,3 +241,20 @@ Not all of these are strict requirements (some are) but are all generally advisa
...
+  - Don't do unexpected things with user data.
+  - Don't expose user data to additional audiences without permission.
+  - Don't collect data your cogs don't need.
+  - Don't store data in unexpected locations.
+    Utilize the cog data path, Config, or if you need something more
+    prompt the owner to provide it.
+
+17. Utilize the data deletion and statement APIs
+
+  - See ``redbot.core.commands.Command.red_delete_data_for_user``
```
Jackenmen: Making this into a link is a bit more helpful + this should be `Cog`, not `Command`
```suggestion
  - See `redbot.core.commands.Cog.red_delete_data_for_user`
```

docs/guide_cog_creation.rst
```diff
@@ -98,6 +98,9 @@ Open :code:`__init__.py`. In that file, place the following:
 
     from .mycog import Mycog
 
+    # you can set this to whatever accurate statement you want to make here.
+    __red_end_user_data_statement__ = "This cog doesn't store any user data"
```
Jackenmen: I'm not entirely sure if including this in a "first cog guide" is really a good idea. Beginners get confused by all kind of things and this could very well become one of them.

redbot/cogs/customcom/customcom.py
```diff
@@ -40,15 +41,35 @@ class OnCooldown(CCError):
...
+
+        for guild_id in all_guilds.keys():
+            await asyncio.sleep(0)
+            async with self.config.guild_from_id(guild_id).commands() as all_commands:
+                async for com_name, com_info in AsyncIter(all_commands.items(), steps=100):
+                    if not com_info:
+                        continue
+
+                    if com_info.get("author", {}).get("id", 0) == user_id:
+                        com_info["author"]["id"] = 0xDE1
```
Jackenmen, line 65: Maybe you could add some special handling (i.e. not showing the placeholder ID as you did in modlog) when this is used?

redbot/cogs/permissions/permissions.py
```diff
@@ -114,6 +115,49 @@ def __init__(self, bot: Red):
...
+        count = 0
+
+        _uid = str(user_id)
+
+        # The dict as returned here as string keys. Above is for comparison,
+        # there's a below recast to int where needed for guild ids
+
+        for typename, getter in ((COG, self.bot.get_cog), (COMMAND, self.bot.get_command)):
+
+            async with self.config.custom(typename).all() as cog_rules:
```
Jackenmen: Considering that you use `_remove_rule()` below, does this need to be using ctx manager? Especially that it could result in something getting overridden if there would be a context switch while handling the deletion request (although that does seem rather unlikely).

redbot/core/commands/commands.py
```diff
@@ -66,6 +70,10 @@
 DisablerDictType = MutableMapping[discord.Guild, Callable[["Context"], Awaitable[bool]]]
 
 
+class RedUnhandledAPI(Exception):
```
Jackenmen, line 74: This hasn't been added to `__all__` or imported in `redbot/core/commands/__init__.py`, if that's intentional then it should probably be made private (although I can technically see use case for this if Cog Creator would want to raise for any `requester` that isn't supported by cog, i.e. if we would add a new one in future, see below for more info on that though).

redbot/core/bot.py
```diff
@@ -1040,7 +1048,7 @@ async def load_extension(self, spec: ModuleSpec):
 
     def remove_cog(self, cogname: str):
         cog = self.get_cog(cogname)
-        if cog is None:
+        if cog is None or isinstance(cog, Core):
```
Jackenmen: Isn't this supposed to be checking for `_RuleDropper` (although in terms of cogs this is practically equivalent)?
```suggestion
        if cog is None or isinstance(cog, commands.commands._RuleDropper):
```

redbot/core/bot.py
```diff
@@ -1394,6 +1405,114 @@ async def shutdown(self, *, restart: bool = False):
         await self.logout()
         sys.exit(self._shutdown_mode)
 
+    async def _core_data_deletion(
```
Jackenmen, line 1410: This doesn't handle `autoimmune_ids` from guild scope and it can contain user IDs.

redbot/core/bot.py
```diff
@@ -1394,6 +1405,114 @@ async def shutdown(self, *, restart: bool = False):
...
+    ):
+        if requester != "discord_deleted_user":
+            return
+
+        await self._config.user_from_id(user_id).clear()
+        all_members = await self._config.all_members()
+
+        async for guild_id, guild_data in AsyncIter(all_members.items(), steps=100):
+            if user_id in guild_data:
+                await self._config.member_from_ids(guild_id, user_id).clear()
```
Jackenmen: I can't see any defaults registered for member group in `RedBase.__init__`, am I missing something?

redbot/core/bot.py
```diff
@@ -1394,6 +1405,114 @@ async def shutdown(self, *, restart: bool = False):
...
+        """
+        This tells each cog and extension, as well as any APIs in Red
+        to go delete data
+
+        Calling this should be limited to interfaces designed for it.
+
+        Parameters
+        ----------
+        requester
+        user_id
```
Jackenmen, line 1450: This docstring should at least refer to `red_delete_data_for_user()` for description of parameters so that devs know how to call it.

redbot/core/bot.py
```diff
@@ -1394,6 +1405,114 @@ async def shutdown(self, *, restart: bool = False):
...
+                log.warning(f"{stype}.{sname} did not handle data deletion ")
+                failures["unhandled"].append(sname)
+            except Exception as exc:
+                log.exception(f"{stype}.{sname} errored when handling data deletion ")
+                failures[stype].append(sname)
+
+        handlers = [
+            *(wrapper(coro, "extension", name) for name, coro in extension_handlers.items()),
+            *(wrapper(coro, "cog", name) for name, coro in cog_handlers.items()),
+            *(h(requester=requester, user_id=user_id) for h in special_handlers),
```
Jackenmen: I think that handling failures of special handlers wouldn't be a bad idea. I know that it's part of core deletions so it *shouldn't break™*, but it seems a bit safer to just handle it in case anything goes wrong.

redbot/core/commands/commands.py
```diff
@@ -809,6 +818,107 @@ def help(self):
...
+        This should be overridden by all cogs.
+
+        Overridden implementations should return a mapping of filenames to io.BytesIO
+        containing a human-readable version of the data
+        the cog has about the specified user_id or an empty mapping
+        if the cog does not have end user data.
+
+        The data should be easily understood for what it represents to
+        most users of age to use Discord.
+        You may include multiple files by providing an archive,
```
Jackenmen: Now that this can return dictionary, the part about archive is probably unneeded.

redbot/core/commands/commands.py
```diff
@@ -809,6 +818,107 @@ def help(self):
...
+        be done to indicate that this has been considered.
+
+        Parameters
+        ----------
+        requester: Literal["discord_deleted_user", "owner", "user", "user_strict"]
+
+            Note: you should safely handle
+            any string value (log a warning if needed)
+            as additional requester types may be added
+            in the future without prior warning.
```
Jackenmen: What "safely handle" would mean here? For me as CC, raising `RedUnhandledAPI` in such cases (or technically just call to `super()`) sounds like the best idea as I can't really know what kind of requester type gets added (and therefore can't make any assumptions about what to do with it), but I am not sure if this is what you had in mind.

redbot/core/commands/commands.py
```diff
@@ -809,6 +818,107 @@ def help(self):
...
+                as user nicknames if receiving a request of this nature.
+            user:
+                The request was made by a user,
+                the bot settings allow a user to request their own data
+                be deleted, and the bot is configured to let cogs keep
+                data needed for operation.
+                Under this case, you may elect to retain data which is
+                essential to the functionality of the cog. This case will
+                only happen if the bot owner has opted into keeping
+                minimal EUD needed for cog functionality.
```
Jackenmen: The method is currently showing like this:
image

I propose to make it look like this:
image

The relevant rst code:
```suggestion
            This method can currently passed one of these strings:

            - ``"discord_deleted_user"``

                The request should be processed as if
                Discord has asked for the data removal
                This then additionally must treat the
                user ID itself as something to be deleted.
                The user ID is no longer operational data
                as the ID no longer refers to a valid user.

            - ``"owner"``

                The request was made by the bot owner.
                If removing the data requested by the owner
                would be an operational hazard
                (such as removing a user id from a blocked user list)
                you may elect to inform the user of an alternative way
                to remove that ID to ensure the process can not be abused
                by users to bypass anti-abuse measures,
                but there must remain a way for them to process this request.

            - ``"user_strict"``

                The request was made by a user,
                the bot settings allow a user to request their own data
                be deleted, and the bot is configured to respect this
                at the cost of functionality.
                Cogs may retain data needed for anti abuse measures
                such as IDs and timestamps of interactions,
                but should not keep EUD such
                as user nicknames if receiving a request of this nature.

            - ``"user"``

                The request was made by a user,
                the bot settings allow a user to request their own data
                be deleted, and the bot is configured to let cogs keep
                data needed for operation.
                Under this case, you may elect to retain data which is
                essential to the functionality of the cog. This case will
                only happen if the bot owner has opted into keeping
                minimal EUD needed for cog functionality.
```

If you have some ideas how to make it look better, this can be done differently, it's just what I was able to come up with.

redbot/cogs/permissions/converters.py
```diff
@@ -138,12 +138,19 @@ class CogOrCommand(NamedTuple):
...
-            return cls(type="COMMAND", name=cmd.qualified_name, obj=cmd)
+        ret = None
+        if cog := ctx.bot.get_cog(arg):
+            ret = cls(type="COG", name=cog.qualified_name, obj=cog)
+
+        elif cmd := ctx.bot.get_command(arg):
+            ret = cls(type="COMMAND", name=cmd.qualified_name, obj=cmd)
+
+        if ret:
+            if isinstance(ret.obj, commands.commands._RuleDropper):
```
Jackenmen, line 149: Now that this is checked here, the `isinstance()` checks for `_AlwaysAvailableCommand` in `[p]permissions addglobalrule`, `[p]permissions addguildrule`, `[p]permissions setdefaultguildrule`, and `[p]permissions setdefaultglobalrule` can be removed.

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,86 @@
...
+=====================
+
+Notes for everyone
+******************
+
+What data Red collects
+----------------------
+
+Red and the cogs included with it collect some amount of data
+about users for it's normal operations.
```
Jackenmen:
```suggestion
about users for their normal operation.
```

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,86 @@
...
+What data Red collects
+----------------------
+
+Red and the cogs included with it collect some amount of data
+about users for it's normal operations. 
+
+In particular the bot will keep track of a short history of usernames/nicknames
+which actions refer to your discord account (such as creating a playlist)
+as well as the content of specific messages used directly as commands for the bot
+(such as reports sent to servers).
```
Jackenmen: This could use some commas

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,86 @@
...
+to view statements about how extensions use your data made by the authors of 
+the specific 3rd party extensions an instance of Red has installed.
+
+How can I delete data Red has about me?
+---------------------------------------
+
+The command ``[p]mydata forgetme`` provides a way for users to remove
+large portions of their own data from the bot. This command will not
+remove operational data, such as a record that your
+discord account was the target of a moderation action.
```
Jackenmen:
```suggestion
Discord account was the target of a moderation action.
```

docs/red_core_data_statement.rst
```diff
@@ -0,0 +1,86 @@
...
+
+3rd party extensions to Red are able to delete data when this command
+is used as well, but this is something each extension must implement.
```
Jackenmen, line 38: > but this is something each extension must implement

Maybe we could say that user will be informed about it after deletion or something along that meaning?

redbot/core/settings_caches.py
```diff
@@ -125,136 +127,184 @@ def __init__(self, config: Config):
         self._config: Config = config
         self._cached_whitelist: Dict[Optional[int], Set[int]] = {}
         self._cached_blacklist: Dict[Optional[int], Set[int]] = {}
+        # because of discord deletion
+        # we now have sync and async access that may need to happen at the
+        # same time.
+        # blame discord for this.
+        self._access_lock = asyncio.Lock()
+
+    async def discord_deleted_user(self, user_id: int):
```
Jackenmen, line 136: Considering that we don't load **all** whitelist/blacklist data on bot load, we can't rely on completeness of this cache.

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+        """ Commands which interact with the data [botname] has about you """
+
+    # 1/10 minutes. It's a static response, but the inability to lock
+    # will annoy people if it's spammable
+    @commands.cooldown(1, 600, commands.BucketType.user)
+    @mydata.command(cls=commands.commands._AlwaysAvailableCommand, name="whatdata")
+    async def mydata_whatdata(self, ctx: commands.Context):
+        """ Find out what type of data [botname] stores and why """
+
+        ver = "latest" if VersionInfo.from_str(__version__).dev_release else "stable"
```
Jackenmen: meh
```suggestion
        ver = "latest" if red_version_info.dev_release else "stable"
```

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+        link = f"https://docs.discord.red/en/{ver}/red_core_data_statement.html"
+        await ctx.send(
+            _(
+                "This bot stores some data about users as necessary to function. "
+                "This is mostly the ID your user is assigned by discord, linked to "
```
```suggestion
                "This is mostly the ID your user is assigned by Discord, linked to "
```

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+                "may not store additional things. "
+                "You can use `{prefix}mydata 3rdparty` "
+                "to view the statements provided by each 3rd-party addition."
+            ).format(link=link, prefix=ctx.clean_prefix)
+        )
+
+    # 1/30 minutes. It's not likely to change much and uploads a standalone webpage.
+    @commands.cooldown(1, 1800, commands.BucketType.user)
+    @mydata.command(cls=commands.commands._AlwaysAvailableCommand, name="3rdparty")
+    async def mydata_3rd_party(self, ctx: commands.Context):
```
Jackenmen, line 517: Is this command supposed to have absolutely no usage of i18n?

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+        formatted_statements = []
+
+        no_statements = []
+
+        for ext_name, statement in sorted(statements.items()):
+            if not statement:
+                no_statements.append(ext_name)
+            else:
+                formatted_statements.append(
+                    f"### {entity_transformer(ext_name)}" f"\n\n{entity_transformer(statement)}"
```
Jackenmen: From usage of `entity_transformer()`, I'm guessing that end user data statements aren't supposed to support markdown? I don't really see anything wrong about it (maybe links or emphasis would be *somewhat* useful but it doesn't seem all that important if at all), just wanted to confirm.
Also, these strings can be made into one:
```suggestion
                    f"### {entity_transformer(ext_name)}\n\n{entity_transformer(statement)}"
```

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+                check=lambda m: m.channel.id == ctx.channel.id and m.author.id == ctx.author.id,
+                timeout=30,
+            )
+        except asyncio.TimeoutError:
+            await ctx.send(_("Did not get confirmation, cancelling."))
+        else:
+            if message.content.strip() == confirm_token:
+                return True
+            else:
+                await ctx.send(_("Did not get a matching confimration, cancelling."))
```
Jackenmen: Code suggestion for what Flame said [here](https://github.com/Cog-Creators/Red-DiscordBot/pull/4045/files#r460463811):
```suggestion
                await ctx.send(_("Did not get a matching confirmation, cancelling."))
        return False
```

redbot/cogs/audio/core/events/red.py
```diff
@@ -19,3 +19,12 @@ async def on_red_api_tokens_update(
             self.api_interface.spotify_api.update_token(api_tokens)
         elif service_name == "audiodb":
             self.api_interface.global_cache_api.update_token(api_tokens)
+
+    async def red_delete_data_for_user(
```
Jackenmen, line 24: From my look at Audio, it seems that there's one more place that store user IDs:
- Custom config scope called "EQUALIZER" which should look somewhat like this example:
```py
{
  "EQUALIZER": {
    GUILD_ID: {
      "eq_presets": {
        PRESET_NAME: {
          "bands": ...,
          "author": ID_HERE
        },
        ...
      }
    },
    ...
  }
}
```

redbot/cogs/alias/alias_entry.py
```diff
@@ -90,6 +90,32 @@ def __init__(self, config: Config, cache_enabled: bool = True):
...
+                    break
+            else:
+                continue
+            # basically, don't build a context manager wihout a need.
+            async with self.config.guild_from_id(guild_id).entries as entry_list:
+                for a in entry_list:
+                    if a.get("creator", 0) == user_id:
+                        # fmt: off
+                        a["creator"] = 0xde1
+                        # fmt: on
```
Jackenmen: Seems a bit weird to do this in this place but no other.
```suggestion
                        a["creator"] = 0xde1
```

redbot/core/commands/commands.py
```diff
@@ -917,24 +1032,72 @@ async def disabler(ctx: "Context") -> bool:
         return disabler
 
 
-# This is intentionally left out of `__all__` as it is not intended for general use
-class _AlwaysAvailableCommand(Command):
+# The below are intentionally left out of `__all__`
+# as they are not intended for general use
+class _AlwaysAvailableMixin:
```
Jackenmen, line 1069: The usage of `_AlwaysAvailableCommand` in `isinstance` check in `[p]command disable` (2 occurrences) needs to be replaced with this mixin (or possibly `_RuleDropper` as it seems to be used for everything else).

Jackenmen (changes requested):
Code (and docs) review done, everything should be in review comments.

I'll be doing functional testing tomorrow.

redbot/core/core_commands.py
mikeshardmind: I didn't want to guarantee markdown in future at the current time.

redbot/core/commands/commands.py
mikeshardmind: I have mixed feelings on this one, and will take the time to write it out in more detail later.

redbot/cogs/permissions/permissions.py
mikeshardmind: context manager use was intentional, though I'll double-check that there's no race potential later.

redbot/cogs/audio/core/events/red.py
Drapersniper: :aaaaaaaaaaa: forgot equilizers existed, my bad sinbad...

redbot/cogs/audio/core/events/red.py
```diff
@@ -19,3 +20,34 @@ async def on_red_api_tokens_update(
...
```
Jackenmen, line 24: Some of data deletion (`self.playlist_api`) in Audio cog depends on the cog being ready, so this most likely needs to wait for that event first.

Jackenmen (changes requested):
Appendix 1 to my code review.

redbot/core/commands/commands.py
mikeshardmind: Lack of being in `__all__` already prevents the name from being exported. I don't want to document this in details yet, but I do intend to make this public when implementing the data *fetching* API later on in case a cog wants to send this in particular cases it doesn't handle yet on that, when it does handle other cases (since fetching and making a display has more to it than just deleting what's appropriate to delete)

redbot/core/core_commands.py
Jackenmen: The comment link I added was actually pointing to the `return False` which hasn't been added here yet.

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+            for ext in no_statements:
+                parts.append(f"\n - {entity_transformer(ext)}")
+
+        generated = markdown.markdown("\n".join(parts), output_format="html")
+
+        html = "\n".join((PRETTY_HTML_HEAD, generated, HTML_CLOSING))
+
+        fp = io.BytesIO(html.encode())
+
+        await ctx.send(file=discord.File(fp, filename="3rd-party.html"))
```
Jackenmen: Possibly some short text could be added? Bot just attaching the file made seem a bit weird to the user.

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+            await ctx.send(
+                _(
+                    "I've deleted any non-operational data about you "
+                    "(that I know how to delete) {mention}"
+                ).format(mention=ctx.author.mention)
+            )
+
+        if results.unhandled:
+            await ctx.send(
+                _("{mention} The following cogs did not handle deletion:\n {cogs}").format(
```
```suggestion
                _("{mention} The following cogs did not handle deletion:\n{cogs}").format(
```

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+                    "these cogs and modules.\n"
+                    "Note: Outside of these failures, data should have been deleted."
+                ).format(modules=humanize_list(results.failed_modules))
+            )
+        else:
+            await ctx.send(_("I've deleted all data about that user that I know how to delete."))
+
+        if results.unhandled:
+            await ctx.send(
+                _("{mention} The following cogs did not handle deletion:\n {cogs}").format(
```
```suggestion
                _("{mention} The following cogs did not handle deletion:\n{cogs}").format(
```

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+            await ctx.send(
+                _(
+                    "I've deleted all non-operational data about that user "
+                    "that I know how to delete."
+                )
+            )
+
+        if results.unhandled:
+            await ctx.send(
+                _("{mention} The following cogs did not handle deletion:\n {cogs}").format(
```
```suggestion
                _("{mention} The following cogs did not handle deletion:\n{cogs}").format(
```

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+                ).format(modules=humanize_list(results.failed_modules))
+            )
+        else:
+            await ctx.send(
+                _("I've deleted all data about that user " "that I know how to delete.")
+            )
+
+        if results.unhandled:
+            await ctx.send(
+                _("{mention} The following cogs did not handle deletion:\n {cogs}").format(
```
```suggestion
                _("{mention} The following cogs did not handle deletion:\n{cogs}").format(
```

redbot/core/core_commands.py
```diff
@@ -443,6 +482,495 @@ async def uptime(self, ctx: commands.Context):
...
+    @mydata_owner_management.command(name="deleteforuser")
+    async def mydata_user_deletion_request_by_owner(self, ctx, user_id: int):
+        """ Delete data [botname] has about a user for a user. """
+        if not await self.get_serious_confirmation(
+            ctx,
+            _(
+                "This will cause the bot to get rid of or disassociate "
+                "a lot of non-operational data from the "
+                "specified user. Users have access to "
+                "different command for this unless they can't interact with the bot at all."
```
```suggestion
                "different command for this unless they can't interact with the bot at all. "
```

redbot/core/core_commands.py
mikeshardmind: Handled, this was functionally not an issue, but the type-hint is correct now.

redbot/cogs/alias/alias_entry.py
```diff
@@ -90,6 +90,30 @@ def __init__(self, config: Config, cache_enabled: bool = True):
...
+
+        all_guilds = await self.config.all_guilds()
+        async for guild_id, guild_data in AsyncIter(all_guilds.items(), steps=100):
+            for a in guild_data["entries"]:
+                if a.get("creator", 0) == user_id:
+                    break
+            else:
+                continue
+            # basically, don't build a context manager wihout a need.
+            async with self.config.guild_from_id(guild_id).entries as entry_list:
```
Jackenmen:
```suggestion
            async with self.config.guild_from_id(guild_id).entries() as entry_list:
```

redbot/cogs/warnings/warnings.py
```diff
@@ -45,6 +46,21 @@ def __init__(self, bot: Red):
         self.bot = bot
         self.registration_task = self.bot.loop.create_task(self.register_warningtype())
 
+    async def red_delete_data_for_user(
```
Jackenmen, line 50: After testing, I found out that warning data also has "mod" key.

redbot/cogs/alias/alias.py
```diff
@@ -38,24 +42,107 @@ class Alias(commands.Cog):
...
+
+            await asyncio.sleep(0)
+            # control yielded per loop since this is most likely to happen
+            # at bot startup, where this is most likely to have a performance
+            # hit.
+
+        await self.config.handled_string_creator.set(True)
+
+    def sync_init(self):
+        t = asyncio.create_task(self._initialize)
```
Jackenmen:
```suggestion
        t = asyncio.create_task(self._initialize())
```

redbot/cogs/alias/alias.py
```diff
@@ -38,24 +42,107 @@ class Alias(commands.Cog):
...
+
+        def done_callback(fut: asyncio.Future):
+            try:
+                t.result()
+            except Exception as exc:
+                log.exception("Failed to load alias cog", exc_info=exc)
+                self.bot.remove_cog(self)
```
Jackenmen: a) `remove_cog()` requires cog name, not cog object
b) what about unloading of the extension itself? I don't think you can do that fully synchronously, but not doing that is odd too; maybe this shouldn't be done at all here, I'm not sure

Jackenmen (changes requested):

redbot/cogs/warnings/warnings.py
```diff
@@ -57,10 +58,31 @@ async def red_delete_data_for_user(
...
-        async for guild_id, guild_data in AsyncIter(all_members.items(), steps=100):
-            if user_id in guild_data:
+        c = 0
+
+        for guild_id, guild_data in all_members.items():
+            c += 1
+            if not c % 100:
+                await asyncio.sleep(0)
+
+            if str(user_id) in guild_data:
```
Jackenmen: casting to int is done by `config.all_members()`
```suggestion
            if user_id in guild_data:
```

redbot/cogs/warnings/warnings.py
```diff
@@ -57,10 +58,31 @@ async def red_delete_data_for_user(
...
+                    await asyncio.sleep(0)
+
+                for warn_id, warning in user_warns.get("warnings", {}):
+                    c += 1
+                    if not c % 100:
+                        await asyncio.sleep(0)
+
+                        if warning.get("mod", 0) == user_id:
+                            await self.config.member_from_ids(guild_id, remaining_user).set_raw(
+                                "warnings", warn_id, "mod", value=0xDE1
```
Jackenmen: Could do special handling (i.e. not showing ID for deleted user) for this case in `[p]warnings` and `[p]mywarnings`, but this is not critical.

Jackenmen (commented):
comments for the warnings commit

redbot/cogs/warnings/warnings.py
mikeshardmind: *sigh* had that one right before, and missed that on revisiting.

redbot/cogs/warnings/warnings.py
mikeshardmind: compared to the other changes, this one's a small deal, but I'll look at it in a sec

redbot/core/commands/commands.py
mikeshardmind: It's in all and sparsely documented now

docs/red_core_data_statement.rst
mikeshardmind: commas were removed by request from an earlier review...

redbot/core/commands/commands.py
```diff
@@ -59,6 +59,7 @@
     "command",
     "group",
     "RESERVED_COMMAND_NAMES",
+    "RedUnhandledAPI",
```
Jackenmen, line 62: This should be re-exported in `redbot/core/commands/__init__.py`

redbot/cogs/warnings/warnings.py
```diff
@@ -45,6 +47,42 @@ def __init__(self, bot: Red):
...
+
+            if user_id in guild_data:
+                await self.config.member_from_ids(guild_id, user_id).clear()
+
+            for remaining_user, user_warns in guild_data.items():
+                c += 1
+                if not c % 100:
+                    await asyncio.sleep(0)
+
+                for warn_id, warning in user_warns.get("warnings", {}):
```
```suggestion
                for warn_id, warning in user_warns.get("warnings", {}).items():
```

redbot/cogs/reports/reports.py
```diff
@@ -60,6 +60,39 @@ def __init__(self, bot: Red):
...
+        paths = []
+
+        # this doesn't use async iter intentionally due to the nested iterations
+        for guild_id_str, tickets in all_reports.items():
+            for ticket_number, ticket in tickets.items():
+                steps += 1
+                if not steps % 100:
+                    await asyncio.sleep(0)  # yield context
+
+            if ticket.get("user_id", 0) == user_id:
```
Jackenmen: actual data of ticket is in `ticket["report"]`

redbot/cogs/permissions/permissions.py
```diff
@@ -114,6 +115,52 @@ def __init__(self, bot: Red):
...
+                obj = getter(obj_name)
+
+                for guild_id, guild_rules in rules_dict.items():
+
+                    count += 1
+                    if not count % 100:
+                        await asyncio.sleep(0)
+
+                    if _uid in guild_rules:
+                        if obj:
```
Jackenmen, line 157: This is wrong - if getter doesn't return any object, it should still remove the rule.

redbot/cogs/warnings/warnings.py
```diff
@@ -45,6 +47,42 @@ def __init__(self, bot: Red):
...
+
+                for warn_id, warning in user_warns.get("warnings", {}).items():
+                    c += 1
+                    if not c % 100:
+                        await asyncio.sleep(0)
+
+                        if warning.get("mod", 0) == user_id:
+                            await self.config.member_from_ids(guild_id, remaining_user).set_raw(
+                                "warnings", warn_id, "mod", value=0xDE1
+                            )
```
Jackenmen: This is overintended.

redbot/core/bot.py
```diff
@@ -1395,6 +1407,121 @@ async def shutdown(self, *, restart: bool = False):
...
+        for details about the parameters and intent.
+
+        Parameters
+        ----------
+        requester
+        user_id
+
+        Returns
+        -------
+        DataDeletionResults
```
Jackenmen, line 1454:
```suggestion
        DataDeletionResults
            A named tuple ``(failed_modules, failed_cogs, unhandled)``
            containing lists with names of failed modules, failed cogs,
            and cogs that didn't handle data deletion request.
```

redbot/core/commands/commands.py
```diff
@@ -809,6 +821,123 @@ def help(self):
...
+            .. note::
+                you should safely handle
+                any string value (log a warning if needed)
+                as additional requester types may be added
+                in the future without prior warning.
+                (see what this method can raise for details)
+
+
+            - ``"discord_deleted_user"``:
+                The request should be processed as if
```
Jackenmen: This list doesn't show at all, most likely because there are two new lines before it rather than just one (which makes Sphinx treat it as not part of explanation for `requester` parameter)

redbot/core/core_commands.py
```diff
@@ -443,6 +482,502 @@ async def uptime(self, ctx: commands.Context):
...
+
+        # Can't check this as a command check, and want to prompt DMs as an option.
+        if not ctx.channel.permissions_for(ctx.me).attach_files:
+            ctx.command.reset_cooldown(ctx)
+            return await ctx.send(_("I need to be able to attach files (try in DMs?)"))
+
+        statements = {
+            ext_name: getattr(ext, "__red_end_user_data_statement__", None)
+            for ext_name, ext in ctx.bot.extensions.items()
+            if not ext.__package__.startswith("redbot.")
```
Jackenmen: Probably unnecessary, but this is a safe-guard for the case where `ext.__package__` would be None (as it's technically documented that it can be `None`)
```suggestion
            if not (ext.__package__ and ext.__package__.startswith("redbot."))
```

Jackenmen (commented):
Just few smaller enhancements left, everything works functionally just fine.

redbot/core/commands/commands.py
```diff
@@ -894,13 +895,16 @@ async def red_delete_data_for_user(
```
Jackenmen, line 892:
```suggestion
            This method can currently passed one of these strings:

```

Jackenmen (approved):
Looks good to me, thanks for all the help with making this happen!

This PR is temporarily blocked, look in Discord for more info.

PR #4046, opened 2020-07-07
Allow creating an instance non-interactively

### LINKED ISSUE #2396: Allow redbot-setup to be run in headless mode
See https://cogboard.red/t/allow-headless-setup-command/250

Necessary for Docker deployment of v3, otherwise an [Expect script](https://forums.docker.com/t/dockerfile-how-to-answer-install-question-from-application/5240) needs to be written to answer the GUI questions for redbot-setup.

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

resolves 2396 and makes future work with containers more bearable.

example use
```
redbot somename --make-instance
```

further modification supported by use of edit, this just creates with defaults (allowing it to be used in conjunction with non-interactive editing of existing instances via edit for headless setup)

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -15,8 +15,8 @@ import sys
 from argparse import Namespace
 from copy import deepcopy
 from pathlib import Path
-from typing import NoReturn
 
+import appdirs
 import discord
 
 # Set the event loop policies here so any subsequent `new_event_loop()`
@@ -29,7 +29,7 @@ _update_event_loop_policy()
 import redbot.logging
 from redbot.core.bot import Red, ExitCodes
 from redbot.core.cli import interactive_config, confirm, parse_cli_flags
-from redbot.setup import get_data_dir, get_name, save_config
+from redbot.setup import get_data_dir, get_name, save_config, load_existing_config
 from redbot.core import data_manager, drivers
 from redbot.core._sharedlibdeprecation import SharedLibImportWarner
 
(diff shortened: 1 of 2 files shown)


LGACode:
YES. I have to use Expect right now to answer the interactive setup for Docker because Docker does not play nice with interactive command prompts. I've had to update the expect script every time for the last 3 Red updates because the wording in the sentence about what characters are allowed in the instance name was slightly changed every time, and Expect is pretty harsh about how it detects the questions and answers...

A headless way of running redbot-setup would solve all these issues for Docker and we would no longer have to rely on things like Expect. Maybe even paving a way towards full Docker support?

Drapersniper:
Just because this is a PR, doesn't mean Docker is supported.

Eventually when Docker support is a priority we will work on it, until then 🤷 as we have higher priority stuff going on at the moment.

mikeshardmind:
Just a heads up, while Docker may not be supported by this PR, this PR should give people tools to go do their own thing more easily until we do support it. I have no intention of this being part of a larger PR set, just acknowledged in the description that it happens to help with other goals too.

Kowlin:
Well this beats my solution of just manually placing an config file with the bot 😄

Jackenmen:
> further modification supported by use of edit

While this isn't in scope of this PR, something worth noting is that PostgreSQL driver currently does not support headless configuration of storage details.

mikeshardmind:
No, but Postgres does support env vars for providing these details (should likely support setting this non-interactively later on, but it's a problem that doesn't require something like an expect script to work around at least)

PR #4047, opened 2020-07-07
An initiative to make codebase racially neutral

owocado:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature

Through this PR, I'm putting forth and proposing the idea/belief to make the codebase of Red-DiscordBot project (as much) racially neutral (as possible), also less commonly known as the idea of inclusive terminology. 

**Terms such as “blacklist” and “whitelist” generally reinforce the notion that black==bad and white==good.** 

I personally believe that any code should be racially neutral. These terms such as "whitelist" and "blacklist" can usually be replaced by “denylist” and “allowlist” without changing their meanings, but particular instances may need other replacements.

Many popular open-source projects have proposed to implement similar initiatives, including [MySQL](https://mysqlhighavailability.com/mysql-terminology-updates/), [Twitter](https://twitter.com/TwitterEng/status/1278733305190342656), [Chromium Project](https://chromium.googlesource.com/chromium/src/+/master/styleguide/inclusive_code.md) and [GitHub](https://twitter.com/natfriedman/status/1271253144442253312) (twitter post by GitHub CEO). Similar idea is even proposed to [Linux Kernel](https://lkml.org/lkml/2020/7/4/229). I believe it is a good time if we can also implement the similar initiative for this project.

While there are advantages to this idea, there are cons as well, one I could think of is that replacing commonly used terms with less well-known synonyms could itself be a challenge for non-English users. But **I strongly believe the pros greatly outweigh the cons.**
I agree finding consensus on such matters is not easy, but if adopted, the Red-DiscordBot will be one of many projects where inclusive code is encouraged.

To quote an excerpt from The Chromium Project's [Code of Conduct](https://www.chromium.org/conduct):

> Each of us has the right to enjoy our experience and participate without fear of discrimination, or condescension, whether blatant or subtle.

To add emphasis to above: exclusive code is discriminatory and condescending, and reading biased code isn't enjoyable. **Words matter in our meetings, our conversations, and the documents we write.**

While I am already expecting some amount of possible backlash and ridicule reactions towards this PR, I don't mind if this PR gets turned down by the majority, but it opens up the discussion/conversation to let everyone know what is the stance or viewpoint of the members and contributors of this project on this topic. I politely urge everyone to comment your opinion in this thread. All constructive critisisms are wholeheartedly welcome. 
While I can agree this PR is poorly made, but I humbly request you to focus more on the idea/belief I'm trying to propose through this PR. This is but a template PR. Hope everyone will take it positively.

### DIFF
--- a/redbot/cogs/audio/core/__init__.py
+++ b/redbot/cogs/audio/core/__init__.py
@@ -74,8 +74,8 @@ class Audio(
             use_external_lavalink=False,
             restrict=True,
             localpath=str(cog_data_path(raw_name="Audio")),
-            url_keyword_blacklist=[],
-            url_keyword_whitelist=[],
+            url_keyword_denylist=[],
+            url_keyword_allowlist=[],
             **self._default_lavalink_settings,
         )
 
@@ -103,8 +103,8 @@ class Audio(
             vote_enabled=False,
             vote_percent=0,
             room_lock=None,
-            url_keyword_blacklist=[],
-            url_keyword_whitelist=[],
+            url_keyword_denylist=[],
+            url_keyword_allowlist=[],
             country_code="US",
         )
         _playlist: Mapping = dict(id=None, author=None, name=None, playlist_url=None, tracks=[])
--- a/redbot/cogs/audio/core/commands/audioset.py
+++ b/redbot/cogs/audio/core/commands/audioset.py
@@ -31,57 +31,57 @@ class AudioSetCommands(MixinMeta, metaclass=CompositeMetaClass):
     @command_audioset.group(name="restrictions")
     @commands.mod_or_permissions(manage_guild=True)
     async def command_audioset_perms(self, ctx: commands.Context):
-        """Manages the keyword whitelist and blacklist."""
+        """Manages the keyword allowlist and denylist."""
 
     @commands.is_owner()
     @command_audioset_perms.group(name="global")
     async def command_audioset_perms_global(self, ctx: commands.Context):
-        """Manages the global keyword whitelist/blacklist."""
+        """Manages the global keyword allowlist/denylist."""
 
-    @command_audioset_perms_global.group(name="whitelist")
-    async def command_audioset_perms_global_whitelist(self, ctx: commands.Context):
-        """Manages the global keyword whitelist."""
+    @command_audioset_perms_global.group(name="allowlist")
+    async def command_audioset_perms_global_allowlist(self, ctx: commands.Context):
+        """Manages the global keyword allowlist."""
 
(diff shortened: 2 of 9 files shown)


Vexed01:
I am mostly in support of this, I have seen the "allowlist" and "denylist" idea floating around recently.
As you have said this is going to be a big change.

I think that for commands white/backlist should be a build-in alias so it still works (possibly for removal around 3.5), and this would mean allow/denylist is shown in help to gently nudge users but white/blacklist would still work so it's not too sudden.

Also owo, ideally something like thing this would be in an issue first to better facilitate discussion, then if/when approved moved to a PR :stuck_out_tongue:

Jackenmen:
Ignoring the reasoning behind this idea for now, this PR just blindly replaces all usage of "whitelist" and "blacklist" without any additional handling. If this were to be done, this either needs a config migration or it needs to keep the old names in the backend.

Drapersniper:
+1 in favour of this PR.

However theres a couple of things to note here.

1. This should not be done as one bulk PR. Each cog/core should be done as its individual PR for easier review regression catches
2. Data magration needs to be done.
3. Without looking in depth some of config defaults may need to remain inplace for migration purposes.

I suggest we Close this PR and track the progress is a informational Issue where we can monitor PRs for specific areas.

Twentysix26:
> to make the codebase of Red-DiscordBot project (as much) racially neutral (as possible), also less commonly known as the idea of inclusive terminology.  

> Terms such as “blacklist” and “whitelist” generally reinforce the notion that black==bad and white==good.

-1

Those words [have nothing to do with race](https://en.wikipedia.org/wiki/Blacklisting#Origins_of_the_term), nor are they used in a racist context. Changing them means replacing terms that are standard in this industry, which results in no benefit other than confusing users for no reason.

owocado:
> +1 in favour of this PR.
> 
> However theres a couple of things to note here.
> 
> 1. This should not be done as one bulk PR. Each cog/core should be done as its individual PR for easier review regression catches
> 2. Data magration needs to be done.
> 3. Without looking in depth some of config defaults may need to remain inplace for migration purposes.
> 
> I suggest we Close this PR and track the progress is a informational Issue where we can monitor PRs for specific areas.

That sounds a better proposal than mine. 
I apologize for any inconvenience I may have caused to anyone. Please feel free to close this PR or take any action you would think would be right in context.

Flame442:
-1 not only for the reasons 26 has pointed out, but also because it would also cause issues for users who are used to the current commands and language. I would be perfectly fine adding aliases to commands, but changing every instance of the language is just not something I agree with personally.

Drapersniper:
Closing this in favour of #4048.

PR #4049, opened 2020-07-07
Add per-command embed settings (`embed_requested`)

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This uses the fact that `embed_requested()` already accepts `command` argument and adds per-command embed settings that can be modified with `[p]embedset command`. There's a global per-command setting as well as guild-specific per-command setting.

Currently, guild-specific command setting is checked after channel setting and before guild setting while global command setting is checked right before checking the global setting.
Full hierarchy looks as follows:

a) Guild context
  1. Channel override
  2. Guild command override
  3. Guild override
  4. Global command override
  5. Global setting

b) DM context:
  1. User override
  2. Global command override
  3. Global setting

Whether guild override and global command override in guild context should be in opposite order or not is imo debatable but I chose what made sense to me.

I can also add embed settings cache, but I would prefer to do that in separate PR and if needed (because of possible performance loss), this PR could be changed to depend on it.

---

This was brought up before, but somehow it didn't make its way into an issue so this is the place where this feature will be tracked.

// I have not tested this PR *yet*.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -53,6 +53,7 @@ from .utils import common_filters, AsyncIter
 from .utils._internal_utils import send_to_owners_with_prefix_replaced
 
 CUSTOM_GROUPS = "CUSTOM_GROUPS"
+COMMAND_SCOPE = "COMMAND"
 SHARED_API_TOKENS = "SHARED_API_TOKENS"
 
 log = logging.getLogger("red")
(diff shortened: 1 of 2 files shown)


Jackenmen:
This might also need some additional handling for command groups.

mikeshardmind:
I have mixed feelings about this in terms of the appearances of a consistent UX. The ordering shown here experiences it the least, but having it take into account parent commands feels like it's going to open a can of worms on potentially inconsistent user-facing behavior even in the same context, so that's a -1 from me to extend it to that unless the parent command being set means subcommands can't set a different override.

Jackenmen:
> unless the parent command being set means subcommands can't set a different override.

I don't think we should be limiting customisation possibilities to users here just because *we* think it could make UX of some subcommands less consistent if the users use it wrongly so I would like to avoid *such* limitation. But I can see how taking parent command into account could actually lower the customisation possibilities too as it removes some of the granularity. Taking parent commands into account could be convenient in some cases but it's probably not worth it considering the possible problems now that I think of it. But I wanted to at least bring it up though as I haven't thought of group commands at all when making this PR.

TrustyJAID (approved):
This looks good to me.

PR #4051, opened 2020-07-08
[Docs] Add --edit to set Bot owner under Discord Teams.

xBlynd:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

When Discord Bot is a Team application, by default Red has no owner set.  This does not restrict the use however has at times locked users out.  Added a note section in two locations in the documentation to potential good locations.

This is known information from the Discord server, however, After a few installs in the last couple months for friends and others.  I have found myself running into this quite often.  Seems like it shows value in being noted during install or at least in the getting started section.  I leave it to the group to decide.  No hurt feelings regardless.

### DIFF
--- a/docs/autostart_pm2.rst
+++ b/docs/autostart_pm2.rst
@@ -12,11 +12,11 @@ Installing PM2
 
 Start by installing Node.JS and NPM via your favorite package distributor.
 
-- Using Ubuntu run the following command.
+Using Ubuntu run the following command.
 :code:`curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash -
 sudo apt install -y nodejs`
 
-- Using Debian, run the following command as root.
+Using Debian, run the following command as root.
 :code:`curl -sL https://deb.nodesource.com/setup_14.x | bash -
 apt install -y nodejs`
 
--- a/docs/autostart_pm2.rst
+++ b/docs/autostart_pm2.rst
@@ -12,15 +12,15 @@ Installing PM2
 
 Start by installing Node.JS and NPM via your favorite package distributor.
 
-- Using Ubuntu run the following command:
+- Using Ubuntu run the following command.
 :code:`curl -sL https://deb.nodesource.com/setup_14.x | sudo -E bash -
 sudo apt install -y nodejs`
 
-- Using Debian, run the following command as root:
+- Using Debian, run the following command as root.
 :code:`curl -sL https://deb.nodesource.com/setup_14.x | bash -
 apt install -y nodejs`
 
-From there run the following command:
+From there run the following command.
 :code:`npm install pm2 -g`
 
 After PM2 is installed, run the following command to enable your Red instance to be managed by PM2. Replace the brackets with the required information.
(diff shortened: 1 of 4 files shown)


Drapersniper:
I'm not entirely okay mentioning the coowner and team owned flags without some huge red flags in the docs. People could run them and forget them or set them without fully understand the downside of doing so specially when running the bot in their local PC.

yamikaitou:
What about the Windows page?

xBlynd:
Draper -  Do you suggest going into more detail?  I tried to keep it as simple and not so verbose.  Can you elaborate on the concerns?

NeuroAssassin:
In my opinion, it would make sense to include the piece about team owners, but personally I feel the co-owners flag should not be included.  That feature is hidden for a reason, and I do not think it should be publicly documented, as that suggests it may be officially supported, or makes users aware of the fact that it exists, without them explicitly wanting that feature.

mikeshardmind:
I think it needs to be made clear that anyone considered an owner can essentially run arbitrary code on the host machine (barring other restrictions like SELinux) if this is to be included. (which is also part of why the default behavior is not to consider team members as owners)

xBlynd:
> I think it needs to be made clear that anyone considered an owner can essentially run arbitrary code on the host machine (barring other restrictions like SELinux) if this is to be included. (which is also part of why the default behavior is not to consider team members as owners)

This was my intent with co-owner.  I have team members that i dont want full access to certain applications.

Drapersniper:
Neuro pretty much said what I meant. What he said also applies to the team flag as it will make every single member a bot owner. Which access to all bot level commands and as full access to the machine the bot is running on. So making someone a co owner by accident is not something we should suggest preemptively.

I feel telling them about redbot edit instance is enough in this case.  If they have the need for multiple owners then joining support where they can ask and be educated about the risks is a good trade off imo.

xBlynd:
Okay then we can separate the conversation into ~two~ three(I can count)points.

1. Owner
In my opinion needs to be included(I dont have any weight i know lol).  By default my bot is on Team and by default Red doesnt set a owner on team applications.
2. Co-Owner
I understand the concerns and could be noted somewhere else for someone that wants to expand the development of bot.  not included in install
3. Team
I understand the concerns and could be noted somewhere else for someone that wants to expand the development of bot. Not included in install


image

xBlynd:
Okay so as a comment for reviewers I don't want to push a change yet but essentially..
`.. note::

    If your Discord application(bot) is tied to a "Team" in the Discord 
    Developer Console, by default Red will have no set Owner.  To set 
    bot owner follow the steps below:
    
    To set owner manually, launch Red using 
          ``redbot --edit instance_name``

    For more information on setting up Teams with Red please visit
    (idk-somelink?)
`

Jackenmen:
This PR has been brought up for discussion more than once and there was no clear consensus which generally means status quo wins. I think it is fairer to you to close this rather than keep it in limbo possibly forever. If we reach a consensus about this at some point, we'll update the docs. Nonetheless, thanks for bringing this up, and sorry that this PR has not been touched for so long.

PR #4062, opened 2020-07-09
Make use of d.py menus across core

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

- [x] redbot/cogs/customcom/customcom.py
- [x] redbot/cogs/general/general.py
- [x] redbot/cogs/modlog/modlog.py
- [x] redbot/cogs/warnings/warnings.py
- [x] redbot/core/commands/help.py
- [x] redbot/cogs/economy/economy.py
- [x] redbot/cogs/alias/alias.py
- [x] helper functions to emulate existing behaviour of `menu()` - Partially implemented `dpymenu()` supports same fingerprint as `menu()` however it does not accept custom controls.


redbot/core/utils/menus.py
```diff
@@ -224,3 +233,442 @@ async def task():
...
+
+class SimpleSource(_dpy_menus.ListPageSource):
+    def __init__(self, pages: Iterable[Union[List[str], List[discord.Embed]]]):
+        super().__init__(pages, per_page=1)
+
+    async def format_page(
+        self, menu: SimpleHybridMenu, page: Union[List[str], List[discord.Embed]]
+    ) -> Union[discord.Embed, str]:
+        """Sends thee specified page to the menu."""
+        return page
```
mikeshardmind: These subclasses can't be added here. discord.py's menus aren't stable yet, placing them in this file would add a guarantee to them.

redbot/core/commands/help.py
```diff
@@ -707,13 +708,10 @@ async def _delete_delay_help(
...
         else:
-            # Specifically ensuring the menu's message is sent prior to returning
-            m = await (ctx.send(embed=pages[0]) if embed else ctx.send(pages[0]))
-            c = menus.DEFAULT_CONTROLS if len(pages) > 1 else {"\N{CROSS MARK}": menus.close_menu}
-            # Allow other things to happen during menu timeout/interaction.
-            asyncio.create_task(menus.menu(ctx, pages, c, message=m))
-            # menu needs reactions added manually since we fed it a messsage
-            menus.start_adding_reactions(m, c.keys())
+
+            await SimpleHybridMenu(
```
mikeshardmind, line 857: This significantly changes the help menu behavior. I don't think this is a good thing. The 3 minute timeout also probably necessitates a max concurrency decorator on the help command and then also ensuring that separate invocations from context are also limited, which adds a lot more to what 3rd party replacements would also have to be aware of. I'd prefer this particular menu use not be included.

redbot/core/utils/help.py
```diff
@@ -0,0 +1,23 @@
...
+from redbot.core.i18n import Translator
+from redbot.core.utils.chat_formatting import box
+
+from redbot.core.utils.menus import SimpleHybridMenu
+from redbot.vendored.discord.ext import menus
+
+_ = Translator("Alias", __file__)
+
+
+class HelpSource(menus.ListPageSource):
```
mikeshardmind: Placement of this class is problematic for version guarantees

mikeshardmind (changes requested):
I didn't review how this would interact with any of the cogs using it in this PR, but as far as help goes, and where some of the classes are placed, this is a problem.

redbot/core/utils/menus.py
```diff
@@ -5,17 +5,108 @@
...
+from ..i18n import Translator
 
 _ReactableEmoji = Union[str, discord.Emoji]
 
+_ = Translator("Menus", __file__)
+
+log = logging.getLogger("red.menus")
+
 
 async def menu(
```
mikeshardmind, line 24: This breaks existing users of this menu, it would be better to leave the existing menu as-is and just tell people to wait for stable or import it from vendor with a warning that it's not stable from there.

mikeshardmind (changes requested):

PR #4064, opened 2020-07-10
[Mod] `[p]mute|unmute voice` now take action instantly

### LINKED ISSUE #4028: [Mod] Voicemutes are broken
# Other bugs

Voice muting a user

The user to get muted

#### What actually happened?
Modlog entry created. Nothing else happened

Voice mute someone.

Not sure if this is worth fixing as someone with manage channel perms can just mute people. If we don't want to fix it we might want to remove the voicemute command though

### LINKED ISSUE #4048: [Tracking] An initiative to make codebase racially neutral
Making this issue to take place instead of #4047.

There is a few things to ensure while taking these steps.

- This should not be done as one bulk PR
  - Each cog that has this applied to it should have individual PRs
  - An Individual PR for core changes should exist
- We shouldn't make any breaking changes/changes that would confuse the existing user base
- Before (if at all) removing all mentioned of `blacklist` and `whitelist` we should give users some grace period, because if not this would be a breaking change, that's not compliant with our guarantees.
- When changing internals we should ensure everything they change is also changed.
- When changing config defaults/existing values, we should ensure a data migration is performed.

Currently the following uses either `whitelist` or `blacklist`:
- Audio Cog (redbot/cogs/audio/core/__init__.py)
  - Audioset commands (redbot/cogs/audio/core/commands/audioset.py)
  - Audio Utils (redbot/cogs/audio/core/utilities/validation.py)
  - Audio Lavalink Manager (redbot/cogs/audio/manager.py)
- Bot Core (redbot/core/bot.py)
- Bot Core commands (redbot/core/core_commands.py)
- Bot Global checks (redbot/core/global_checks.py)
- Bot Setting cache (redbot/core/settings_caches.py)
- Permissions converters (redbot/cogs/permissions/converters.py)

*Docs should auto update once the above are update.*

The permission cogs currently allows "blacklist" and "whitelist" as a user input for it's converters I don't think this should ever be changed, although we should add the new keywords to this converter to also accept them.

---

We should focus on changing front facing commands first:
- [ ] Audioset commands (redbot/cogs/audio/core/commands/audioset.py)
- [x] Bot Core commands (redbot/core/core_commands.py)
- [ ] Permissions converters (redbot/cogs/permissions/converters.py)

*Note: We should rename the user facing commands only at this stage, while aliasing the old names*

---

Once the above are worked on we should then look at potentially implementing this in the backend and config.
- [ ] Audio Cog (redbot/cogs/audio/core/__init__.py)
  - [ ] Audioset commands (redbot/cogs/audio/core/commands/audioset.py)
  - [ ] Audio Utils (redbot/cogs/audio/core/utilities/validation.py)
  - [ ] Audio Lavalink Manager (redbot/cogs/audio/manager.py)
- [ ] Bot Core (redbot/core/bot.py)
- [x] Bot Core commands (redbot/core/core_commands.py)
- [ ] Bot Global checks (redbot/core/global_checks.py)
- [ ] Bot Setting cache (redbot/core/settings_caches.py)

---
Note: Due to audio being in a major rewrite currently I would ask that if you choose to pick up this PR to work on it, don't touch audio. as I will be following the above in Audio anyways as part of the rewrite.

Also if I missed anything on the summary above please feel free to flag it.

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

When `[p]mute|unmute voice` is run, now the changes will take into effect instantly, instead of when the user rejoins. This is done via the move members permission, moving the user into the channel they are already in. Fixes #4028. However, as jack said in #4028:

>  I don't know if it's intended, but as long as nobody in Discord says it isn't, we could just use that.

### DIFF
--- a/redbot/cogs/mod/mutes.py
+++ b/redbot/cogs/mod/mutes.py
@@ -207,6 +207,14 @@ class MuteMixin(MixinMeta):
             await ctx.send(
                 _("Muted {user} in channel {channel.name}").format(user=user, channel=channel)
             )
+            try:
+                await user.move_to(channel)
+            except discord.Forbidden:
+                await ctx.send(
+                    _(
+                        "Because I don't have the move members permission, this will take into effect when the user rejoins."
+                    )
+                )
         else:
             await ctx.send(_("Mute failed. Reason: {}").format(issue))
 
@@ -322,6 +330,14 @@ class MuteMixin(MixinMeta):
             await ctx.send(
                 _("Unmuted {user} in channel {channel.name}").format(user=user, channel=channel)
             )
+            try:
+                await user.move_to(channel)
+            except discord.Forbidden:
+                await ctx.send(
+                    _(
+                        "Because I don't have the move members permission, this will take into effect when the user rejoins."
+                    )
+                )
         else:
             await ctx.send(_("Unmute failed. Reason: {}").format(message))
 
--- a/redbot/cogs/mod/mutes.py
+++ b/redbot/cogs/mod/mutes.py
@@ -216,7 +216,7 @@ class MuteMixin(MixinMeta):
                     )
                 )
         else:
-            await ctx.send(_("Mute failed. Reason: {}").format(issue))
+            await ctx.send(issue)
 
     @mute.command(name="channel")
     @commands.guild_only()
--- a/redbot/cogs/mod/mutes.py
+++ b/redbot/cogs/mod/mutes.py
@@ -208,7 +208,7 @@ class MuteMixin(MixinMeta):
                 _("Muted {user} in channel {channel.name}").format(user=user, channel=channel)
             )
         else:
-            await ctx.send(issue)
+            await ctx.send(_("Mute failed. Reason: {}").format(issue))
 
     @mute.command(name="channel")
     @commands.guild_only()
--- a/redbot/cogs/mod/mutes.py
+++ b/redbot/cogs/mod/mutes.py
@@ -208,11 +208,14 @@ class MuteMixin(MixinMeta):
                 _("Muted {user} in channel {channel.name}").format(user=user, channel=channel)
             )
             try:
-                await user.move_to(channel)
-            except discord.Forbidden:
+                if channel.permissions_for(ctx.me).move_members:
+                    await user.move_to(channel)
+                else:
+                    raise RuntimeError
+            except (discord.Forbidden, RuntimeError):
                 await ctx.send(
                     _(
-                        "Because I don't have the move members permission, this will take into effect when the user rejoins."
+                        "Because I don't have the Move Members permission, this will take into effect when the user rejoins."
                     )
                 )
         else:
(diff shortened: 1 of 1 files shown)


redbot/cogs/mod/mutes.py
```diff
@@ -207,6 +207,14 @@ async def voice_mute(self, ctx: commands.Context, user: discord.Member, *, reaso
             await ctx.send(
                 _("Muted {user} in channel {channel.name}").format(user=user, channel=channel)
             )
+            try:
+                await user.move_to(channel)
+            except discord.Forbidden:
+                await ctx.send(
+                    _(
+                        "Because I don't have the move members permission, this will take into effect when the user rejoins."
```
Jackenmen: Permission name should be title-cased same as it is presented in Discord's UI.
```suggestion
                        "Because I don't have the Move Members permission, this will take into effect when the user rejoins."
```

redbot/cogs/mod/mutes.py
```diff
@@ -322,6 +330,14 @@ async def unmute_voice(
             await ctx.send(
                 _("Unmuted {user} in channel {channel.name}").format(user=user, channel=channel)
             )
+            try:
+                await user.move_to(channel)
+            except discord.Forbidden:
+                await ctx.send(
+                    _(
+                        "Because I don't have the move members permission, this will take into effect when the user rejoins."
```
Jackenmen: Permission name should be title-cased same as it is presented in Discord's UI.
```suggestion
                        "Because I don't have the Move Members permission, this will take into effect when the user rejoins."
```

redbot/cogs/mod/mutes.py
```diff
@@ -207,6 +207,14 @@ async def voice_mute(self, ctx: commands.Context, user: discord.Member, *, reaso
             await ctx.send(
                 _("Muted {user} in channel {channel.name}").format(user=user, channel=channel)
             )
+            try:
+                await user.move_to(channel)
+            except discord.Forbidden:
```
Jackenmen: This should be in an `if` checking whether the bot has that permission so that we don't make API request if we already know that it's gonna fail.
```suggestion
                if channel.permissions_for(ctx.me).move_members:
                    await user.move_to(channel)
                else:
                    raise RuntimeError
            except (discord.Forbidden, RuntimeError):
```

redbot/cogs/mod/mutes.py
```diff
@@ -322,6 +330,14 @@ async def unmute_voice(
             await ctx.send(
                 _("Unmuted {user} in channel {channel.name}").format(user=user, channel=channel)
             )
+            try:
+                await user.move_to(channel)
+            except discord.Forbidden:
```
Jackenmen: This should be in an `if` checking whether the bot has that permission so that we don't make API request if we already know that it's gonna fail.
```suggestion
                if channel.permissions_for(ctx.me).move_members:
                    await user.move_to(channel)
                else:
                    raise RuntimeError
            except (discord.Forbidden, RuntimeError):
```

Jackenmen (changes requested):
Looks mostly good, just small string changes needed and preemptive checks.

There's one thing that *could* be done differently, although *I don't think* it's really a better way (just thought, it's worth mentioning) - the call to `_voice_perm_check()` in line 179 and 302 could have `move_members=True` argument passed and the command would just early-fail in such case. But since the moving isn't entirely necessary, it probably doesn't make that much sense to do in this way.

redbot/cogs/mod/mutes.py
Vexed01: 👍 As you mentioned `_voice_perm_check()` isn't especially better here as it sends a message in the text channel that could be misleading/otherwise require "unnecessary" perms.
_I couldn't get permissions_for to work when I was doing it..._

Jackenmen (approved):
Looks good to me. Thanks for the fix!

PR #4077, opened 2020-07-16
[Core] Add `Red.message_eligible_as_command`

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

  - This is a small utility function which makes it easy for cog
  creators to treat non-command messages as commands correctly.

  - This also modifies `Red.ignored_channel_or_guild`'s signature to
  explicitly add support for passing a message object (the only needed
  attributes are entirely shared with context)

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -3,7 +3,6 @@ import inspect
 import logging
 import os
 import platform
-import re
 import shutil
 import sys
 import contextlib
@@ -19,7 +18,6 @@ from typing import (
     Dict,
     NoReturn,
     Set,
-    Coroutine,
     TypeVar,
     Callable,
     Awaitable,
@@ -30,7 +28,6 @@ from types import MappingProxyType
 import discord
 from discord.ext import commands as dpy_commands
 from discord.ext.commands import when_mentioned_or
-from discord.ext.commands.bot import BotBase
 
 from . import Config, i18n, commands, errors, drivers, modlog, bank
 from .cog_manager import CogManager, CogManagerUI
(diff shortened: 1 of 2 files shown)


redbot/core/bot.py
```diff
@@ -406,14 +406,57 @@ async def allowed_by_whitelist_blacklist(
...
+
+        Parameters
+        ----------
+        message
+            The message object to check
+
+        Returns
+        -------
+        bool
+            Whether or not the message is eligible to be trated as a command.
```
```suggestion
            Whether or not the message is eligible to be treated as a command.
```

Jackenmen (changes requested):
This method could also have check for `author.bot` and I think that we should also replace bot's global checks with one global check that calls this method to avoid code repetition and make sure we won't possibly forget about this method in future *if* some other global check would be added.

redbot/core/bot.py
mikeshardmind: This can't be used to reduce code repetition, unfortunately. This does not interact with permissions, and cannot properly order all the checks to pretend it does.

mikeshardmind:
This can't be used to reduce code repetition, unfortunately. This does not interact with permissions, and cannot properly order all the checks to pretend it does.

redbot/core/bot.py
```diff
@@ -406,14 +406,57 @@ async def allowed_by_whitelist_blacklist(
 
         return True
 
-    async def ignored_channel_or_guild(self, ctx: commands.Context) -> bool:
+    async def message_is_eligible_as_command(self, message: discord.Message) -> bool:
```
Jackenmen: This method name would be better - it's common convention and existing methods in `Red` use same pattern:
```suggestion
    async def is_message_eligible_as_command(self, message: discord.Message) -> bool:
```

Jackenmen: Alternatively the name from PR title could be used since there's some precedence of that in `Red` too (although I find it a bit less nice):
```suggestion
    async def message_eligible_as_command(self, message: discord.Message) -> bool:
```

mikeshardmind:
`global_checks.py` was simplified, I was mistaken on a detail. I went with the name from the PR title.

Jackenmen (changes requested):
L473 and L474 need to use `self` instead of `ctx.bot`
Update: Other than the above, PR looks and works fine.

Jackenmen (approved):
Looks good to me.

PR #4081, opened 2020-07-18
Windows autostart docs

palmtree5:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Supersedes #2726


docs/autostart_win.rst
```diff
@@ -0,0 +1,105 @@
...
+
+.. code-block:: none
+
+    choco install nssm -y
+
+------------------------
+Manually installing NSSM
+------------------------
+
+Go to `<https://nssm.cc/download>`_ and download the latest release. Once the download
```
Jackenmen: It might be good to point at latest **pre** release (or at least pre-release build 2.2.4-101) as otherwise users may have problems with starting service on Windows 10 1703+. Technically we avoid this issue with `AppNoConsole=1` but choco already installs that build anyway and this build also comes with handy `nssm dump` command that may help make certain things easier for Support Squad.

Update: From only 2 support cases, I already found `nssm dump` to be a very valuable debug command that allows me to help people with nssm issues.

docs/autostart_win.rst
```diff
@@ -0,0 +1,105 @@
...
+Setting up the service
+----------------------
+
+To create the service, run the following command:
+
+.. note:: If Red was installed in a venv, please activate it before running this command
+
+.. code-block:: none
+
+    nssm install redbot $(Get-Command redbot).Definition <instance name> --no-prompt
```
Jackenmen: Considering that virtual environments are now required, including command that activates venv as part of the instructions makes sense:
```suggestion
.. code-block:: none

    . $env:USERPROFILE\redenv\Scripts\activate.ps1
    nssm install redbot $(Get-Command python).Definition -O -m redbot <instance name> --no-prompt
```
Also, I put the `-O` flag same as we do in systemd and pm2 guides.

docs/autostart_win.rst
```diff
@@ -0,0 +1,105 @@
...
+
+Then run the following set of commands one at a time:
+
+.. code-block:: none
+
+    nssm set redbot AppExit 0 Exit
+    nssm set redbot AppStopMethodConsole 10000
+    nssm set redbot AppRestartDelay 15000
+    nssm set redbot AppStdout $Env:USERPROFILE\red-stdout.log
+    nssm set redbot AppStderr $Env:USERPROFILE\red-stderr.log
```
Jackenmen: I think it would be better to use one log file for both stdout and stderr (nssm support that), it's easier to read like this.
```suggestion
    nssm set redbot AppStdout $Env:USERPROFILE\redbot.log
    nssm set redbot AppStderr $Env:USERPROFILE\redbot.log
```

docs/autostart_win.rst
```diff
@@ -0,0 +1,105 @@
...
+- Disabling the console window
+
+Now that the service has been configured, run the following command:
+
+.. code-block:: none
+
+    nssm set redbot ObjectName $Env:COMPUTERNAME\$Env:USERNAME <password>
+
+replacing <password> with the password you use to login to your computer. This
+sets the service to run under your user account so it can find your instances.
```
Jackenmen: I personally wouldn't want to put my password in cmd/powershell as part of a command (one of the reason being it saving in command history), I think there should at least be mentioned an alternative way to set that password through `services.msc`.

Also, both PowerShell and cmd have certain characters that have to be escaped, currently provided solution assumes that user's password won't contain such characters. I think that a better solution here would be using something that will prompt for password, e.g:
```
nssm set redbot ObjectName $Env:COMPUTERNAME\$Env:USERNAME $(Read-Host Password)
```

This isn't perfect - such prompt would still keep the password visible (as `input("Password: ")` does in Python), but it would at least always get the password right. The alternative that would show typed password with asterixes can be done in `Read-Host` by using `-AsSecureString` but it would result in this awfully long command, so the first solution might be good enough:
```
nssm set redbot ObjectName $Env:COMPUTERNAME\$Env:USERNAME $([System.Net.NetworkCredential]::new("", $(Read-Host -AsSecureString "Password")).Password)
```

---

One more issue that user may run into is having no password on the account - in such case, they won't be able to run service on an account with a blank password as it could be considered a security issue (see: <https://stackoverflow.com/a/1048691>)
While we probably shouldn't provide solution for it, it might be worth mentioning in the document.

Jackenmen (dismissed):
My review comments for this PR, mostly copied from #2726.

docs/autostart_win.rst
palmtree5: I'm thinking 
```pwsh
$c = Get-Credential
nssm set redbot ObjectName $c.Username $c.Password
```
might work here but I need to test it

Jackenmen: Same as the `Read-Host -AsSecureString`, `$c.Password` is also a secure string which sadly requires doing something like this:
```
$([System.Net.NetworkCredential]::new("", $c.Password)).Password
```
to get the actual raw string

I guess we will just have to call the solution I posted above as good enough then.

Lmmb74:
Hello. I'm actually having trouble using these instructions. The service never starts completely, it pauses during the process, so I'm getting the following error:
`redbot: Unexpected status SERVICE_PAUSED in response to START control.`

I'm executing PowerShell as Administrator, and tried using both the Windows default one and the PowerShell 7 Preview. I know I should wait until this pull is done, but maybe it would serve as a "What to do when..." answer for the guide. Thanks in advance.

P.S.: I had to use "" for my password since it uses special characters (#). And I tried using @jack1142 prompt too.

Jackenmen:
@Lmmb74 `nssm dump` should show more information that might be helpful to solve this issue. I'm not sure whether it's that good idea to ask for help here though, might be better to do it in [our discord server](https://discord.gg/red) (even though it's still not officially supported)

Lmmb74:
Here is the [dump](https://hastebin.com/gexikequno.sql). Also, I asked for help in Discord and I was given a solution involving a batch file.

Jackenmen:
Solved this in Discord server. The issue wasn't related to nssm instructions - Red simply wasn't installed in the venv.
While it might be worth to mention that Event Viewer can be helpful for debugging issues like "Unexpected status SERVICE_PAUSED in response to START control.", I wouldn't consider it essential for this PR.

Crossedfall:
Wouldn't it be preferable to use the built in task scheduler to auto run the bot? It reduces the amount of bloat the user needs to install onto their system, while also being generally more reliable since it's a core feature of the operating system.

palmtree5 (commented):
Went with the longer option for password handling

Kowlin:
Superseded by #5996.

PR #4082, opened 2020-07-18
Mac autostart docs

palmtree5:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Supersedes #2726 

This one definitely needs testing by someone with a Mac as I do not have a Mac


Ryan5453:
I tested this a few days ago and wanted to note that I couldn't get it to work. Maybe I was just doing it wrong but even at that point I think the docs need to be tweaked a little bit because they're a little confusing and I wouldn't expect the end user to be able to figure this out.

I also wanted to point out that the [PM2 guide for Linux](https://docs.discord.red/en/stable/autostart_pm2.html) does work completely on Mac, with no changes needed. That does require node & npm to be installed though, so I still think these docs are a good idea but still need some work.

aikaterna:
Hi there, I finally was successful in using a plist to start Red on a Catalina machine (10.15.7). I ran into plenty of fun while trying to diagnose this and the other previous PR for Mac autostarting as plist file errors are not very easy to debug.

* Please include instructions on how to save and exit vi. i.e. using `:x`
* I was unable to launch this plist having the `UserName` be the same as `GroupName` as the username did not exist as a group name. If the `GroupName` key is removed it will be set to the default group of the user which is much more ideal than us trying to guess the group the user belongs to.
* Maybe consider adding error logs, the path is just a suggestion. If we use any path that the `UserName` cannot access, the plist file will fail silently/show error 78. 

```
        <key>StandardOutPath</key>
        <string>/tmp/red_out.log</string>
        <key>StandardErrorPath</key>
        <string>/tmp/red_err.log</string>
```

* Be more explicit about what all to replace in the plist file when creating, like:

Replace:

`/path/to/redbot` with your redbot executable path
`instance-name` with your instance name
`username` with your Mac user name

palmtree5:
> Hi there, I finally was successful in using a plist to start Red on a Catalina machine (10.15.7). I ran into plenty of fun while trying to diagnose this and the other previous PR for Mac autostarting as plist file errors are not very easy to debug.
> 
> * Please include instructions on how to save and exit vi. i.e. using `:x`

Honestly don't know that I'm a fan of having the user edit the plist file with `vi` tbh. It tends to be a little complicated for anyone unfamiliar with it. What happens if we do `EDITOR=/Applications/TextEdit.app/Contents/MacOS/TextEdit sudo -e /Library/LaunchDaemons/red.plist`? I'm thinking if we can have the user edit with a GUI app that they may be a little more familiar with, it may simplify the editing part.

> * I was unable to launch this plist having the `UserName` be the same as `GroupName` as the username did not exist as a group name. If the `GroupName` key is removed it will be set to the default group of the user which is much more ideal than us trying to guess the group the user belongs to.

I'll get this one fixed

> * Maybe consider adding error logs, the path is just a suggestion. If we use any path that the `UserName` cannot access, the plist file will fail silently/show error 78.
> 
> ```
>         <key>StandardOutPath</key>
>         <string>/tmp/red_out.log</string>
>         <key>StandardErrorPath</key>
>         <string>/tmp/red_err.log</string>
> ```
> 

Good call, it'll help with support to have the logs

> * Be more explicit about what all to replace in the plist file when creating, like:
> 
> Replace:
> 
> `/path/to/redbot` with your redbot executable path
> `instance-name` with your instance name
> `username` with your Mac user name

Fair point

aikaterna:
I had some halfway success by using these instructions:
```
user@iMac / % export EDITOR=/System/Applications/TextEdit.app/Contents/MacOS/TextEdit
user@iMac / % export VISUAL="$EDITOR"                                                
user@iMac / % sudo -e /Library/LaunchDaemons/red.plist  
```

This does attempt to open the file in TextEdit, and launches the program, but then it said
Screen Shot 2021-03-02 at 12 34 34 PM

After hitting OK and closing the TextEdit process, the terminal window where the original commands were run is "hung" in the idea that it requires a ctrl-c to get back to the command prompt.

Jackenmen:
This guide should be running Red with `python -O -m redbot` rather than `redbot` executable directly - same as the systemd guide for Linux.

As for the `sudo -e`, I think Mac has `nano` too (which we should have been doing in systemd guide for long time tbh), so we could just use that, assuming it actually works on Mac ofc.

aikaterna:
`sudo nano` works fine on Mac for this situation instead of `sudo -e`. I also set `python` to be the bot's venv's python as Jack suggested with `<string>/Users/mac_username/venv_name/bin/python -O -m /Users/mac_username/venv_name/bin/redbot</string>` and that worked out ok as well.

Ryan5453:
idk if this is worthy to add, but while I was testing it I changed 
`<string>--no-prompt</string>` 
to 
`<string>--no-prompt --team-members-are-owners</string>` 
assuming I could add additional flags, but that did not work (the bot didn't start) and I needed to change it to 
`<string>--no-prompt</string>`
`<string>--team-members-are-owners</string>`

Boxck:
Ja takk
tir. 2. mar. 2021, 21:37 skrev aikaterna <notifications@github.com>:

palmtree5:
> `sudo nano` works fine on Mac for this situation instead of `sudo -e`. I also set `python` to be the bot's venv's python as Jack suggested with `<string>/Users/mac_username/venv_name/bin/python -O -m /Users/mac_username/venv_name/bin/redbot</string>` and that worked out ok as well.

Got it, will switch it to `sudo nano` as `nano` is way easier to work with than `vi` if the user is unfamiliar

> idk if this is worthy to add, but while I was testing it I changed
> `<string>--no-prompt</string>`
> to
> `<string>--no-prompt --team-members-are-owners</string>`
> assuming I could add additional flags, but that did not work (the bot didn't start) and I needed to change it to
> `<string>--no-prompt</string>`
> `<string>--team-members-are-owners</string>`

I'll make sure there's a note in about this.

palmtree5:
Ok, all requested changes have been made. Would someone just confirm that running `[p]restart` for the bot works? Not sure anyone's tested that specifically. Also may want to confirm that `which python` with the venv activated does indeed get the venv's python and not something else

Ryan5453:
I can confirm that `[p]restart` does work, and `which python` returned  `/Users/<my_user>/redenv/bin/python` for me (which worked)

Ryan5453:
One thing that might be worthy to add is how to view the logs, maybe `nano  /tmp/red_out.log`?

palmtree5:
Ooh, good suggestion! Will get that added at some point

palmtree5:
Added a note about the logging

aikaterna:
One thing I thought of while testing this again was about the eventual size of the log files because information would be appended to the end of each log in /tmp every time the bot is restarted. I found some information that was rather old (10.6.x) that claims that /tmp is emptied every 3 days but I'm not finding much for Catalina regarding this. This is the only concern I have left before approving this item and am willing to approve it anyway, but we may get people with very large log files if there is no automatic cleaning of /tmp in Catalina.

aikaterna:
Thanks to a link Palm provided I was able to find some info on, and confirm, that Catalina (10.15.x) should delete those files every 3 days or so. I would expect Big Sur does the same.

aikaterna (approved):

PR #4043, opened 2020-07-06
Create cog disabling API

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

### Description of the changes

Allow disabling cogs on a per guild basis (resolves #3945)

PR #4085, opened 2020-07-18
Document prefix preservation

mikeshardmind:

Reserves a prefix so that things like #4043 will not be a breaking addition in the future.

### DIFF
--- a/docs/version_guarantees.rst
+++ b/docs/version_guarantees.rst
@@ -20,6 +20,11 @@ which is not private (even if not documented) should not break without notice.
 Anything in the ``redbot.cogs`` and ``redbot.vendored`` modules or any of their submodules is specifically
 excluded from being guaranteed.
 
+Method names and names of attributes of classes, functions, extensions, and modules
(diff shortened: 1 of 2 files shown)


Jackenmen:
Considering that you said that #4045 might also allow to get/delete data on module-level (as additional, non-standard way) we need to at least expand this preservation to cog packages.

I partially want to say that we should expand this to *everything*, as in make it generic to all subclassable classes not just `Cog` and `Command` (+ what's said above), but to be fair it seems like unnecessary future-proofing, considering there's nothing else in Red that's meant to be subclassable (Help Formatter API is one more thing that is based on subclassing but that one is gonna be provisional first anyway) and therefore realistically supported by guarantees.

mikeshardmind:
Help Formatter is also reasonably supported by guarantees in that changing that one after provisional status would definitely need to be breaking either way since it's just a container class for methods with an abc that defines what's currently needed to function.

Jackenmen (approved):
Looks good to me, this PR is blocked until Red 3.4

PR #4089, opened 2020-07-20
[Announcer] Don't die with a lack of channels

### LINKED ISSUE #4088: Announcer doesn't handle the case where the server has no text channels
**What were you trying to do?**
Using the core's announcer cog.

**What were you expecting to happen?**
Sending the message to all the server the bot is in

**What actually happened?**
Some servers get the message and some not. Even if the bot has admin perm.

**How can we reproduce this issue?**
By this screenshot, the issue may be the fact that one of the server the bot is in has no txt channel.
So the fix would be ignoring a server without any txt channels and send a message to the bot owner specifying what server has no txt channel so we can force the bot to quit it.

287b728dc2301a7d862c09708cf4dd57

mikeshardmind:
### Type

- [X] Bugfix
- [x] Enhancement
- [ ] New feature

fixes #4088 by not trying to guess channels
This also assumes that a channel not being set means the server doesn't want an announcement, and isn't a failure.

### DIFF
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -75,9 +75,7 @@ class Admin(commands.Cog):
         self.config.register_global(serverlocked=False)
 
         self.config.register_guild(
-            announce_ignore=False,
-            announce_channel=None,  # Integer ID
-            selfroles=[],  # List of integer ID's
+            announce_channel=None, selfroles=[],  # Integer ID  # List of integer ID's
         )
 
         self.__current_announcer = None
@@ -320,7 +318,7 @@ class Admin(commands.Cog):
     async def announceset_channel(self, ctx, *, channel: discord.TextChannel = None):
         """
         Change the channel where the bot will send announcements.
-        
+
         If channel is left blank it defaults to the current channel.
         """
         if channel is None:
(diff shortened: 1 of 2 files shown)


Drapersniper:
Im not 100% okay with killing `announce_ignore` I could have set a channel then  set ignore.

and after this PR i will start getting the announcements again>

Can we also get a human redable message to `[p]announceset clearchannel` please. as a simple tick is not very friendly.

Drapersniper (changes requested):
*sigh*

Jackenmen:
> Im not 100% okay with killing announce_ignore I could have set a channel then set ignore.
> 
> and after this PR i will start getting the announcements again>

perhaps some "migration" should be done that would just clear the channel if `announce_ignore` is `True`? It seems a bit unnecessary to have separate setting for ignoring if it's already gonna be ignored **unless** you have set `announce_channel`.

This is a behavior change anyway though, so there's gonna be a note about it at the top of changelog no matter what.

mikeshardmind:
I added a migration to this which prefers the prior existance of ignoring announcements over having a channel set.

Drapersniper (changes requested):
Thanks for adding the migration, looking good now.

Can we also add a human friendly message to `[p]announceset clearchannel` instead of just a `await ctx.tick()` please.

Also wouldn't a context manager in the migration reduce the number of calls to config? if so can we please use the context manager on line 110 and 111 please.

mikeshardmind:
Can't use a context manager over all guilds.

Drapersniper:
@mikeshardmind 

> Can't use a context manager over all guilds.

i meant
```py
        for guild_id, guild_data in all_guilds.items():
            if guild_data.get("announce_ignore", False):
                await self.config.guild_from_id(guild_id).all() as guild_data:
                    del guild_data["announce_channel"]
                    del guild_data["announce_ignore"]
```

mikeshardmind:
I've switched this to a config manager, but I have no intention of changing this to not use `ctx.tick` This is a simple success and is what the method is designed for. Not using this will create a non-uniform experience across the bot and annoy users who requested less noisy feedback on basic things which led to that being added originally.

Drapersniper (approved):
After an internal conversation, I'm willing to approve this to not hold up the fix that this PR implements.

However going forward I will not be approving any further PR that uses `await ctx.tick()` as the sole response to users.

One proposal that was raised in the discussion was a method that accepts a message, and the bot owner/server can toggle wether the message or tick is displayed at the completion of the command, which would be a significant improvement over the sole react on the command, I've filed #4091 to track it.

PR #4095, opened 2020-07-21
Remove unused try except blocks in `modlog.create_case()` usage

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

`modlog.create_case()` doesn't raise since 3.0.0b15 (#1717) but these try except block were never removed.


Drapersniper (approved):
LGTM as we catch the RuntimeError inside create_case nowdays, this was never being caught in the places it was removed.

PR #4096, opened 2020-07-21
Use NONE privilege level in has_permissions check

PredaaA:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Currently a command that doesn't have any privilege level will return a None, while we already have an NONE value in PrivilegeLevel enum, that PR fixes that.

### DIFF
--- a/redbot/core/commands/requires.py
+++ b/redbot/core/commands/requires.py
@@ -553,9 +553,10 @@ class Requires:
             if user_perms.administrator or user_perms >= self.user_perms:
                 return True
 
-        privilege_level = await PrivilegeLevel.from_ctx(ctx)
-        if privilege_level >= self.privilege_level:
-            return True
+        if self.privilege_level is not PrivilegeLevel.NONE:
+            privilege_level = await PrivilegeLevel.from_ctx(ctx)
+            if privilege_level >= self.privilege_level:
+                return True
 
         return False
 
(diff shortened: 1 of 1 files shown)


redbot/core/commands/requires.py
```diff
@@ -553,10 +553,9 @@ async def _verify_user(self, ctx: "Context") -> bool:
             if user_perms.administrator or user_perms >= self.user_perms:
                 return True
 
-        if self.privilege_level is not None:
-            privilege_level = await PrivilegeLevel.from_ctx(ctx)
-            if privilege_level >= self.privilege_level:
-                return True
+        privilege_level = await PrivilegeLevel.from_ctx(ctx)
+        if privilege_level >= self.privilege_level:
+            return True
```
Jackenmen: This still needs to check if privilege is not `PrivilegeLevel.NONE` cause we don't want to allow people to use the command if developer used `@bot.has_permissions(...)`

Jackenmen:
After looking a bit closer, `PrivilegeLevel.NONE` and `None` **are** entirely different things and it is all intended.
To describe the differing behavior:
- `has_permissions()` is supposed to **only** accept discord permissions check and otherwise DENY access, therefore it's using `None`.
- default behavior of command (when there's no checks like `has_permissions()`, `mod()`, or `mod_or_permissions()`) is supposed to ALLOW any user to use the command, therefore it's using `PrivilegeLevel.NONE`

So all of the current behavior that I know of is correct and if something doesn't handle the case of `Requires.privilege_level` then that's where the bug is.

Unless there's something else that I'm missing here, I'm gonna close this issue later.

Jackenmen:
Looks like "later" ended up being almost 2 months.

PR #4098, opened 2020-07-22
Update chocolatey install commands per `chocolatey.org`

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

See https://chocolatey.org/install


Drapersniper (approved):
I think we should link them to choco install instructions instead of maintaining our own.

Sure that would cause users to switch tabs more often, but at least they would have up to date instructions (Not blocking the PR just stating my feeling about us duplicating install instructions.

For example When Docker support is added we shouldnt give them the instruction on how to install docker/compose on their desired OS... thats a headache you don't want.

PR #4109, opened 2020-07-24
Use mod/admin_or_permissions instead of just mod/admin.

flaree:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Commands locked to just mod role can be confusing at first especially if those don't understand them, this brings streams in line with a lot of other core cogs which uses _or_permissions.

This also updates the streamset docstring as that does more than just managing tokens.

### DIFF
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -280,7 +280,7 @@ class Streams(commands.Cog):
 
     @commands.group()
     @commands.guild_only()
-    @checks.mod()
+    @checks.mod_or_permissions(manage_channels=True)
     async def streamalert(self, ctx: commands.Context):
         """Manage automated stream alerts."""
         pass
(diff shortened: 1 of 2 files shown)


Jackenmen:
Could we do the same for commands using `.admin()` permission check where applicable?
There are currently 4 of those, and at least 2 should have the permissions alternative:
- `[p]set serverprefix` (`@commands.admin_or_permissions(manage_guild=True)`)
- `[p]set nickname` (`@commands.admin_or_permissions(manage_nicknames=True)`)

These 2 might not have a good permissions alternative, but I'm mentioning them here so the list of commands using `.admin()` permission check is complete and in case someone has some idea for perm to use:
- `[p]warnings`
- `[p]audioset dailyqueue`

flaree:
I've added both nickname/serverprefix but I couldn't think of any that suits warnings/dailyqueue.

Kowlin (approved):

PR #4111, opened 2020-07-24
[Trivia] whosthatpokemon - Gen 1 - replace "pokemon" with "Pokémon"

owocado:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

as discussed with aika in main server.


aikaterna (approved):
Thanks for doing this, we should use this going forward on the other gens too.

PR #4120, opened 2020-07-27
[Economy] Prevent forbidden error when blocked by user

MeatyChunks:
Stops `[p]payouts` throwing a console error if the user has blocked the bot. Probably too spammy to put the payout message in chat.

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/cogs/economy/economy.py
+++ b/redbot/cogs/economy/economy.py
@@ -532,7 +532,10 @@ class Economy(commands.Cog):
     @guild_only_check()
     async def payouts(self, ctx: commands.Context):
         """Show the payouts for the slot machine."""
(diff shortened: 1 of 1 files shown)


Jackenmen:
I don't really see a good reason to not just send it in the channel command was called from, but I'll let others say something first before requesting this change.

Zephyrkul:
imo it should have a per-channel cooldown in that case - I can't speak to why it initially was DM'd but it's a medium-sized, very colorful message so some users might get annoyed if it gets spammed.

Kowlin:
For now this seems fine, if we want to move it out of people's DMs we can do so at a later date.

Kowlin (approved):

PR #4121, opened 2020-07-27
Propose more info on simple serverinfo

madebylydia:
- [X] Enhancement

### DIFF
--- a/redbot/cogs/general/general.py
+++ b/redbot/cogs/general/general.py
@@ -270,7 +270,7 @@ class General(commands.Cog):
             data.set_footer(
                 text=_("Server ID: ")
                 + str(guild.id)
-                + _('  •  Use {command} for more info on the server.').format(
+                + _("  •  Use {command} for more info on the server.").format(
                     command=f"{ctx.clean_prefix}serverinfo 1"
                 )
             )
--- a/redbot/cogs/general/general.py
+++ b/redbot/cogs/general/general.py
@@ -267,7 +267,11 @@ class General(commands.Cog):
             data.add_field(name=_("Voice Channels"), value=voice_channels)
             data.add_field(name=_("Roles"), value=humanize_number(len(guild.roles)))
             data.add_field(name=_("Owner"), value=str(guild.owner))
-            data.set_footer(text=_("Server ID: ") + str(guild.id))
+            data.set_footer(
+                text=_("Server ID: ")
+                + str(guild.id) 
+                + _(" | Use `{prefix}serverinfo t` for more info on the server.")
+            )
             if guild.icon_url:
                 data.set_author(name=guild.name, url=guild.icon_url)
                 data.set_thumbnail(url=guild.icon_url)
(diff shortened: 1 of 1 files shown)


Drapersniper:
~~if you going to add this to embed, please also add it to non embeded version.~~

Please ignore me all the different PRs/realities are merging into one.

redbot/cogs/general/general.py
```diff
@@ -267,7 +267,11 @@ async def serverinfo(self, ctx, details: bool = False):
             data.add_field(name=_("Voice Channels"), value=voice_channels)
             data.add_field(name=_("Roles"), value=humanize_number(len(guild.roles)))
             data.add_field(name=_("Owner"), value=str(guild.owner))
-            data.set_footer(text=_("Server ID: ") + str(guild.id))
+            data.set_footer(
+                text=_("Server ID: ")
+                + str(guild.id) 
+                + _(" | Use `{prefix}serverinfo t` for more info on the server.")
```
Dav-Git: You're missing a ``.format(prefix=ctx.clean_prefix)`` at the end.

Dav-Git (changes requested):

redbot/cogs/general/general.py
Drapersniper: `clean_prefix` 😅 we don't want the bot to ping stuff.

redbot/cogs/general/general.py
```diff
@@ -267,7 +267,15 @@ async def serverinfo(self, ctx, details: bool = False):
...
             data.add_field(name=_("Owner"), value=str(guild.owner))
-            data.set_footer(text=_("Server ID: ") + str(guild.id))
+            data.set_footer(
+                text=_("Server ID: ")
+                + str(guild.id)
+                + _(
+                    ' | Use "{prefix}serverinfo 1" for more info on the server.'.format(
+                        prefix=ctx.clean_prefix
+                    )
+                )
```
Jackenmen: 1. `.format()` needs to be used *after* the string gets translated (so `_("...").format(...)` rather than `_("...".format(...))`)
2. We're trying to make an effort to not include command names in strings to translate, so using `{command}` replacement field for whole command would be better here.

```suggestion
                + _(' | Use "{command}" for more info on the server.').format(
                    command=f"{ctx.clean_prefix}serverinfo 1"
                )
```

Jackenmen (changes requested):
I think this a nice QoL addition, but there are 2 smaller issues that need a fix first.

redbot/cogs/general/general.py
```diff
@@ -267,7 +267,13 @@ async def serverinfo(self, ctx, details: bool = False):
             data.add_field(name=_("Voice Channels"), value=voice_channels)
             data.add_field(name=_("Roles"), value=humanize_number(len(guild.roles)))
             data.add_field(name=_("Owner"), value=str(guild.owner))
-            data.set_footer(text=_("Server ID: ") + str(guild.id))
+            data.set_footer(
+                text=_("Server ID: ")
+                + str(guild.id)
+                + _(' | Use "{command}" for more info on the server.').format(
```
Jackenmen: Another suggestion after discussion/voting in Discord.
```suggestion
                + _('  •  Use {command} for more info on the server.').format(
```

Jackenmen (changes requested):

Jackenmen (approved):

Jackenmen:
Thanks for the PR!

PR #4131, opened 2020-07-30
[Mod] Correctly log and store the previous (nick)name

Kowlin:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

So it appears we had this bug in Red for the past 3 years, where we didn't store the correct past (nick)name. Lets fix that.

### DIFF
--- a/redbot/cogs/mod/events.py
+++ b/redbot/cogs/mod/events.py
@@ -104,10 +104,10 @@ class Events(MixinMeta):
             async with self.config.user(before).past_names() as name_list:
                 while None in name_list:  # clean out null entries from a bug
                     name_list.remove(None)
-                if after.name in name_list:
+                if before.name in name_list:
                     # Ensure order is maintained without duplicates occurring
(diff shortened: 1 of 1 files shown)


mikeshardmind:
While it was added to V3 3 years ago, it's historically accurate with the 4 year old V2 behavior. Hurray for consistency? <https://github.com/Cog-Creators/Red-DiscordBot/commit/8b52bd56af7904afef7ad49b3392d7dc1cc7bcd3>

mikeshardmind (changes requested):
To avoid not having as full a record as is possible if someone leaves, name changes, then rejoins, this should probably store both the before and after (with nick still checking against None) rather than just one or the other.

Jackenmen (commented):
@Kowlin Works as intended, do you want to add logging of "after" (nick)names here as well?

Kowlin:
Not at the moment, no.

Jackenmen (approved):

PR #4132, opened 2020-07-30
[Drivers] Add a msgpack driver

mikeshardmind:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

  - Smaller files than json (less blocking)
  - Easy review (it's the same as json, but with a different (de)serializer)
  - msgpack is an incredibly [stable, compact serialization](https://msgpack.org/)
  - Makes msgpack the default for new instances for performance reasons.

### DIFF
--- a/redbot/core/drivers/__init__.py
+++ b/redbot/core/drivers/__init__.py
@@ -5,6 +5,7 @@ from .. import data_manager
 from .base import IdentifierData, BaseDriver, ConfigCategory
 from .json import JsonDriver
 from .postgres import PostgresDriver
+from .msgpack import MsgpackDriver
 
 __all__ = [
     "get_driver",
@@ -12,6 +13,7 @@ __all__ = [
     "IdentifierData",
     "BaseDriver",
     "JsonDriver",
+    "MsgpackDriver",
     "PostgresDriver",
     "BackendType",
 ]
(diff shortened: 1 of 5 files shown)


redbot/core/drivers/__init__.py
```diff
@@ -109,7 +117,7 @@ def get_driver(
         try:
             storage_type = BackendType(data_manager.storage_type())
         except RuntimeError:
-            storage_type = BackendType.JSON
+            storage_type = BackendType.MSGPACK
```
Drapersniper: Let's please not make this the default driver on same PR it is introduced

Drapersniper (changes requested):
Happy for this to be introduced as it will give users an alternative driver that is in theory better than JSON. However I'm not okay with it being made the default driver in the PR that introduces it, and would like some soak time before we switch it to be the default driver.

redbot/core/drivers/__init__.py
mikeshardmind: as in discord, this isn't a default quite in that way. Completely understandable how that could happen, but we force the user to actually make a choice, and I'm unsure when this would ever hit this exception handling, but did not want to check if this one was safe to remove as part of this.

mikeshardmind:
This is currently in use on a couple of bots. I intend to be using this from here forward (for now anyhow). This is stable enough that I have only mild reservations about deploying this a few hours before I'm unavailable for a few days, but the immediate benefits of this are already evident.

PR #4134, opened 2020-07-30
[RPC] Shutdown properly

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This appears to solve the hanging issues seen by @NeuroAssassin, but I'd prefer confirmation of that working in that specific case.

### DIFF
--- a/redbot/core/rpc.py
+++ b/redbot/core/rpc.py
@@ -75,7 +75,9 @@ class RPC:
         accepting queries.
         """
         await self._runner.setup()
-        self._site = web.TCPSite(self._runner, host="127.0.0.1", port=port)
+        self._site = web.TCPSite(
+            self._runner, host="127.0.0.1", port=port, shutdown_timeout=0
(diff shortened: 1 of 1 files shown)


NeuroAssassin:
Solved the issue for me 👍

Jackenmen (approved):
Looks go to me.
[According to aiohttp documentation](https://docs.aiohttp.org/en/stable/web_reference.html#aiohttp.web.run_app):
> `shutdown_timeout` [is] a delay to wait for graceful server shutdown before disconnecting all open client sockets hard way.
> A system with properly [Graceful shutdown](https://docs.aiohttp.org/en/stable/web_advanced.html#aiohttp-web-graceful-shutdown) implemented never waits for this timeout but closes a server in a few milliseconds.

so maybe this needs something like that, however this is anyway a big improvement over the bot hanging for over a minute to shutdown.

PR #4138, opened 2020-07-31
[Core] Make core_commands.py comply with #4048

### LINKED ISSUE #4048: [Tracking] An initiative to make codebase racially neutral

Dav-Git:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This PR starts work on #4048 but only takes care of ``redbot/core/core_commands.py`` renaming all instances of "blacklist" to "denylist" and "whitelist" to "allowlist".


Drapersniper:
Thanks!
For the sake of conflicts and to not risk delaying #4045 any further. Going to mark this PR as blocked by that one,  (since this bad boy will cause a lot of other PRs to become conflicting once it is merged)

Dav-Git:
Hey look at that, no conflicts! Or do I need to do some magic now the DATA API is merged?

Drapersniper:
Can I be a complete and utter ass ...  at least for myself `blocklist` makes more sense to me than `denylist`

So can you change it please ... (Before you do give it .. 24 hours and see if anyone speaks up agaisnt/in favour  though)

Dav-Git:
Sorry Draper, I totally didn't forget about this at all. Got it switched out (properly TM) but added ``denylist`` as an alias because I like the word. Does that work for you?

Drapersniper:
In the surface I'm fine with it staying as an alias ... however I do have concerns about it clogging the namespace for commands (that's why we avoid aliases to bare commands and tend to apply them only to sub commands that way a single core command isn't reserving multiple commands in the namespace).

If no one is Has an issue with it I'm fine with it in this case.

Dav-Git:
The thought process here was that anything implementing similar allow/blocklist functionality is already limited by the unavailability of the allowlist command as well as the danger of being confused. Most 3rd party cogs would solve this by putting such commands behind a settings command group. I think while I generally agree that the command namespace should be kept as clear as possible this does not take away a large chunk of the practically usable spectrum.

Drapersniper (dismissed):
LGTM...

I will say though, im not a fan of the "Local blocklist management commands.' usage, specifically `Local ` here as without context I'm not sure what local means. IMO it should be `server`  however in other places in the bot we do use local so won't block or nitpick on this PR because of it.

Dav-Git:
Do NOT merge this, just found a bunch of instances of denylist my search and replace didn't catch and it's inconsistent in some places.

To adress latest review comment, I chose local in the strings where I did due to the function names. Let me know if you want that changed.

redbot/core/core_commands.py
```diff
@@ -1922,11 +1922,11 @@ async def allowlist_clear(self, ctx: commands.Context):
         await self.bot._whiteblacklist_cache.clear_whitelist()
         await ctx.send(_("Allowlist has been cleared."))
 
-    @commands.group(aliases=["blacklist", "denylist"])
+    @commands.group(aliases=["blacklist", "list"])
```
Drapersniper: Did you mean to do this here and in line 1994?

Dav-Git: Close but not quite. One sec

redbot/core/core_commands.py
```diff
@@ -1945,36 +1945,36 @@ async def blacklist_add(self, ctx: commands.Context, *users: Union[discord.Membe
...
+    async def blocklist_list(self, ctx: commands.Context):
         """
-        Lists blacklisted users.
+        Lists users on the blocklist.
         """
         curr_list = await self.bot._whiteblacklist_cache.get_blacklist(None)
 
         if not curr_list:
-            await ctx.send("Blacklist is empty.")
+            await ctx.send("list is empty.")
```
```suggestion
            await ctx.send("Blocklist is empty.")
```

Drapersniper (dismissed):
Only change needed here

Drapersniper (approved):
LGTM, Thanks Dav.

PR #4141, opened 2020-07-31
[Trivia] whosthatpokemon - Generation III

owocado:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Credits and big thanks to @aikaterna for most of the heavy work.


Jackenmen (commented):
I checked that each URL here works, I'm assuming the answers should be good, but @aikaterna is more involved in this than me, so I'm not approving this before I hear about it

Jackenmen (approved):

PR #4142, opened 2020-07-31
Remove provisional note from `NoParseOptional`, fix and start using `UserInputOptional`

Jackenmen:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

This PR:
- Removes provisional note from `NoParseOptional`
- Fixes type checking version of `UserInputOptional` (per the note in its docstring, it wasn't working correctly)
    - Small note here: this is not an ideal solution as the type hint can be passed multiple types (it will fail at runtime, but type checkers won't be able to detect that it's gonna fail), but this is the best thing that can be achieved without writing mypy plugin; this is also noted in docstring
- Adds some usage of `UserInputOptional` to core cogs

I also wanted to remove provisional note from `ignore_optional_for_conversion` kwarg in `Command`, but apparently there was no such note to begin with.

### DIFF
--- a/docs/framework_commands.rst
+++ b/docs/framework_commands.rst
@@ -40,10 +40,13 @@ extend functionalities used throughout the bot, as outlined below.
 
 .. automodule:: redbot.core.commands.converter
     :members:
-    :exclude-members: convert
+    :exclude-members: UserInputOptional, convert
     :no-undoc-members:
 
     .. autoclass:: APIToken
+
+    .. autodata:: UserInputOptional
+        :annotation:
 
 ******************
 Help Functionality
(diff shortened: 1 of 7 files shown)


Drapersniper:
Its working fine but docs are conflicting.
(Why aren't we merging all the 
```py
if not TYPE_CHECKING:
```
under a since place rather than calling this check multiple times in the same module?

Jackenmen:
> Why aren't we merging all the `if not TYPE_CHECKING:` under a single place

I wanted to say for readability purposes - this way we have both versions of the classes right next to each other, but I also just figured that we have the re-exports at the top, so the readability fails a bit for those cases. I could probably move the re-export imports to the place the types are defined in for that though.

Drapersniper (approved):
LGTM

PR #4143, opened 2020-07-31
[RPC] Fix for non-rpc users

mikeshardmind:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

  - RPC probably needs rewriting.
  - Also, I noticed some extremely sharp edges
  and a potential crash point (unrelated to the fixed issue)
  on windows that the side effects from have been mitigated here
  partially.

Also, this needs testing by a windows user not using rpc


Jackenmen (approved):
Looks good to me, thanks for the follow up!

PR #4156, opened 2020-08-04
Fix amount of messages in log message of `[p]cleanup message`

### LINKED ISSUE #4155: [Cleanup] bot reports incorrect number of messages it mass deleted
# Command bugs

#### Command name

[p]cleanup messages [>100]

Cleanup

Bot should report the exact or correct amount of messages it deleted from the given channel in console/logs.

#### What actually happened?

Bot reports incorrect amount of messages deleted in console. I think it passes the same number we give as input.

Load cleanup cog, if haven't.
Try to delete more than 100 messages in a given channel with `[p]cleanup messages`. Confirm when prompted to. 
Try to pass an integer larger than the total number of messages of last 2 weeks which exists in the target channel. <- This step is essential to reproduce this bug.
Observe in your console/logs the amount the bot returns as deleted messages.
image
(^^for context in this screenshot, I used `[p]cleanup messages 500` in a channel.)

Now check your guild audit logs and see how many messages the bot actually deleted and cross-verify with the output bot provided in console/logs.
image
You should be able to confirm that bot reports incorrect number of messages it mass deleted.

-----
While I have only observed and tested this with `[p]cleanup messages` subcommand, I assume the same bug should occur with other cleanup subcommands becuase of how `check_100_plus()` staticmethod is handled, though I could be wrong, but please verify it with other `cleanup` subcommands.

MeatyChunks:
Fixes #4155

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -378,7 +378,7 @@ class Cleanup(commands.Cog):
         to_delete.append(ctx.message)
 
         reason = "{}({}) deleted {} messages in channel {}.".format(
-            author.name, author.id, number, channel.name
+            author.name, author.id, len(to_delete), channel.name
         )
         log.info(reason)
 


Jackenmen (approved):
Looks good to me.

Vexed01 (changes requested):
All the other commands in cleanup use `humanize_number(len(to_delete), override_locale="en_US")` to do the number in the log. Why not here?

Edit - just to clarify what happened, [jack didn't see this](https://discordapp.com/channels/133049272517001216/718148684629540905/740150637710278747). It's not too important, don't worry.

PR #4158, opened 2020-08-05
[CustomCom] Implement user_allowed check

Zephyrkul:
# Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

This now exists in the form of `bot.message_eligible_as_command`


Zephyrkul:
May or may not be superseded by another PR I'm working on. Marking as draft until then.

Zephyrkul:
Closing in favor of #4159. While 4159 will still listen to blacklisted / ignored users to a certain extent, CCs won't trigger under 4159 for them. Rather, CC will generate the command and cache it regardless of blacklists / ignorelists, and won't run it. If a blacklisted user wants to speed up a CC call for the next person to call it, especially since the bot won't waste further processing on that command, so be it.

PR #4169, opened 2020-08-07
EUD things for Downloader

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Adds support for `end_user_data_statement` key in `info.json`.
What this functionality brings:
- end user data statement is shown by `[p]cog info` command
- user is informed about being able to check end user data statement with `[p]cog info` during cog install
- if end user data statement changes, user will be informed during cog update

// diff is small, but the `_update_cogs_and_libs` got a little bit messy (in my view anyway); it should all work, but I don't think I can make the code much better without rewriting more stuff

**Things in this PR may still change.**

### DIFF
--- a/docs/guide_publish_cogs.rst
+++ b/docs/guide_publish_cogs.rst
@@ -55,6 +55,9 @@ Keys common to both repo and cog info.json (case sensitive)
 Keys specific to the cog info.json (case sensitive)
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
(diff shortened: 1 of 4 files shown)


Jackenmen:
I also have a question here - is there anything else that we should be doing towards this in Downloader?

mikeshardmind:
Since the end-user data statement is new + is most likely to be the longest part of the info, should it be at the end of the cog info? Not sure if this will make it look better in the typical cases or not.

Kowlin (approved):

PR #4172, opened 2020-08-08
[Mute] Indicate that a guild mute/unmute is currently being processed.

### LINKED ISSUE #4066: Indicate when a server-wide mute is in the process of being applied
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality

While the great overhaul for the muting system is being awaited, the current one works fine most of the time, except that it scales particularly poorly to large servers by taking up to two minutes to finish applying each of the channel permission overrides whenever you have an extremely large number of channels.

Because of this, moderators unfamiliar with the command will attempt to use it, see no response from the bot, assume they made a mistake, and then repeatedly try again (perhaps with slight variations to the command). After a minute of trial-and-error, they will throw up their hands in defeat and frustration, and then a minute later, the bot will send five separate confirmations that the user has been muted.

Currently, the easiest solution to this would just be for the bot to send a message whenever a mute command has been successfully entered announcing that the user-to-be-muted is in the process of being muted, and that the moderator should be patient. After it has completed, this message could be edited to the usual "User has been muted in this server".

This sort of feedback would definitely make it more accessible for people unfamiliar with the bot by clearly and immediately communicating the effect of the command to the moderator.

Dav-Git:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Added a channel.typing() block around the for loop applying the mute per channel to indicate to a user that the bot is in the process of doing something. This closes #4066

### DIFF
--- a/redbot/cogs/mod/mutes.py
+++ b/redbot/cogs/mod/mutes.py
@@ -254,22 +254,23 @@ class MuteMixin(MixinMeta):
         audit_reason = get_audit_reason(author, reason)
 
         mute_success = []
-        for channel in guild.channels:
-            success, issue = await self.mute_user(guild, channel, author, user, audit_reason)
-            mute_success.append((success, issue))
-            await asyncio.sleep(0.1)
-        await modlog.create_case(
-            self.bot,
-            guild,
-            ctx.message.created_at,
-            "smute",
-            user,
-            author,
-            reason,
-            until=None,
-            channel=None,
-        )
-        await ctx.send(_("User has been muted in this server."))
+        async with ctx.channel.typing():
+            for channel in guild.channels:
+                success, issue = await self.mute_user(guild, channel, author, user, audit_reason)
+                mute_success.append((success, issue))
+                await asyncio.sleep(0.1)
+            await modlog.create_case(
+                self.bot,
+                guild,
+                ctx.message.created_at,
+                "smute",
+                user,
+                author,
+                reason,
+                until=None,
+                channel=None,
+            )
+            await ctx.send(_("User has been muted in this server."))
 
     @commands.group()
     @commands.guild_only()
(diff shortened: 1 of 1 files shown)


redbot/cogs/mod/mutes.py
```diff
@@ -254,22 +255,22 @@ async def guild_mute(self, ctx: commands.Context, user: discord.Member, *, reaso
...
-            "smute",
-            user,
-            author,
-            reason,
-            until=None,
-            channel=None,
-        )
-        await ctx.send(_("User has been muted in this server."))
+        async with ctx.channel.typing():
+            async for channel in AsyncIter(guild.channels):
```
mikeshardmind: This `AsyncIter` use appears to be wasteful and further slows down how long mutes take. Each channel mute will already yield control of the event loop during communication with Discord

redbot/cogs/mod/mutes.py
```diff
@@ -254,22 +255,22 @@ async def guild_mute(self, ctx: commands.Context, user: discord.Member, *, reaso
...
-            ctx.message.created_at,
-            "smute",
-            user,
-            author,
-            reason,
-            until=None,
-            channel=None,
-        )
-        await ctx.send(_("User has been muted in this server."))
+        async with ctx.channel.typing():
```
fixator10:
```suggestion
        async with ctx.typing():
```

redbot/cogs/mod/mutes.py
Dav-Git: Will the mute itself do that or the ``asyncio.sleep(0.1)`` that was here previously. If the mute does it on it's own, I don't see why we were sleeping in the first place here.

mikeshardmind: the .1 sleep that was there previously was similarly unneeded. discord.py already handles any actual rate limit issues this could otherwise cause already, and each set of channel overwrites we want to set involves a network request. Under the hood, when the request is made, a context switch will happen while awaiting a response. There's also nothing preventing this from being done as a gather operation other than that nobody decided to write it this way.

Drapersniper (approved):
LGTM

PR #4179, opened 2020-08-13
REUSE Compliance and remove changes to the GPLv3 license document

nfitzen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature

- Use SPDX identifiers in relevant documents and Debian's DEP5 format for bulk-licensing the entire project as well as vendored ones
    - Use the GPLv3 document in `COPYING` verbatim and move other projects' licenses to individual files/dep5
    - Adding SPDX license information to every file would be preferred, but it could be breaking, so it's not done here. That's for a different commit.
- Rename the `LICENSE` file to `COPYING` akin to the GNU standard, although that's probably a bad idea in all honesty since external websites probably link to this project's LICENSE directly. It might be preferred to change it back.

#### Rationale
According to the copyright notice in the [GPLv3 license document](https://www.gnu.org/licenses/gpl-3.0.en.html), "everyone is permitted to copy and distribute verbatim copies of this license document, but changing it is not allowed."
As such, I think modifying even the "How to Apply These Terms to Your New Programs" is a copyright violation since that section is a part of the document itself (although IANAL).

Either way, using SPDX identifiers is probably a good idea, and keeping the GPLv3 document in its own file (without other licenses) can improve license detection.

The [REUSE](https://reuse.software/) project uses a combination of [SPDX](//spdx.dev/) identifiers and [Debian's DEP5 format](//www.debian.org/doc/packaging-manuals/copyright-format/1.0/) to make licensing easy. This would be an easy way to state copyright for vendored or reused code and would make it easier for automated tools.


Kowlin:
Hey there, thank you for your contribution. However we are not accepting any changes to our licenses.

nfitzen:
@Kowlin it's not a change to the license, though? Unless it was licensed under GPLv3 or later, in which case that needs to be specified in the README.

Currently, the GPLv3 license document has been modified, which a) could be a copyright violation, b) makes it harder for automated tools to work with it, and c) [isn't recommended by SPDX](//spdx.org/licenses/GPL-3.0-only.html). The copyright notice should be in the README and/or the code itself, not the license file, when applying the GPL.

I hope you might reconsider at some point.

Twentysix26:
As stated before we are currently not accepting changes related to our licenses. This means that we won't consider *any* type of change to how our current licenses are presented. Thank you for your understanding.

PR #4182, opened 2020-08-14
Add allowed_mentions and mention everyone check to stream alerts

kablekompany:
Allows role mentions for discord.py 1.4+

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Referencing line `747` in commit. Gives permissions to mention role on stream announcement.

### DIFF
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -747,11 +747,6 @@ class Streams(commands.Cog):
                             allowed_mentions=discord.AllowedMentions(roles=True, everyone=True),
                         )
                         stream._messages_cache.append(m)
-                        guild = channel.guild
-                        can_mention_everyone = guild.me.guild_permissions.mention_everyone
-                        if can_mention_everyone:
-                            await self.save_streams()
-                            return
                         if edited_roles:
                             for role in edited_roles:
                                 await role.edit(mentionable=False)
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -765,10 +765,6 @@ class Streams(commands.Cog):
             mentions.append("@here")
         can_manage_roles = guild.me.guild_permissions.manage_roles
         can_mention_everyone = channel.permissions_for(guild.me).mention_everyone
-        if can_mention_everyone:
-            for role in guild.roles:
-                mentions.append(role.mention)
-                return
         for role in guild.roles:
             if await self.config.role(role).mention():
                 if not can_mention_everyone and can_manage_roles and not role.mentionable:
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -764,7 +764,7 @@ class Streams(commands.Cog):
         if await settings.mention_here():
             mentions.append("@here")
         can_manage_roles = guild.me.guild_permissions.manage_roles
-        can_mention_everyone = guild.me.guild_permissions.mention_everyone
+        can_mention_everyone = channel.permissions_for(guild.me).mention_everyone
         if can_mention_everyone:
             for role in guild.roles:
                 mentions.append(role.mention)
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -773,7 +773,7 @@ class Streams(commands.Cog):
         if can_mention_everyone:
             for role in guild.roles:
                 mentions.append(role.mention)
-            return
+                return
         for role in guild.roles:
             if await self.config.role(role).mention():
                 if can_manage_roles and not role.mentionable:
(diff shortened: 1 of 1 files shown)


Jackenmen (commented):
While this PR solves the issue with mentions not actually pinging, do you think you could also make this not edit the roles if bot has mention everyone perms (as that perm now also allows to ping non-mentionable roles)?

kablekompany:
## Lines for reference:

- `[750-754]`:
```py
guild = channel.guild
can_mention_everyone = guild.me.guild_permissions.mention_everyone
if can_mention_everyone:
  await self.save_streams()
  return # if bot can mention everyone already, let's stop here 
```

- `[772-776]`: 
```py
can_mention_everyone = guild.me.guild_permissions.mention_everyone
if can_mention_everyone:
  for role in guild.roles:
    mentions.append(role.mention)
    return  # if bot can mention everyone already, let's stop here
```


Hopefully, this is what you had in mind? Humbly admit I may not have a business in biting this piece but looking forward to feedback on the best methodology for the wanted outcome. Sorry for the redraws on simple mistakes...

Thanks!

redbot/cogs/streams/streams.py
```diff
@@ -741,8 +741,17 @@ async def check_streams(self):
...
+                            content,
+                            embed=embed,
+                            allowed_mentions=discord.AllowedMentions(roles=True, everyone=True),
+                        )
                         stream._messages_cache.append(m)
+                        guild = channel.guild
+                        can_mention_everyone = guild.me.guild_permissions.mention_everyone
+                        if can_mention_everyone:
+                            await self.save_streams()
+                            return
```
Jackenmen: This can all be removed if you apply my other suggestion.
```suggestion
```

redbot/cogs/streams/streams.py
```diff
@@ -760,6 +769,11 @@ async def _get_mention_str(self, guild: discord.Guild) -> Tuple[str, List[discor
         if await settings.mention_here():
             mentions.append("@here")
         can_manage_roles = guild.me.guild_permissions.manage_roles
+        can_mention_everyone = guild.me.guild_permissions.mention_everyone
+        if can_mention_everyone:
+            for role in guild.roles:
+                mentions.append(role.mention)
+                return
```
Jackenmen: This can be all removed if you apply my other suggestion.
```suggestion
```

redbot/cogs/streams/streams.py
```diff
@@ -760,6 +769,11 @@ async def _get_mention_str(self, guild: discord.Guild) -> Tuple[str, List[discor
...
         for role in guild.roles:
             if await self.config.role(role).mention():
                 if can_manage_roles and not role.mentionable:
```
Jackenmen: This is all that needs to be done here:
```suggestion
                if not can_mention_everyone and can_manage_roles and not role.mentionable:
```

redbot/cogs/streams/streams.py
```diff
@@ -760,6 +769,11 @@ async def _get_mention_str(self, guild: discord.Guild) -> Tuple[str, List[discor
         if await settings.mention_here():
             mentions.append("@here")
         can_manage_roles = guild.me.guild_permissions.manage_roles
+        can_mention_everyone = guild.me.guild_permissions.mention_everyone
```
Jackenmen: It would be better if this were checking for permissions in channel:
```suggestion
        can_mention_everyone = channel.permissions_for(guild.me).mention_everyone
```
This requires you to also add a `channel` argument to this function and pass it in the place where it's called.

Jackenmen (changes requested):
The added changes make the bot mention all roles in the guild, which is not the same behavior as there was before.

This can also be simplified a lot, I sent the review comments which should make it clear what should be done here.

kablekompany:
committed suggestions and added channel iteration for `can_mention_everyone` https://github.com/kablekompany/Red-DiscordBot/blob/8eb21913bb402b634af7a22a0414ede7c3014eb4/redbot/cogs/streams/streams.py#L767

redbot/cogs/streams/streams.py
```diff
@@ -760,9 +764,11 @@ async def _get_mention_str(self, guild: discord.Guild) -> Tuple[str, List[discor
         if await settings.mention_here():
             mentions.append("@here")
         can_manage_roles = guild.me.guild_permissions.manage_roles
+        for channel in guild.channels:
+            can_mention_everyone = channel.permissions_for(guild.me).mention_everyone
```
Jackenmen: As I said before, you need to add `channel` param to `_get_mention_str()` method so that the bot checks the permission for the channel it will send the message in. Iterating through all channels here doesn't make much sense - we only care about permissions in *that specific channel* the message will be sent in.

kablekompany: understood. param added, iteration removed

Jackenmen (changes requested):
Looks alright now, there's just one thing left - it looks like you forgot to pass the channel argument to `_get_mention_str()` call in line 716.

kablekompany:
big time appreciate your patience and guidance @jack1142

Jackenmen (approved):
Looks good to me, and thanks so much for including new functionality related to Mention Everyone perm as well ❤️

PR #4184, opened 2020-08-15
[Economy] Add embed to leaderboard

### LINKED ISSUE #4104: [Economy] Add embed to leaderboard
- [ ] Cog
- [x] Command
- [ ] API functionality

add embed support to leaderboard.

TurnrDev:
# Enhancement request

#### Describe the enhancement

add embed support to leaderboard as requested on #4104

Yes

Closes #4104

### DIFF
--- a/redbot/cogs/economy/economy.py
+++ b/redbot/cogs/economy/economy.py
@@ -484,6 +484,7 @@ class Economy(commands.Cog):
         guild = ctx.guild
         author = ctx.author
         max_bal = await bank.get_max_balance(ctx.guild)
+        embed_requested = await ctx.embed_requested()
         if top < 1:
             top = 10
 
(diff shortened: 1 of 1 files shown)


TurnrDev:
Embed: image

The embed author is set to the guild name and icon, unless global, when it's set to the bots name and icon

No embed: 
image

redbot/cogs/economy/economy.py
```diff
@@ -536,15 +543,58 @@ async def leaderboard(self, ctx: commands.Context, top: int = 10, show_global: b
...
-                highscores.append(box(temp_msg, lang="md"))
+                if embed_requested:
+                    embed = base_embed.copy()
+                    embed.description = box(temp_msg, lang="md")
+                    embed.set_footer(
+                        text=_(
+                            "You are {author_pos}/{len_board}. Page {page_num}/{page_len}."
+                            " Use {left} and {right} to navigate."
+                        ).format(
+                            author_pos=bank.get_leaderboard_position(author),
```
Drapersniper: this is missing an await on `bank.get_leaderboard_position` however please dont do this, the this to a variable outside the loop and then reference this variable here and in L578

Drapersniper (changes requested):
Please do the same thing and assign variables rather than making a list and slicing it on like 558,559, 582 and 583

Line 552 and 576 will tell the users what reaction to use to move, but it will show even if there's only 1 page, I would reccomend this part is not added to the the embed at all

redbot/cogs/economy/economy.py
TurnrDev: Thank you for spotting that! I've now fixed this in 516cf91fc6310968848f431410764debbe4d828d :)

TurnrDev:
> Please do the same thing and assign variables rather than making a list and slicing it on like 558,559, 582 and 583
> 
> Line 552 and 576 will tell the users what reaction to use to move, but it will show even if there's only 1 page, I would reccomend this part is not added to the the embed at all

16a4b7e0331fe0143428c8ce3265b9f38fb68c23
Made changes :) 

Removed instructions as instructions aren't added anywhere else on the bot and should be the scope of a separate PR.

Drapersniper (approved):
Awesome its looking smexy... thanks for this.
image


Once I'll further tweak the UI in #4062 as part of the dpy implementation, but stils already looking pretty darn good!

Drapersniper (changes requested):
so turns out ad i looked at the screenshot a second time ... the Position is off `25/10`

Drapersniper (approved):
Thansk for the PR

TurnrDev:
No no, thank you for the challenge @Drapersniper :)

PR #4188, opened 2020-08-16
Don't send message about DMing ACL when command's called from DM

### LINKED ISSUE #4178: [Permissions] DM notification for command in DM
# Command bugs

#### Command name

`permissions acl getglobal`

`Permissions`

Get YAML without notification about DM

#### What actually happened?

Got YAML file and `I've just sent the file to you via DM.` message in DM

Execute `[p]permissions acl getglobal` in DM

douglascdev:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

`[p]permissions acl getglobal` will only send a message telling the user the file will be send via DM if the command message didn't come from a DM, closes #4178

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -332,7 +332,7 @@ class Permissions(commands.Cog):
         except discord.Forbidden:
             await ctx.send(_("I'm not allowed to DM you."))
         else:
-            if not isinstance(ctx.channel, discord.channel.DMChannel):
+            if not isinstance(ctx.channel, discord.DMChannel):
                 await ctx.send(_("I've just sent the file to you via DM."))
         finally:
             file.close()
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -333,7 +333,8 @@ class Permissions(commands.Cog):
         except discord.Forbidden:
             await ctx.send(_("I'm not allowed to DM you."))
         else:
-            await ctx.send(_("I've just sent the file to you via DM."))
+            if not isinstance(ctx.channel, discord.channel.DMChannel):
+                await ctx.send(_("I've just sent the file to you via DM."))
         finally:
             file.close()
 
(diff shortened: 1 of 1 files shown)


redbot/cogs/permissions/permissions.py
```diff
@@ -333,7 +332,8 @@ async def permissions_acl_getglobal(self, ctx: commands.Context):
         except discord.Forbidden:
             await ctx.send(_("I'm not allowed to DM you."))
         else:
-            await ctx.send(_("I've just sent the file to you via DM."))
+            if not isinstance(ctx.channel, discord.channel.DMChannel):
```
Jackenmen: It would be a bit safer to use `discord.DMChannel` name rather than `discord.channel.DMChannel` as discord.py doesn't cover the latter by the version guarantees (however unlike it is that this would get changed).

Or alternatively, you could check if `ctx.guild` is `None` instead.

Jackenmen (dismissed):
One nit-pick, PR looks good otherwise 👍

redbot/cogs/permissions/permissions.py
douglascdev: Thank you for the review, I've changed it to `discord.DMChannel`. For future PR's, what should I've done to find that out by myself? I know now that the docs tell you to use `discord.DMChannel`(should have looked that up from the start), but is that how you know that `discord.channel.DMChannel` is not covered by version guarantees? Because it's not specifically mentioned in the docs?

Jackenmen: Per [discord.py's version guarantees](https://discordpy.readthedocs.io/en/stable/version_guarantees.html), only documented things are part of the public API that is covered by them.

// FYI, we are doing some things in Red that are not supported by d.py but only when it's required for what we try to achieve (e.g. there's no supported way to do so) and we try to mark such things in code so that we can check them when updating dependencies.

Jackenmen (approved):
LGTM, thanks for the fix!

redbot/cogs/permissions/permissions.py
douglascdev: Got it, thank you jack

PR #4189, opened 2020-08-16
Shorten the audit log reason appropriately, remove unused vars

### LINKED ISSUE #4187: [Mod] Receiving "wrong" error when banning with more than 512 characters.
# Other bugs

Trying to ban a member with over 512 characters.

Receiving a well-made error.

#### What actually happened?

Bot say `An unexpected error occurred.`

Ban someone with more than 512 characters.

madebylydia:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This close #4187.


Kowlin (dismissed):
Not really happy with the solution here, I'd rather truncate the messages when its being written to modlog and keep the full message within the configs.

Jackenmen:
I think it would be better, if `get_audit_reason()` were updated with a new keyword-only argument called `shorten` (defaulting to `False`) that would automatically shorten the audit reason to 512 characters, if needed. This would allow us to avoid repeating the same code as the current state of the PR does.

If someone from org gives `+1` to this, I can update the PR appropriately myself, otherwise I don't want to because it's a bit more than a minor change.

Flame442:
+1 assuming `shorten` says in the docs that it will shorten to fit the spec of discord audit logs

Jackenmen (approved):
This should be ready to go now, thanks.

PR #4195, opened 2020-08-17
Actually change the version before release!

Kowlin:
Did you know that changing version is helpful with not making our CI fail :D


Jackenmen (approved):
what an awesome PR!

PR #4198, opened 2020-08-18
 [Drivers] Add SQLite3 local driver support

### LINKED ISSUE #3832: Deploy SQLite (apsw - JSON)  driver

### LINKED ISSUE #3922: [Drivers] Add a SQL driver using JSON1

Drapersniper:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

fixed #3922
fixes #3832

Look... I know it aint pretty... yes it is monkey patched... but it works.

There are some improvements i need to do here but time for me to put my legs up and go watch some anime.

### Key Points
1. This driver will always be avaliable for Red instances just the JSON driver, as it requires no extra dependencies.
2. This driver will not support external SQL servers, and will store files in the local host.
3. Each cog identifier get its own database.
4. Each Scope gets its own table.
5. this driver makes use of `json_set` and `json_type` JSON1 functions to only write the necessary data, instead of writting the scope file (unlike the JSON driver)

Not allowing me to reopen #3934 so making a new one


Drapersniper:
Lets not use APSW for this due to this blocking potential - Remake using an async sql backend.

PR #4326, opened 2020-08-20
Accept `discord.Object`s for cases, add `last_known_username` param in `create_case()`

TrustyJAID:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

`create_case` has missmatching types for user where it should match the Case object typehints.

### DIFF
--- a/redbot/core/modlog.py
+++ b/redbot/core/modlog.py
@@ -240,13 +240,13 @@ class Case:
         guild: discord.Guild,
         created_at: int,
         action_type: str,
-        user: Union[discord.abc.User, int],
-        moderator: Optional[Union[discord.abc.User, int]],
+        user: Union[discord.Object, discord.abc.User, int],
+        moderator: Optional[Union[discord.Object, discord.abc.User, int]],
         case_number: int,
         reason: str = None,
         until: int = None,
         channel: Optional[Union[discord.TextChannel, discord.VoiceChannel, int]] = None,
-        amended_by: Optional[Union[discord.abc.User, int]] = None,
+        amended_by: Optional[Union[discord.Object, discord.abc.User, int]] = None,
         modified_at: Optional[int] = None,
         message: Optional[discord.Message] = None,
         last_known_username: Optional[str] = None,
@@ -286,9 +286,12 @@ class Case:
         data.pop("case_number", None)
         # last username is set based on passed user object
         data.pop("last_known_username", None)
-
         for item in list(data.keys()):
-            setattr(self, item, data[item])
+            if isinstance(item, discord.Object):
+                # probably expensive to call but meh should capture all cases
+                setattr(self, item, data[item].id)
+            else:
+                setattr(self, item, data[item])
 
         # update last known username
         if not isinstance(self.user, int):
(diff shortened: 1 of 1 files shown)


mikeshardmind:
Repeating what I said in the linked pull for context: modlog should probably accept `discord.Object` (and pull out the id and use it as it already accepts the id) since it's already doing type detection and it's possible for things like message authors to be a`discord.Object` in some rare cases (this is documented in discord.py) so if moderation happens based on one of these cases, it will fail to create a case.

TrustyJAID:
I have changed this to accept discord.Object objects as a potential type and adjusted the logic for what to do when dealing with an Object rather than an int or User object. I also added an extra optional arg for `create_case` to allow for users to provide the `last_known_username` when creating cases since it was easy to add.

Jackenmen:
This is modifying current API of the `Case ` by breaking existent expectations where non-cached users are presented as an `int`.
I would suggest to instead just cast `discord.Object` to `int` if that's what's passed to `user`, `moderator`, or `amended_by`.

Jackenmen (changes requested):
I should have done the comment above as a "request changes" review since PR is in current state a breaking change. Fixing that with this review.

TrustyJAID:
@jack1142 this does not remove the option to still provide an int, just extends that it may be an int, a fake user object with missing attributes, or a valid user object.

Jackenmen:
@TrustyJAID I meant that the resulting `Case` object may now be of `discord.Object` type which changes user's expectations. While it's not that big of a deal considering this will only be returned like this from `modlog.create_case()` (so developer would need to explicitly pass `discord.Object` to `create_case()`) and using `modlog.get_case()` will still return a `Case` object with an `int`, it does break consistency.

And on top of that, this imo complicates the logic unnecessarily when it's possible to just cast it to `int` in `Case`'s `__init__`.

Jackenmen (dismissed):
Type hints don't have `discord.Object` as supported type, and the types in docstrings are still inaccurate. Also, `Case.edit()` still does not support `discord.Object` with this PR which would be good to fix for consistency.

redbot/core/modlog.py
```diff
@@ -821,16 +829,18 @@ async def create_case(
...
+    moderator: Optional[Union[discord.Object, discord.abc.User, int]]
         The moderator who took the action
     reason: Optional[str]
         The reason the action was taken
     until: Optional[datetime]
         The time the action is in effect until
     channel: Optional[discord.TextChannel]
         The channel the action was taken in
+    last_known_username: Optional[str]
+        The last known username of the user
```
Jackenmen, line 843: Might be worth noting that this will be ignored, if passed `user` parameter implements `discord.abc.User`.

Jackenmen (changes requested):
Looks good to me, I just have one comment about documentation improvement for the new `last_known_username` that would be nice to address.

redbot/core/modlog.py
```diff
@@ -280,9 +286,12 @@ async def edit(self, data: dict):
...
         # last username is set based on passed user object
         data.pop("last_known_username", None)
-
         for item in list(data.keys()):
-            setattr(self, item, data[item])
+            if isinstance(item, discord.Object):
+                # probably expensive to call but meh should capture all cases
+                setattr(self, item, data[item].id)
+            else:
+                setattr(self, item, data[item])
```
Jackenmen:
```suggestion
            value = data[item]
            if isinstance(value, discord.Object):
                # probably expensive to call but meh should capture all cases
                setattr(self, item, value.id)
            else:
                setattr(self, item, value)
```

Jackenmen (changes requested):
Looks like there's a bug here I missed

Jackenmen (approved):
Looks good to me, thanks!

PR #4330, opened 2020-08-21
[Mod] Config setting for days of messages to delete when banning for mention spam

Flame442:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Instead of using the dpy default of 1 day, use the setting configured for Mod.


Twentysix26:
I have mixed feelings about this change: for example in my use case I like default [p]ban days at 0, since I believe that preserving history is important unless it contains NSFW or similar undesirable content. However, for something like mention spam I wouldn't care as much, as it's usually done by raiders and cleaning the mess up in this case is good (also due to multiple people being pinged).
Those two different mechanics using the same setting is problematic, since it would not allow me to customize it how I want. Perhaps a separate setting would be better?

Flame442:
This PR has been updated to use a new config value for the number of days worth of messages to delete when banning for mention spam rather than piggybacking on the existing `delete_days` setting.

Jackenmen:
This should probably support the new `delete_message_seconds` rather than just the day count *but* the specifics of how this should be approached depend at least partially on #5922 so I think this PR should be held off until a resolution for that issue is decided.

PR #4364, opened 2020-08-24
Added length validation for [p]set name and [p]set nickname

### LINKED ISSUE #4363: [Core] Set name and nickname should handle length errors in a user friendly way
# Command bugs
#### Command name
`[p]set name` and `[p]set nickname`
Core
The bot to friendly say the name was too long
#### What actually happened?
`[p]set name` sent "Failed to change name. Remember that you can only do it up to 2 times an hour.", which is an unrelated error message
`[p]set nickname` fully errored
Use the commands above with a really long name

hisztendahl:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Return error message if [p]set name or [p]set nickname string is too long

Fixes #4363

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1919,7 +1919,12 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
     async def _username(self, ctx: commands.Context, *, username: str):
         """Sets [botname]'s username."""
         try:
-            await self._name(name=username)
+            if len(username) > 32:
+                await ctx.send(_("Failed to change name. Must be 32 characters or fewer."))
+                return
+            else:
+                print("USERNAME LENGTH OK")
+                await self._name(name=username)
         except discord.HTTPException:
             await ctx.send(
                 _(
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -1919,7 +1919,12 @@ async def stream(self, ctx: commands.Context, streamer=None, *, stream_title=Non
     async def _username(self, ctx: commands.Context, *, username: str):
         """Sets [botname]'s username."""
         try:
-            await self._name(name=username)
+            if len(username) > 32:
+                await ctx.send(_("Failed to change name. Must be 32 characters or fewer."))
+                return
+            else:
+                print("USERNAME LENGTH OK")
```
TrustyJAID: We don't use `print` statements for this. So this should either be removed or changed to `log.debug` if you want to keep it although I don't really see a need for it beyond testing the function worked.

TrustyJAID (changes requested):

redbot/core/core_commands.py
hisztendahl: Woops good catch, forgot to remove it

TrustyJAID (dismissed):
Looks good to me!

Dav-Git (dismissed):

NeuroAssassin (dismissed):
Thanks a lot!

madebylydia (changes requested):
Can you remove the else statement? They're not a real necessity in here.

madebylydia (approved):

Drapersniper (approved):
Reapproved Trusty's approval after the change Predeactor suggested.

PR #4366, opened 2020-08-25
[Dev] Add repl pause

NeuroAssassin:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Adds a command to pause the REPL in the current channel, so that you can peacefully write codeblocks without your bot sending an error, and yet retain the variables of the REPL instead of having to restart.

### DIFF
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -238,13 +238,24 @@ class Dev(commands.Cog):
         }
 
         if ctx.channel.id in self.sessions:
-            await ctx.send(
-                _("Already running a REPL session in this channel. Exit it with `quit`.")
-            )
+            if self.sessions[ctx.channel.id]:
+                await ctx.send(
+                    _("Already running a REPL session in this channel. Exit it with `quit`.")
+                )
+            else:
+                await ctx.send(
+                    _(
+                        "Already running a REPL session in this channel. Resume the REPL with `{}repl resume`."
+                    ).format(ctx.prefix)
+                )
             return
 
         self.sessions[ctx.channel.id] = True
-        await ctx.send(_("Enter code to execute or evaluate. `exit()` or `quit` to exit."))
+        await ctx.send(
+            _(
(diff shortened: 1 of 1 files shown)


Dav-Git (dismissed):
I don't quite get the usecase but looks good and works.

redbot/core/dev_commands.py
```diff
@@ -305,6 +309,22 @@ async def repl(self, ctx):
             except discord.HTTPException as e:
                 await ctx.send(_("Unexpected error: `{}`").format(e))
 
+    @repl.command()
+    async def pause(self, ctx, toggle: Optional[bool] = None):
+        """Pauses the REPL running in the current channel"""
+        if not ctx.channel.id in self.sessions:
```
Jackenmen:
```suggestion
        if ctx.channel.id not in self.sessions:
```

redbot/core/dev_commands.py
```diff
@@ -305,6 +309,22 @@ async def repl(self, ctx):
             except discord.HTTPException as e:
                 await ctx.send(_("Unexpected error: `{}`").format(e))
 
+    @repl.command()
+    async def pause(self, ctx, toggle: Optional[bool] = None):
```
Jackenmen, line 324: We shouldn't be using the behavior of d.py's `Optional` converter here. You can either remove the `Optional` or have the proper type hint using `NoParseOptional` from `redbot.core.commands` (`from redbot.core.commands import NoParseOptional as Optional`)

redbot/core/dev_commands.py
```diff
@@ -242,17 +243,20 @@ async def repl(self, ctx):
             )
```
Jackenmen: Maybe you could mention that you can "unpause" the REPL using `[p]repl pause`, if it's paused, so that it's less confusing to the user why they can't use it even though it's running.

Jackenmen (changes requested):
Mostly good, seems like a useful addition (if only I used REPL :P)

Jackenmen (dismissed):
LGTM

Jackenmen (approved):

PR #4369, opened 2020-08-26
Cleanup reactions

phenom4n4n:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Adds a `[p]cleanup reactions <number>` command which searches the messages provided and clears reactions from them if there are any.

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -586,7 +586,7 @@ class Cleanup(commands.Cog):
         """Clear reactions from the specified number of messages."""
         msgs = 0
 
-        async for message in ctx. channel.history(limit=positive_int):
+        async for message in ctx.channel.history(limit=number):
             if message.reactions:
                 await message.clear_reactions()
                 msgs += 1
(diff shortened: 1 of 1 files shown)


TurnrDev (commented):
Surely `str.format()` is better than `%-formatting`? Certainly a lot more readable.

Zephyrkul:
Python's logging module uses %-formatting.

phenom4n4n:
^ this logging is based off how the rest of the cog logs.

TurnrDev:
> Python's logging module uses %-formatting.

> ^ this logging is based off how the rest of the cog logs.

Oh, ignore me then :+1:

Vexed01:
`%-formatting` is what `cleanup_spam` uses. The rest of the cleanup cog sets a `reason` var with `str.format()`, then does `log.info(reason)`.

Drapersniper:
Okay so, this is gonna be extremely API intensive... to the point you could start hitting the global ratelimits if you pass large numbers.

One solution is to enforce it to clear reaction only on the last 100 messages, this of course would severally restrict usability of this command.

phenom4n4n:
Another solution may be to limit the number of messages searched in channel history to 300, and limit the number of messages that are actually cleared to 100. If it reaches than 100 messages, the loop will break. Then put a 5s channel cooldown on the command. 
These numbers are examples, I'm not sure which numbers are best for use in production.

phenom4n4n:
New commit reflects the idea proposed here:
> Another solution may be to limit the number of messages searched in channel history to 300, and limit the number of messages that are actually cleared to 100. If it reaches than 100 messages, the loop will break. Then put a 5s channel cooldown on the command.

PR #4370, opened 2020-08-26
[Core] add `[p]set api list` and `[p]set api remove`

Vuks69:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

Added new subcommands: `[p]set api list` and `[p]set api remove`.
`[p]set api list` causes the bot to send a message containing all currently set API services, without exposing tokens. The message is formatted similarly to `[p]repo list`.

`[p]set api remove` removes the provided services from config, including their associated tokens.

Should be useful for managing your API keys along with `[p]set api` and especially when removing api keys, since there were no easy-to-use tools for that.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1057,6 +1057,25 @@ class RedBase(
             for name in token_names:
                 group.pop(name, None)
 
+    async def remove_shared_api_services(self, *service_names: str):
+        """
+        Removes shared API services, as well as keys and tokens associated with them.
+
+        Parameters
+        ----------
+        service_names: str
+            the services to remove
+
+        Examples
+        ----------
+        Removing the youtube service
+
+        >>> await ctx.bot.remove_shared_api_services("youtube")
+        """
+        async with self._config.custom(SHARED_API_TOKENS).all() as group:
+            for service in service_names:
+                group.clear_raw(service)
+
     async def get_context(self, message, *, cls=commands.Context):
         return await super().get_context(message, cls=cls)
 
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -986,9 +986,9 @@ class RedBase(
         """
         return await self._config.guild(discord.Object(id=guild_id)).mod_role()
 
-    async def get_shared_api_keys(self) -> List[str]:
+    async def get_shared_api_services(self) -> List[str]:
         """
-        Gets the shared API keys (services).
+        Gets the shared API service names.
         """
         return list((await self._config.custom(SHARED_API_TOKENS).all()).keys())
 
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -26,6 +26,7 @@ from typing import (
     Any,
     Literal,
     MutableMapping,
+    overload,
 )
 from types import MappingProxyType
 
@@ -986,20 +987,32 @@ class RedBase(
         """
         return await self._config.guild(discord.Object(id=guild_id)).mod_role()
 
-    async def get_shared_api_tokens(self, service_name: Optional[str] = None) -> Dict[str, str]:
+    @overload
+    async def get_shared_api_tokens(self, service_name: str = ...) -> Dict[str, str]:
+        ...
+
+    @overload
+    async def get_shared_api_tokens(self, service_name: None = ...) -> Dict[str, Dict[str, str]]:
+        ...
+
+    async def get_shared_api_tokens(
+        self, service_name: Optional[str] = None
+    ) -> Union[Dict[str, Dict[str, str]], Dict[str, str]]:
         """
         Gets the shared API tokens for a service, or all of them if no argument specified.
 
         Parameters
         ----------
-        service_name: str
-            The service to get tokens for.
+        service_name: str, optional
+            The service to get tokens for. Leave empty to get tokens for all services.
 
         Returns
         -------
-        Dict[str, str]
+        Dict[str, Dict[str, str]] or Dict[str, str]
             A Mapping of token names to tokens.
             This mapping exists because some services have multiple tokens.
+            If ``service_name`` is `None`, this method will return
+            a mapping with mappings for all services.
         """
         if service_name is None:
             return await self._config.custom(SHARED_API_TOKENS).all()
@@ -1060,7 +1073,7 @@ class RedBase(
 
         Parameters
         ----------
-        service_names: str
+        *service_names: str
             The services to remove.
 
         Examples
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -26,6 +26,7 @@ from typing import (
     Any,
     Literal,
     MutableMapping,
+    Iterable,
 )
 from types import MappingProxyType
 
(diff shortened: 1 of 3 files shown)


redbot/core/core_commands.py
```diff
@@ -2059,7 +2060,7 @@ async def custominfo(self, ctx: commands.Context, *, text: str = None):
         else:
             await ctx.send(_("Text must be fewer than 1024 characters long."))
 
-    @_set.command()
+    @_set.group(invoke_without_command=True)
     @checks.is_owner()
     async def api(self, ctx: commands.Context, service: str, *, tokens: TokenConverter):
         """Set various external API tokens.
```
Drapersniper: There is no mention subcomments of `[p]help set`, I would reccomend this command is broken up into 3 subcommands.

redbot/core/core_commands.py
```diff
@@ -2077,6 +2078,32 @@ async def api(self, ctx: commands.Context, service: str, *, tokens: TokenConvert
...
+
+        This command does not expose keys or tokens associated with services,
+        but may expose them if tokens have been incorrectly set as services."""
+
+        services = await ctx.bot.get_shared_api_services()
+        if not services:
+            await ctx.send(_("No API services have been set yet."))
+            return
+
+        sorted_services = sorted(services, key=lambda r: str.lower(r))
```
```suggestion
        sorted_services = sorted(services, key=str.lower)
```

redbot/core/core_commands.py
```diff
@@ -2077,6 +2078,32 @@ async def api(self, ctx: commands.Context, service: str, *, tokens: TokenConvert
...
+        joined = _("Set API services:\n")
+        for key in sorted_services:
+            joined += "{}\n".format(key)
+        for page in pagify(joined, ["\n"], shorten_by=16):
+            await ctx.send(box(page.lstrip(" "), lang="diff"))
```
```suggestion
            joined += "+ {}\n".format(key)
        for page in pagify(joined, ["\n"], shorten_by=16):
            await ctx.send(box(page.lstrip(" "), lang="diff"))
```
So that it applies some syntax highlighting
```diff
Set API services:
+ test
+ test2
```

redbot/core/core_commands.py
```diff
@@ -2077,6 +2078,32 @@ async def api(self, ctx: commands.Context, service: str, *, tokens: TokenConvert
...
+
+    @api.command(name="remove")
+    async def api_remove(self, ctx: commands.Context, *services: str):
+        """Remove the given services with all their keys and tokens."""
+        services = list(services)
```
```suggestion
```
This is not needed

Drapersniper (dismissed):
Pausing the review as @Vuks69  said he needs to look into something else.

redbot/core/core_commands.py
Jackenmen: I think it would be better to just mention in `[p]set api`'s docstring that it's also for listing and deleting (in whatever way that would be worded) if there's an issue with clarity, so that all the commands related to API tokens are still grouped together.

Vuks69:
From my side the PR is now ready for review.

redbot/core/core_commands.py
Vuks69: I've updated the `[p]set api` docstring to reflect the new subcommands.

redbot/core/core_commands.py
```diff
@@ -2077,6 +2078,31 @@ async def api(self, ctx: commands.Context, service: str, *, tokens: TokenConvert
         await ctx.bot.set_shared_api_tokens(service, **tokens)
         await ctx.send(_("`{service}` API tokens have been set.").format(service=service))
 
+    @api.command(name="list")
+    async def api_list(self, ctx: commands.Context):
+        """Show all external API services that have been set.
+
+        This command does not expose keys or tokens associated with services,
```
Jackenmen: Hmm, it would (IMO) be useful to also know the key names that were set for the API services (i.e. `twitch: client_id, client_secret`), but I'm not gonna request it here - it's up to you whether you want to include it.

redbot/core/bot.py
```diff
@@ -1051,6 +1058,25 @@ async def remove_shared_api_tokens(self, service_name: str, *token_names: str):
...
                 group.pop(name, None)
 
+    async def remove_shared_api_services(self, service_names: Iterable[str]):
+        """
+        Removes shared API services, as well as keys and tokens associated with them.
+
+        Parameters
+        ----------
+        service_names: str
+            the services to remove
```
```suggestion
            The services to remove.
```

redbot/core/core_commands.py
```diff
@@ -2077,6 +2078,31 @@ async def api(self, ctx: commands.Context, service: str, *, tokens: TokenConvert
...
+        but may expose them if tokens have been incorrectly set as services."""
+
+        services = await ctx.bot.get_shared_api_services()
+        if not services:
+            await ctx.send(_("No API services have been set yet."))
+            return
+
+        sorted_services = sorted(services, key=str.lower)
+        joined = _("Set API services:\n")
```
Jackenmen: Plural support:
```suggestion
        joined = _("Set API services:\n") if len(services) > 1 else _("Set API service:\n")
```

Jackenmen (changes requested):
Aside from review comments, you also need to add the new methods (`Red.remove_shared_api_services`, `Red.get_shared_api_services`) at the end of `docs/framework_apikeys.rst` file. It should be easy to figure out from other methods there how you need to add the new ones but do let me know if you run into any issues with that or anything else.

redbot/core/core_commands.py
Vuks69: I will definitely look into that, but at a later date.

Jackenmen: Based on a conversation in Discord, this will be made as part of this PR due to needed changes in the API methods that are added by this PR.

redbot/core/bot.py
```diff
@@ -1051,6 +1058,25 @@ async def remove_shared_api_tokens(self, service_name: str, *token_names: str):
             for name in token_names:
                 group.pop(name, None)
 
+    async def remove_shared_api_services(self, service_names: Iterable[str]):
```
Jackenmen: I've just noticed that unlike `remove_shared_api_tokens()`, this method is using an iterable as a parameter. I think it would be better for consistency if this did it similarly to the aforementioned method:
```suggestion
    async def remove_shared_api_services(self, *service_names: str):
```

// I'm just posting the code suggestion for the method signature; the method usage and its docstring also need to be updated and I haven't made the code suggestion for that.

Sorry for missing this earlier :(

Jackenmen (approved):
Looks good to me after doing few changes.

@Vuks69 I said that I'm gonna take care of the rest here, but since you might want to look at what I've done, I put a full list at the end of this and I'm gonna wait with merging this until tomorrow.

---

**Improve readability of `[p]set api list`**

It's a lot clearer when the sublists are indented,
especially on mobile where code blocks aren't colored at all.

New look:
https://user-images.githubusercontent.com/6032823/94614944-3bbd7c80-02a7-11eb-89e4-64bdf06c650b.png

---

**Improve error message**

Updated message to be more specific
(used phrases: "None of the services" and "had any keys set")
and a bit more nice to the user (judging word "anyway" removed)

---

**Add plural support**

---

**Use `.format()` after translation is applied**

If `.format()` is used before `_()`, the search for translation is done
with the string that already has the dynamic text added,
which results in no matches.

---

**Use `humanize_list()` rather than `str.join()`**

This is done for consistency within Red
and it makes the list have the last element joined
with `and` (or its equivalent in chosen locale)

---

**Improve docstrings and type hints**

- added overloads to help out developers a bit more
  * this wasn't necessary, but development tools work better
    with this information
- fixed type hint to use Union as now
  its return type depends on whether `service_name` is passed
- updated docstrings to contain information about the added behavior

PR #4371, opened 2020-08-26
Bump black to 20.8b1 (and reformat)

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature
- [x] Dependency update

Bumps black version and reformats with it. The biggest fix (but also style/behavior change) in Black 20.8b is proper handling of "magic trailing comma", which you can notice in the diffs.

So Black will now explode collections and any bracket pair **if** a trailing comma is present in it.
Since previous version had some issues with how it handled this, you can see in the diffs that there are places where Black now explodes the collection/bracket pair. There were few places where exploding didn't make sense so I removed the wrongly placed commas there.

As for other changes, Black now formats docstrings and handles some multiline things a bit differently which is what rest of the things in the diff are about.

// It might be worth to wait few days with this as it's very much possible that Black will need some hotfix releases.
There's not that much of reformatting and I don't see anything that could make existing PRs conflict so there's probably no need to postpone this more than aforementioned few days.

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -319,7 +319,7 @@ def handle_edit(cli_flags: Namespace):
 async def run_bot(red: Red, cli_flags: Namespace) -> None:
     """
     This runs the bot.
-    
+
     Any shutdown which is a result of not being able to log in needs to raise
     a SystemExit exception.
 
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -76,7 +76,8 @@ class Admin(commands.Cog):
         self.config.register_global(serverlocked=False, schema_version=0)
 
         self.config.register_guild(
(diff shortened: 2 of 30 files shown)


Drapersniper (approved):
LGTM, the mypy-extension dep is for future plans?

Jackenmen:
> LGTM, the mypy-extension dep is for future plans?

No, it's part of Black's dependencies and we currently include both primary dependencies as well as dependencies that come from primary dependencies.

Dav-Git:
I'm guessing this will update itself with ``make syncenv`` (for dev) and ``pip install`` on my machine (or any one else deving with red)on next release?

Jackenmen:
@Dav-Git yep, [commit `0ff83dd`](https://github.com/Cog-Creators/Red-DiscordBot/pull/4371/commits/0ff83dd905e921b2577c9bd4cd3d4c443efa00b8) handles that

PR #4375, opened 2020-08-27
JSON schemas for cogs and repos

fixator10:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Add an JSON schemas for `info.json`s (cogs and repos)


schema/red_cog.schema.json
```diff
@@ -0,0 +1,97 @@
...
+      ]
+    }
+  },
+  "definitions": {
+    "required_cog": {
+      "type": "object",
+      "patternProperties": {
+        ".*": {
+          "type": "string",
+          "format": "uri"
```
fixator10, line 92: Should be right, but not tested, since most of validators seem to be implement this as "annotation", or doesn't implement at all.
[7.2. Implementation Requirements](https://json-schema.org/draft/2019-09/json-schema-validation.html#rfc.section.7.2)

schema/red_cog.schema.json
```diff
@@ -0,0 +1,97 @@
...
+      "description": "A short description of the cog or repo. For cogs, this info is displayed when a user executes [p]cog list"
+    },
+    "end_user_data_statement": {
+      "type": "string",
+      "description": "A statement explaining what end user data the cog is storing. This is displayed when a user executes [p]cog info. If the statement has changed since last update, user will be informed during the update."
+    },
+    "min_bot_version": {
+      "type": "string",
+      "description": "Min version number of Red in the format MAJOR.MINOR.MICRO",
+      "pattern": "\\d+\\.\\d+\\.\\d+"
```
Jackenmen: Here's the pattern that should be used to properly match any supported version string (I haven't escaped it for JSON):
```
^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)((a|b|rc)(0|[1-9][0-9]*))?(\.post(0|[1-9][0-9]*))?(\.dev(0|[1-9][0-9]*))?$
```

schema/red_cog.schema.json
```diff
@@ -0,0 +1,97 @@
...
+    },
+    "max_bot_version": {
+      "type": "string",
+      "description": "Max version number of Red in the format MAJOR.MINOR.MICRO, if min_bot_version is newer than max_bot_version, max_bot_version will be ignored",
+      "pattern": "\\d+\\.\\d+\\.\\d+"
```
Jackenmen: Here's the pattern that should be used to properly match any supported version string (I haven't escaped it for JSON):
```
^(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)((a|b|rc)(0|[1-9][0-9]*))?(\.post(0|[1-9][0-9]*))?(\.dev(0|[1-9][0-9]*))?$
```

schema/red_cog.schema.json
```diff
@@ -0,0 +1,97 @@
...
+  "definitions": {
+    "required_cog": {
+      "type": "object",
+      "patternProperties": {
+        ".*": {
+          "type": "string",
+          "format": "uri"
+        }
+      },
+      "additionalProperties": false
```
Jackenmen, line 95: I think this is unneeded; we don't reject info.json files if they have some keys we don't support.

schema/red_cog.schema.json
```diff
@@ -0,0 +1,97 @@
...
+        "COG",
+        "SHARED_LIBRARY"
+      ]
+    }
+  },
+  "definitions": {
+    "required_cog": {
+      "type": "object",
+      "patternProperties": {
+        ".*": {
```
Jackenmen: Cog name can't be an empty string. We could probably also have more specific that only matches if cog name is valid Python identifier (`^[a-zA-Z_][a-zA-Z0-9_]*$`), but I don't think it's really that important, so just removing `*` should work:
```suggestion
        ".": {
```

schema/red_cog.schema.json
```diff
@@ -0,0 +1,97 @@
...
+    },
+    "requirements": {
+      "type": "array",
+      "description": "List of required libraries that are passed to pip on cog install.",
+      "items": {
+        "type": "string"
+      }
+    },
+    "tags": {
+      "type": "array",
```
Jackenmen, line 70: Could we make the items in the "tags" array unique using [`uniqueItems`](https://tools.ietf.org/html/draft-handrews-json-schema-validation-01#section-6.4.5)?

Jackenmen (changes requested):
Haven't tested this yet, but seems mostly good by looking at the schemas.

schema/red_cog.schema.json
fixator10: I changed it to `.+`, because it seems more optimal
image
image

schema/red_cog.schema.json
```diff
@@ -0,0 +1,97 @@
...
+      "type": "string",
+      "description": "Optional, defaults to COG. Must be either COG or SHARED_LIBRARY. If SHARED_LIBRARY then hidden will be True.",
+      "enum": [
+        "COG",
+        "SHARED_LIBRARY"
+      ]
+    }
+  },
+  "definitions": {
+    "required_cog": {
```
Jackenmen, line 87: Soooo, I'm dumb - I thought that you used `additionalProperties` on the whole schema, not just the `required_cog`... You did it right the first time and my review comment saying you should remove it was wrong.

schema/red_cog.schema.json
```diff
@@ -0,0 +1,97 @@
+{
+  "$id": "https://raw.githubusercontent.com/Cog-Creators/Red-DiscordBot/V3/develop/schema/red_cog.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Red-DiscordBot Сog",
```
Jackenmen:
```suggestion
  "title": "Red-DiscordBot Сog metadata file",
```

schema/red_cog_repo.schema.json
```diff
@@ -0,0 +1,27 @@
+{
+  "$id": "https://raw.githubusercontent.com/Cog-Creators/Red-DiscordBot/V3/develop/schema/red_cog_repo.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Red-DiscordBot Сog Repo",
```
Jackenmen:
```suggestion
  "title": "Red-DiscordBot Сog Repo metadata file",
```

Jackenmen (changes requested):
Two small enhancements to schema titles and one review comment about me being dumb.

Everything else seems good and worked properly when I tested it in VS Code.

Jackenmen (approved):
Looks good to me. Let me know when you make a PR to schema store (or if you want me to do it).

Jackenmen:
PR to include theses schemas in SchemaStore is here: https://github.com/SchemaStore/schemastore/pull/1226

PR #4372, opened 2020-08-26
Expose public methods to internal apis.

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

##Note
If you currently feel like you would benefit from a public method let me know and ill look into adding it to this PR.

### Description of the changes
- Add some methods to interact with the blocklist/allowlist.
- Restore the old `bot.counter` property with several safeguards

PR #4376, opened 2020-08-28
Ensure Bank API only works with Integers.

xBlynd:
### Type

- [X ] Bugfix
- [ ] Enhancement
- [ ] New feature

Per PR #4372  from @Drapersniper in Testing for Red Discord. https://discord.com/channels/133049272517001216/171665724262055936/748529319759970325
Change bank.set_balance amount typehint to float and cast to int later.
https://github.com/Cog-Creators/Red-DiscordBot/pull/4372/commits/6199653670ed25387da51ff6b9da5cc29535e9f1#diff-88a9ebbf67d6343dee1a7d2255998245R367

### DIFF
--- a/redbot/core/bank.py
+++ b/redbot/core/bank.py
@@ -203,16 +203,16 @@ async def can_spend(member: discord.Member, amount: int) -> bool:
     return await get_balance(member) >= amount
 
 
-async def set_balance(member: Union[discord.Member, discord.User], amount: float) -> int:
+async def set_balance(member: Union[discord.Member, discord.User], amount: Union[int, float]) -> int:
     """Set an account balance.
 
     Parameters
     ----------
     member : Union[discord.Member, discord.User]
         The member whose balance to set.
-    amount : int
-        The amount to set the balance to.
-
+    amount : Union[int, float]
+        The amount to set the balance to. [This gets converted to an integer]
+    
     Returns
     -------
     int
--- a/redbot/core/bank.py
+++ b/redbot/core/bank.py
@@ -203,7 +203,7 @@ async def can_spend(member: discord.Member, amount: int) -> bool:
     return await get_balance(member) >= amount
 
 
-async def set_balance(member: Union[discord.Member, discord.User], amount: int) -> int:
+async def set_balance(member: Union[discord.Member, discord.User], amount: float) -> int:
     """Set an account balance.
 
     Parameters
@@ -229,6 +229,7 @@ async def set_balance(member: Union[discord.Member, discord.User], amount: int) 
         ``bank._MAX_BALANCE``.
 
     """
+    amount = int(amount)
     if amount < 0:
         raise ValueError("Not allowed to have negative balance.")
     guild = getattr(member, "guild", None)
--- a/redbot/core/bank.py
+++ b/redbot/core/bank.py
@@ -191,6 +191,11 @@ async def can_spend(member: discord.Member, amount: int) -> bool:
     amount : int
         The amount the member wants to spend.
 
+    Raises
+    ------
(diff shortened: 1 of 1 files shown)


Zephyrkul (changes requested):
This function doesn't accept floats, and in fact forces `amount` to be an int, so the typehint should remain as int. The point of typehints in Python is to show what the function is intended to take, not what all it could possibly take. d.py does some extra things with typehints but that only affects commands, not functions.

redbot/core/bank.py
```diff
@@ -229,6 +229,7 @@ async def set_balance(member: Union[discord.Member, discord.User], amount: int)
         ``bank._MAX_BALANCE``.
 
     """
+    amount = int(amount)
```
```suggestion
    if not isinstance(amount, int):
        raise TypeError("Amount must be of type int, not {}.".format(type(amount)))
```

Drapersniper: Can you please also apply this isinstance check to 


Line 847 in the `set_default_balance` function
replace
```py
amount = int(amount)
```
with
```py
if not isinstance(amount, int):
    raise TypeError("Amount must be of type int, not {}.".format(type(amount)))
```

Could you please also apply this to `set_max_balance` and `can_spend` 

This will ensure all bank methods only work with integers.

redbot/core/bank.py
```diff
@@ -203,16 +203,16 @@ async def can_spend(member: discord.Member, amount: int) -> bool:
...
     Parameters
     ----------
     member : Union[discord.Member, discord.User]
         The member whose balance to set.
-    amount : int
-        The amount to set the balance to.
-
+    amount : Union[int, float]
+        The amount to set the balance to. [This gets converted to an integer]
+
```
```suggestion
    amount : int
        The amount to set the balance to.
```

redbot/core/bank.py
```diff
@@ -203,16 +203,16 @@ async def can_spend(member: discord.Member, amount: int) -> bool:
     return await get_balance(member) >= amount
 
 
-async def set_balance(member: Union[discord.Member, discord.User], amount: int) -> int:
+async def set_balance(member: Union[discord.Member, discord.User], amount: Union[int, float]) -> int:
```
```suggestion
async def set_balance(member: Union[discord.Member, discord.User], amount: int) -> int:
```

Drapersniper (changes requested):
Hey @xBlynd,

Let me develop on what Zeph said here, for the intent of this PR how it is right now using the `Union[int, float]` typehint here is the correct behaviour, as it indicates that this function will accept both int and floats, the fact that it is cast to int inside the function does not affect the typehint.

After some internal discussion we decided that the best approach here will be to restrict the function from accepting floats like every other function in bank.


Also please make sure to run `make.bat reformat` or `make reformat` on your local copy of the repo before commiting this will ensure that style checks passes.

Drapersniper:
Blynd this comment is not for you, its just so its not forgotten, once this PR is merged we need to also look into making a data migration PR for bank, to ensure there are no floats currently saved in config.

xBlynd:
Enforced type checking on bank functions by checking the type when the function is called. Also updated documentation in each function to correspond to the changes.

redbot/core/bank.py
```diff
@@ -905,6 +923,8 @@ def cost(amount: int):
 
     Other exceptions will propagate and will be handled by Red's (and/or
     any other configured) error handling.
+
+    TODO: Add documentation for input/output/exceptions
     """
```
```suggestion
    """
    # TODO: Add documentation for input/output/exceptions
```

Since this will show on the docs, lets not add it to docstrings, add a comment to the code itself.

Drapersniper (approved):
LGTM, thanks for this contribution!

PR #4377, opened 2020-08-28
[Mod] Dots and properly capitalise Discord

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

All da dots for kickin' 'n' banin' as well as dem capital Ds.

Oh, also I haven't tested this on a live bot.
smirk

### DIFF
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -172,7 +172,7 @@ class KickBanMixin(MixinMeta):
                                 guild_tempbans.remove(uid)
             await asyncio.sleep(60)
 
-    @commands.command
+    @commands.command()
     @commands.guild_only()
     @commands.bot_has_permissions(kick_members=True)
     @checks.admin_or_permissions(kick_members=True)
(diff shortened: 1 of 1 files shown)


redbot/cogs/mod/kickban.py
```diff
@@ -478,7 +478,7 @@ async def tempban(
                 server_name=guild.name, date=unban_time.strftime("%m-%d-%Y %H:%M:%S")
             )
             if invite:
-                msg += _(" Here is an invite for when your ban expires: {invite_link}").format(
+                msg += _("Here is an invite for when your ban expires: {invite_link}.").format(
```
Drapersniper: The space here is needed otherwise there will be no space between this string and `"You have been temporarily banned from {server_name} until {date}."`

Also because discord is weird ... I wouldn't add the dot here, or if you do i would add `<>` around the invite link.  because on some clients I've seen discord trying to add following symbols to the URL.

Drapersniper (changes requested):

redbot/cogs/mod/kickban.py
Vexed01: Ah yes, reading the code does usually help.

So after some [#advanced-testing](https://discordapp.com/channels/133049272517001216/171665724262055936/748850189426491423) it does seem like invites are super super picky. I'll just leave those lines as they are.

Drapersniper (approved):
LGTM, thanks for yet another contrib Vexed

PR #4385, opened 2020-08-31
Adds a small util to validate data.

Drapersniper:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Adds a small util that can be used to validate data agaisnt stricted restrictions, which will be enforced in Config2.0:tm:

Using this before saving something to config lets you know if what you are saving can be safely used with the new Config class, one of example of this, is in the Redis Driver numbers outside if 64bit integers are not supported, and saving such a number would fail when trying to migrate the data to redis, from the existing drivers.


redbot/core/utils/__init__.py
```diff
@@ -498,3 +505,19 @@ def map(self, func: Callable[[_T], Union[_S, Awaitable[_S]]]) -> AsyncIter[_S]:
...
+
+    Parameters
+    ----------
+    value: Any
+        The object to be checked.
+
+    Returns
+    -------
+    bool
+        Whether or not ``value`` respect the stricter data boundaries.
```
Stonedestroyer, line 521: whatever

redbot/core/utils/__init__.py
```diff
@@ -498,3 +505,19 @@ def map(self, func: Callable[[_T], Union[_S, Awaitable[_S]]]) -> AsyncIter[_S]:
             raise TypeError("Mapping must be a callable.")
         self._map = func
         return self
+
+
+def is_safe_for_strict_config(value: Any) -> bool:
+    """Checks ``value`` agaisn't a stricter ruleset which will be enforced in a future iteration on ``Config``.
```
Stonedestroyer, line 511: against

Stonedestroyer:
```suggestion
    """Checks ``value`` against a stricter ruleset which will be enforced in a future iteration on ``Config``.
```

Stonedestroyer (changes requested):

PR #4398, opened 2020-09-04
[Dev] Automatically import modules

Zephyrkul:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

Allows Dev's commands to attempt to import modules before throwing NameErrors.

There are a couple minor, unrelated changes that are in this PR; feel free to let me know if you'd rather theses be in a separate PR.
- Docstrings have been updated with the fact that `asyncio` and `aiohttp` are included in the environment (this was missed in #3508)
- repl will now take in _last_result when first called for easily continuing from a debug / eval statement. It does not set _last_result.

### DIFF
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -68,9 +68,12 @@ class EnvImporter(dict):
     async def send_imports(self, ctx: commands.Context):
         if not self.imported:
             return
-        message = "\n".join(f"import {key}" for key in self.imported)
+        message = [
+            "# These modules were automatically imported when running your code:",
+            *map("import {}".format, self.imported),
+        ]
         self.imported.clear()
-        for page in pagify(message, shorten_by=10):
+        for page in pagify("\n".join(message), shorten_by=10):
             await ctx.send(box(page, lang="py"))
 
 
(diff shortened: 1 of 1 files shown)


Zephyrkul:
Summarizing a short discussion I had with Flame in #testing here.
Currently, the output looks like this:
<details>
<summary>Screenshot</summary>

(Small note, if you want a different format, feel free to let me know)
Screenshot
</details>

Currently, debug and eval have aiohttp and discord already imported. Since the command replies with what it imported, I suggested not having aiohttp or discord imported by the commands and instead leave those to the autoimporter, so that the command replies with all imports (except commands, which the autoimporter can't handle by itself, and asyncio, which is required by repl).

<details>
<summary>Flame's thoughts</summary>

> idk that's a weird debate to me
cause on the one hand there's stuff we all expect to be imported by default in the background
but on the other, if some things post the import and not others, that's weird
I don't have a strong opinion one way or the other to be honest, but draper or jack might
I see advantages to both sides to be honest
</details>

Zephyrkul:
I have been asked to make this behavior off by default, and instead an option that can be enabled. As such, to fulfill that request, I will instead be publishing this as a cog in the near-ish future. Thank you for your consideration.

PR #4403, opened 2020-09-05
Add `bot.get_or_fetch_user/member` methods

### LINKED ISSUE #4402: Add an method to the bot object to allow for fetch_user to be cached
# Feature request

- [ ] Cog
- [ ] Command
- [ x ] API functionality

Add a simple method to the bot object that can be used instead of bot.fetch_user, this would cache the user object so that repeated uses on this on same id would return the cached object, this could also be expanded to anything that uses fetch (fetch channel, fetch guild, fetch message)

Drapersniper:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

add `get_or_fetch_channel`, `get_or_fetch_member` and `get_or_fetch_user` to the bot object.
closes #4402


Drapersniper:
I was considering doing one for guild also, but it would result in inconsistent behaviout as `fetch_guild` doesn't return a full guild object, but `get_guild` does... it would cause it to return more than `fetch_guild` currently does if  the guild is in the cache ... but still inconsistent, if you feel like it would be a nice to have to lmk and ill add it too

redbot/core/bot.py
```diff
@@ -665,6 +665,92 @@ async def get_embed_color(self, location: discord.abc.Messageable) -> discord.Co
...
+
+        Returns
+        --------
+        discord.Member
+            The user you requested.
+        """
+
+        if (user := guild.get_user(member_id)) is not None:
+            return user
+        return await guild.fetch_user(member_id)
```
```suggestion
        if (member := guild.get_member(member_id)) is not None:
            return member
        return await guild.fetch_member(member_id)
```

Jackenmen (approved):

PR #4406, opened 2020-09-07
Let the bot owners know what the origin of `send_to_owners` is.

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Since `send_to_owners` and its variations can be used from anywhere and by anything i can easily spam users with not very useful messages, this PR ensure that the user is always aware of where the message is originating from so that they can take appropriate action.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -50,7 +50,7 @@ from .settings_caches import (
 
 from .rpc import RPCMixin
 from .utils import common_filters, AsyncIter
-from .utils._internal_utils import send_to_owners_with_prefix_replaced
+from .utils._internal_utils import get_cog_by_call, send_to_owners_with_prefix_replaced
 
 CUSTOM_GROUPS = "CUSTOM_GROUPS"
 SHARED_API_TOKENS = "SHARED_API_TOKENS"
@@ -1401,6 +1401,8 @@ class RedBase(
 
         This logs failing sends
         """
+        cog_name = get_cog_by_call(self, "send_to_owners")
+
         destinations = await self.get_owner_notification_destinations()
 
         async def wrapped_send(location, content=None, **kwargs):
@@ -1408,12 +1410,17 @@ class RedBase(
                 await location.send(content, **kwargs)
             except Exception as _exc:
                 log.error(
-                    "I could not send an owner notification to %s (%s)",
+                    "I could not send an owner notification from cog '%s' to %s (%s)",
+                    cog_name,
                     location,
                     location.id,
                     exc_info=_exc,
                 )
 
+        content = "Message from {cog_name}\n{content}".format(
+            cog_name=cog_name, content=content if content is not None else ""
+        )
+        content = content[:1999]
         sends = [wrapped_send(d, content, **kwargs) for d in destinations]
         await asyncio.gather(*sends)
 
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1407,6 +1407,7 @@ class RedBase(
 
         async def wrapped_send(location, content=None, **kwargs):
             try:
+                await location.send(f"The following message is from {cog_name}")
                 await location.send(content, **kwargs)
             except Exception as _exc:
                 log.error(
@@ -1417,9 +1418,6 @@ class RedBase(
                     exc_info=_exc,
                 )
 
-        content = "Message from {cog_name}:\n{content}".format(
-            cog_name=cog_name, content=content if content is not None else ""
-        )
         sends = [wrapped_send(d, content, **kwargs) for d in destinations]
         await asyncio.gather(*sends)
 
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1417,7 +1417,7 @@ class RedBase(
                     exc_info=_exc,
                 )
 
-        content = "Message from {cog_name}\n{content}".format(
+        content = "Message from {cog_name}:\n{content}".format(
             cog_name=cog_name, content=content if content is not None else ""
         )
         content = content[:1999]
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1393,7 +1393,7 @@ class RedBase(
 
         return destinations
 
-    def send_to_owners(self, content=None, **kwargs):
+    def send_to_owners(self, content=None, show_origin_message: bool = True, **kwargs):
         """
         This sends something to all owners and their configured extra destinations.
 
@@ -1401,10 +1401,15 @@ class RedBase(
 
         This logs failing sends
         """
-        cog_name = get_cog_by_call(self, "send_to_owners")
-        return self._send_to_owners(cog_name, content, **kwargs)
+        if show_origin_message:
+            cog_name = get_cog_by_call(self, "send_to_owners")
+        else:
+            cog_name = None
+        return self._send_to_owners(
+            cog_name, content, show_origin_message=show_origin_message, **kwargs
+        )
 
-    async def _send_to_owners(self, cog_name: str, content=None, **kwargs):
+    async def _send_to_owners(self, cog_name: Optional[str], content=None, **kwargs):
         """
         This sends something to all owners and their configured extra destinations.
 
@@ -1417,11 +1422,11 @@ class RedBase(
         async def wrapped_send(location, content=None, **kwargs):
             try:
                 await location.send(content, **kwargs)
-                await location.send(f"The message above came from: {cog_name}")
+                if cog_name:
+                    await location.send(f"The message above came from: {cog_name}")
             except Exception as _exc:
                 log.error(
-                    "I could not send an owner notification from cog '%s' to %s (%s)",
-                    cog_name,
+                    "I could not send an owner notification to %s (%s)",
                     location,
                     location.id,
                     exc_info=_exc,
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1407,8 +1407,8 @@ class RedBase(
 
         async def wrapped_send(location, content=None, **kwargs):
             try:
-                await location.send(f"The following message is from {cog_name}")
                 await location.send(content, **kwargs)
+                await location.send(f"The message above came from: {cog_name}")
             except Exception as _exc:
                 log.error(
                     "I could not send an owner notification from cog '%s' to %s (%s)",
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1393,7 +1393,7 @@ class RedBase(
 
         return destinations
 
-    async def send_to_owners(self, content=None, **kwargs):
+    def send_to_owners(self, content=None, **kwargs):
         """
         This sends something to all owners and their configured extra destinations.
 
@@ -1402,7 +1402,16 @@ class RedBase(
         This logs failing sends
         """
         cog_name = get_cog_by_call(self, "send_to_owners")
+        return self._send_to_owners(cog_name, content, **kwargs)
 
+    async def _send_to_owners(self, cog_name: str, content=None, **kwargs):
+        """
+        This sends something to all owners and their configured extra destinations.
+
+        This takes the same arguments as discord.abc.Messageable.send
+
+        This logs failing sends
+        """
         destinations = await self.get_owner_notification_destinations()
 
         async def wrapped_send(location, content=None, **kwargs):
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1420,7 +1420,6 @@ class RedBase(
         content = "Message from {cog_name}:\n{content}".format(
             cog_name=cog_name, content=content if content is not None else ""
         )
-        content = content[:1999]
         sends = [wrapped_send(d, content, **kwargs) for d in destinations]
         await asyncio.gather(*sends)
 
--- a/redbot/core/utils/_internal_utils.py
+++ b/redbot/core/utils/_internal_utils.py
@@ -241,7 +241,7 @@ async def create_backup(dest: Path = Path.home()) -> Optional[Path]:
 # this might be worth moving to `bot.send_to_owners` at later date
 
 
-async def send_to_owners_with_preprocessor(
+def send_to_owners_with_preprocessor(
     bot: Red,
     content: str,
     *,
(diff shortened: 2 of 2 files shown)


redbot/core/bot.py
```diff
@@ -1401,19 +1401,26 @@ async def send_to_owners(self, content=None, **kwargs):
...
                 log.error(
-                    "I could not send an owner notification to %s (%s)",
+                    "I could not send an owner notification from cog '%s' to %s (%s)",
+                    cog_name,
                     location,
                     location.id,
                     exc_info=_exc,
                 )
 
+        content = "Message from {cog_name}\n{content}".format(
```
Dav-Git: Can you please add ``:`` after ``Message from {cog_name}`` ?
My poor soul hurts without them.

Flame442:
-1 from me.

As far as I see, there is no point in adding extra overhead to that function to try and stop some kind of spam (or other attack) when similar (and even more dangerous) things can be done by a cog if the creator so chooses. I understand adding things to keep developers from hurting themselves (IE preventing mentions in sends unless explicitly requested by the dev), however adding extra code to "deter" (at best) a malicious dev solves nothing.

Drapersniper:
This PR is not necessarely about stoping a malicious cog, it is to ensure the Bot owners who receive the message always know where the message is coming from ensuring they take the appropriate action.

This PR doesn't stop anything it simply adds a prefix to the existing message saying "Here is the cog who is sending this message"

Dav-Git:
I think I like this. It's certainly not all bad to let bot owners know where the heck their messages are coming from. If you don't want this info included in your messages for whatever reason good or bad, we know that there are ways to do it.

Only caveat imo is that we will no longer be able to pass messages with len(2000) to send_to_owners()

Jackenmen:
-1, even though this may improve bot owner's experience a bit by making it a bit more obvious what cog sent the message, it takes away control from cog developers on what is sent to the user as this just unconditionally prepends text to the message. Yes, the cog developers could do it differently if they need to, but then they need to add the logic for sending in each owner notification destination themselves. It takes away the universality of this seemingly simple method. More so, if cog developer is already clear with users on the source of message, this will just hurt readability of his message.

Also, using `content[:1999]` is changing the behavior, even if the method doesn't error can cut important part of the message, so this is still breaking in my opinion.

Flame442:
On top of the fact dav pointed out, it also crops out the end of the passed string to a length that is not clear or documented. While I still am not for this PR, if this PR is to be merged that will need to be handled better than it currently is.

Drapersniper:
This is a few examples od cog that give no context from where a message is being sent from.

https://github.com/LordHighfixxer/FixxerV3Cogs/blob/146c183a10e2bbb08003e1de97ca522ab0a53ecf/rss/__init__.py#L13
https://github.com/Ianardo-DiCaprio/IanardoCogs/blob/de3b806f0ef5bc5485bff4cd669840859cbef067/bugreport/bugreport.py#L95
https://github.com/mon5termatt/Snapcog/blob/57b5cb100181f8d38f1b42bc4b9d096c89ad62ba/general/__init__.py#L9
https://github.com/mon5termatt/Snapcog/blob/57b5cb100181f8d38f1b42bc4b9d096c89ad62ba/anticommandspam/__init__.py#L13
https://github.com/mon5termatt/Snapcog/blob/57b5cb100181f8d38f1b42bc4b9d096c89ad62ba/rss/__init__.py
https://github.com/mon5termatt/Snapcog/blob/57b5cb100181f8d38f1b42bc4b9d096c89ad62ba/rolemanagement/__init__.py
https://github.com/Cog-Creators/Red-DiscordBot/blob/1d80fe9aecbde48d55cb32fb53807ed70ade6b02/redbot/cogs/admin/announcer.py#L73

This is Cogs with users in control, given that we are providing them the ability to message all owners/destination channels set, we should at least ensure we tell the user who is sending the message so its not a random message that just appear in my dms as a secondary owner.

Drapersniper:
> On top of the fact dav pointed out, it also crops out the end of the passed string to a length that is not clear or documented. While I still am not for this PR, if this PR is to be merged that will need to be handled better than it currently is.

If this is a problem sure i can ensure the cropping doesnt happen.

Flame442:
Those examples are all* of unapproved repos that I have not seen recommended in cog support, therefor I can only assume they are personal repos or users who have not had much user feedback. Obviously developers cannot think out the entirety of what a user will do or expect since they see things from their own perspective, so it is up to developers to consider the user when they are coding.

*All except for 1 which is Ian's, however that is a bug report message from a cog the bot owner would set up and install, knowing when they get a bug report it is probably from that cog.

Also, that change you just did is still breaking since now the command will error (see: post a log message in console) for things that it would not have previously.

Drapersniper:
Core: 
  - https://github.com/Cog-Creators/Red-DiscordBot/blob/1d80fe9aecbde48d55cb32fb53807ed70ade6b02/redbot/cogs/admin/announcer.py#L73

Fox-V3:
  - https://github.com/bobloy/Fox-V3/blob/master/stealemoji/stealemoji.py#L250

PCXCogs: 
  - https://github.com/PhasecoreX/PCXCogs/blob/master/updatenotify/updatenotify.py#L394

Flare-Cogs:
  - https://github.com/flaree/Flare-Cogs/blob/master/forward/forward.py#L35
  - https://github.com/flaree/Flare-Cogs/blob/master/botlistspost/botlistspost.py#L49


Sorry Bobloy, Phase and Flare didn't want to exactly call any of you by name but since Flame mentioned that the examples that everything was just unapproved.


Anyways, point is its not just unapproved cogs, sure Bugreport from Ian is done via a command, but thats assuming a singular bot owner, any bots with multiple owners would result in otu of context random messages being sent when one owner request something and a send_to_owners is used,  this wouldnt be an issue if the messages had full context of where they came from.

Flame442:
> Core

That is from a fail state of a BOT OWNER ONLY command. While it is fair that an instance with 2 bot owners would have 1 slightly confused person, that's an issue to be resolved IN THAT MESSAGE not in `send_to_owners`.

> https://github.com/bobloy/Fox-V3/blob/master/stealemoji/stealemoji.py#L250

Again, this is a message that only happens when all these criteria are met
1. The bot owner installed the cog
2. The bot owner turned on "autobank" ("Toggles automatically creating new guilds as emoji banks")
3. The bot owner turned on "on" ("Toggles whether emoji's are collected or not")

Should that message do a better job explaining where that invite came from? YES! However, I still do not think that's a job for `send_to_owners` to handle.

> https://github.com/PhasecoreX/PCXCogs/blob/master/updatenotify/updatenotify.py#L394

Again, this is a message that only happens the user installs a cog who's purpose is to notify you when your bot needs an update ("updatenotify"). This one I really do not think needs to say where it comes from because the user installed a cog to notify them of updates and they are being notified of an update, simple as that.

> https://github.com/flaree/Flare-Cogs/blob/master/forward/forward.py#L35

I'm not going to dive deep in to this one because I am tired of this, but this is another cog the user sets up and expects the output they get ("Forward messages sent to the bot **to the bot owner** or in a specified channel").

---

I hope you understand that picking random cases where `send_to_owner` is used does not prove anything to me. Either the cogs are currently handling their `send_to_owner` calls in an adequate way or they *should* be and it is not the job of the `send_to_owner` func to handle that.

flaree:
As a cog creator and a bot user I really like this change. It's definitely helpful to diagnose where a message is coming from without getting massively confused or having to dig through code or your cogs just to find the one.

I currently take advantage of send to owner is a few cogs(not just ones on the approved repo) and I try to make it clear where it's coming from but not everyone else does the same.

Drapersniper:
Given that its an util provides BY RED, to allow cogs to message ALL owners, i think ensuring the utility is clear and ensuring it tells the users where a message is originating from is something that should be done by Red, if thats not the case that util shouldnt exist and cogs should handle their own way to send stuff to bot owners.

And as a small FYI, `updatenotify` has caused confusion in the past as people weren't sure if it was coming from it or core.

Note that my point here is Owners, everything you said is great if the bot only has one owner but we support multiple owners.

Twentysix26:
Mixed feelings about this.
IMO it's trying to solve a problem that should be solved by the individual developers (poorly written messages).
However, as long as it's not mangling the message the CC is trying to send (which would be a hard -1 from me but I see it's no longer the case) it can at least be of benefit for the end user in certain cases.

madebylydia:
It's a +1 for me.

I like this change from my point of view of developer and bot owner, even if I'm not getting much messages from 3rd party cogs (And Red in general), it can be useful for some certain context (That I may not have in mind)

What Flame said don't really disappoint me, honestly, it is absolutely not a problem for me, but I'm kinda nervous of what it could show. Still not a problem.

fixator10:
± from me

really mixed feelings about that. Its kinda can be useful sometimes to know where message comes from, but in some cases it can be an real pain in the ass. E.g. if someone sends an message thats >2k char length, most of devs will somehow handle it. This PR makes this handling much less predictable, especially if it appends to the start of the same message. If it doesnt, it make trigger another problem: ratelimits. Imagine i sent 5 notifications in row. With message about message origin before message it will be 10 messages and highly likely will hit ratelimit

Drapersniper:
> ± from me
> 
> really mixed feelings about that. Its kinda can be useful sometimes to know where message comes from, but in some cases it can be an real pain in the ass. E.g. if someone sends an message thats >2k char length, most of devs will somehow handle it. This PR makes this handling much less predictable, especially if it appends to the start of the same message. If it doesnt, it make trigger another problem: ratelimits. Imagine i sent 5 notifications in row. With message about message origin before message it will be 10 messages and highly likely will hit ratelimit

Drapersniper:
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa

Drapersniper:
God damn it shit UX from github aaaaaaaaaaa, 

> ± from me
> 
> really mixed feelings about that. Its kinda can be useful sometimes to know where message comes from, but in some cases it can be an real pain in the ass. E.g. if someone sends an message thats >2k char length, most of devs will somehow handle it. This PR makes this handling much less predictable, especially if it appends to the start of the same message. If it doesnt, it make trigger another problem: ratelimits. Imagine i sent 5 notifications in row. With message about message origin before message it will be 10 messages and highly likely will hit ratelimit



anyways the first part of your message should be resolved now, this will not send a message unless the message is sent by `send_to_owners`, as it will show after it is sent now.
image


The ratelimit is something i considered when adding an additional message, the additional message was added so that there would be no editing of the original message, as that was a large part of the concern, the downside of this is that it will require 1 extra message per destination for the message being sent, given that when looking through core and approved cogs for the usage of `send_to_owners` I couldnt find a place where `send_to_owners` is used in a loop which would cause it to trigger the ratelimits and even if it happened over in a while its not like it will be a regular occourance so I have the mindset that in this case the ratelimit is not a big concern.

Jackenmen:
I know this could potentially make this less effective (although I really doubt that considering what would the default be) and be used in wrong situations, but I think it might be worth adding a kwarg for cog developer to disable this if they have a reason not to send this (assuming this PR goes through at all ofc). One place in core where this could make some sense is IMO `[p]contact` command. It doesn't actually use `send_to_owners` due to it having to handle embed settings, but I just wanted to use it as an example use case for such kwarg.

Drapersniper:
Although I'm not entirely happy with giving CC the option to not let the user know where this is originating from ... I do see a benefit, here as a good CC could set this new kwarg to false and let the user know where this is coming from keeping in mind the multi bot owner scenario,the downside of this kwarg is that bad CCs and unapproved repos could choose to just optout of this, so if bot owners think allowing cogs to optout of this message is not a good idea do say so.

ltzmax:
+1 from me.
i find it useful and helpfully, it would also be helpfully for alot of people i believe, without getting confused where all is coming from.

Twentysix26:
> the downside of this kwarg is that bad CCs and unapproved repos could choose to just optout of this

If they're so keen on hiding the source of their message they can just bypass this entirely by not using the utility. I agree that a kwarg would be positive and gives some more freedom to CC who know what they're doing. People with bad intentions will find a thousand other ways to bypass this change and there is not much we can do about it.

redbot/core/utils/_internal_utils.py
```diff
@@ -316,3 +369,39 @@ async def fetch_latest_red_version_info() -> Tuple[Optional[VersionInfo], Option
         required_python = data["info"]["requires_python"]
 
         return release, required_python
+
+
+def get_cog_by_call(bot: Red, func: str) -> str:
+    try:
+        stack = inspect.stack()
```
Jackenmen, line 376: Usage of `inspect.stack()` is very inefficient, it would be much preferred to use `inspect.currentframe()` and traversing the frames one by one using `frame.f_back`.
Using the code object from the frame, you can get the function name: `frame.f_code.co_name`

By using this way, you will avoid a lot of file io that has to happen in order to get `FrameInfo` objects with file names, line numbers and *especially* source file context for each frame.

Here's a code snippet of how it would look (untested):
```py
frame = inspect.currentframe()
while frame is not None:
    if frame.f_code.co_name == func:
        # we're interested in the function call that happened in outer frame
        frame = frame.f_back
        break
    frame = frame.f_back
if frame is not None:  # replaces `if index >= 0`
    ...
```

redbot/core/utils/_internal_utils.py
```diff
@@ -316,3 +369,39 @@ async def fetch_latest_red_version_info() -> Tuple[Optional[VersionInfo], Option
...
+        valid = False
+        frm = None
+        for index, item in enumerate(stack):
+            if item.function == func:
+                break
+        else:
+            index = -2
+        if index >= 0:
+            frm = stack[index + 1]
+            full_caller_page = inspect.getmodule(frm[0]).__package__
```
Jackenmen, line 387: `inspect.getmodule()` may return `None` which would make this raise an AttributeError.

redbot/core/utils/_internal_utils.py
```diff
@@ -316,3 +369,39 @@ async def fetch_latest_red_version_info() -> Tuple[Optional[VersionInfo], Option
...
+            if full_caller_page.startswith("asyncio"):
+                return "A task, unable to pinpoint origin."
+            elif full_caller_page.startswith("redbot.cogs."):
+                cog_name = full_caller_page.split(".")[2]
+                return f"Red Core - `{cog_name}`"
+            elif full_caller_page.startswith("redbot"):
+                return "Red Core"
+            caller_package = full_caller_page.split(".")[0]
+            for cog_name, cog in bot._BotBase__cogs.items():
+                cog_package = inspect.getmodule(cog).__package__.split(".")[0]
```
Jackenmen: nitpick, but using the `__module__` directly looks simpler and achieves same thing.
```suggestion
                cog_package = cog.__module__.split(".")[0]
```

redbot/core/utils/_internal_utils.py
```diff
@@ -240,7 +241,52 @@ async def create_backup(dest: Path = Path.home()) -> Optional[Path]:
...
+    Other Parameters
+    ----------------
+    content_preprocessor: Optional[Callable[[Red, discord.abc.Messageable, str], Awaitable[str]]]
+        Optional async function that takes
+        bot object, owner notification destination and message content
+        and returns the content that should be sent to given location.
+    """
+    try:
+        stack = inspect.stack()
+        if stack[1].function == "send_to_owners_with_prefix_replaced":
```
Jackenmen: Big performance boost, see the other review comment for more info.
We could probably gain even more, if we just had a kwarg in which we would pass the function name, since this utility function is not public anyway.
```suggestion
        frame = inspect.currentframe().f_back
        if frame.f_code.co_name == "send_to_owners_with_prefix_replaced":
```

Jackenmen (changes requested):
Code review focusing on technicalities, I have not tested this PR.
I still don't know whether this PR will go in, but I would not want this to be missed at later review, if it were to go in.

PR #4415, opened 2020-09-09
[Modlog API] Handle deleted channels in `Case.message_content()`

Dav-Git:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Modlog now doesn't error in non-embedded version when a channel doesn't exist anymore.


Jackenmen (approved):

PR #4416, opened 2020-09-09
Add custom groups tutorial to Config's docs

### LINKED ISSUE #2896: [Docs] Add custom groups to Config tutorial's "Advanced Usage" section
Currently the custom groups aren't mentioned in Config tutorial at all and you can only find information about custom groups in api reference. New cog creators may think that the only solution when data doesn't fit in regular discord models, is using global group, for example this place of tutorial doesn't even mention there is such thing as `custom()`:
image

Generally the whole "Advanced Usage" section promotes use of `.get_raw()` when you can't easily fit data in one of the standard categories. Also, it's not obvious (at least until you get the error when trying custom groups without it) that you have to use `init_custom` to be able to use the custom group, there's only a small mention that the method "must be called first" in this method's docstring, but it doesn't explain what "first" means here and it also isn't mentioned in `register_custom` method where some people could expect such non-regular information to be. I think putting some info about this in the tutorial ("Advanced Usage" section to be specific) could really make people use it more and understand how they work.

Last thing that *I think* no longer works is the "V2 Data Usage" section, I don't see the `init_custom` anywhere and from what I know it's required.

bobloy:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Updates the config docs to be more modern, and adds a tutorial to use `Config.custom`, `Config.init_custom` and `Config.register_custom`

Resolves #2896

### DIFF
--- a/docs/framework_config.rst
+++ b/docs/framework_config.rst
@@ -21,8 +21,9 @@ Basic Usage
 .. code-block:: python
 
     from redbot.core import Config
+    from redbot.core import commands
 
-    class MyCog:
+    class MyCog(commands.Cog):
         def __init__(self):
             self.config = Config.get_conf(self, identifier=1234567890)
 
@@ -52,7 +53,7 @@ Then, in the class's :code:`__init__` function, you need to get a config instanc
 
 .. code-block:: python
 
-    class MyCog:
+    class MyCog(commands.Cog):
         def __init__(self):
             self.config = Config.get_conf(self, identifier=1234567890)
 
@@ -68,7 +69,7 @@ After we've gotten that, we need to register default values:
 
 .. code-block:: python
 
-    class MyCog:
+    class MyCog(commands.Cog):
         def __init__(self):
             self.config = Config.get_conf(self, identifier=1234567890)
             default_global = {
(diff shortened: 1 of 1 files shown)


docs/framework_config.rst
```diff
@@ -21,8 +21,9 @@ Basic Usage
 .. code-block:: python
 
     from redbot.core import Config
+    from redbot.core import commands
```
Bakersbakebread, line 24:
```suggestion
    from redbot.core import Config, commands
```

bobloy: This was intentionally left as two lines because shortly after there is an explanation of Config that goes line by line. Leaving it as two lines made the explanation easier to understand.

See line 49/50 : https://github.com/Cog-Creators/Red-DiscordBot/pull/4416/files#diff-249132a9932c4d5c2a61eceab315ecb1R50

For the remainder of the document where it is not explaining imports, the suggested format should be used.

docs/framework_config.rst
```diff
@@ -188,19 +189,120 @@ features within Config to enable developers to work with data how they wish.
...
+    async def checkbaz(self, ctx):
+        # baz_val = await self.config.guild(ctx.guild).baz()
+
+        baz_val = await self.config.custom("CustomGuildGroup", ctx.guild.id).baz()
+        await ctx.send("The value of baz is {}".format("True" if baz_val else "False"))
+
+Here we used :py:meth:`Config.custom` to access our custom group much like we would have used :py:meth:`Config.guild`.
+Since it's a custom group, we need to use :code:`id` attribute of guild to get a unique identifier.
+
+Now lets show and example that uses multiple identifiers:
```
Jackenmen:
```suggestion
Now let's see an example that uses multiple identifiers:
```

docs/framework_config.rst
```diff
@@ -188,19 +189,120 @@ features within Config to enable developers to work with data how they wish.
...
+- :py:meth:`Config.register_custom`
+- :py:meth:`Config.custom`
 - :py:meth:`Group.get_raw`
 - :py:meth:`Group.set_raw`
 - :py:meth:`Group.clear_raw`
 
+
+Custom Groups
+^^^^^^^^^^^^^
+While Config has built in groups for the common discord objects,
```
Jackenmen:
```suggestion
While Config has built-in groups for the common discord objects,
```

docs/framework_config.rst
```diff
@@ -188,19 +189,120 @@ features within Config to enable developers to work with data how they wish.
...
+                "blah": [],
+                "baz": 1234567890
+            }
+
+            # self.config.register_guild(**default_guild)
+
+            self.config.init_custom("CustomGuildGroup", 1)
+            self.config.register_custom("CustomGuildGroup", **default_guild)
+
+In the above we registered the custom group named "CustomGuildGroup" to contain the same defaults
```
Jackenmen:
```suggestion
In the above, we registered the custom group named "CustomGuildGroup" to contain the same defaults
```

docs/framework_config.rst
```diff
@@ -188,19 +189,120 @@ features within Config to enable developers to work with data how they wish.
...
+that :code:`self.config.guild` normally would. First we initialized the group "CustomGuildGroup" to
```
Jackenmen:
```suggestion
that :code:`self.config.guild` normally would. First, we initialized the group "CustomGuildGroup" to
```

docs/framework_config.rst
```diff
@@ -188,19 +189,120 @@ features within Config to enable developers to work with data how they wish.
...
+accept one identifier by calling :py:meth:`Config.init_custom` with the argument :code:`1`. Then we used
+:py:meth:`Config.register_custom` to register the default values.
+
+.. important::
+
+    :py:meth:`Config.init_custom` **must** be called prior to using a custom group.
+
+
+Now lets use this custom group:
```
Jackenmen:
```suggestion
Now let's use this custom group:
```

Jackenmen (changes requested):
I really like how this is written! There's just a few grammar comments that need to be addressed before merging.

Jackenmen (approved):
Big thanks for making our docs better! LGTM.

PR #4423, opened 2020-09-13
Update to d.py 1.5.1, explicitly request privileged intents 

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This PR updates our d.py version to 1.5.1, which includes changing usage of `bot.request_offline_members()` with `guild.chunk()` and adding re-exports for new objects in d.py's commands extension.
On top of that, this PR makes Red explicitly request privileged intents and makes it fail to start otherwise.

From other things:
- added a test checking that we're not missing re-exports from d.py version we're using to avoid potential bugs in future
- added **unsupported** `--disable-intent` option that can be used to disable given intent (can be used multiple times)
- `[p]debuginfo` will now show intents that have been disabled by the user despite the "unsupported note" on the flag that does, so that supporters can still easily diagnose weird issues related to disabled intents
    - this may not be always helpful as disabling some intents will make the bot outright not listen to messages, but let's hope that's not gonna be hard to diagnose


redbot/core/cli.py
```diff
@@ -237,6 +238,15 @@ def parse_cli_flags(args):
     parser.add_argument(
         "--no-message-cache", action="store_true", help="Disable the internal message cache."
     )
+    parser.add_argument(
+        "--disable-intent",
```
TrustyJAID, line 241: If this is set this way are we going to be applying all intents by default?

Jackenmen: Yes.

Drapersniper: I guess it makes sense since the majority of users are under 100 servers, but given that majority of users also dont read. I would honestly feel more confortable if this was off by default and its a new flag we tell users to start red with based on what they have enabled.

TrustyJAID: The current way this PR works Draper is it ensure the bot will at least start for most of our users. If we don't have all intents enabled once the bot starts reaching more guilds it will increase start time for many users.

Drapersniper: But the intent are all off by default and need to be enabled via the portal, this is what im referencing most users not reading.

I expect the vast majority of live users today to not have the intents enabled in the portal.

flaree:
We have had an announcement in the past also telling users to enable it. So both ways, people will be affected.

redbot/core/cli.py
Jackenmen: The error for when people try to use privileged intents will contain information on how to enable them and I will most likely also update docs appropriately. While I do want to add some functionality related to intents (be it checks, metadata or whatever TBD), currently `--disable-intent` is just there so people with big bots can at least disable the intents. Currently all 1st-party and 3rd-party cogs expect to receive all events and it does not really make sense to change the default and bother a big majority of users to change the default.

I just want to remind this is purposefully just a draft, and the better errors as well as doc changes are still to be done.

TrustyJAID: ~~Something to note is that currently this does not allow disabling members or presences intent.~~ Nevermind this does work as expected.

Jackenmen: That's not true, `**dict.fromkeys(cli_flags.disable_intent, False)` that is lower in the dict literal will 
 just replace them.

Drapersniper:
Can we add 4 new checks that commands can be wrapped with
```py
@bot_has_intent(**kwargs)
@bot_has_presence_intent()
@bot_has_member_intent()
@bot_has_all_intents()
```

Jackenmen:
This PR is ready for review now. It's been a bit repurposed and there isn't a lot of intents-focused features other than what's required for Red to work with d.py 1.5.1. The intents-focused feature will come in some later version when we're ready to commit to that.

@Drapersniper could you move your comment into a separate new issue that will gather things we want to provide in order to fully support intents in Red?

Jackenmen:
Closing and reopening for tests to rerun.

redbot/core/cli.py
```diff
@@ -236,6 +237,17 @@ def parse_cli_flags(args):
...
+        action="append",
+        choices=list(discord.Intents.VALID_FLAGS),  # DEP-WARN
+        default=[],
+        help="Unsupported option that allows to disable the given intent."
```
Drapersniper:
```suggestion
        help="Unsupported flag that allows disabling the given intent."
```

redbot/core/cli.py
```diff
@@ -236,6 +237,17 @@ def parse_cli_flags(args):
...
+        " Currently NOT SUPPORTED as Red is not prepared to work without all intents.\n"
+        f"Go to https://discordpy.readthedocs.io/en/v{discord_version}/api.html#discord.Intents"
+        " to see what each intent does.\n"
+        "Option can be used multiple times to specify multiple intents.",
```
Drapersniper:
```suggestion
        "This flag can be used multiple times to specify multiple intents.",
```

Drapersniper (changes requested):
I've completed a code review everything is looking fine, other than the 2 optional wording suggestions, I haven't had the time to do a functionality test as been busy but hoping to complete it over the next few days.

Jackenmen:
I actually had "flag" there before, but I figured I should probably use terminology where "argument" is positional, "option" is a named argument that is passed with some value and "flag" is an option without a value.
i.e.
```
redbot instance_name_is_parameter --option-has value --flag-does-not-have-value
```

We don't really have precedence here so I'm up to whatever you think makes sense, just let me know.

Drapersniper:
> I actually had "flag" there before, but I figured I should probably use terminology where "argument" is positional, "option" is a named argument that is passed with some value and "flag" is an option without a value.
> i.e.
> 
> ```
> redbot instance_name_is_parameter --option-has value --flag-does-not-have-value
> ```
> 
> We don't really have precedence here so I'm up to whatever you think makes sense, just let me know.

This comes from my professions background, but for me a flag is either a boolean_store or a specified value, since we have no precedence here I personally would prefer it to be called flags, that said dont feel strongly enough about it to argue either way.

Drapersniper:
```
(Red-DiscordBot-QA) C:\Users\guyre\PycharmProjects\Red-DiscordBot>redbot intents --dev  --disable-intent presences --disable-intent members
c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\client.py:250: DeprecationWarning: fetch_offline_members is deprecated, use chunk_guilds_at_sta
rtup instead
  self._connection = self._get_state(**options)
Unexpected exception (<class 'ValueError'>):
  File "c:\users\guyre\pycharmprojects\red-discordbot\redbot\__main__.py", line 497, in main
    red = Red(
  File "c:\users\guyre\pycharmprojects\red-discordbot\redbot\core\bot.py", line 209, in __init__
    super().__init__(*args, help_command=None, **kwargs)
  File "c:\users\guyre\pycharmprojects\red-discordbot\redbot\core\rpc.py", line 130, in __init__
    super().__init__(**kwargs)
  File "c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\ext\commands\bot.py", line 99, in __init__
    super().__init__(**options)
  File "c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\ext\commands\core.py", line 1096, in __init__
  File "c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\shard.py", line 290, in __init__
    super().__init__(*args, loop=loop, **kwargs)
  File "c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\client.py", line 250, in __init__
    self._connection = self._get_state(**options)
  File "c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\shard.py", line 311, in _get_state
    return AutoShardedConnectionState(dispatch=self.dispatch,
  File "c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\state.py", line 1103, in __init__
  File "c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\state.py", line 173, in __init__
    raise ValueError('Intents.members must be enabled to chunk guilds at startup.')
ValueError: Intents.members must be enabled to chunk guilds at startup.
``` 
Not a super fan of this error, so just wanted to point out it could be made slightly more user friendly but hey ho.

```
(Red-DiscordBot-QA) C:\Users\guyre\PycharmProjects\Red-DiscordBot>redbot intents --dev  --disable-intent presences
c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\client.py:250: DeprecationWarning: fetch_offline_members is deprecated, use chunk_guilds_at_sta
rtup instead
  self._connection = self._get_state(**options)
Red requires all Privileged Intents to be enabled.
You can find out how to enable Privileged Intents with this guide:
https://docs.discord.red/en/stable/bot_application_guide.html#enabling-privileged-intents
[2020-10-21 14:31:44] [INFO] red.main: Shutting down with exit code: 1
[2020-10-21 14:31:44] [INFO] red.main: Please wait, cleaning up a bit more


```

```
c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\client.py:250: DeprecationWarning: fetch_offline_members is deprecated, use chunk_guilds_at_sta
rtup instead
  self._connection = self._get_state(**options)
``` 

We should silence this deprication warning.


That said disabling members makes red unusable ... If it is not avaliable to the user due to them not being approved or only being approved/applied to presences, Red will become unusable.

Given that this is the case I personally do not feel confortable with this change going into a micro release, as it has the potential to break any Red user in more than 100 guilds (or that been in more than 100 guild at some point since the verification process started) *note: Users below 100 guilds can just toggle in the dev portal *
image


With the above stated I would suggest this to be punted to the 3.5 release to give devs further time to apply/enable the intents, and with that make an @everyone announced in the support server and a Pinned issue in the repo.

Due to not having analytics in the project we are unable to say how many users will be affected by this and have an unusable bot that will require a downgrade, until they get hold of the required intent (if they are able to)

redbot/core/core_commands.py
```diff
@@ -2560,6 +2560,14 @@ async def debuginfo(self, ctx: commands.Context):
             osver = "Could not parse OS, report this on Github."
         user_who_ran = getpass.getuser()
         driver = storage_type()
+        disabled_intents = (
+            ", ".join(
```
Drapersniper, line 2564: Nit Pick here, please use humanize list.

redbot/__main__.py
```diff
@@ -391,6 +391,13 @@ async def run_bot(red: Red, cli_flags: Namespace) -> None:
                 print("Token has been reset.")
                 sys.exit(0)
         sys.exit(1)
+    except discord.PrivilegedIntentsRequired:
```
Drapersniper, line 394: This error is raised if I'm missing one of the intents in the portal and haven't used the flag...
the exact message is:

```
(Red-DiscordBot-QA) C:\Users\guyre\PycharmProjects\Red-DiscordBot>redbot intents --dev
c:\users\guyre\.virtualenvs\red-discordbot-qa\lib\site-packages\discord\client.py:250: DeprecationWarning: fetch_offline_members is deprecated, use chunk_guilds_at_startup instead
  self._connection = self._get_state(**options)
Red requires all Privileged Intents to be enabled.
You can find out how to enable Privileged Intents with this guide:
https://docs.discord.red/en/stable/bot_application_guide.html#enabling-privileged-intents
[2020-10-21 14:45:03] [INFO] red.main: Shutting down with exit code: 1
[2020-10-21 14:45:03] [INFO] red.main: Please wait, cleaning up a bit more

```
.... We should direct them to the new flag, or allow red to boot up without presence. at least imo... presense was sparcelly given to folks.

Drapersniper (changes requested):

Drapersniper:
Okay to add to this, after a conversation in AC, We have decided to carry on adding this to 3.4.1, however we will be doing an announcement ahead of the released so that the user base is aware of the potential problem that they may face if they update to 3.4.1 without the correct intents avaliable and what they can do in that case, and the release will follow a few days after (Hopefully).

redbot/core/core_commands.py
Jackenmen: I purposefully didn't use humanize_list here - `[p]debuginfo` is debug information that should help debug issues and nothing in it is translated.

redbot/__main__.py
Jackenmen: The printed link explains everything needed already (just change `stable` to `latest` when using it for now, because we haven't released yet). As for directing to the flag, I think it will only cause us trouble with no benefit - it is already shown in `redbot --help` and since it is currently not supported, we shouldn't really advertise it more than that.

Drapersniper (approved):
So based in our conversations, and that we will be updating users via an announcement prior to release and the clarification added post my initial review I can happly approve this.

PR #4425, opened 2020-09-13
Add `quote()` function to chat formatting utilities

Zephyrkul:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

Note that, unlike other methods in this module, this method does *not* escape formatting by default at current.

### DIFF
--- a/redbot/core/utils/chat_formatting.py
+++ b/redbot/core/utils/chat_formatting.py
@@ -327,17 +327,13 @@ def underline(text: str, escape_formatting: bool = True) -> str:
     return "__{}__".format(text)
 
 
-def quote(text: str, escape_formatting: bool = False) -> str:
+def quote(text: str) -> str:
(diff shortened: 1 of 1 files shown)


Jackenmen:
Is there much reason to add option that escapes markdown here? In methods like `underline()`, this is needed so that the underlined text is underlined exactly as is. This problem doesn't exist for quoting as far as I know, so I'm interested to know, if there's other reason you decided to include it here.

Zephyrkul:
Solely for the reason that it's present in other methods, which escape *all* formatting by default.

Jackenmen:
I think it can be removed if that's the only reason then. It fits well with the `box()` method that also doesn't offer the option to escape formatting.

Jackenmen (approved):
LGTM, thanks for PRing.

PR #4433, opened 2020-09-17
Added support for getting leaderboard size back with position

### LINKED ISSUE #4200: [Bank API] Make it so `get_leaderboard_position` returns a tuple with user position and total entries in leaderboard
# Feature request

- [ ] Cog
- [ ] Command
- [ ] API functionality

As the title suggest, this would allow making displays without having to make more expensive calls.

Make it into an kwarg so that it is not breaking and allows existing behaviour when not used.

Alternativelly add a new API call that returns this tuple

MrBillson:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Added support for getting leaderboard size back with position. You can now pass in an optional param that is a flag to return the overall leaderboard size or not. This is the enhancement requested in issue 4200. This is a backwards compatible non-breaking change.

closes #4200


redbot/core/bank.py
```diff
@@ -559,8 +559,8 @@ async def get_leaderboard(positions: int = None, guild: discord.Guild = None) ->
 
 
 async def get_leaderboard_position(
-    member: Union[discord.User, discord.Member]
-) -> Union[int, None]:
+    member: Union[discord.User, discord.Member], *args, **kwargs
+) -> Union[int, None] or List[tuple]:
```
```suggestion
) -> Union[Optional[int], Tuple[Optional[int], int]]:
```

redbot/core/bank.py
```diff
@@ -559,8 +559,8 @@ async def get_leaderboard(positions: int = None, guild: discord.Guild = None) ->
 
 
 async def get_leaderboard_position(
-    member: Union[discord.User, discord.Member]
-) -> Union[int, None]:
+    member: Union[discord.User, discord.Member], *args, **kwargs
```
```suggestion
    member: Union[discord.User, discord.Member], return_lb_size: bool = False
```

redbot/core/bank.py
```diff
@@ -569,17 +569,27 @@ async def get_leaderboard_position(
     member : `discord.User` or `discord.Member`
         The user to get the leaderboard position of
 
+    returnLeaderboardSize : Boolean to return the size of the
```
```suggestion
    return_lb_size : bool
        Whether or not to return a tuple with position and size of leaderboard
```

redbot/core/bank.py
```diff
@@ -569,17 +569,27 @@ async def get_leaderboard_position(
...
+        overall leaderboard or not
```
```suggestion
```

redbot/core/bank.py
```diff
@@ -569,17 +569,27 @@ async def get_leaderboard_position(
...
+        The position of the user on the leaderboard followed
+        by the size of the leaderboard
+
     Raises
     ------
     TypeError
         If the bank is currently guild-specific and a `discord.User` object was passed in
 
     """
+    returnLeaderboardSize = kwargs.get("returnLeaderboardSize")
```
```suggestion
```

redbot/core/bank.py
```diff
@@ -591,9 +601,13 @@ async def get_leaderboard_position(
     else:
         pos = discord.utils.find(lambda x: x[1][0] == member.id, enumerate(leaderboard, 1))
         if pos is None:
-            return None
+            returnPos = None
         else:
-            return pos[0]
+            returnPos = pos[0]
+    if returnLeaderboardSize:
```
```suggestion
    if return_lb_size is True:
```

redbot/core/bank.py
```diff
@@ -569,17 +569,27 @@ async def get_leaderboard_position(
...
+    returnLeaderboardSize : Boolean to return the size of the
+        overall leaderboard or not
+
+
     Returns
     -------
     `int`
         The position of the user on the leaderboard
 
+    'int', 'int'
```
```suggestion
   Tuple[Optional[int], int]
```

redbot/core/bank.py
```diff
@@ -591,9 +601,13 @@ async def get_leaderboard_position(
     else:
         pos = discord.utils.find(lambda x: x[1][0] == member.id, enumerate(leaderboard, 1))
         if pos is None:
-            return None
+            returnPos = None
```
```suggestion
            return_pos = None
```

redbot/core/bank.py
```diff
@@ -591,9 +601,13 @@ async def get_leaderboard_position(
...
         else:
-            return pos[0]
+            returnPos = pos[0]
```
```suggestion
            return_pos = pos[0]
```

redbot/core/bank.py
```diff
@@ -591,9 +601,13 @@ async def get_leaderboard_position(
...
+    if returnLeaderboardSize:
+        return returnPos, len(leaderboard)
```
```suggestion
        return return_pos, len(leaderboard)
```

redbot/core/bank.py
```diff
@@ -591,9 +601,13 @@ async def get_leaderboard_position(
...
+    else:
+        return returnPos
```
```suggestion
        return return_pos
```

Drapersniper (changes requested):
Just a quick initial review from my phone, also I noticed something like 578 states the return type in an int, that's incorrect since it's an Optional[int], would you mind tweaking that while working on this pr? If not it's fine it can be done in a separate PR.

You will notice I've renamed some of the variables you have added, this is to ensure we are following PEP and keeping variable naming motion consistent across the project.

MrBillson:
Great feedback. I have a better idea on contributing to this project now!

MrBillson:
I think all of the requested changes are done. Let me know if not. Thanks!

Drapersniper (approved):
This is awkward, I forgot to assign it to me so I missed your changed, Approving for now.

@jack1142 this one is on the milestone for 3.4.11, but just in case merging would make your life harder, please go ahead and merge when it is a good time.

redbot/core/bank.py
```diff
@@ -559,27 +559,36 @@ async def get_leaderboard(positions: int = None, guild: discord.Guild = None) ->
 
 
 async def get_leaderboard_position(
-    member: Union[discord.User, discord.Member]
-) -> Union[int, None]:
+    member: Union[discord.User, discord.Member], return_lb_size: bool = False
```
Drapersniper, line 607:
```suggestion
    member: Union[discord.User, discord.Member], *, return_lb_size: bool = False
```

Jack suggested this should be changed into a keyword-only arg, so if you can do this too it would be awesome.

Drapersniper (changes requested):
Once the pending suggestion is addressed I can merge this one for you @MrBillson

Jackenmen:
I'm not sure that I like returning a tuple or an integer depending on a kwarg value here instead of just having a new function. Or maybe even not doing either, because one can just use `get_leaderboard()` and manually get the data. It's not like `get_leaderboard_position()` has some optimisations. Marking for discussion.

PR #4437, opened 2020-09-19
Handle aiohttp errors in `[p]set avatar`

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1776,8 +1776,13 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
                 url = url[1:-1]
 
             async with aiohttp.ClientSession() as session:
-                async with session.get(url) as r:
-                    data = await r.read()
+                try:
+                    async with session.get(url) as r:
+                        data = await r.read()
+                except aiohttp.InvalidURL:
+                    return await ctx.send(_("That URL is invalid."))
+                except aiohttp.ClientConnectionError:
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -1776,8 +1776,13 @@ async def avatar(self, ctx: commands.Context, url: str = None):
...
 
             async with aiohttp.ClientSession() as session:
-                async with session.get(url) as r:
-                    data = await r.read()
+                try:
+                    async with session.get(url) as r:
+                        data = await r.read()
+                except aiohttp.InvalidURL:
+                    return await ctx.send(_("That URL is invalid."))
+                except aiohttp.ClientConnectionError:
```
Jackenmen: Request can raise subclasses of any of these [1]: `aiohttp.ClientPayloadError`, `aiohttp.ClientResponseError`, `aiohttp.ClientConnectionError`.

If you want to, you can make more specific error messages for each, but I don't think it matters at all, so just using `aiohttp.ClientError`, which is a superclass of all of the exception mentioned above will work just fine here:
```suggestion
                except aiohttp.ClientError:
```

[1] aiohttp reference - https://docs.aiohttp.org/en/stable/client_reference.html#hierarchy-of-exceptions

Jackenmen (changes requested):

Jackenmen (approved):
Looks good to me, thanks!

PR #4440, opened 2020-09-20
Owners can bypass cooldowns with non persistent toggle

flaree:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Allow owners to bypass cooldowns if the dev cogs is loaded.


Jackenmen:
I can definitely see how this may be useful, however personally I don't want my bot to ignore cooldowns for me when I use `--dev` flag just because I'm bot owner. I do agree it can ease up testing commands with cooldown, but it also makes it impossible for the developer to check whether the cooldown actually works (without switching to another non-owner account or disabling the `--dev` flag) as well as it can confuse developers if they forget (or just don't know about it as it's imo quite surprising) about this special casing for `--dev` flag.

So personally, I would be much more happier if this were a separate flag or other kind of switch (maybe a command in Dev cog?) so that this behavior can be controlled separately from `--dev` flag.

I'm looking forward to opinions from other org members on this, so we'll see how that ends up.

TrustyJAID:
You have some good points there Jack. I am also in favour of this being added to core but if I'm developing cogs with dev enabled I likely want to ensure that the cooldown is actually working instead of bypassing and then explicitly ignore it later. Maybe as a bot level toggle in `[p]set ownercooldown`?

Jackenmen:
> Maybe as a bot level toggle in [p]set ownercooldown?

IMO this would not be a bad solution, however, I think it should say that it's a development setting (or maybe it should just be only available when running with `--dev` flag?), because I think there are cogs for which it's somewhat important that bot owner is included in people that get the cooldown. I'm thinking out loud here so maybe I'm overthinking it, not sure.

TrustyJAID:
I see it being more versatile as a bot owner to debug things in prod rather than during development tbh so having it tied to the `--dev` flag is not ideal because then it would make me want to enable the dev flag in prod more often just to bypass cooldowns when attempting to fix a command for someone. I think of this as something akin to the new sudo and unsudo PR. Actually maybe blocking this PR behind that one would be better so that it only works when in sudo mode.

Jackenmen:
Putting this behind sudo is not ideal as it makes it not possible to use this feature when you don't have sudo functionality enabled at all (the default). I, for example, will definitely not have sudo functionality enabled on my development bots.

TrustyJAID:
hmm I see I didn't realize sudo and unsudo was another CLI flag option. Well in that case this would be better suited as another CLI option separate from dev imo. Keeps the option there for bots that want to utilize it without being tied directly to the `--dev` flag.

Twentysix26:
Having it bypass cooldowns when --dev is enabled is not good, at best I would be mildly annoyed because I can't properly test cooldowns without disabling dev mode, at worst I wouldn't know about it in the first place and think that my code is faulty.
I also don't think it should be an option *outside* dev mode: cooldowns are usually in place for very good reasons and having it outside dev mode gives the end user the impression that it's ok to disable them.
IMO this would best be implemented as a command in the dev cog and without any sort of persistence: when the bot restarts the cooldowns will start working again.

flaree:
I've gone ahead and changed this to a non persistent attribute which a command in dev can toggle

Jackenmen:
I'll be moving this PR forward based on what's been suggested by Twentysix (and what is also the current implementation in this PR) if I don't hear any feedback opposing the proposed solution by December 17 (Thursday).

Jackenmen (changes requested):
Currently, due to use of `ctx.reinvoke()` this can cause max concurrency check to be skipped completely if the command is on cooldown.
Before and after invoke hooks also don't get ran if you don't use `call_hooks=True`, but because there's no way to make `reinvoke()` run the max concurrency check, I think we will need to use `ctx.reset_cooldown()` + `bot.invoke(ctx)` combo instead.

Jackenmen (approved):
Looks good to me, sorry it took so long to get in the review.

PR #4443, opened 2020-09-22
Add bundled install extras.

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Allow red and all its dependencies to be installed with `pip install Red-DiscordBot[all]`

### DIFF
--- a/redbot/core/events.py
+++ b/redbot/core/events.py
@@ -119,15 +119,14 @@ def init_events(bot, cli_flags):
                 if expected_version(current_python, py_version_req):
                     installed_extras = []
                     for extra, reqs in red_pkg._dep_map.items():
-                        if extra is None:
+                        if extra is None or extra in {"dev", "all"}:
                             continue
                         try:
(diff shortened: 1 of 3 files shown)


Jackenmen (changes requested):
As discussed on Discord, the bot's update message needs to be changed so that it doesn't check for `all` and `dev` extras. Otherwise,  users that just happened to install all extras covered by `all` rather than explicitly asking for `all` would get `all` extra in the update message.

Drapersniper:
> As discussed on Discord, the bot's update message needs to be changed so that it doesn't check for `all` and `dev` extras. Otherwise, users that just happened to install all extras covered by `all` rather than explicitly asking for `all` would get `all` extra in the update message.

Done

Jackenmen (approved):
I made an optimization that makes this not call `pkg_resources.require()` on `dev` and `all` extras at all, so LGTM now.

PR #4449, opened 2020-09-28
Remove protecting commands and cogs being removed with bot_remove_func

Stonedestroyer:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Changes ruledropper so commands can always be removed and cogs as well, doesn't touch permisssions or disabling cogs. So functionality.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1113,7 +1113,7 @@ class RedBase(
 
     def remove_cog(self, cogname: str):
         cog = self.get_cog(cogname)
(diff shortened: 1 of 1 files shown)


TrustyJAID (commented):
This is a good change to prevent unexpected behavior on removing commands even if they shouldn't be removed. Currently the bot allows anyone to create a command that cannot be removed simply and this fixes that in a way  that is expected.

Drapersniper (approved):
LGTM no issues found here.

PR #4463, opened 2020-10-01
Handle verified bots, improve errors in `[p]set username`

Kowlin:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

Stop the bot from returning an false error and add a fail safe for verified bots.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1949,10 +1949,10 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
                 _(
                     "Failed to change the username. "
                     "Discord returned the following error:\n"
-                    "``{}``\n\n"
+                    "{error_message}\n\n"
                     "Remember that you can only do it up to 2 times an hour."
-                    " Use nicknames if you need frequent changes. `{}set nickname`"
-                ).format(error_string, ctx.clean_prefix)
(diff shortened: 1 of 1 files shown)


Jackenmen (dismissed):
It's not perfect that we're showing an error message like this, but I guess it will have to do. With that said, could you have the exception handler check if `e.code` is equal to `50035`?

Kowlin:
Well here is the thing. I have. And it's no where near perfect since EVERYTHING that errors out with change nickname returned a 50035... And that's honestly driving me insane

Jackenmen:
@Kowlin I went ahead and updated the code myself, could you say if it looks okay to you? I used this occasion of making more specific errors in `[p]set username` to also add a timeout to setting username so that the error about being able to change username up to 2 times an hour could make some sense.

Here's the commit with the changes I made:
[`d1a8bed` (#4463)](https://github.com/Cog-Creators/Red-DiscordBot/pull/4463/commits/d1a8bedb14b680378b43e9f3d0449cbf9ea684d9)

Jackenmen (approved):
LGTM, I got a go-ahead from Kowlin too:
image

Jackenmen:
And here are some tanks to make up for Kowlin not sending one:
tenor (1)

PR #4465, opened 2020-10-04
Raise on an uncompliant message in Context.maybe_send_embed()

### LINKED ISSUE #4383: Maybe_send_embed inconsistent behavior for cog developers
# Other bugs

Use `maybe_send_embed` to send an empty string

Consistent behavior with an error, failing to send an empty message per:
`discord.errors.HTTPException: 400 Bad Request (error code: 50006): Cannot send an empty message`

#### What actually happened?

Only caused an error when embeds were off, no error when embeds were on.

`[p]eval await ctx.maybe_send_embed("")`
`[p]embedset global`
`[p]eval await ctx.maybe_send_embed("")`

Dav-Git:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #4383

### DIFF
--- a/redbot/core/commands/context.py
+++ b/redbot/core/commands/context.py
@@ -248,8 +248,11 @@ class Context(DPYContext):
             see `discord.abc.Messageable.send`
         discord.HTTPException
             see `discord.abc.Messageable.send`
+        ValueError
+            when messages length is not between 1 and 2000.
         """
-
+        if not message or len(message) > 2000:
+            raise ValueError("Message lenght must be between 1 and 2000")
         if await self.embed_requested():
             return await self.send(
                 embed=discord.Embed(description=message, color=(await self.embed_colour()))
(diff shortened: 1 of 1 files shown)


redbot/core/commands/context.py
```diff
@@ -248,8 +248,11 @@ async def maybe_send_embed(self, message: str) -> discord.Message:
             see `discord.abc.Messageable.send`
         discord.HTTPException
             see `discord.abc.Messageable.send`
+        ValueError
+            when messages length is not between 1 and 2000.
```
Flame442:
```suggestion
            when message's length is not between 1 and 2000 characters.
```

redbot/core/commands/context.py
```diff
@@ -248,8 +248,11 @@ async def maybe_send_embed(self, message: str) -> discord.Message:
...
         """
-
+        if not message or len(message) > 2000:
+            raise ValueError("Message lenght must be between 1 and 2000")
```
Flame442:
```suggestion
            raise ValueError("Message length must be between 1 and 2000 characters.")
```

Flame442 (changes requested):

Flame442 (approved):

PR #4467, opened 2020-10-04
Add a setting for adding a tick reaction when help is DMed

### LINKED ISSUE #4075: [Help] Give confirmation on help being sent to users DMs
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality

A modification to the `[p]help` command when it is set to only send the help messages via DMs. 

#### What would you like it do do?
I think either responding in the channel it was called with a message like "Help was successfully sent to DMs" or even a non-obtrusive method like applying a checkmark reaction to the [p]help message upon sending successfully. If the first option is doable, make it an option in `[p]helpset`.

#### The reasoning
I feel like the users that don't call the command get confused when no help is shown in the channel when the command is called by another user. I don't really want to disable sending help via dm because the command list is so big. I am aware of menus but I kind of don't want to use them because of their timeout (I know this is soon to be fixed however). This was originally discussed on Discord here: https://discordapp.com/channels/133049272517001216/387398816317440000/732765573221580851

palmtree5:
Fixes #4075

Opted to do this just by adding a tick after help is sent to whatever destination it's sent to but if we want to do it differently, we can discuss


phenom4n4n:
This should be added within Red's Help Formatter implementation and not this command so that it is overridable.

Drapersniper:
I personally would like to see a toggle under helpset to enable/disable the core behaviour on this too.

Jackenmen:
Yeah, I think the toggle would be reasonable. I also think it would be better if we only ticked when the help is sent to DMs rather than always.

palmtree5:
I've implemented this as a toggleable setting, defaulting to being off. It should only tick if help was sent to DMs and this setting is enabled

Jackenmen (approved):
LGTM

PR #4468, opened 2020-10-05
[Core] Strip commas on cog command when pagified

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

If you use [p]cogs when the loaded or unloaded content is pagified, the pages will start with `, ` if you are using embeds. This uses the same stripping that non-embed cog pages have already.


Jackenmen (approved):
It's a bit unfortunate that `pagify()` splits in a way that makes the delimiter show up at the beginning rather than the end, but we can't really do anything about it *in this PR* so LGTM.

PR #4469, opened 2020-10-05
[Core] Pagify cog unload output properly

aikaterna:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

When you unload a large amount of cogs (close to message limit on characters), the character count is easily pushed over 2k characters on the command response with the cog names each being wrapped in 2x backticks. The current implementation breaks in the middle of cog names or wherever else it feels like it. 1600 might have been a safe value for splitting but at 1500 I don't see how multiple short cog names could make it break at least.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1385,7 +1385,9 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
 
         if output:
             total_message = "\n\n".join(output)
-            for page in pagify(total_message):
+            for page in pagify(total_message, delims=[", "], page_length=1500):
+                if page.startswith(", "):
+                    page = page[2:]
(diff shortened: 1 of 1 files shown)


Jackenmen:
I don't think the page length should really matter that much as the issue was that this didn't split on commas. I felt like mentioning it but I don't think it hurts to have a bit lower page length either :)

Anyway, I have a suggestion: this could have both `\n` and `,` set as delimiters and it could also use the `priority=True` kwarg so that it favors `\n` over `,`. What do you think?

Jackenmen (approved):
It's a bit unfortunate that `pagify()` splits in a way that makes the delimiter show up at the beginning rather than the end, but we can't really do anything about it *in this PR* so LGTM.

PR #4472, opened 2020-10-06
Use `bot.get_user()` instead of `get_all_members()` in `[p]dm`

PredaaA:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This just avoid iterate all guild members while we can just use get_user directly.


Jackenmen (approved):
LGTM

PR #4474, opened 2020-10-08
[General] Fix error on long titles in `[p]urban`

MeatyChunks:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

There's a few titles that are greater than 256 characters which causes a console error. The example I saw was scrolling through the pages of `[p]urban Jackin' Off` which leads to http://synonyms-chokin-the-chicken-spanking-the-monkey-flogging-the-do.urbanup.com/3169366


Jackenmen (approved):
Looks good to me, thanks!

PR #4479, opened 2020-10-10
Add `delete()` method to `Context`

phenom4n4n:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Adds a `ctx.delete()` method which suppresses discord.NotFound and other exceptions, for ease of use.

### DIFF
--- a/redbot/core/commands/context.py
+++ b/redbot/core/commands/context.py
@@ -138,7 +138,7 @@ class Context(DPYContext):
         Returns
         -------
         bool
-            :code:`True` if adding the reaction succeeded.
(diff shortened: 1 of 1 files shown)


Jackenmen:
This should be called something like `delete_quietly()` rather than just `delete()` to avoid potential confusion as well as the possibility of the need to add `delete()` for something else (possibly even by upstream d.py commands extension which we can never completely rule out).

I don't yet know if this is worth including in Red Core (for example, how often could the code be made simpler with an addition like this realistically), but I just wanted to get what I said in the first paragraph out of the way.

PR #4480, opened 2020-10-10
Update Windows instructions to not install Python 3.9

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Tested on Win 10 Pro VM.
This will obviously fail if someone installed Python 3.9 already with Chocolatey already, but we can't really do much about that (other than passing `--allow-downgrade` flag which isn't a good idea...)


ARealWant:
Maybe you should trying to use an older Python Version, they're definetly some packages, which arent compatible.

Drapersniper (approved):
LGTM

PR #4481, opened 2020-10-10
Fix `[p]cleanup self` not working in DMs for non-owners

### LINKED ISSUE #4408: `[p]cleanup self` doesn't work in DMs if you're not a bot owner
# Command bugs

#### Command name

`[p]cleanup self`

Cleanup

The bot to clean one message.

#### What actually happened?

Nothing.

1. `[p]load cleanup` from bot owner account.
2. Use `[p]ping` and then `[p]cleanup self 1` from regular account in DMs with the bot

madebylydia:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

This fix #4408.

### DIFF
--- a/redbot/cogs/cleanup/converters.py
+++ b/redbot/cogs/cleanup/converters.py
@@ -1,6 +1,6 @@
 from typing import NewType, TYPE_CHECKING
 
-from redbot.core.commands import BadArgument, Context, Converter, check
+from redbot.core.commands import BadArgument, Context, Converter, permissions_check
 from redbot.core.utils.mod import is_mod_or_superior, check_permissions
 from redbot.core.i18n import Translator
 from redbot.core.utils.chat_formatting import inline
@@ -41,4 +41,4 @@ def check_self_permissions():
             return True
         return False
 
-    return check(predicate)
+    return permissions_check(predicate)
--- a/redbot/cogs/cleanup/converters.py
+++ b/redbot/cogs/cleanup/converters.py
@@ -1,6 +1,7 @@
 from typing import NewType, TYPE_CHECKING
 
-from redbot.core.commands import BadArgument, Context, Converter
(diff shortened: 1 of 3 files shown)


Jackenmen (changes requested):
This PR allows any user to clean bot's messages in the guilds which shouldn't be the case.

redbot/cogs/cleanup/converters.py
```diff
@@ -1,6 +1,7 @@
 from typing import NewType, TYPE_CHECKING
 
-from redbot.core.commands import BadArgument, Context, Converter
+from redbot.core.commands import BadArgument, Context, Converter, check
```
Jackenmen:
```suggestion
from redbot.core.commands import BadArgument, Context, Converter, permissions_check
```

redbot/cogs/cleanup/converters.py
```diff
@@ -28,3 +29,16 @@ def positive_int(arg: str) -> int:
...
+    async def predicate(ctx: Context):
+        if not ctx.guild:
+            return True
+        if await check_permissions(ctx, {"manage_messages": True}) or await is_mod_or_superior(
+            ctx.bot, ctx.author
+        ):
+            return True
+        return False
+
+    return check(predicate)
```
Jackenmen:
```suggestion
    return permissions_check(predicate)
```

Jackenmen (changes requested):
Seems good now, but one thing I would want to ask here is to use `permission_check()` rather than `check()` here. The difference between them is that `permissions_check()` can be overridden by Permissions cog, which is similar to how `mod_or_permissions()` works for example.

Usage of `permissions_check()` is exactly the same as the usage of `check()`, so this is just a matter of replacing the two occurrences as presented in code suggestions below.

Jackenmen (approved):
LGTM

PR #4483, opened 2020-10-12
[Cleanup]adds handling for messages cont. emojis/links

kablekompany:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

`[p]cleanup emoji`
`[p]cleanup links`

each default to searching for 10 messages respectively

Blocked by #4728 (`[p]cleanup stickers`)


kablekompany:
Have made this change on my own red, and can confirm no false positives as of yet on link or emoji clearing (unicode emojis of course are not purged). Initially I had an alias with `cleanup text ":" to sort of target either but couldn't settle for one or the other, and it would always clean both. This has been tested in a moderation environment where chat speed is variable of 1-2msgs per second with enforced slowmode of 30s (so you can imagine the spam, ads, emoji overuse encountered). 

Just wanted to validate positive test to date and provide operational stroke of proposed addition, regardless if there is intent or desire for use. 

Thanks!

redbot/cogs/cleanup/cleanup.py
```diff
@@ -10,12 +10,18 @@
...
 from redbot.core.utils.mod import slow_deletion, mass_purge
 from redbot.core.utils.predicates import MessagePredicate
+from .checks import check_self_permissions
 from .converters import PositiveInt, RawMessageIds, positive_int
 
 _ = Translator("Cleanup", __file__)
 
 log = logging.getLogger("red.cleanup")
 
+LINKS = re.compile(r"http[s]?://(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*(),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+")
```
Jackenmen: We have a regex for links in `redbot.core.utils.common_filters.URL_RE` which could be used here.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -115,13 +121,13 @@ def message_filter(message):
         return collected
 
     @commands.group()
-    @checks.mod_or_permissions(manage_messages=True)
```
Jackenmen: Could you merge current `V3/develop` into this branch so that these differences don't show up in this diff?

Jackenmen (commented):
This is not a review, I just had two things I wanted to mention. Feel free to ignore this until actual review comes up.

kablekompany:
applied the noted changes (i think) 
let me know if that diff is fucky (git is confusing for me at times 😓 )

kablekompany:
moved emoji regex to core and added unicode emoji regex extracting to core as well
Can revert if these properties are not sufficient for core otherwise. 

making the unicode regex accessible could open up use cases in automod filters or further expanding cleanup functions.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -517,7 +512,6 @@ def check(m):
         await mass_purge(to_delete, channel)
 
     @cleanup.command(name="self")
-    @check_self_permissions()
```
Drapersniper: I dont think this was intentional, this is the reason that `    @checks.mod_or_permissions(manage_messages=True)` isnt added to the group like you done in this PR. I dont know enough about this check to comment on why just something i saw when skimming through it.

Jackenmen: Looks to me like a merge that went south. The checks in cleanup cog were touched somewhat recently and this seems to be accidentally reverting it.

kablekompany: 'Twas a revert to remove the diff from PR #4481 from @Predeactor as git was not attributing the correct blame. Thought reverting the commit of adding those and merging in would fix but ig not...
oh boi guise i'm unsure how to fix without closing and reopening with a with fresh V3/develop base 😟

kablekompany: Let me know if that's best course, and I'll dip this PR, burn the branch, base from current V3/develop, write in the changes, and open new PR.

Sorry! ha

kablekompany:
successfully got the git blame back on track

kablekompany:
Added support for cleanup stickers 

Cause those were a fantastic idea and definitely not spammed :XtoDoubt:

kablekompany:
Will discontinue pulling in base unless otherwise expressed this is a potential wanted merge. Just to save us all some email notifs 😅 

will rectify a final updated upstream when/if that review is requested. Cheers!

PR #4484, opened 2020-10-12
Update install_windows.rst

Turphin:
change all mentions of python with a version number to python 3.9
line 114 needed this change if someone installed redbot without any prior knowledge of different python versions or how to use python commands

### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature


flaree:
Python 3.9 isnt supported as of yet. 3.8.1 is still the lowest version required.

Jackenmen:
Red currently does NOT support Python 3.9 and there is already a PR that updates Windows install instructions to always install Python 3.8 (#4480).

PR #4485, opened 2020-10-13
owner check

phenom4n4n:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Adds a guild owner check to `is_mod_or_superior` and `is_admin_or_superior`.

### DIFF
--- a/redbot/core/utils/mod.py
+++ b/redbot/core/utils/mod.py
@@ -145,9 +145,7 @@ async def is_mod_or_superior(
     else:
         raise TypeError("Only messages, members or roles may be passed")
 
-    if await bot.is_owner(user):
-        return True
-    if await bot.is_mod(user):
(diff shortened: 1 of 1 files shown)


redbot/core/utils/mod.py
```diff
@@ -222,9 +220,7 @@ async def is_admin_or_superior(
     else:
         raise TypeError("Only messages, members or roles may be passed")
 
-    if await bot.is_owner(user):
-        return True
-    if await bot.is_admin(user):
+    if user == user.guild.owner or await bot.is_owner(user) or await bot.is_admin(user):
         return True
 
     return False
```
bobloy, line 226:
```suggestion
    return user == user.guild.owner or await bot.is_owner(user) or await bot.is_admin(user)
```

Looks like each of these could be simplified to just return the bool

PR #4486, opened 2020-10-13
Move modlogset/bankset commands to core

### LINKED ISSUE #3709: Move `is_owner_if_bank_global()` check to `redbot.core.bank`
# Feature request

- [ ] Cog
- [ ] Command
- [x] API functionality

`is_owner_if_bank_global()` check is currently residing in `redbot.cogs.bank`:
https://github.com/Cog-Creators/Red-DiscordBot/blob/2fe6ae1d0601232b7992fe5a2c14a16a506d3bbd/redbot/cogs/bank/bank.py#L12-L31

It's proven to be useful in multiple cogs in core (we import it in Bank, Economy and Trivia) and some cog creators (even though it's not guaranteed by our version guarantees) also use it in their 3rd-party cogs (for example - Adventure cog) so it might be a good addition to Bank API.
It's really simple for cog creators to implement it themselves but since it could be considered generally useful, it might be a good idea to have it in core api.

We could also add some variations of it as well, but first we need to decide whether we want it at all.

### LINKED ISSUE #4128: Move commands for changing Bank and ModLog API settings to `[p]set`
To be specific:
- `[p]modlogset` from `modlog` cog should be moved to `[p]set modlog`
- `[p]bankset` from `bank` cog should be moved to `[p]set bank` (as part of this, `bank` cog would be removed)
- `[p]economyset registeramount` from `economy` cog should be moved to `[p]set bank registeramount`
- (this is a maybe) `[p]bank reset` and `[p]bank prune` should be made part of `[p]set bank`

IMO, they shouldn't be part of any cog since they allow to change settings of a core API, same as `[p]autoimmune`, `[p]whitelist`, or `[p]embedset` that also allow users to change API settings.

Relevant discussion can be found in: #4067 

This is currently planned for 3.5.0 release as it's going to break people's usage of permissions.

palmtree5:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Fixes #4128 
Fixes #3709

This accomplishes everything on the list in the mentioned issue, though I did this by simply moving the commands over. The only renames happening here are:

- `[p]economyset registeramount` -> `[p]bankset registeramount`
- `[p]bank reset` -> `[p]bankset reset`
- `[p]bank prune` -> [p]bankset prune`

This obviously removes the `Bank` cog. The `Modlog` cog remains with 3 commands (`[p]case`, `[p]casesfor`, and `[p]reason`)

### Other changes

- added `RawUserIdConverter` to `redbot.core.commands`
- added `is_owner_if_bank_global` to `redbot.core.bank` (#3709)


Dav-Git (changes requested):
I don't think this move serves it's intended purpose without grouping all commands under ``set``. The idea was to have all settings in one place which this PR doesn't really accomplish now.
If the commands were moved into the ``set`` command group we would have achieved that goal.

Drapersniper:
There's such a thing as having too many commands, when the audio api is done for example it will not be under set due to the number of commands it will introduce. I disbelieve that every single global configuration command must be under set... that becomes a mess of a user experience as there's too many commands under a single group to deal with

palmtree5:
I agree with Draper. Commands for configuring APIs on the bot (like modlog and bank) should absolutely be in the core cog, but I also feel they should not all be under `[p]set` as that will result in a further mess (which is already the case there meaning we may wish to do some regrouping of `[p]set`'s subcommands either by creating command groups under it or splitting things out of `[p]set` into their own groups to group relevant commands together)

Dav-Git:
That's a good point. A regrouping of set commands to sort them by core API /feature they belong to makes sense. However with these groups any commands that control core or API settings need to be in ``set``

palmtree5:
So this PR was already a breaking change for the command moves and renames but initially there was an issue with tests failing, necessitating moving `redbot.cogs.mod.converters.RawUserIds`. I wound up moving that to `redbot.core.commands.converter.RawUserIds` and getting rid of `redbot.cogs.mod.converters` meaning that particular converter would be covered in our version guarantees moving forward.

palmtree5:
Merge conflicts have been fixed on this one. Should hopefully have not broken anything either but still probably needs testing to be sure of that

palmtree5:
Ok, NOW the merge conflicts have actually been fixed (after I managed to miss a couple trying to resolve in vscode lol).

redbot/core/core_commands.py
```diff
@@ -1969,6 +1971,343 @@ async def _restart(self, ctx: commands.Context, silently: bool = False):
...
+        )
+
+    @bank.is_owner_if_bank_global()
+    @checks.guildowner_or_permissions(administrator=True)
+    @bankset.command()
+    async def registeramount(self, ctx: commands.Context, creds: int):
+        """Set the initial balance for new bank accounts.
+
+        Example:
+            - `[p]economyset registeramount 5000`
```
aikaterna: This should be `[p]bankset registeramount 5000`

aikaterna:
There was some talk in AC this morning about some setting info that is duplicated in both bankset showsettings and economyset showsettings (max balance and default initial balance) where they should be removed from the economy settings and just have them available in bank settings.

https://discord.com/channels/133049272517001216/718148684629540905/868174816082006036
https://discord.com/channels/133049272517001216/718148684629540905/868181565878960168

Jackenmen:
@palmtree5 If you have some time to resolve the conflicts that have shown up here since May, I think I could go ahead with getting this reviewed and merged soon.
Let me know if you won't have time for this and I'll try finding some time to resolve these conflicts myself.

palmtree5:
I'll try to look at this in the next couple days

palmtree5:
@jack1142 conflicts have been resolved on this and checks have passed

Jackenmen (approved):
LGTM

PR #4492, opened 2020-10-15
Audio Fixes

### LINKED ISSUE #4489: Audio bug
Fix the bug

### LINKED ISSUE #4491: Help of `[p]audioset globalapi toggle` is wrong
# Command bugs

#### Command name

`[p]help audioset globalapi toggle`

Audio

I expected default to be OFF according to help.

#### What actually happened?

Default is ON according to help.
image

`[p]help audioset globalapi toggle`

### LINKED ISSUE #4495: Audio doesn't work with i18n
# Other bugs

Use Audio while having a non-English locale set.

Strings to be translated (if the translation is available of course, but it is in this case).

#### What actually happened?

Strings are not translated.

1. Set the bot to some locale with big coverage, e.g. French, German, Polish, Russian
2. Try using few commands in Audio.
3. See that strings aren't getting translated.

---

The reason for this is the way `Translator` object is created in Audio cog vs how the po files are actually structured. Audio creates a single `Translator` object at `redbot.core.audio.core.cog_utils` that refers to `po` files in `redbot/core/audio/locales` folder. Meanwhile, the actual translation can exist in any of:
```
redbot/cogs/audio/locales
redbot/cogs/audio/apis/locales
redbot/cogs/audio/core/locales
redbot/cogs/audio/core/commands/locales
redbot/cogs/audio/core/events/locales
redbot/cogs/audio/core/tasks/locales
redbot/cogs/audio/core/utilities/locales
```
depending on where the string actually resides. The solution can for example be creating a `Translator` object in each file that needs translations (as do other cogs) rather than rely on a single one, since in such case its path is inaccurate.

Drapersniper:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

resolves #4491
Typing indicators on playlist commands.
Fix a bs edge case that i can't reproduce (in theory) as now I ensure the player lock is always released upon an exception/error
Fix yet another bs edge case, where setting a new token doesnt update audio properly for the global api.
resolves #4495

Theoretically this resolves #4489 as I cannot reproduce it even with the exact steps the one user who hit this said they done.

### DIFF
--- a/redbot/cogs/audio/apis/api_utils.py
+++ b/redbot/cogs/audio/apis/api_utils.py
@@ -3,18 +3,21 @@ import json
 import logging
 from collections import namedtuple
 from dataclasses import dataclass, field
+from pathlib import Path
 from typing import List, MutableMapping, Optional, Union
 
 import discord
 import lavalink
 
 from redbot.core.bot import Red
+from redbot.core.i18n import Translator
 from redbot.core.utils.chat_formatting import humanize_list
 
 from ..errors import InvalidPlaylistScope, MissingAuthor, MissingGuild
 from ..utils import PlaylistScope
 
 log = logging.getLogger("red.cogs.Audio.api.utils")
+_ = Translator("Audio", Path(__file__))
 
 
 @dataclass
--- a/redbot/cogs/audio/apis/global_db.py
+++ b/redbot/cogs/audio/apis/global_db.py
@@ -38,8 +38,9 @@ class GlobalCacheWrapper:
         self._token: Mapping[str, str] = {}
         self.cog = cog
 
-    def update_token(self, new_token: Mapping[str, str]):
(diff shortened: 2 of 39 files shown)


Jackenmen:
The solution used for #4495 is correct, I did quick testing of it and it worked as intended. I haven't looked if each file with the usage of `_()` uses its own Translator object now, but considering there are no ImportErrors, I can suspect it's all good.

Drapersniper:
Commit sucks ... 
image

aikaterna (commented):
This is a provisional approval for everything in this PR. There may be some additional error handling added after this comment, which is why this isn't a true approval.

Drapersniper:
Updated Errors as discussed with Aika.
image

aikaterna (approved):

PR #4493, opened 2020-10-15
[Core] added some missing dots.

ltzmax:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Added some missing dots some places.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3238,7 +3238,7 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         if await ctx.bot.is_automod_immune(user_or_role):
             await ctx.send(_("They are immune."))
         else:
-            await ctx.send(_("They are not Immune."))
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -3236,9 +3236,9 @@ async def autoimmune_checkimmune(
         """
 
         if await ctx.bot.is_automod_immune(user_or_role):
-            await ctx.send(_("They are immune"))
+            await ctx.send(_("They are immune."))
         else:
-            await ctx.send(_("They are not Immune"))
+            await ctx.send(_("They are not Immune."))
```
Stonedestroyer: This should probably be lower-case i and not upper-case.

ltzmax: fixed. 👍

Drapersniper (approved):
DOOOOOOOOOOOOOOOOTS, lgtm thanks Max!

PR #4501, opened 2020-10-17
Improve grammar in max concurrency error

phenom4n4n:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Adds proper casing for max concurrency reached instead of a lazy "(s)" and proper grammar for when the concurrency is set to users or members.


redbot/core/events.py
```diff
@@ -297,16 +297,22 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
             if error.per is commands.BucketType.default:
                 msg = _(
                     "Too many people using this command."
                     " It can only be used {number} time(s) concurrently."
-                ).format(number=error.number)
+                ).format(number=number)
+            if error.per in (commands.BucketType.user, commands.BucketType.member):
+                msg = _(
+                    "You are using this command too many times."
+                    " It can only be used {number} time(s) per {type} concurrently."
```
Drapersniper:
```py
"time(s)"
``` 
is not really translation friendly, that said it is already present in existing code.

```py
"You are using this command too many times."
``` 
feels odd. Its not that they are using it too many times, its that they are using it too many times before the previous complete, the sentence above could also apply to cooldowns etc.

Jackenmen (approved):
With a little help from Bread and Predeactor, I changed the messages a bit more and this should now be ready to go, thanks for the PR :)

PR #4488, opened 2020-10-14
[Docs] Cleanup Cog Guide

bobloy:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

### Description of the changes

Adds the cleanup cog guide to the docs. Based on the descriptions provided in the commands.
Personalization may be necessary, feedback on that would be helpful.

PR #4511, opened 2020-10-19
[Docs] Downloader cog guide

bobloy:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Adds programmatically generated docs using the same method as #4488
A step towards the #1734 goal


docs/cog_guides/downloader.rst
```diff
@@ -0,0 +1,547 @@
...
+.. code-block:: none
+
+    [p]cog reinstallreqs 
+
+**Description**
+
+This command will reinstall cog requirements and shared libraries for all installed cogs.
+
+Red might ask user to use this when it clears contents of lib folder
+because of change in minor version of Python.
```
Flame442:
```suggestion
because of a change in the minor version of Python.
```

docs/cog_guides/downloader.rst
```diff
@@ -0,0 +1,547 @@
...
+**Description**
+
+List information about a single cog.
+
+Example:
+    - ``[p]cog info 26-Cogs defender``
+
+**Arguments**
+
+- ``<repo_name>`` The repo to get cog info from..
```
Flame442:
```suggestion
- ``<repo_name>`` The repo to get cog info from.
```

docs/cog_guides/downloader.rst
```diff
@@ -0,0 +1,547 @@
...
+
+Install a cog from the specified revision of given repo.
+
+Example:
+    - ``[p]cog installversion Broken-Repo e798cc268e199612b1316a3d1f193da0770c7016 cog_name``
+
+**Arguments**
+
+- ``<repo_name>`` The name of the repo to install cogs from.
+- ``<revision>`` The revision to install from.
```
Flame442, line 152: Expanding what a "revision" is could be useful since end users probably don't know much about git.

docs/cog_guides/downloader.rst
```diff
@@ -0,0 +1,547 @@
...
+
+.. code-block:: none
+
+    [p]cog reinstallreqs 
+
+**Description**
+
+This command will reinstall cog requirements and shared libraries for all installed cogs.
+
+Red might ask user to use this when it clears contents of lib folder
```
Flame442:
```suggestion
Red might ask the owner to use this when it clears contents of the lib folder
```

docs/cog_guides/downloader.rst
```diff
@@ -0,0 +1,547 @@
...
+
+**Syntax**
+
+.. code-block:: none
+
+    [p]cog unpin <cogs>
+
+**Description**
+
+Unpin cogs - this will remove update lock from cogs.
```
Flame442:
```suggestion
Unpin cogs - this will remove the update lock from those cogs.
```

docs/cog_guides/downloader.rst
```diff
@@ -0,0 +1,547 @@
...
+
+Update all cogs, or ones of your choosing.
+
+Examples:
+    - ``[p]cog update``
+    - ``[p]cog update defender``
+
+**Arguments**
+
+- ``<cogs>`` The cog or cogs to update. If omitted, all cogs are updated.
```
Flame442: Should this be using square brackets? Is this a mistake of the auto-generator?

docs/cog_guides/downloader.rst
```diff
@@ -0,0 +1,547 @@
...
+
+    [p]cog updatetoversion <repo_name> <revision> [cogs]
+
+**Description**
+
+Update all cogs, or ones of your choosing to chosen revision of one repo.
+
+Note that update doesn't mean downgrade and therefore revision
+has to be newer than the one that cog currently has. If you want to
+downgrade the cog, uninstall and install it again.
```
Flame442:
```suggestion
Note that update doesn't mean downgrade and therefore `revision`
has to be newer than the version that cog currently has installed. If you want to
downgrade the cog, uninstall and install it again.
```

docs/cog_guides/downloader.rst
```diff
@@ -0,0 +1,547 @@
...
+
+Example:
+    - ``[p]cog updatetoversion Broken-Repo e798cc268e199612b1316a3d1f193da0770c7016 cog_name``
+
+**Arguments**
+
+- ``<repo_name>`` The repo or repos to update all cogs from.
+- ``<revision>`` The revision to update to.
```
Flame442, line 345: Same note about "revision" as above.

docs/cog_guides/downloader.rst
```diff
@@ -0,0 +1,547 @@
...
+
+**Syntax**
+
+.. code-block:: none
+
+    [p]repo update [repos...]
+
+**Description**
+
+Update all repos, or ones of your choosing.
```
Flame442, line 526: Maybe add a note that this does **not** update the installed cogs?

Flame442 (changes requested):
A few things that need fixing, but this is looking good otherwise!
Not sure what is up with the argument brackets, but they don't seem consistent.
All of the changes in the rst need to be applied to the cog as well where applicable.

docs/cog_guides/downloader.rst
Flame442: While I do think this grammar change should be done in the docstring, this command should **not** be included in the docs  as it is marked as hidden.

Jackenmen: The command is not actually hidden, I've mistakenly told Flame that it is.
Nonetheless, I still don't think the command should really be shown in the user guide as it's generally only supposed to be used if Red asks you to and it's not a command that the user really needs to learn about from this guide.

bobloy: Would it be more correct to make the command hidden? I will remove from the guide regardless.

docs/cog_guides/downloader.rst
bobloy: Added an explanation of revision with a link and an image. It still makes assumptions that the user know what they're doing though.

docs/cog_guides/downloader.rst
bobloy: Also added a comment that this command shouldn't be used unless Red asks for it.

Jackenmen: Seems fine to me to hide it.

docs/cog_guides/downloader.rst
bobloy: Swapped to square brackets and reviewed all other bracket to make sure the match the `usage`

docs/cog_guides/downloader.rst
bobloy: Added a comment pointing to `cog installversion` to get an explanation of revision

docs/cog_guides/downloader.rst
bobloy: Comment added below this line

bobloy:
Requested changes have been made.

docs/cog_guides/downloader.rst
bobloy: Added PR to hide this in #4590

Flame442 (approved):
Thanks for your help!

PR #4512, opened 2020-10-19
[Streams] Remove the `__del__`

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This appears to no longer be needed, it also in fact does not process the `cog_unload` function.

### DIFF
--- a/redbot/cogs/streams/__init__.py
+++ b/redbot/cogs/streams/__init__.py
@@ -4,8 +4,3 @@ from .streams import Streams
 def setup(bot):
     cog = Streams(bot)
     bot.add_cog(cog)
-
-
-def teardown(bot):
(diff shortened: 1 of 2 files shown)


palmtree5 (commented):
Went ahead and added a teardown function to call that cog unload function as that function is used to cancel the looping task for stream alerts.

Rapptz:
There's no need to do this since cog_unload is automatically called by the library, calling again won't really accomplish much.

palmtree5:
> There's no need to do this since cog_unload is automatically called by the library, calling again won't really accomplish much.

yeah, I somehow managed to forget about that detail, thanks for the reminder. Reverted my commit

Drapersniper (approved):

PR #4517, opened 2020-10-21
Allow `menu()` to listen to both adding and removing reactions

Drapersniper:
This will reduce the number of API calls the bot does when it has `manage_message` permissions as it will no longer make API calls to unreact.
The menu will accept unreact as a valid input in the menu, bringing it in like with Toby and Dpy behaviour.

### DIFF
--- a/redbot/core/utils/menus.py
+++ b/redbot/core/utils/menus.py
@@ -88,11 +88,20 @@ async def menu(
             return
 
     try:
-        react, user = await ctx.bot.wait_for(
-            "reaction_add",
-            check=ReactionPredicate.with_emojis(tuple(controls.keys()), message, ctx.author),
-            timeout=timeout,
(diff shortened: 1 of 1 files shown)


Jackenmen (commented):
I'll look into this a bit more tomorrow, but I'm giving this a "soft approval" for now.

Jackenmen (dismissed):

Jackenmen (approved):
Looks good to me, but sadly we'll have to wait with this until Red 3.5.

While the change for breakage is rather unlikely, it is not entirely impossible. For example, it is possible that something in the bot removing the user's reaction will trigger this and cause the menu to switch to the previous/next page without actual interaction from the user.

Drapersniper:
Keeping this one open  as jack already spent time on it.

PR #4525, opened 2020-10-23
[Mod] Fix + more info for modset defaultduration

aikaterna:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This fixes a text response invoked via the modset defaultduration command.
The input was also not consuming rest and would not accept values with spaces like `1 week 2 days`.
Also provides the user a more friendly response in the docstring to help figure out what's expected for the input.

### DIFF
--- a/redbot/cogs/mod/settings.py
+++ b/redbot/cogs/mod/settings.py
@@ -377,13 +377,22 @@ class ModSettings(MixinMeta):
     async def defaultduration(
         self,
         ctx: commands.Context,
-        duration: Optional[commands.TimedeltaConverter] = timedelta(days=0),
+        *,
+        duration: commands.TimedeltaConverter,
     ):
-        """Set the default time to be used when a user is tempbanned."""
+        """Set the default time to be used when a user is tempbanned.
+
+        Accepts: seconds, minutes, hours, days, weeks
+        Use `0` with a unit of measure to set the time to zero.
+        Examples:
+            `[p]modset defaultduration 0 minutes`
+            `[p]modset defaultduration 7d12h10m`
+            `[p]modset defaultduration 7 days 12 hours 10 minutes`
+        """
         guild = ctx.guild
         await self.config.guild(guild).default_tempban_duration.set(duration.total_seconds())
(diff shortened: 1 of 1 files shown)


redbot/cogs/mod/settings.py
```diff
@@ -377,13 +377,22 @@ async def defaultdays(self, ctx: commands.Context, days: int = 0):
...
+        Examples:
+            `[p]modset defaultduration 0 minutes`
+            `[p]modset defaultduration 7d12h10m`
+            `[p]modset defaultduration 7 days 12 hours 10 minutes`
+        """
         guild = ctx.guild
         await self.config.guild(guild).default_tempban_duration.set(duration.total_seconds())
         await ctx.send(
-            _("The default duration for tempbanning a user is now {duration}.").format(
+            _(f"The default duration for tempbanning a user is now {duration}.").format(
```
Jackenmen: That should not be an f-string, this format field is already handled by `.format()` + f-strings do not work with translations

Jackenmen (changes requested):

redbot/cogs/mod/settings.py
```diff
@@ -377,13 +377,22 @@ async def defaultdays(self, ctx: commands.Context, days: int = 0):
...
         ctx: commands.Context,
-        duration: Optional[commands.TimedeltaConverter] = timedelta(days=0),
+        *,
+        duration: commands.TimedeltaConverter,
     ):
-        """Set the default time to be used when a user is tempbanned."""
+        """Set the default time to be used when a user is tempbanned.
+
+        Accepts: seconds, minutes, hours, days, weeks
+        Use `0` with a unit of measure to set the time to zero.
```
Jackenmen: Should we allow to set 0 as the time? There's already softban for situations where you want to ban user and immediately unban them.
The converter allows setting the minimum duration, so this should be fairly easy and would actually simplify a few things (i.e. docstring, and the currently needed `or` in the `.format()`).

Jackenmen (approved):
Looks good to me.

PR #4529, opened 2020-10-25
[Audio] New stuff from RLL 0.7.0

Drapersniper:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Depends on https://github.com/Cog-Creators/Red-Lavalink/pull/83

This will allow for better persistent queues with RLL 0.7.0

RLL Only uses the port set with `[p]llset wsport`, `[p]llset restport` has been removed from the bot.


setup.cfg
```diff
@@ -53,7 +53,7 @@ install_requires =
     python-Levenshtein-wheels==0.13.1
     pytz==2020.1
     PyYAML==5.3.1
-    Red-Lavalink==0.6.0
+    Red-Lavalink==0.7.0
```
Drapersniper:
```suggestion
    Red-Lavalink==0.7.1
```
Bumping this to 0.7.1, ahead of the release of https://github.com/Cog-Creators/Red-Lavalink/pull/84

redbot/cogs/audio/core/commands/llset.py
```diff
@@ -188,31 +188,6 @@ async def command_llsetup_password(self, ctx: commands.Context, password: str):
                 ),
             )
 
-    @command_llsetup.command(name="restport")
```
aikaterna, line 191: Please take a look at this a little closer. I have a detailed traceback I will pass to you, but setting the wsport to something other than 2333 on an internal server causes a traceback when trying to connect with play or other audio commands.

Drapersniper: Should be resolved, issue appeart to originate from having old instance of nodes in the node pool of RLL, because `llset` commands didnt do cleanup here (nor does RLL). I added some cleanup and havent been able to  reproduce.

Drapersniper:
**VERY IMPORTANT**, with audio 2.3.0 we updated our `application.yml` in the release announcement we should point this to users and ask that users with an external instance update theirs, and ensure that the port in this file is set via `[p]llset wsport` otherwise they will face connection issues.

aikaterna (approved):

PR #4532, opened 2020-10-26
[checks] add not_scheduled check

Zephyrkul:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

Add a not_scheduled check which disallows running commands if `ctx.assume_yes` is set.

### DIFF
--- a/redbot/core/checks.py
+++ b/redbot/core/checks.py
@@ -14,6 +14,7 @@ from .commands import (
     admin_or_permissions,
     mod,
     mod_or_permissions,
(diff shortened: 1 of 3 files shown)


bobloy:
Could this PR also implement the check into some of the relevant core command or cogs?
For example: `cc create random` could use this.

Zephyrkul:
I do not currently have the time to spare to scan all the core cogs and make such judgment calls.
Besides which, such commands already have this check inside the command body itself.

PR #4536, opened 2020-10-26
Fix auto channel unmute still applying after bot restart

### LINKED ISSUE #4535: Channel mute data is only cleared in cache on error
I was waiting for the automatic unmute to trigger and once it did, it sent the error:
```
I am unable to unmute KslibLivxud#1317 in #bhc-casual for the following reason:
Failed to mute or unmute user. I need the Manage Permissions permission.
```
I then waited for next loop cycle and the message was NOT sent again.
However, after I restarted the bot, it sent the same error again.

It seems that channel mute data is only cleared in the cache if the automatic unmute errors and therefore the error message gets resent on each bot restart.

TrustyJAID:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This closes #4535 by removing users from config regardless of status of success in auto unmute.


Jackenmen (approved):
Looks good to me, thanks.

PR #4537, opened 2020-10-26
Replace use of asyncio.gather with bounded_gather

### LINKED ISSUE #4530: "[p]muteset makerole" can error out and fail if ratelimited
# Command bugs

#### Command name

`[p]muteset makerole`

Mutes

The command to successfully create a new role with overrides configured for a server's channels and categories

#### What actually happened?

It causes an exception, leaving a partially-configured role and posting an error in chat. The traceback ends with `discord.ext.commands.errors.CommandInvokeError: Command raised an exception: HTTPException: 429 Too Many Requests (error code: 0): You are being rate limited.`

1. Start with a server that has no mute role configured, and a considerably large number of channels.
2. Run `[p]muteset makerole [role]`
3. If Discord ratelimits Red while it's applying overrides, the command will error out leaving the role only partially configured.

TrustyJAID:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This fixes #4530 ~~but makes it a lot slower by comparison in order to avoid rate limiting the bot on large servers. This is just temporary until a better solution is found and needed. We *are* enforcing role based mutes instead.~~

Thanks to @zephyrkul for pointing out `bounded_gather` util in Red to have some improvements in speed.

### DIFF
--- a/redbot/cogs/mutes/mutes.py
+++ b/redbot/cogs/mutes/mutes.py
@@ -1380,7 +1380,7 @@ class Mutes(VoiceMutes, commands.Cog, metaclass=CompositeMetaClass):
         else:
             results = []
             for channel in guild.channels:
-                result.append(await self.channel_unmute_user(guild, channel, author, user, reason))
+                results.append(await self.channel_unmute_user(guild, channel, author, user, reason))
             for task in results:
                 if not task["success"]:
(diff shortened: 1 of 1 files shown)


Jackenmen (approved):
LGTM

PR #4541, opened 2020-10-27
Modlog account for very long case reasons

TrustyJAID:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This adds some logic to allow very long modlog cases to still post and provide as much detail as possible.

### DIFF
--- a/redbot/core/modlog.py
+++ b/redbot/core/modlog.py
@@ -15,6 +15,7 @@ from .utils.common_filters import (
     filter_urls,
     escape_spoilers,
 )
+from .utils.chat_formatting import pagify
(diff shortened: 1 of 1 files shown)


redbot/core/modlog.py
```diff
@@ -425,6 +435,18 @@ async def message_content(self, embed: bool = True):
...
+                reason = (
+                    next(
+                        pagify(
+                            reason,
+                            delims=[" ", "\n"],
+                            page_length=1000,
+                        )
+                    )
+                    + "..."
+                )
```
Jackenmen: This needs to be in an `if self.reason:`

Jackenmen (approved):
LGTM

PR #4551, opened 2020-10-27
Update the copyright symbol in the `[p]info` command

TurnrDev:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Switched from `(c)` to `©` in the info command. Red-DiscordBot requires at least Python 3.7, so has support for Unicode. I think the use of the proper Unicode copyright symbol would be a good shout.


Kowlin:
Thanks for the PR, however we're not gonna accept this on purpose due the fact that we want to ensure this is render-able for every client. And the current implementation ensures that this will remain that way.

TurnrDev:
That makes sense! No worries :+1:

PR #4563, opened 2020-10-28
[Mutes] Add ability to send DMs on mute/unmute

### LINKED ISSUE #3752: Configurable mute messages
# Feature request

Functionality

- [ ] Have the option of sending a DM on mute to the user being muted just like the Warnings cog or ``[p]kick`` and ``[p]ban`` already do.

- [ ] Add toggle to sending mute/Unmute messages to the channel the command was executed in. This is especially useful for those using 3rd-party tempmutes cogs, however it might be a desired feature for others too, since the modlog entry is sometimes sufficient as a confirmation/logging message.

I would have made a PR but the mute feature branch is not in a working state and I didn't want to just PR to dev.

Dav-Git:
#### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Closes #3752


Dav-Git:
For some reason local tox keeps failing, however these tests pass... review with caution

redbot/cogs/mutes/mutes.py
```diff
@@ -74,6 +74,8 @@ def __init__(self, bot: Red):
             "notification_channel": None,
             "muted_users": {},
             "default_time": 0,
+            "dm": False,
```
Bakersbakebread, line 77: Would it be beneficial to change this to `dm_on_mute` ? To be inline with other settings such as [dm_on_kickban](https://github.com/Cog-Creators/Red-DiscordBot/blob/7bb6e60c52134ed87b43841fabf0167a604cfe36/redbot/cogs/mod/mod.py#L57)

Dav-Git: Iirc this is in line with the warning cog's config key, however I am open for either.

Jackenmen (changes requested):
Code and functionality seem to be working mostly great.

There are two things that need addressing:
1. DM notifications can't be dependent on the modlog cases, because it breaks when you disable the case type in the modlog (and not sending the DM if the case type is disabled is not a good idea either)
2. DM notifications aren't sent on voice mutes/unmutes, is that intentional?

Dav-Git:
Oh... I totally missed voicemutes... yeah. I'll add that in a bit here.

redbot/cogs/mutes/mutes.py
```diff
@@ -640,7 +656,13 @@ async def on_guild_channel_update(
                     )
                     log.debug("created case")
                     if await self.config.guild(after.guild).dm():
-                        await self._send_dm_notification(modlog_case)
+                        await self._send_dm_notification(
+                            after,
+                            None,
+                            after.guild,
+                            _("(Voice-)Channel unmute"),
```
Jackenmen: This should differ between voice and channel mute, there's even code checking for the channel type just before creating the case in here, so you could choose a proper string there.

redbot/cogs/mutes/mutes.py
```diff
@@ -1097,7 +1119,9 @@ async def mute(
                         channel=None,
                     )
                     if await self.config.guild(user.guild).dm():
-                        await self._send_dm_notification(modlog_case)
+                        await self._send_dm_notification(
+                            user, author, guild, _("Server Mute"), reason, duration
```
Jackenmen: Casing consistency
```suggestion
                            user, author, guild, _("Server mute"), reason, duration
```

redbot/cogs/mutes/voicemutes.py
```diff
@@ -233,3 +240,67 @@ async def unmute_voice(
...
+                if duration
+                else ""
+            )
+            message += _("\n**Guild**: {guild_name}\n**Channel**: {channel_name}").format(
+                guild_name=guild.name, channel_name=channel.name
+            )
+            try:
+                await user.send(message)
+            except discord.Forbidden:
+                pass
```
Jackenmen: This should not be here at all, it's already defined in the resulting Mutes class so no need to repeat this code here. You can put the abstract method for this in `MixinMeta` if your editor is complaining, but if you don't I'll do that myself before merging.

redbot/cogs/mutes/mutes.py
```diff
@@ -462,44 +468,50 @@ async def _auto_channel_unmute_user(
...
+        mute_type: str,
+        reason: str,
+        duration=None,
+    ):
+        show_mod = await self.config.guild(guild).show_mod()
+        title = bold(mute_type)
+        duration = _(" {duration}").format(duration=humanize_timedelta(timedelta=duration))
+        until = datetime.utcfromtimestamp(datetime.now(timezone.utc) + duration).strftime(
+            "%Y-%m-%d %H:%M:%S UTC"
+        )
```
Jackenmen: This doesn't seem to handle the case when `duration` is `None`?

redbot/cogs/mutes/mutes.py
```diff
@@ -462,44 +468,50 @@ async def _auto_channel_unmute_user(
...
+
+        if moderator is None:
+            moderator_string = "Unknown"
```
Jackenmen:
```suggestion
            moderator_string = _("Unknown")
```

redbot/cogs/mutes/mutes.py
```diff
@@ -462,44 +468,50 @@ async def _auto_channel_unmute_user(
...
+        else:
+            moderator_string = f"{moderator.name}#{moderator.discriminator}"
```
Jackenmen:
```suggestion
            moderator_string = str(moderator)
```

Jackenmen (changes requested):

redbot/cogs/mutes/mutes.py
```diff
@@ -454,6 +467,72 @@ async def _auto_channel_unmute_user(
...
+                em.add_field(name=_("Duration"), value=duration_str)
+            em.add_field(name=_("Guild"), value=guild.name, inline=False)
+            if show_mod:
+                em.add_field(name=_("Moderator"), value=moderator_str)
+            try:
+                await user.send(embed=em)
+            except discord.Forbidden:
+                pass
+        else:
+            message = f"{title}\n------------------\n"
```
Jackenmen: I think it might look better with this formatting, but I'm not sure:
```suggestion
            message = f"{title}\n>>> "
```

What do you think?
*Before*
image
*After*
image

Jackenmen (commented):
This looks good to me now, but I have one suggestion **to consider** about non-embed version of the DM before I approve

Dav-Git:
Love your idea

Jackenmen (approved):
As said earlier, this already looked good to me, so this is just a formality 😄 
Thank you for PRing.

PR #4577, opened 2020-11-04
[Core] Rich logging for not so rich people

### LINKED ISSUE #3936: Remove the extras from Red's splash screen (or only show postgres)
The extras shown in our splash screen are useless - Voice is always True since 3.0.0, Docs and Tests extras are irrelevant when running Red (they are only needed to run the external tools for development like sphinx or pytest):
image

I say we either should just get rid of that or only put Postgres extra in there (and more extras in future if we would have more)
Also, if we were to go with the latter it could somehow be made clearer what the ticks in that box mean.

Kowlin:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Added rich logging and a flag to forcefully enable it within red. Rich logging will beautify your terminal so it all looks shiny and new.

image

image

closes #3936


madebylydia:
Guess `Loaded X cogs with Y commands` could be put under `Unique Users` as it's a part of the bot's statistics, even if not directly related to Discord, I think it would fit better to me.
Maybe putting the Invite URL on the middle could be nice too. Feel free to correct me.

Jackenmen (approved):
If this breaks, you can blame *Rich* people for not sponsoring us :)

https://patreon.com/Red_Devs

PR #4583, opened 2020-11-06
Warn when new slotmin/slotmax value will cause slots to not work

bobloy:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Slotmax wasn't checking if the bank was global or not when comparing to slotmin.

Slotmin wasn't checking slotmax at all.

It was easy to get into a state where no bid was accepted in the slots


redbot/cogs/economy/economy.py
```diff
@@ -751,11 +751,18 @@ async def economyset_showsettings(self, ctx: commands.Context):
...
-        if bid < 1:
-            await ctx.send(_("Invalid min bid amount."))
+        is_global = await bank.is_global()
+        if is_global:
+            slot_max = await self.config.SLOT_MAX()
+        else:
+            slot_max = await self.config.guild(ctx.guild).SLOT_MAX()
+        if bid < 1 or bid > slot_max:
+            await ctx.send(
+                _("Invalid maximum bid amount. Must be less than or equal to the maximum amount.")
```
bobloy:
```suggestion
                _("Invalid minimum bid amount. Must be less than or equal to the maximum amount.")
```

bobloy: Proofreading is a good skill. Someday I will have it.

Jackenmen (commented):
I think splitting errors for `bid < 1` and `bid > slot_max`/`bid < slot_min` would result in more user-friendly messages (i.e. saying "Bid amount must be a positive number." for `bid < 1` case).

I also wonder if trying to set slot min/max to a value that doesn't make sense with the current slot max/min should outright fail or just warn the user about it so that they aren't forced to use the commands in the right order (similarly to what we're currently doing with commands in `[p]modset mentionspam` group). I'm not really sure about it though.

bobloy:
To your first comment, I could see using a different converter than just `int` that specifies greater than 0.

Jackenmen:
Here's one that you could just copy:
https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/cleanup/converters.py#L19-L30

We might want to make it into a core converter later on, but that would need to be a separate PR (+ someone would need to first make an issue asking for it) so it's gonna be simpler to duplicate code for now.

bobloy:
Added the positive_int converter, and adjusted it to allow maxes and mins that clash with a warning.

Jackenmen (approved):
Looks good to me, thanks.

PR #4590, opened 2020-11-10
[Downloader] Hide `[p]cog reinstallreqs`

bobloy:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This command is only meant to be run when Red asks for it by name. 
Thus, being found in the help will only lead to confusion for users. See conversation in #4511


Jackenmen (approved):
LGTM, thanks.

PR #4593, opened 2020-11-11
Fix wrong words in docstrings of reaction and message predicates

madebylydia:
### Type

- [X] Typofix


Jackenmen (approved):
Thanks for the fix.

PR #4603, opened 2020-11-13
[Utils] [Chat Formatting] New rows to embeds util

### LINKED ISSUE #4600: Add a util that generates a list of embeds based on a list of rows
# Feature request

- [ ] Cog
- [ ] Command
- [x] API functionality

The util (to be placed under `chat_formatting`) should be a method that converts a list of "rows" (`List[str]`) into a list of embeds. It does so by putting distributing the rows across fields, where each field consists of a certain amount of rows. The `field value` would then be equivalent to `"\n".join(rows[start:end]`. 
When more than (by default) 2 fields are needed to display all rows, additional embeds are made.

This util should be versatile, and be able to configure several parameters, such as:
- Amount of fields per embed
- Amount of rows per field
- Whether to equally spread the rows across the fields, or to "greedily" push the rows as much to the first embeds as possible.

#### Use case
There are several applications in which this util would be helpful. Personally, I already use derivations of this util several times. Some possible use cases:
- A command to list all roles on a server and their user count ([like in my personal repository](https://github.com/FixedThink/HashCogs/blob/master/fixedvarious/fixedvarious.py#L171)).
- Leaderboards (e.g. for Trivia), which could display e.g. 10 people (and their scores) per field.
- Search results, sorted on relevance.

This list is non-exhaustive, and I'm sure there are many more applications in which a generic row-list to embed-list converter would be beneficial.

This util introduces new function, whose output can be used with other utils at the choice of cog developers.

FixedThink:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

- Added `rows_to_embeds()` util in `chat_formatting`.
- Added two unit tests for the util.

Resolves #4600.


redbot/core/utils/chat_formatting.py
```diff
@@ -283,6 +284,103 @@ def pagify(
...
+    Returns
+    -------
+    embed_list : `list` of `discord.Embed`
+        A list of embeds, based on the base embed, with all rows spread across the embeds.
+
+    Raises
+    ------
+    ValueError
+        When either ``embed_max_fields`` or ``field_max_rows`` is not positive.
+    """
```
Drapersniper, line 388: We should cap this to the api limited value too

redbot/core/utils/chat_formatting.py
```diff
@@ -283,6 +284,103 @@ def pagify(
...
+    if always_page_number or embed_cnt > 1:
+        foot_str = "{_page} of {_total}"  # _a and _z to be filled in the loop.
+        if footer_static:
+            foot_str += " — {}".format(footer_static)
+    elif footer_static:
+        foot_str = footer_static
+    # Initialise for nested loop.
+    embed_list: List[discord.Embed] = []
+    start = 0
+    f = 0
```
Drapersniper, line 420: This doesn't appear to be used anywhere

redbot/core/utils/chat_formatting.py
```diff
@@ -283,6 +284,103 @@ def pagify(
...
+    always_page_number: bool = False,
+) -> List[discord.Embed]:
+    """
+    Parameters
+    ----------
+    rows : `list` of `str`
+        The strings that form a row in the embed
+    base_embed : `discord.Embed`
+        The embed to use as a base for the list of embeds
+    embed_max_fields : `int`, optional
```
Drapersniper: The optional typing here should be Optiona[type] just to follow the existing format, this also applies to the other instances in the docstring

Drapersniper (changes requested):
Just some things I've noticed at a glance ... I'm not a fan of the util name but can't think of an alternative either

redbot/core/utils/chat_formatting.py
FixedThink: - `embed_list` is used in L380 (to append)
- `start` is used several times (lines 371, 373m 376)
- `f` is incremented in L379

Drapersniper: I meant specifically f, it's increments but I don't see it being used anywhere other then in the line which it gets incremented

FixedThink:
Updated the files based on the comments above. I renamed the function to `embeds_from_rows`, maybe that sounds a bit better? At least the first word now makes clear what the function returns.

redbot/core/utils/chat_formatting.py
```diff
@@ -284,7 +284,7 @@ def pagify(
             yield in_text
 
 
-def rows_to_embeds(
+def embeds_from_rows(
```
bobloy, line 342:
```suggestion
def embedify_list(
```
Alternative name option. Follows the naming scheme of `pagify`

tests/core/test_utils.py
```diff
@@ -203,7 +203,7 @@ def test_embed_list():
     base_embed = Embed(title="Test")
     rows = list(str(i) for i in range(11))
 
-    embed_list = chat_formatting.rows_to_embeds(
+    embed_list = chat_formatting.embeds_from_rows(
```
bobloy, line 206:
```suggestion
    embed_list = chat_formatting.embedify_list(
```

tests/core/test_utils.py
```diff
@@ -223,7 +223,7 @@ def test_embed_list_2():
...
```
bobloy, line 226:
```suggestion
    embed_list = chat_formatting.embedify_list(
```

bobloy (commented):
Alternative naming suggestion: `embedify_list`

Jackenmen:
Sorry that this has been open for so long without any feedback...
While this may have some use, I don't really​ want to commit to a design without a bunch of concrete usage examples *with* code snippets and screenshots. I'll close this and the corresponding issue because I don't think it makes to keep this in limbo forever... *Maybe* this would be useful, but until aforementioned concrete usage examples are provided, this wasn't going to progress anyway...

PR #4615, opened 2020-11-16
[Streams] Support Youtube stream schedules

### LINKED ISSUE #3691: Scheduled YouTube streams are treated as non-streams
# Other bugs

I scheduled a YouTube stream and started it some time later.

I expected a stream alert when I started the stream.

#### What actually happened?

There was no alert neither when I made the schedule nor when I actually started the stream. This happens because we check if the stream is actually live when it shows up in RSS feed and if it's not, we treat the scheduled stream as non-stream and no longer check for it.

1. Add stream alert for your YouTube channel
2. Schedule a stream and wait until it shows up in the RSS (`https://www.youtube.com/feeds/videos.xml?channel_id=CHANNEL_ID_HERE`)
3. Wait 10 minutes for the bot to check the updated RSS (you can lower the needed waiting time with `[p]streamset timer`)
4. Start stream and wait until it shows up in the RSS
5. Wait 10 minutes for the bot to check the updated RSS (you can lower the needed waiting time with `[p]streamset timer`)
6. Stream alert won't show up

---

There's no perfect way for solving this problem, from discussion with Draper, the possible solutions we came up with are:

a) Post info about the scheduled stream as soon as we get info about it (so after number `3` in repro steps) and add info in embed that it's a scheduled stream, not live stream
b) Save info about the scheduled stream and send the alert at the time when it was supposed to start
c) Same as b) but before sending the alert, check that the stream isn't rescheduled/check that it's live (this will make it use more quota which we might really want to avoid)
d) Do both a) and b) (or a) and c))
e) Something we haven't thought of yet?

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Add support for YouTube stream schedules.

- The `[p]youtubealert` command will show the start time of a scheduled stream
- The bot will perform two announcements, one when we receive the schedule, a second one when the stream actually start.
- If autodelete is enabled, the bot will delete the message announcing the stream schedule before announcing the stream
- This can be disabled with `[p]streamset ignoreschedule`, similar to `[p]streamset ignorereruns`

### To do

- ~~I just noticed the bot gets stuck in a loop deleting its message and reannouncing when the stream is online :eyes: not much to fix, omw~~ Fixed
- Maybe add a setting for a different message when the announcement is a schedule?

Fixes #3691

### DIFF
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -58,6 +58,7 @@ class Streams(commands.Cog):
         "live_message_mention": False,
         "live_message_nomention": False,
         "ignore_reruns": False,
+        "ignore_schedule": False,
     }
 
     role_defaults = {"mention": False}
@@ -208,9 +209,7 @@ class Streams(commands.Cog):
         await self.maybe_renew_twitch_bearer_token()
         token = (await self.bot.get_shared_api_tokens("twitch")).get("client_id")
         stream = TwitchStream(
-            name=channel_name,
-            token=token,
-            bearer=self.ttv_bearer_cache.get("access_token", None),
+            name=channel_name, token=token, bearer=self.ttv_bearer_cache.get("access_token", None),
         )
         await self.check_online(ctx, stream)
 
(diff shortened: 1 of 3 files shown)


Kowlin (approved):
Working like a charm.

PR #4620, opened 2020-11-17
Some crazy changelog things

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

You'll get to know more later ;)


Jackenmen:
An idea (from Zephyrkul):
divide "Core Bot" into smaller sections, like "Core" (the cog), "Dependencies", "Support" (for that Fedora 32 removal for example), etc.

**Be sure to split it off into a new issue *if* I don't end up doing it in here (which is not that unlikely)**

Jackenmen:
Superseded by #5602 and #5603.

PR #2455, opened 2019-02-15
[V3 Alias] Customize Parameters

Zephyrkul:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

### Description of the changes
Allows customizing where parameters are placed in aliases using positional formatting arguments. Like with #2051, the bot will verify and correct newly created aliases if it can, and return a user-friendly error if it can't.
Parsing follows discord.py's parsing style. Quoted phrases are considered one argument and will be inserted appropriately.
Unused parameters are appended to the end.

Examples:

`[p]alias add oneban ban {0} 1`
`[p]oneban @zephyrkul too much fluff`
turns into: `[p]ban @zephyrkul 1 too much fluff`

`[p]alias add banhammer ban {1} 7`
corrected into: `[p]ban {0} 7`
`[p]banhammer @zephyrkul trolling`
turns into: `[p]ban @zephyrkul 7 trolling`

`[p]alias add badban ban {0} {2}`
returns: `Arguments must be sequential. Missing arguments: 1`
No alias is created.

`[p]alias add nonum ban {} 1`
returns: `Arguments must be specified with a number.`
No alias is created.

PR #4632, opened 2020-11-20
[Alias] Alter `call_alias()` to prevent stripping of newlines

goettner:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

As detailed in issue #2704 when an alias for a command is created, and this alias is passed input, any newlines in the input are stripped. This is reproducible with the [p]say command (an example can be seen in the issue documentation).

To fix this, I altered the method get_extra_args_from_alias() in redbot/cogs/alias/alias_entry.py to not skip whitespace in between arguments, maintaining newlines if they exist. When these arguments are then formatted into the content of the message that is passed along to the bot, whitespace is not added back in. This should have the effect of better maintaining the passed message, while still allowing customized parameters to be inserted into the input.
 
I also added a test case for this issue to tests/cogs/test_alias.py. This required the splitting up of the call_alias in redbot/cogs/alias/alias.py into two functions, in order to return a value that could be tested. It also required the creation of two new fixtures in redbot/pytest/core.py, newline_message and prefix. In the case that more tests are eventually added to test_alias, these might prove useful.

### DIFF
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -173,6 +173,15 @@ class Alias(commands.Cog):
         raise ValueError("No prefix found.")
 
     async def call_alias(self, message: discord.Message, prefix: str, alias: AliasEntry):
+        new_message = translate_alias_message(message, prefix, alias)
+        await self.bot.process_commands(new_message)
+
+    def translate_alias_message(self, message: discord.Message, prefix: str, alias: AliasEntry):
+        """
+        Translates a discord message using an alias
+        for a command to a discord message using the
+        alias' base command.
+        """
         new_message = copy(message)
         try:
             args = alias.get_extra_args_from_alias(message, prefix)
@@ -187,7 +196,7 @@ class Alias(commands.Cog):
 
         # noinspection PyDunderSlots
         new_message.content = "{}{} {}".format(prefix, command, arg_string)
-        await self.bot.process_commands(new_message)
+        return new_message
 
     async def paginate_alias_list(
         self, ctx: commands.Context, alias_list: List[AliasEntry]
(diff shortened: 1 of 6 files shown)


goettner:
Also not sure why I'm getting the Lint Python error, if anyone knows why let me know and I'll be happy to rectify it.

Jackenmen:
> Also not sure why I'm getting the Lint Python error, if anyone knows why let me know and I'll be happy to rectify it.

`translate_alias_message()` is a method of `Alias` instance, so you need to use it with `self`:
```py
new_message = self.translate_alias_message(message, prefix, alias)
```

goettner:
> > Also not sure why I'm getting the Lint Python error, if anyone knows why let me know and I'll be happy to rectify it.
> 
> `translate_alias_message()` is a method of `Alias` instance, so you need to use it with `self`:
> 
> ```python
> new_message = self.translate_alias_message(message, prefix, alias)
> ```

Gotcha, will do.

Zephyrkul (changes requested):
This PR doesn't play well with [alias parameters](https://github.com/Cog-Creators/Red-DiscordBot/pull/2455).

If parameters are used, they are inserted into the new message content *and* appended to the content, creating [results like this](https://cdn.discordapp.com/attachments/718148684629540905/779190681100353616/unknown.png).

goettner:
> This PR doesn't play well with [alias parameters](https://github.com/Cog-Creators/Red-DiscordBot/pull/2455).
> 
> If parameters are used, they are inserted into the new message content _and_ appended to the content, creating [results like this](https://cdn.discordapp.com/attachments/718148684629540905/779190681100353616/unknown.png).

Huh, okay. I wasn't aware of that feature. I'll see if I can find a way to still allow for that while maintaining newlines. Thanks for the feedback!

goettner:
Alright, customized parameters should work now, alongside newlines. Let me know what you think.

goettner:
I accidentally pushed some unrelated changes to this PR, and couldn't figure out a way to revert them. I'll open a new PR when I fix it.

PR #4636, opened 2020-11-24
bot.remove_command returns a dpy command object instead of None

npc203:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

d.py returns a command object for the remove_command, red doesn't. Hence made this small PR for Red to be ever so slightly closer to d.py. I hope it comes under the bug fix category.  

Edit: changed category to feature

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1374,7 +1374,7 @@ class RedBase(
                 if permissions_not_loaded:
                     subcommand.requires.ready_event.set()
 
-    def remove_command(self, name: str) -> dpy_commands:
+    def remove_command(self, name: str):
         command = super().remove_command(name)
         if not command:
             return
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1374,7 +1374,7 @@ class RedBase(
                 if permissions_not_loaded:
                     subcommand.requires.ready_event.set()
 
-    def remove_command(self, name: str) -> None:
+    def remove_command(self, name: str) -> dpy_commands:
         command = super().remove_command(name)
         if not command:
             return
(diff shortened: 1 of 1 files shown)


Drapersniper:
So there's an empty return here, which would make the return typehint Optional, 
That said you typehinted the command module rather then the command/group objects.

Optional[Union[dpy_commands.Command, dpy_commands.Group]] is more what i'd expect with this PR (I haven't checked this is right so don't just copy paste blindly).

that said Is there an issue discussing this PR? I'm interested in seeing the reasoning for this.

npc203:
Ye, i think i should've made this an issue at first (sorry for the bad code). So what i was trying to do was, i have conflicting commands in different cogs and i would like to do  

something = bot.remove_command("command")
something.name = "new_name"
bot.add_command(something)

Then it didnt work as expected, checked the dpy cogs, it was fine, and after sometime, checked the source of Red, found the culprit. Would prefer red commands to be closer to dpy as it makes life easier.

redbot/core/bot.py
```diff
@@ -1374,14 +1374,15 @@ def add_command(self, command: commands.Command) -> None:
                 if permissions_not_loaded:
                     subcommand.requires.ready_event.set()
 
-    def remove_command(self, name: str) -> None:
+    def remove_command(self, name: str):
         command = super().remove_command(name)
         if not command:
             return
```
bobloy:
```suggestion
        if command is None:
            return command
```
This checks the results of the `super` call more correctly and causes the return types to be exactly that of the `super`

Jackenmen:
There's not really a point in Red differing from d.py here, it looks like the command object being returned just has been unintentionally omitted when we overrode this method. Of course, you can get around it, but there's no reason to not fix it.

Jackenmen (approved):
Looks good to me, thanks for the contribution!

PR #4644, opened 2020-11-30
Add support for pip 20.3 (a.k.a. use old dependency resolver) and disallow pip 21

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This makes use of `-use-deprecated=legacy-resolver` for pip 20.3.x to avoid using the new dependency resolver in Downloader until we ensure proper support for how it works. Until then, Red will forbid using pip 21 to avoid problems that may occur because of it.


Jackenmen:
Turns out it's not really all that important because even with the new resolver, pip still ignores dependencies when `--target` is used.

PR #4647, opened 2020-12-01
Add `on_red_before_identify` dispatch

fixator10:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Adds a dispatch of `on_red_before_identify` event in `before_identify_hook`

Allows detecting `IDENTIFY` without editing core

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -381,7 +381,7 @@ class RedBase(
     async def before_identify_hook(self, shard_id, *, initial=False):
         """A hook that is called before IDENTIFYing a session.
         Same as in discord.py, but also dispatches "on_red_identify" bot event."""
-        self.dispatch("red_identify", shard_id, initial)
+        self.dispatch("red_before_identify", shard_id, initial)
         return await super().before_identify_hook(shard_id, initial=initial)
 
     @property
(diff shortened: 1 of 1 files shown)


Jackenmen:
This should use `red_` prefix to avoid potential conflict.

PredaaA:
I don't think it should be called `on_red_identify`. Because as that hook is named, it's *before* IDENTIFY happen that it is called, and also because d.py already dispatch `on_connect` and `on_shard_connect` events, which is when IDENTIFY happens.

https://discordpy.readthedocs.io/en/latest/api.html#discord.Client.before_identify_hook
https://discordpy.readthedocs.io/en/latest/api.html?highlight=on_connect#discord.on_connect
https://discordpy.readthedocs.io/en/latest/api.html?highlight=on_connect#discord.on_shard_connect

So I guess it should be called `on_red_before_identify` instead.

Jackenmen:
I would love to hear some sample use case(s) for this. While it's not really hurting us to have this dispatch, some justification is always nice to have :)

Rapptz:
This could be potentially useful for tracking how many times a shard has disconnected. I do some basic tracking like this in my bot to see how many identifies and resumes I've had per shard in a 24 hour period.

Jackenmen (approved):
Let's get this merged then, thanks!

PR #4649, opened 2020-12-02
[Trivia] Add support for payout to multiple trivia winners

### LINKED ISSUE #3931: [Trivia] When a trivia game results in a tie, bot picks a random user between the 2 tied players for the payout
# Command bugs

#### Command name

It doesn't have a command name, because this happens when the trivia ends in a tie.

Trivia

I was expecting the bot to announce that "This trivia session has ended in a tie." Or something like that.

#### What actually happened?

The bot randomly picks a winner between the two winners for the economy payout and "Congratulations, ..... for coming first" announcement. 

First set the payout multiplier to your desired value other than the default 0 value with `[p]triviaset payout <value>` command. 
Start and play the trivia with someone else or a group of people. Play in such a way that all participating players ends up winning equal score. When trivia ends, the bot picks a random winner from the players who have tied with equal score for the payout and for "Congratulations, ..... for coming first" message.

leblancg:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This PR is for issue #3931. These changes allow for more than one winner of the trivia game. Each winner will receive an even split of the payout. This also adds a new message to congratulate the winner(s).

### DIFF
--- a/redbot/cogs/trivia/session.py
+++ b/redbot/cogs/trivia/session.py
@@ -304,7 +304,7 @@ class TriviaSession:
                 if score == top_score:
                     winners.append(player)
                 num_humans += 1
-        if len(winners) == 0 or num_humans < 3:
+        if not winners or num_humans < 3:
             return
         payout = int(top_score * multiplier / len(winners))
         if payout <= 0:
@@ -315,14 +315,23 @@ class TriviaSession:
                 await bank.deposit_credits(winner, payout)
             except errors.BalanceTooHigh as e:
                 await bank.set_balance(winner, e.max_balance)
-        await self.ctx.send(
-            "Congratulations {users}! You have {plural}received {num} {currency} for winning!".format(
-                users=" and ".join(bold(winner.display_name) for winner in winners),
-                plural="each " if len(winners) > 1 else "",
+        if len(winners) > 1:
+            msg = _(
+                "Congratulations {users}! You have each received {num} {currency} for winning!"
+            ).format(
+                users=humanize_list([bold(winner.display_name) for winner in winners]),
                 num=payout,
                 currency=await bank.get_currency_name(self.ctx.guild),
             )
-        )
+        else:
+            msg =  _(
+                "Congratulations {user}! You have received {num} {currency} for winning!"
+            ).format(
+                user=bold(winners[0].display_name),
+                num=payout,
+                currency=await bank.get_currency_name(self.ctx.guild),
+            )
+        await self.ctx.send(msg)
 
 
 def _parse_answers(answers):
--- a/redbot/cogs/trivia/session.py
+++ b/redbot/cogs/trivia/session.py
@@ -261,7 +261,7 @@ class TriviaSession:
             await self.send_table()
         multiplier = self.settings["payout_multiplier"]
         if multiplier > 0:
-            await self.pay_winner(multiplier)
+            await self.pay_winners(multiplier)
         self.stop()
 
     async def send_table(self):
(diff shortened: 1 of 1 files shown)


leblancg:
I've tested it out on my own, but is any additional QA recommended? Not really sure how to unit test this using pytest.

Jackenmen:
There's no need for making unit tests here, PRs are reviewed by one of our Core Dev/QA members.

redbot/cogs/trivia/session.py
```diff
@@ -282,41 +282,47 @@ def force_stop(self):
...
+            return
+        top_score = self.scores.most_common(1)[0][1]
+        winners = []
+        num_humans = 0
+        for (player, score) in self.scores.items():
+            if not player.bot:
+                if score == top_score:
+                    winners.append(player)
+                num_humans += 1
+        if len(winners) == 0 or num_humans < 3:
```
Jackenmen: According to our style guide, the fact that empty sequences evaluate to false should be used instead of `len()` (you can see PEP 8 for more information)
```suggestion
        if not winners or num_humans < 3:
```

redbot/cogs/trivia/session.py
```diff
@@ -282,41 +282,47 @@ def force_stop(self):
...
+            except errors.BalanceTooHigh as e:
+                await bank.set_balance(winner, e.max_balance)
+        await self.ctx.send(
+            "Congratulations {users}! You have {plural}received {num} {currency} for winning!".format(
+                users=" and ".join(bold(winner.display_name) for winner in winners),
+                plural="each " if len(winners) > 1 else "",
+                num=payout,
+                currency=await bank.get_currency_name(self.ctx.guild),
+            )
+        )
```
Jackenmen: There's a few things with how this message is defined that need to be changed:
1. `_()` is a call to Translator object defined at the top of the file and is needed in order for the string to be translated when there's a translated version available.
2. Instead of `.join()`, we have a `humanize_list()` utility function that formats a list according to formatting rules of the locale in use.
3. To support plural rules properly with translations, the string for a single user should be a separate from the string for multiple users as it allows translators to make more appropriate translations. While that is not perfect (Red currently doesn't support tools that would be more appropriate for this), it's still better for translating than using a single string with a `{plural}` format field.
```suggestion
        if len(winners) > 1:
            msg = _(
                "Congratulations {users}! You have each received {num} {currency} for winning!"
            ).format(
                users=humanize_list([bold(winner.display_name) for winner in winners]),
                num=payout,
                currency=await bank.get_currency_name(self.ctx.guild),
            )
        else:
            msg =  _(
                "Congratulations {user}! You have received {num} {currency} for winning!"
            ).format(
                user=bold(winners[0].display_name),
                num=payout,
                currency=await bank.get_currency_name(self.ctx.guild),
            )
        await self.ctx.send(msg)
```

redbot/cogs/trivia/session.py
```diff
@@ -282,41 +282,47 @@ def force_stop(self):
         LOG.debug("Force stopping trivia session; #%s in %s", channel, channel.guild.id)
 
     async def pay_winner(self, multiplier: float):
```
Jackenmen: Since this is not a public API, it would be clearer if this got renamed to `pay_winners()` and the usage of it got updated appropriately.
```suggestion
    async def pay_winners(self, multiplier: float):
```

Jackenmen (changes requested):
This is a really clean solution, the code also seems more readable than what's been there before 😄 

The new functionality appears to work fine however, I got a few review comments before I can approve this.

Jackenmen (approved):
I went ahead and made the appropriate changes on my own and this is now ready to be merged.
Thanks for your contribution!

Issue #4391, opened 2020-09-03
Bank API should support discord.Object when getting Balance


TrustyJAID:
# Feature request

#### Select the type of feature you are requesting:

- [ ] Cog
- [ ] Command
- [x] API functionality

#### Describe your requested feature

Cores Bank API should accept discord.Object since we cannot always ensure that it's safe to acquire a full user/member object when accessing things as trivial as an economy balance.

Stonedestroyer:
I will be doing this in the upcoming days.

goettner:
I'd like to take a look at this issue, but I'm not totally sure how to go about accepting a discord.Object as a parameter for balance() in economy.py instead of a Member. 
Would it be a decent solution to extract User.id, and pass this along to bank.get_balance() instead (which could possible be used in other calls, like transfer()), or am I missing something there? Thanks!

Drapersniper:
The ideas is you'd accept either a member object, user object or discord object, which all 3 have the .id attribute.

Essentially extracting that attribute in the method it self when the id is needed (for config calls)

goettner:
Okay, I understand, I'll take a crack at it. Thanks for the clarification

PR #4654, opened 2020-12-03
[Bank] Add discord.Object support to bank.get_balance

goettner:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

As detailed in #4391, much of the Core Bank API functions only accepted discord.Member or discord.User input types for routine functions such as get_balance(). This is problematic, given the unnecessary and possibly sensitive information that is included with a Member or User object. This can be improved by changing the calls' input types to enable passing it a discord.Object instead, in order to prevent sensitive information being unnecessarily acquired.

I also added test cases to tests/cogs/test_economy.py, in order to test the ability of these API calls to function correctly with a  discord.Object as input, and the necessary infrastructure in pytest/core.py

### DIFF
--- a/redbot/core/bank.py
+++ b/redbot/core/bank.py
@@ -203,7 +203,7 @@ def _decode_time(time: int) -> datetime:
     return datetime.utcfromtimestamp(time)
 
 
-async def get_balance(member: discord.Member) -> int:
+async def get_balance(member: Union[discord.Member, discord.User, discord.Object]) -> int:
     """Get the current balance of a member.
 
     Parameters
@@ -599,7 +599,7 @@ async def get_leaderboard_position(
             return pos[0]
 
 
-async def get_account(member: Union[discord.Member, discord.User]) -> Account:
+async def get_account(member: Union[discord.Member, discord.User, discord.Object]) -> Account:
     """Get the appropriate account for the given user or member.
 
     A member is required if the bank is currently guild specific.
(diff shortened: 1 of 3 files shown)


Drapersniper:
You've misunderstood my last statement I think, discord objects only have the `.id` attributes and not the rest of the other attributes that a user or member has, as such special handling is required here.

The remaining bank apis should also accept the discord.Object input.

goettner:
Okay, I see. With that in mind then, how do you propose handling situations where other object members are used? Specifically with this block in bank.get_account() in mind:
Screen Shot 2020-12-03 at 11 20 48 AM

Drapersniper:
> Okay, I see. With that in mind then, how do you propose handling situations where other object members are used? Specifically with this block in bank.get_account() in mind:
> Screen Shot 2020-12-03 at 11 20 48 AM

For one discord.Object could only ever replace users and not members, since members have an additional guild attribute. This would mean objects can ever only be used in global banks

For the attribute access, I would personally add placeholders, such as "John doe" for the display name access since that doesn't exist on an Object

goettner:
> For the attribute access, I would personally add placeholders, such as "John doe" for the display name access since that doesn't exist on an Object

Pushed some changes based on your feedback, let me know what you think.

redbot/core/bank.py
```diff
@@ -620,12 +630,12 @@ async def get_account(member: Union[discord.Member, discord.User, discord.Object
     else:
         all_accounts = await _config.all_members(member.guild)
```
bobloy, line 631: This line will and should fail error with `discord.Object`. Even if it were to succeed with `Object`, `get_default_balance` below would throw a runtime error

bobloy (changes requested):

PR #4659, opened 2020-12-05
[Trivia] Validate custom trivia file upload using schema

### LINKED ISSUE #4571: Trivia should have some validation checks for the trivia files
We should be checking the validity of the trivia list in `[p]trivia` command and `[p]triviaset custom upload` command. I don't know how much checks we need (could use schema verification, although that can result in poor errors sometimes), but our format looks like this (written in a form of a dictionary):
```py
# AUTHOR and CONFIG are special keys that are optional
{
    "AUTHOR": "some string",  # type: str
    "CONFIG": {
        # all keys in here are optional
        "max_score": 123,  # type: int
        "timeout": 123.0,  # type: float
        "delay": 123.0,  # type: float
        "bot_plays": False,  # type: bool
        "reveal_answer": False,  # type: bool
        "payout_multiplier": 123.0,  # type: float
    },
    "question 1 (string)": [
        # list of answers; can be a combination of any of the types below
        "answer 1",  # type: str
        123,  # type: int (gets implicitly converted by Trivia)
        True,  # type: bool (gets implicitly converted by Trivia;
               # this one can be a source of issues,
               # but it probably needs to remain to avoid breakage,
               # at least until next minor version)
        1.69,  # type: float (gets implicitly converted by Trivia;
               # this one can be a source of issues,
               # but it probably needs to remain to avoid breakage,
               # at least until next minor version)
    ],
}
```

As an example, this is one of the errors that can currently occur when starting the trivia with an invalid format, that happens to be valid YAML.
```
Exception in command 'trivia'
  File "/Users/choure/redenv/lib/python3.8/site-packages/discord/ext/commands/core.py", line 85, in wrapped
  File "/Users/choure/redenv/lib/python3.8/site-packages/redbot/cogs/trivia/trivia.py", line 320, in trivia
    trivia_dict.update(dict_)
ValueError: dictionary update sequence element #0 has length 1; 2 is required
```

leblancg:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Fixes #4571. This PR adds schema validation when the user attempts to upload a custom trivia file. I also added a test case that mirrors the other trivia test case to verify that schema validation works for all core trivia lists.

### DIFF
--- a/redbot/cogs/trivia/trivia.py
+++ b/redbot/cogs/trivia/trivia.py
@@ -4,6 +4,7 @@ import math
 import pathlib
 from collections import Counter
 from typing import List, Literal
+from schema import Schema, Optional, Or, SchemaError
 
 import io
 import yaml
@@ -671,9 +672,39 @@ class Trivia(commands.Cog):
                 return
 
         buffer = io.BytesIO(await attachment.read())
-        yaml.safe_load(buffer)
-        buffer.seek(0)
+        trivia_dict = yaml.safe_load(buffer)
 
+        schema = Schema(
+            {
+                Optional("AUTHOR"): str,
+                Optional("CONFIG"): {
+                    Optional("max_score"): int,
+                    Optional("timeout"): Or(int, float),
+                    Optional("delay"): Or(int, float),
+                    Optional("bot_plays"): bool,
+                    Optional("reveal_answer"): bool,
+                    Optional("payout_multiplier"): Or(int, float),
+                },
+                str: [
+                    str,
+                    int,
+                    bool,
+                    float,
+                ],
+            }
+        )
+        try:
+            schema.validate(trivia_dict)
+        except SchemaError as e:
+            await ctx.send(
+                _(
+                    "The custom trivia list was not saved. "
+                    "The file does not follow the proper data format.\n{}"
+                ).format(box(e))
+            )
+            return
+
+        buffer.seek(0)
         with file.open("wb") as fp:
             fp.write(buffer.read())
         await ctx.send(_("Saved Trivia list as {filename}.").format(filename=filename))
--- a/redbot/cogs/trivia/trivia.py
+++ b/redbot/cogs/trivia/trivia.py
@@ -3,7 +3,7 @@ import asyncio
 import math
 import pathlib
 from collections import Counter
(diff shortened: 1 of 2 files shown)


redbot/cogs/trivia/trivia.py
```diff
@@ -671,9 +672,39 @@ async def _save_trivia_list(
...
-        yaml.safe_load(buffer)
-        buffer.seek(0)
+        trivia_dict = yaml.safe_load(buffer)
+
+        schema = Schema(
+            {
+                Optional("AUTHOR"): str,
+                Optional("CONFIG"): {
+                    Optional("max_score"): int,
+                    Optional("timeout"): Or(int, float),
```
leblancg: Or(int, float) is required here instead of just float because schema will not interpret an integer value as a float. (e.g. the value 2 is valid but schema will throw an exception)

redbot/cogs/trivia/trivia.py
```diff
@@ -671,9 +672,39 @@ async def _save_trivia_list(
...
+                    Optional("payout_multiplier"): Or(int, float),
+                },
+                str: [
+                    str,
+                    int,
+                    bool,
+                    float,
+                ],
+            }
+        )
```
Jackenmen: This could be a global variable as it's constant.

redbot/cogs/trivia/trivia.py
```diff
@@ -671,9 +672,39 @@ async def _save_trivia_list(
...
+        try:
+            schema.validate(trivia_dict)
```
Jackenmen: This could be put in a separate function which would make it possible to use directly from the existing tests rather than duplicating the schema.

Jackenmen (changes requested):
Sorry for the huge delay on the review of this, we have quite a big backlog right now and a bit limited time.

This PR looks well written, it would be great if you could expand this to validate the schema in `get_trivia_list()` method as well so that it's checked when trying to start the trivia with `[p]trivia` command (and appropriate exception handling in the `[p]trivia` command, probably by giving a generic error and logging the specific schema error as that command can be run by anyone)

And thanks for contribution :)

Jackenmen (approved):
Thanks for the PR!

PR #4663, opened 2020-12-09
[Core] Add sensible character lengths to activity statuses

bobloy:
### Type

- [X] Bugfix
- [X] Enhancement
- [ ] New feature

Activity statuses were inconsistently checking for max lengths. Exceeding the lengths caused the command to fail silently, and sometimes throws an error.

This PR also adds `twitch` as an alias to the `set stream` command since it only works for twitch streaming statuses, and notes that in the description.


Jackenmen (dismissed):
LGTM, thanks for the PR

Jackenmen (dismissed):
Reapproval, I didn't notice the merge conflict.

Jackenmen (approved):
try number 3

PR #4667, opened 2020-12-11
[Dev] Customizable environment values

laggron42:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This adds two new methods for adding custom values to the environement used in Dev commands, which is currently hard-coded.

There is now in the Dev cog a `get_environement` static method, returning the default dict and adding the custom values in `env_extensions`.

`Red` has the two following public methods added:

- `add_dev_env_value(str, value)` for adding a new value, which is a callable with a single argument `ctx`
- `remove_dev_env_value(str)` the name of the value to remove (can't remove the default values obviously)

### Example usage

```py
nice_member_member_id = 204027971516891136
class MyCog:
  def __init__(self, bot):
    self.important_thing = SomeInstanceOfWhatever()
    self.bot = bot
    bot.add_dev_env_value("thing", lambda ctx: self.important_thing)
    bot.add_dev_env_value("nice_member", lambda ctx: ctx.guild.get_member(nice_member_member_id))

  def cog_unload(self):
    self.bot.remove_dev_env_value("thing")
    self.bot.remove_dev_env_value("nice_member")
```

When loading `MyCog`, the `thing` and `nice_member` values will be available with any dev command.

If the callable raises an exception, it is assigned to the value itself.

### Why?

My tournaments cog relies alot on custom objects, allowing me to do live modifications when testing, debugging or even general usage.

However, accessing those objects with dev commands was kinda painful, since I had to do this every time: `!debug bot.get_cog("Tournaments").tournaments[guild.id]...` which got pretty annoying at some points when I have to do a lot of evals quickly, so I ended up looking for an easier solution, first edited the hard coded env myself, but though it could be a good addition for others


Jackenmen:
I think this is really cool. To throw some other ideas, here's how I plan to use this when/if it gets added:
- making a 3rd-party cog allowing developers to add modules to the environment (i.e. `bot.add_dev_env_value("inspect", lambda ctx: inspect`) that aren't already made available by us
    - might end up just using the existing `instantcmd` (especially that it can run just any arbitrary code) rather than make a cog for this, but the rest of the idea still stands
- adding temporary variables to use rather than continuously returning the same object from eval (arguable I guess, cause one could just use `[p]repl` but I never end up doing that...)

I don't think I would personally be using it as presented in the example, but it seems like a perfectly valid use case too and shows that there are plenty of ways this could be beneficial to have.

Overall, +1 from me.

// *About the implementation, I think `add_dev_env_value` should *probably* raise when trying to replace the name that was already added + imo, it should not be possible for the env values to replace the environment passed by Dev cog*

laggron42:
Forgot to comment 👀 
I fixed the issue we had when testing yesterday, and it raises an exception when accessing a used/reserved name.

Jackenmen:
Since there's been no opposition to this, I'll try to find time to review this next week.

redbot/core/dev_commands.py
```diff
@@ -287,7 +270,10 @@ async def repl(self, ctx):
                     await ctx.send(self.get_syntax_error(e))
                     continue
 
-            variables["message"] = response
+            context = await ctx.bot.get_context(response)
+            env = self.get_environement(context)
+            env["__builtins__"] = __builtins__
+            env["_"] = previous
```
Jackenmen: I think this should only be happening when starting the REPL session (so before the whole `while` loop), rather than on each ran code piece. 

In case you want some reasoning:
a) this doesn't actually work as a session anymore as you're replacing any created variable on each code execution
b) this `context` will not actually be valid as it's not generated from a message using bot command
c) dynamically changing the variables in an active REPL session can result in unexpected behavior, as you've got two options - either you leave any variable that already got its value changed (and that is not gonna be easy to achieve and can be a bit confusing) or you override any variables in the environment with the values you get from `get_environement()` which will be annoying as it would take away the control of the values you have under different names in your REPL session

redbot/core/dev_commands.py
```diff
@@ -92,6 +93,28 @@ def sanitize_output(ctx: commands.Context, input_: str) -> str:
         token = ctx.bot.http.token
         return re.sub(re.escape(token), "[EXPUNGED]", input_, re.I)
 
+    def get_environement(self, ctx: commands.Context) -> dict:
```
Jackenmen: This is spelled incorrectly. Remember to also update any usage of this method.
```suggestion
    def get_environment(self, ctx: commands.Context) -> dict:
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
         """
         self._help_formatter = commands.help.RedHelpFormatter()
 
+    def add_dev_env_value(self, name: str, value: Callable):
```
```suggestion
    def add_dev_env_value(self, name: str, value: Callable[[commands.Context], Any]):
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
...
+
+            One your cog is loaded, the ``mycog`` and ``mycogdata`` values will be implemented in
+            dev commands.
+
+        Parameters
+        ----------
+        name: str
+            The name of your env value
+        value: Callable
+            A callable taking exactly one argument, the context.
```
```suggestion
        value: Callable[[commands.Context], Any]
            The function returning the value of the variable.
            It must take a `commands.Context` as its sole parameter
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
         """
         self._help_formatter = commands.help.RedHelpFormatter()
 
+    def add_dev_env_value(self, name: str, value: Callable):
+        """
+        Add a value to the dev environement (debug, eval and repl commands). If dev mode is
+        disabled, nothing will happen.
```
```suggestion
        Add a custom variable to the dev environment (``[p]debug``, ``[p]eval`, and ``[p]repl` commands).
        If dev mode is disabled, nothing will happen.
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
...
+                        self.bot = bot
+                        bot.add_dev_env_value("mycog", lambda ctx: self)
+                        bot.add_dev_env_value("mycogdata", lambda ctx: self.settings[ctx.guild.id])
+
+                    def cog_unload(self):
+                        self.bot.remove_dev_env_value("mycog")
+                        self.bot.remove_dev_env_value("mycogdata")
+
+            One your cog is loaded, the ``mycog`` and ``mycogdata`` values will be implemented in
+            dev commands.
```
```suggestion
            Once your cog is loaded, the custom variables ``mycog`` and ``mycogdata``
            will be included in the environment of dev commands.
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
...
+
+        Parameters
+        ----------
+        name: str
+            The name of your env value
```
```suggestion
            The name of your custom variable.
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
...
+        value: Callable
+            A callable taking exactly one argument, the context.
+
+        Raise
+        -----
+        TypeError
+            The value isn't a callable.
```
```suggestion
            ``value`` argument isn't a callable.
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
...
+        ValueError
+            The callable takes no or more than one argument.
```
```suggestion
            The passed callable takes no or more than one argument.
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
...
+        RuntimeError
+            The name of the value is either reserved (default environement) or already taken by
+            something else.
```
```suggestion
            The name of the custom variable is either reserved by a variable
            from the default environment or already taken by some other custom variable.
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
...
+            "__builtins__",
+        ]:
+            raise RuntimeError(f"The name {name} is reserved for default environement.")
+        if name in dev.env_extensions:
+            raise RuntimeError(f"The name {name} is already used.")
+        dev.env_extensions[name] = value
+
+    def remove_dev_env_value(self, name: str):
+        """
+        Remove a custom dev env value.
```
```suggestion
        Remove a custom variable from the dev environment.
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
...
+
+        Parameters
+        ----------
+        name: str
+            The name of the env value.
```
```suggestion
            The name of the custom variable.
```

redbot/core/bot.py
```diff
@@ -276,6 +276,88 @@ def reset_help_formatter(self):
...
+
+        Raise
+        -----
+        KeyError
+            The value was never set.
```
```suggestion
            The custom variable was never set.
```

Jackenmen (changes requested):
I've proposed a bunch of docstring changes some of which are related to my belief that "custom dev env value" is not a very precise way to call these variables + and also proposed some other clarity improvements.

As for code review, the only issue I've noticed is REPL sessions not actually be sessions due to how the new environment is generated on each code execution in the session.
And there's also a typo in `get_environement()` method, but that's easy to fix :)

redbot/core/dev_commands.py
laggron42: TIL eval actually updates the env argument, that explains a lot of stuff

redbot/core/dev_commands.py
```diff
@@ -287,18 +272,16 @@ async def repl(self, ctx):
                     await ctx.send(self.get_syntax_error(e))
                     continue
 
-            variables["message"] = response
-
```
Jackenmen, line 291: The existing behavior of `[p]repl` command probably shouldn't be changed here.

redbot/core/dev_commands.py
```diff
@@ -92,6 +93,28 @@ def sanitize_output(ctx: commands.Context, input_: str) -> str:
...
+            "discord": discord,
+            "commands": commands,
+            "_": self._last_result,
+            "__name__": "__main__",
+        }
+        for name, value in self.env_extensions.items():
+            try:
+                env[name] = value(ctx)
+            except Exception as e:
+                env[name] = e
```
Jackenmen: Perhaps this isn't very likely to be an issue, but keeping an entire stack frame alive can be a cause of memory leak so it is recommended to clear the local variables of all the stack frames when the thrown exception with the traceback might be kept alive for a longer time.
```suggestion
            except Exception as e:
                traceback.clear_frames(e.__traceback__)
                env[name] = e
```

Jackenmen (changes requested):
Two more rather minor things. I'll update the PR myself if you won't have time for it so it will get into 3.4.6 either way.

Jackenmen (approved):
Looks good to me now, thanks so much for contributing this new cool feature.

PR #4672, opened 2020-12-16
[Trivia] 2020 Trivia List

OofChair:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [X] New feature

There's a 2015, and a 2016 trivia list, so why not make a 2020 one
Currently there are 20 questions
Sources: [Source 1](https://www.nydailynews.com/news/national/ny-2020-timeline-events-20200705-lvu3hluqdnh3nedmvkpteceqcm-story.html), [Source 2](https://nypost.com/list/major-2020-events/), and [Source 3](https://www.onthisday.com/events/date/2020?p=5)

### DIFF
--- a/redbot/cogs/trivia/data/lists/2020.yaml
+++ b/redbot/cogs/trivia/data/lists/2020.yaml
@@ -29,7 +29,7 @@ When did the Bitcoin Twitter hack occur?:
 - July 15
 - Jul 15th
 - July 15th
-When were the murder first spotted in the US?:
+When were the murder hornets first spotted in the US?:
 - May 2
 - May 2nd
 When did the Beirut explosion happen?:
--- a/redbot/cogs/trivia/data/lists/2020.yaml
+++ b/redbot/cogs/trivia/data/lists/2020.yaml
@@ -5,10 +5,10 @@ When did Prince Harry and Meghan Markle leave the royal family?:
 - Jan 8th
 - January 8th
 When did COVID-19 first appear?:
-- Jan 9th
-- Jan 9
-- January 9
-- January 9th
+- Dec 1, 2019
+- Dec 1st, 2019
+- December 1, 2019
+- December 1st, 2019
 When did Kobe Bryant and Gianna Bryant die? (rest in peace):
 - Jan 26
 - Jan 26th
--- a/redbot/cogs/trivia/data/lists/2020.yaml
+++ b/redbot/cogs/trivia/data/lists/2020.yaml
@@ -1,0 +1,97 @@
+AUTHOR: OofChair#0001
+When did Prince Harry and Meghan Markle leave the royal family?:
+- Jan 8
+- January 8
+- Jan 8th
+- January 8th
+When did COVID-19 first appear?:
+- Jan 9th
+- Jan 9
+- January 9
+- January 9th
+When did Kobe Bryant and Gianna Bryant die? (rest in peace):
+- Jan 26
+- Jan 26th
+- January 26
+- January 26th
+When was Donald Trump impeached?:
+- Feb 5
+- Feb 5th
+- Febuary 5th
+- Febuary 5
+When did Harvey Weinstein get his verdict?:
+- Feb 24
+- Febuary 24
+- Feb 24th
+- Febuary 24th
+When did the Bitcoin Twitter hack occur?:
+- Jul 15
+- July 15
+- Jul 15th
+- July 15th
+When were the murder first spotted in the US?:
+- May 2
+- May 2nd
+When did the Beirut explosion happen?:
+- Aug 4
+- August 4
+- Aug 4th
+- August 4th
+When was Kamala Harris, the first black female vice president, announced as Joe Biden's running mate?:
+- Aug 11
+- Aug 11th
+- August 11th
+- August 11
+When did Chadwick Boseman die of colon cancer?:
+- Aug 28
+- Aug 28th
+- August 28
+- August 28th
+When did Ruth Bader Ginsburg die?:
+- Sept 18
+- September 18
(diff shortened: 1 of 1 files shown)


redbot/cogs/trivia/data/lists/2020.yaml
```diff
@@ -0,0 +1,97 @@
+AUTHOR: OofChair#0001
+When did Prince Harry and Meghan Markle leave the royal family?:
+- Jan 8
+- January 8
+- Jan 8th
+- January 8th
+When did COVID-19 first appear?:
```
Jackenmen, line 5: I'm guessing you meant the first appearance outside China?

redbot/cogs/trivia/data/lists/2020.yaml
```diff
@@ -0,0 +1,97 @@
...
+- Feb 24
+- Febuary 24
+- Feb 24th
+- Febuary 24th
+When did the Bitcoin Twitter hack occur?:
+- Jul 15
+- July 15
+- Jul 15th
+- July 15th
+When were the murder first spotted in the US?:
```
aikaterna: What even is this question?

aikaterna:
IMO, this needs more questions, the list is too short... it is too COVID-focused and also politics focused. There are more things that happened in 2020 besides these topics.

redbot/cogs/trivia/data/lists/2020.yaml
OofChair: meant murder hornets, updated

redbot/cogs/trivia/data/lists/2020.yaml
OofChair: First appearance inside China

Jackenmen: Then I don't think that's true, [according to WHO](https://www.who.int/docs/default-source/coronaviruse/situation-reports/20200423-sitrep-94-covid-19.pdf) the first human cases were reported in early December 2019.

OofChair: Ah yes, I am seeing that one of the earliest cases was in December 2019.

NeuroAssassin:
In addition to what aika said, I was wondering if we could have more questions that did not ask about dates?  If you look at the other 2015 and 2016 lists, they are largely diverse on answer choices.

OofChair:
> IMO, this needs more questions, the list is too short... it is too COVID-focused and also politics focused. There are more things that happened in 2020 besides these topics.

I thought it might be a little short, wasnt sure
I do agree, it is a little bit COVID and politics based, was looking more at US things than whole world things so that's probably why. 

> In addition to what aika said, I was wondering if we could have more questions that did not ask about dates? If you look at the other 2015 and 2016 lists, they are largely diverse on answer choices.

I did see that, I was thinking about that. 
I'm going to implement these changes over the next few days, should I close this one and make a new one when I'm ready?

Kreusada:
Agreed, dates aren't the best thing to quiz about. All these major 2020 events occurred, but people will be more aware of the context of these events rather than their time of occurrence. Also, I'd add a lot more questions, focusing on a wider range of events that occurred in 2020, instead of focusing on prominence. Best of luck

Drapersniper:
A much as I am in favour of a new trivia, I do believe this might be a poor choice should it be approved.

It's too US, death and Political centric for my own taste, and as a non US citizen there's a lot of negative/polarising information in there which I wouldn't want distributes in my own servers.

ltzmax:
I find this a bit scary, on how 2020 have been like is actually nothing that I want on my bot with this. I would rather do an 2019 over 2020. This year hasn't been the best. 

But yea more quest and little less dates would be a bit better maybe.

OofChair:
Thanks everyone for your feedback. So here's some things I should do, is this correct?

- Make it less US
- Less death and political things
- Less/no questions with dates

Like @maxbooiii said, I'm thinking about making a 2019 trivia list instead, that would be better.
If this is correct, I will begin reworking on this 👍🏻

Drapersniper:
> Thanks everyone for your feedback. So here's some things I should do, is this correct?
> 
> * Make it less US
> * Less death and political things
> * Less/no questions with dates
> 
> Like @maxbooiii said, I'm thinking about making a 2019 trivia list instead, that would be better.
> If this is correct, I will begin reworking on this 👍🏻

Really up to you, I believe a 2020 is doable with question like:
"In 2020 Steam broke its own record of concurrent online players, how many players were online ?"  And other trivia questions that are not country specific or pandemic specific.

bobloy:
My contribution to this list, some things that aren't all politics:
https://mashable.com/article/what-happened-2020/

OofChair:
> My contribution to this list, some things that aren't all politics:
> https://mashable.com/article/what-happened-2020/

This will help, thank you :)

redbot/cogs/trivia/data/lists/2020.yaml
```diff
@@ -0,0 +1,48 @@
...
+- august
+- August
+When did WHO, or the World Health Organization, announce a public health emergency about COVID?:
+- january
+- January
+When did the worldwide cases of COVID-19 pass 10 million?:
+- june
+- June
+What streaming service had their subscriber count doubled during the first 3 months of 2020?:
+- Netflix
```
ltzmax, line 38: You've added `Netflix` 2 times, one here and below this one, maybe remove one of them? since they does not look different :P

OofChair: Thanks MAX, but I think I'm gonna close this because I have no time to work on this currently because of life things, and I might reopen later.

PR #4694, opened 2020-12-25
[Streams] Use timezone-aware time

### LINKED ISSUE #4693: [Streams] Mismatched stream start times in embed
# Other bugs

Receive YouTube scheduled streams

An embed with some information pertaining to upcoming stream

#### What actually happened?
Received aforementioned embed, but times are off in the description versus the footer. The footer start time is correct, the one in the description is off by 8 hours, so this seems related to time zones.

In my instance and on my Discord client, my time zone is America/Vancouver (PST/GMT-8). Configure a YouTube channel and wait.
image

Injabie3:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This PR fixes #4693 by converting the naive `.now()` time into a timezone-aware time (such as UTC, but any timezone will work) instead of vice-versa, so that the timedelta difference is relatively correct in the scheduled YouTube embed description.

Screenshots below:
image
image


Jackenmen (approved):
While I was not able to reproduce times being shown with a wrong timezone offset, the cog still works properly with this change and the code changes done here should indeed fix this issue for anyone that can actually experience it.

PR #4697, opened 2020-12-26
Wrap `Red-DiscordBot[postgres]` in quotes in Unix install docs

elysweyr:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Just added an "attention" block inside of the Linux/macOS bot installation documention which highlights the necessity of escaping brackets in terminals.

### DIFF
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -446,6 +446,7 @@ Or, to install with PostgreSQL support:
     python -m pip install -U pip setuptools wheel
     python -m pip install -U "Red-DiscordBot[postgres]"
 
+
 .. note::
 
     These commands are also used for updating Red
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -449,7 +449,7 @@ Or, to install with PostgreSQL support:
 
 .. attention:: 
 
-    Depending on your system brackets may need to be escaped. A common escape sequence would look like this: ``\[``
+    Depending on your system, brackets may need to be escaped. A common escape sequence would look like this: ``\[``
 
 .. note::
 
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -444,12 +444,7 @@ Or, to install with PostgreSQL support:
 .. code-block:: none
 
     python -m pip install -U pip setuptools wheel
-    python -m pip install -U Red-DiscordBot[postgres]
-
-
-.. attention:: 
-
-    Depending on your system, brackets may need to be escaped. A common escape sequence would look like this: ``\[``
+    python -m pip install -U "Red-DiscordBot[postgres]"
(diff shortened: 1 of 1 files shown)


Drapersniper:
I would prefer if we add specifics to which of the supported OS need it and which commands need it...

wouldn't want people running `pip install RedDiscordBot\[postgres\]` on systems that don't need it

elysweyr:
>  Depending on your system brackets may need to be escaped.

This seems to be well phrased to me. @Drapersniper

owocado:
> > Depending on your system brackets may need to be escaped.
> 
> This seems to be well phrased to me. @Drapersniper

you're missing a comma between "system" and "brackets". the lack of comma changes the whole meaning of that sentence imo.

also what Draper said below.

Drapersniper:
> > Depending on your system brackets may need to be escaped.
> 
> This seems to be well phrased to me. @Drapersniper

I would disagree as a user this only confuses me as It makes me question if I need it or not for my setup.

If escaping is needed it should be on the OS specific command section and not a blanket warning at the top.

elysweyr:
@Drapersniper So you suggest to separate the macOS section from the generic Linux part since this applies at least to the macOS installation procedure?

Drapersniper:
> @Drapersniper So you suggest to separate the macOS section from the generic Linux part since this applies at least to the macOS installation procedure?

Haven't had a chance to look this in depth so I'm not sure which commands need escaping on which OS. The ambiguity at the moment is what concerns me, if it's just one command on one specific OS, then this should be added directly to the command, if it's not a 100% repo or different MacOS version have the need to it specifying the version where it is required should be the way we go about this.

If it's a unsupported env then no change is needed if we aren't going to support it.

elysweyr:
@Drapersniper 

It's about 
`python -m pip install -U Red-DiscordBot[postgres]`

I experience this problem on my macOS `10.15.7 (19H2)` system using Terminal (`2.10`), iTerm2 (`3.3.9`) and the pre-installed zsh version (` 5.7.1 (x86_64-apple-darwin19.0)`).

This is a common setup and in addition to that there is no limited support stated in the documentation for these versions listed above. 


>  [...] not a blanket warning at the top.

As stated above this is a not a "blanket warning" and directly added to the respective section where this note applies.

Jackenmen:
Could we just replace `python -m pip install -U Red-DiscordBot[postgres]` command with `python -m pip install -U "Red-DiscordBot[postgres]"`? Will that fix the problem for zsh shell?

Jackenmen (approved):
I ensured that `python -m pip install -U "Red-DiscordBot[postgres]"` works on `cmd`, `bash`, and `zsh` (and I'm suspecting it will also work fine on any other shells too) and I've updated this PR appropriately.

Thanks for bringing our attention to this issue!

PR #4714, opened 2021-01-07
[Mod] Make tempban not fail if no vanity url is set up

Dav-Git:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

If a guild has the VANITY_URL feature but doesn't actually have a value set for their vanity url, ``guild.vanity_invite()`` fails and breaks ``[p]tempban``. This PR fixes that.


Jackenmen:
@Dav-Git Any chance you could include the traceback for this?

Dav-Git:
sure
grafik

Jackenmen (approved):
Thanks for the fix

PR #4719, opened 2021-01-09
Fix auto labeler error from invalid yaml -> json

Flame442:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Without quotes, it converts as `null`. Adding quotes allows it to properly convert as a string.


Jackenmen (approved):
thanks

PR #4722, opened 2021-01-10
[Utils] Additional features for chat_formatting

Kreusada:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Added some additional features for chat_formatting which I would like to see and eventually use. These features include returning the text as a spoiler, ~~with a strikethrough~~, with snake_spaces (just_like_this) and with alternating cases (JuSt LiKe ThIs). Please let me know of your thoughts, I used modified docstrings from previous utils, and kept the escape formatting for each of them.

++ Did not realize that strikethrough was already available, so I removed it.

### DIFF
--- a/redbot/core/utils/chat_formatting.py
+++ b/redbot/core/utils/chat_formatting.py
@@ -144,6 +144,92 @@ def italics(text: str, escape_formatting: bool = True) -> str:
     text = escape(text, formatting=escape_formatting)
     return "*{}*".format(text)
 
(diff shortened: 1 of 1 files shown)


Kreusada:
Due to the fact that the tox style check failed _and_ I unintentionally deleted the patch branch, I will probably end up closing this PR, and potentially reopening it if I feel that these functions are still worth being added.

PR #4727, opened 2021-01-12
Fix the negation rules in labeler configuration

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

By default *any* rule must match, so negation rules cause all files except the ones specified in the rule to match. This fixes it by properly requiring the negation rule to match **in addition to** the other rules.


Jackenmen:
Bypassing QA and admin merging to stop this labeling madness.

PR #4730, opened 2021-01-13
Update `actions/labeler` to use `v3` version

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

The more advanced rules that we were trying to use are available since `actions/labeler@v3`, but *someone* thought it's a good idea to use an outdated version of the labeler...


Jackenmen:
clearly doesn't :(

Jackenmen:
I think we can all agree that @Kowlin is to blame :)

PR #4738, opened 2021-01-17
Fix log rotation in Red

### LINKED ISSUE #4405: Properly rotate logging
# Other bugs

Check log files due to no disk space. 12gb for 2 of reds files. The logs are only labelled as part 1 and 9 with nothing in between.

Log files to correctly rotate and not be just labelled as 1 and 9

Just running red with ~100 cogs.

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #4405 
L61 and L89 are not a part of the fix, but I figured I might as well fix the unescaped dots in those regexes while I'm editing the logging handler.

If you want to test, I suggest changing `maxBytes` in L288 and/or L295 to `10_000` (or something around it) and run something to fill the logs, for example:
```py
import logging

log = logging.getLogger("red")

async def func():
    while True:
        log.info("A"*500)
        await asyncio.sleep(0.5)
return asyncio.create_task(func())
```

I've also tested that when you only have `red-part1.log` and `red-part9.log`, Red still successfully manages to eventually rollout `red-part9.log`

### DIFF
--- a/redbot/logging.py
+++ b/redbot/logging.py
@@ -58,7 +58,7 @@ class RotatingFileHandler(logging.handlers.RotatingFileHandler):
         self.baseStem = stem
         self.directory = directory.resolve()
         # Scan for existing files in directory, append to last part of existing log
-        log_part_re = re.compile(rf"{stem}-part(?P<partnum>\d+).log")
+        log_part_re = re.compile(rf"{stem}-part(?P<partnum>\d+)\.log")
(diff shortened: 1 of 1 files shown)


TrustyJAID (dismissed):
Loogs Kood to me!

TrustyJAID (approved):
Still loogs kood to me.

PR #4739, opened 2021-01-17
Add `filterhit` case type to Filter

### LINKED ISSUE #4676: Include the ability to log filter hits
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality

We have a `filterban` casetype, so **within a similar vein** I was wondering if we could get a `filterhit` command - it would allow us to **mark a channel to act as a log** for when someone uses a blacklisted phrase - loosely related, also record it as a modlog case?

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Fixes #4676 

This case type isn't triggered on nickname filtering, I think it would probably make more sense to add a separate case type for it, but it wasn't requested in the issue so I haven't done anything about this.

Note: This PR is untested!

### DIFF
--- a/redbot/cogs/filter/__init__.py
+++ b/redbot/cogs/filter/__init__.py
@@ -2,5 +2,7 @@ from .filter import Filter
 from redbot.core.bot import Red
 
 
-def setup(bot: Red):
-    bot.add_cog(Filter(bot))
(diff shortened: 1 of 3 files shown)


TrustyJAID (approved):
Looks good to me.

PR #4741, opened 2021-01-17
Ensure no role hierarchy bypassing in Mutes

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Before this, it was possible for the admin to set a role higher or equal to their own top role as mute role.
It was also possible for the moderator to assign roles higher or equal to their own top role if admin set such role as a mute role.


TrustyJAID (approved):
Looks good to me!

PR #4742, opened 2021-01-18
Use partial messages in Streams cog to avoid potential leakage

### LINKED ISSUE #4731: Prevent possible leakage from discord.Message objects in Streams
Currently, we're caching whole message objects for the messages made by Streams cog so that we can auto delete them. This is prone to leakage on gateway reconnection so we should instead cache only channel id and message id and use the partial message objects to delete alert messages.
Doing it like this is dependent on d.py 1.6 in Red, so for now this is blocked by #4728

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #4731

### DIFF
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -888,7 +888,7 @@ class Streams(commands.Cog):
                     if _class.__name__ == "YoutubeStream":
                         raw_stream["config"] = self.config
                     raw_stream["token"] = token
-            raw_stream["bot"] = self.bot
+            raw_stream["_bot"] = self.bot
             streams.append(_class(**raw_stream))
 
         return streams
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -209,6 +209,7 @@ class Streams(commands.Cog):
         await self.maybe_renew_twitch_bearer_token()
         token = (await self.bot.get_shared_api_tokens("twitch")).get("client_id")
         stream = TwitchStream(
+            bot=self.bot,
             name=channel_name,
             token=token,
             bearer=self.ttv_bearer_cache.get("access_token", None),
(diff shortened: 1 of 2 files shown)


redbot/cogs/streams/streams.py
```diff
@@ -714,14 +714,23 @@ async def _stream_alerts(self):
...
+        *,
+        is_schedule: bool = False,
     ):
         m = await channel.send(
             content,
             embed=embed,
             allowed_mentions=discord.AllowedMentions(roles=True, everyone=True),
         )
-        stream._messages_cache.append(m)
+        message_data = {"channel": m.channel.id, "message": m.id}
```
TrustyJAID: Can we include the guild ID in here so that when we pull the channel again later it doesn't have to iterate through all the bots guilds to find the channel?

redbot/cogs/streams/streamtypes.py
```diff
@@ -68,14 +69,19 @@ async def is_online(self):
     def make_embed(self):
         raise NotImplementedError()
 
+    def iter_messages(self):
+        for msg_data in self.messages:
+            data = msg_data.copy()
+            channel = self.bot.get_channel(msg_data["channel"])
```
TrustyJAID: This should be replaced with `self.bot.get_guild(data.pop("guild_id"))` and `guild.get_channel(msg_data["channel"]) with the addition of the previous comment.

redbot/cogs/streams/streamtypes.py
```diff
@@ -56,10 +56,11 @@ class Stream:
     token_name: ClassVar[Optional[str]] = None
 
     def __init__(self, **kwargs):
+        self._bot = kwargs.pop("_bot")
```
TrustyJAID, line 59: This change breaks the `[p]twitchstream` and likely `[p]youtubestream` commands with a KeyError.

TrustyJAID (approved):
This looks good to me!

PR #4744, opened 2021-01-21
[Streams] Change streamtypes log name

Injabie3:
### Type
- [x] Bugfix

This small PR changes the log name in `streamtypes.py` to match the one in `streams.py`, as I wasn't seeing the debug level log messages after launching the bot with the `--debug` flag.


Jackenmen (approved):
Thanks!

PR #4552, opened 2020-10-27
Add better error logging for unexpected errors, add error messages for exceeded YouTube quota

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

### Description of the changes
Let me know if you want me to remove the quota parts and I will.

PR #4745, opened 2021-01-21
Fix error handling for API calls to videos endpoint in YouTube alerts

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

I did something similar in #4552, but I've missed calls to the videos API endpoint which can also return the same errors.


TrustyJAID (approved):
This looks to be working as intended.

PR #4747, opened 2021-01-22
Docstring fixes for dev env values

laggron42:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Looks like "Raise" is not yet a valid header in Napoleon on Sphinx 3.1.2 :o 

Also the example looks better as a header instead of an admonition


Jackenmen (approved):
Thanks for the follow-up! LGTM.

PR #4753, opened 2021-01-25
Make `[p]mockmsg` include the attachments in the mocked message

felineturtle67:
### Type
- [x] Enhancement

Updated mockmsg to allow images and also to allow discord.User for those cases where you need it. Change the raise Exception to an exception you like.

### DIFF
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -336,18 +336,12 @@ class Dev(commands.Cog):
 
     @commands.command(name="mockmsg")
     @checks.is_owner()
-    async def mock_msg(self, ctx, user: Union[discord.Member, discord.User, int], *, content: str):
+    async def mock_msg(self, ctx, user: discord.Member, *, content: str):
         """Dispatch a message event as if it were sent by a different user.
 
-        Only reads the raw content of the message. Attachments, embeds etc. are
-        ignored.
+        Embeds are ignored. Attachments are accepted
         """
-        if isinstance(user, int):
-            try:
-                user = await self.bot.fetch_user(user)
-            except discord.NotFound:
-                raise Exception("User not found") #change this to some other Exception
-                
+       
         old_author = ctx.author
         old_content = ctx.message.content
         ctx.message.author = user
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -352,11 +352,10 @@ class Dev(commands.Cog):
         if ctx.message.attachments:
             attachment = ctx.message.attachments[0]
             image = attachment.proxy_url
+            name = attachment.filename
+            type = name.split()[1]
         
         if image:
-            imgname = image.split(".")
-            type = imgname[1]
-            name = imgname[0]
             image = discord.File(image, filename=f"{name}.{type}")
             ctx.message.attachments = []
             ctx.message.attachments.append(image)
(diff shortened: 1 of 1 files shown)


redbot/core/dev_commands.py
```diff
@@ -336,17 +336,33 @@ async def mock(self, ctx, user: discord.Member, *, command):
...
     @commands.command(name="mockmsg")
     @checks.is_owner()
-    async def mock_msg(self, ctx, user: discord.Member, *, content: str):
+    async def mock_msg(self, ctx, user: Union[discord.Member, discord.User, int], *, content: str):
         """Dispatch a message event as if it were sent by a different user.
 
         Only reads the raw content of the message. Attachments, embeds etc. are
         ignored.
         """
+        if isinstance(user, int):
```
Drapersniper: What is the purpose of this? when will a user or member known to the user be just a int?

redbot/core/dev_commands.py
```diff
@@ -336,17 +336,33 @@ async def mock(self, ctx, user: discord.Member, *, command):
...
+            try:
+                user = await self.bot.fetch_user(user)
```
Drapersniper: Please use the built-in using for fetching users and members since those use the bot cache to avoid api spam

redbot/core/dev_commands.py
```diff
@@ -336,17 +336,33 @@ async def mock(self, ctx, user: discord.Member, *, command):
...
+            except discord.NotFound:
+                raise Exception("User not found") #change this to some other Exception
```
Drapersniper: Don't raise bare exceptions inside a command

redbot/core/dev_commands.py
```diff
@@ -336,17 +336,33 @@ async def mock(self, ctx, user: discord.Member, *, command):
...
+                
         old_author = ctx.author
         old_content = ctx.message.content
         ctx.message.author = user
         ctx.message.content = content
-
+        
+        image = False
```
Drapersniper: image should be `None` not `False`

redbot/core/dev_commands.py
```diff
@@ -336,17 +336,33 @@ async def mock(self, ctx, user: discord.Member, *, command):
 
     @commands.command(name="mockmsg")
     @checks.is_owner()
-    async def mock_msg(self, ctx, user: discord.Member, *, content: str):
+    async def mock_msg(self, ctx, user: Union[discord.Member, discord.User, int], *, content: str):
```
Drapersniper: Why should we accept a User in this command?

Drapersniper (changes requested):
Okay a few things first.

1. Make sure to read and follow https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/CONTRIBUTING.md when creating PRs, all the test are failing due to you not doing this.

2. Make sure to test your PRs.

3. Name your PR's appropriately the current name of this one is useless.

redbot/core/dev_commands.py
felineturtle67: uh i made a comment for you to change it becuase im not sure which exceptions you ppl use.

redbot/core/dev_commands.py
felineturtle67: alr ill fix that

redbot/core/dev_commands.py
felineturtle67: yeah i removed that

PR #4754, opened 2021-01-25
[Utils] Spoiler Function for Chat Formatting

Kreusada:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

Re-opened this PR because I deleted the PR branch.

Additional formatting functions including spoilers, snake_spaces and alternating case (ExAmPlE)

I think that spoiler is definitely worthy of being added, but I'm 50/50 on the other two.

### DIFF
--- a/redbot/core/utils/chat_formatting.py
+++ b/redbot/core/utils/chat_formatting.py
@@ -145,55 +145,64 @@ def italics(text: str, escape_formatting: bool = True) -> str:
     return "*{}*".format(text)
 
 def spoiler(text: str, escape_formatting: bool = True) -> str:
-    """
-    Get the given text as a spoiler.
+    """Get the given text as a spoiler.
+    
     Note: By default, this function will escape ``text`` prior to making the text a spoiler.
+    
     Parameters
     ----------
     text : str
         The text to be marked up.
     escape_formatting : `bool`, optional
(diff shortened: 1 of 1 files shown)


Jackenmen:
My thoughts on this are that `snake()` and `alternatingcase()` functions are too specific and it's quite easy to make yourself and therefore I don't think they should be included in our utils.

`spoiler()` on the other hand seems like a reasonable addition and fits well with the rest of the markdown functions we already have in `.utils.chat_formatting` so I think this PR could get in Red if `snake()` and `alternatingcase()` would be excluded from it.

Kreusada:
Sure. I did think from the start that those other two functions were a bit querky, but its always worth getting an opinion. I'll take those out.

Jackenmen (changes requested):
This PR seems fine, but you're failing the style check. Please run black on your code as explained in [our contributing guidelines](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/CONTRIBUTING.md#41-setting-up-your-development-environment). If you need any help, feel free to ask on our Discord server.

Kreusada:
Good to go 👍

Jackenmen (approved):
LGTM, thanks!

PR #4761, opened 2021-01-26
show a message after cleanup with info

felineturtle67:
Shows a message after the cleanup with the amount of messages deleted

### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -120,7 +120,8 @@ class Cleanup(commands.Cog):
                 if number is not None and number <= len(collected):
                     break
 
-        return collected
+        return collected, len(collected)
+    
 
     @commands.group()
     async def cleanup(self, ctx: commands.Context):
@@ -162,7 +163,7 @@ class Cleanup(commands.Cog):
             else:
                 return False
 
-        to_delete = await self.get_messages_for_deletion(
+        to_delete, deleted = await self.get_messages_for_deletion(
             channel=channel,
             number=number,
             check=check,
@@ -181,6 +182,7 @@ class Cleanup(commands.Cog):
         log.info(reason)
 
         await mass_purge(to_delete, channel)
+        await ctx.send(f"Deleted {deleted} messages", delete_after=6)
 
     @cleanup.command()
     @commands.guild_only()
@@ -227,7 +229,7 @@ class Cleanup(commands.Cog):
             else:
                 return False
 
-        to_delete = await self.get_messages_for_deletion(
+        to_delete, deleted = await self.get_messages_for_deletion(
             channel=channel,
             number=number,
             check=check,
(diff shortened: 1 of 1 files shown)


felineturtle67:
ok fixed an issue

redbot/cogs/cleanup/cleanup.py
```diff
@@ -181,6 +182,7 @@ def check(m):
         log.info(reason)
 
         await mass_purge(to_delete, channel)
+        await ctx.send(f"Deleted {deleted} messages", delete_after=6)
```
NeuroAssassin, line 185: This needs to be i18n compatible.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -251,6 +253,7 @@ def check(m):
...
```
NeuroAssassin, line 256: This needs to be i18n compatible.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -292,6 +295,7 @@ async def after(
...
```
NeuroAssassin, line 298: This needs to be i18n compatible.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -339,6 +343,7 @@ async def before(
...
```
NeuroAssassin, line 346: This needs to be i18n compatible.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -391,6 +396,7 @@ async def between(
...
```
NeuroAssassin, line 399: This needs to be i18n compatible.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -429,6 +435,7 @@ async def messages(
...
```
NeuroAssassin, line 438: This needs to be i18n compatible.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -515,6 +522,7 @@ def check(m):
...
```
NeuroAssassin, line 525: This needs to be i18n compatible.

NeuroAssassin (changes requested):
So, I think this is a PR that could prove to have some use in the future, so I wanted to offer a few points that need a few touch ups.

1. There needs to be a toggle to turn off the messages after the cleanup.  What the default is can be left to other people to discuss (but in my personal opinion would be best defaulted to off).
2. Any messages that are sent from the bot need to be i18n (internationalization) compatible, so that they can be translated and localized.  I've marked them below, and feel free to ask on the Red server if you need help with it.

Overall, I'm glad that you have found feature that can prove to be useful 👍.

PR #4769, opened 2021-01-28
Add a note in systemd guide saying working instance is required

Vexed01:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

I noticed this was in PM2 guide, so why not put it in the systemd guide?
This is just a simple copy-paste.


Jackenmen (approved):
Yeah, I guess there's no reason not to put it here, thanks.

PR #4774, opened 2021-01-30
[Streams] Some fixes for Twitch streams.

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This fixes many KeyError that were happening when trying to get informations of Twitch streams.

### DIFF
--- a/redbot/cogs/streams/streamtypes.py
+++ b/redbot/cogs/streams/streamtypes.py
@@ -293,9 +293,8 @@ class TwitchStream(Stream):
                         "https://api.twitch.tv/helix/games", headers=header, params=params
                     ) as r:
                         game_data = await r.json(encoding="utf-8")
-                if game_data:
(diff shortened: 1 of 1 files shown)


Jackenmen:
I was not able to get KeyErrors in any of the places this PR touches.

The best I could find is getting an empty data key when giving invalid IDs but that should not happen with valid API calls and as far as I can tell, we only pass valid IDs that we get straight from the Twitch API just moments before the API call so that should not be the case.

I'm marking this as "Needs Info" as I was not able to confirm that any of these need fixing. If you have more information that could help us reproduce these issues, that would be greatly appreciated, but for now, I sadly can't review this.

PredaaA:
Closing that PR as it is actually invalid. See #4808.
I *may* PR something to address what I raised in the linked issue.

PR #4790, opened 2021-02-07
Allow Cleanup before/after to accept message replies

phenom4n4n:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Allows `[p]cleanup after` to work without message ID's being provided if the invocation message has a reference.

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -122,6 +122,21 @@ class Cleanup(commands.Cog):
 
         return collected
 
+    @staticmethod
+    async def get_message_from_reference(channel: discord.TextChannel, reference: discord.MessageReference) -> Optional[discord.Message]:
+        message = None
+        resolved = reference.resolved
+        if resolved and isinstance(resolved, discord.Message):
+            message = resolved
+        elif (message := reference.cached_message):
+            pass
+        else:
+            try:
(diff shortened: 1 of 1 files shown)


phenom4n4n:
Also updated `[p]cleanup before` to use the same logic to cleanup messages before a replied message.

redbot/cogs/cleanup/cleanup.py
```diff
@@ -300,7 +328,7 @@ async def after(
     async def before(
         self,
         ctx: commands.Context,
-        message_id: RawMessageIds,
+        message_id: Optional[RawMessageIds],
         number: positive_int,
```
bobloy, line 332: I don't think this works when the number of messages you want to delete can be interpreted as a message ID.

bobloy (changes requested):

Kowlin (approved):
Does what it sets out to do.

redbot/cogs/cleanup/cleanup.py
Kowlin: Funnily enough, its fine and Optional just handles it.

Rapptz: In order for this to be an issue the number would have to be 17 digits long so it's unlikely you'd ever want to delete that many messages.

PR #4791, opened 2021-02-07
[Cleanup] Fix error when providing message snowflake above the maximum (9223372036854775807)

NeuroAssassin:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

~~I needed my traditional per-version credit in the documentation~~

Fixes big angry error when you try to delete before, after or between a message that cannot be a snowflake.

Note: untested, but should work tm.

### DIFF
--- a/redbot/cogs/cleanup/converters.py
+++ b/redbot/cogs/cleanup/converters.py
@@ -6,6 +6,7 @@ from redbot.core.utils.chat_formatting import inline
 
 _ = Translator("Cleanup", __file__)
 
+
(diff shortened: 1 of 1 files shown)


Jackenmen (changes requested):
Could you set this as a module-level constant written as `2 ** 63` instead? It would be clearer than having a somewhat "random" number (for the person reading this in the future) in the code base.

Jackenmen (approved):
Looks good now, thank you.

PR #4792, opened 2021-02-08
grammar fixes

felineturtle67:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

i live in America, idk how your grammar works, but i think this is right

### DIFF
--- a/redbot/core/_sharedlibdeprecation.py
+++ b/redbot/core/_sharedlibdeprecation.py
@@ -21,9 +21,9 @@ class SharedLibImportWarner(MetaPathFinder):
         if parts[0] != "cog_shared" or len(parts) != 2:
             return None
         msg = (
-            "One of cogs uses shared libraries which are"
+            "One of the cogs uses shared libraries which are"
             " deprecated and scheduled for removal in the future.\n"
-            "You should inform author of the cog about this message."
+            "You should inform the author of the cog about this message."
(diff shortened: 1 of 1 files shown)


redbot/core/_sharedlibdeprecation.py
```diff
@@ -21,9 +21,9 @@ def find_spec(self, fullname, path, target=None) -> None:
         if parts[0] != "cog_shared" or len(parts) != 2:
             return None
         msg = (
-            "One of cogs uses shared libraries which are"
+            "One of the cogs uses shared libraries which are"
             " deprecated and scheduled for removal in the future.\n"
-            "You should inform author of the cog about this message."
+            "You should inform the author of the cog about this message."
```
madebylydia: It could be multiple authors. I think the right one would be `You should inform author(s) of the cog about this message.`

felineturtle67: gud point

bobloy: > It could be multiple authors. I think the right one would be `You should inform author(s) of the cog about this message.`

`You should inform the author(s) of the cog...` IMO

felineturtle67:
the requested change has been made

Jackenmen (approved):

PR #4794, opened 2021-02-08
Copy info about the bands from `[p]eq set` help to `[p]eq` help

### LINKED ISSUE #4716: Copy info about the bands from help of `[p]eq set` to help of `[p]eq`?
I think that would make it a bit more obvious, especially when you consider that the ℹ️ emoji shows you the help of `[p]eq`, not the help of `[p]eq set`.
Alternatively, we could have `[p]eq explain` or something, but it's probably not worth adding another command in the group.

What do you think?

ltzmax:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This close #4716


Jackenmen (approved):
Thanks, looks great.

PR #4801, opened 2021-02-10
Add meaningful error messages for bad arguments in [p]bank set

### LINKED ISSUE #4789: `SetParser` converter (`[p]bank set` command) should have meaningful error message

`[p]bank set` command

Currently using invalid input with `[p]bank set` gives an error that doesn't tell you what's wrong:
image

We should add some meaningful error messages to the converter used there.

rijusougata13:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

If the user passes some invalid argument, the bot will now give a meaningful error message in reply.
 
Fixes #4789 

### screenshot
Screenshot from 2021-02-10 19-25-19

### DIFF
--- a/redbot/cogs/economy/economy.py
+++ b/redbot/cogs/economy/economy.py
@@ -100,7 +100,7 @@ def guild_only_check():
 class SetParser:
     def __init__(self, argument):
         allowed = ("+", "-")
-        if(argument.isdigit()):
+        if argument.isdigit():
             self.sum = int(argument)
         if argument and argument[0] in allowed:
             if self.sum < 0:
@@ -112,8 +112,8 @@ class SetParser:
             self.sum = abs(self.sum)
         elif argument.isdigit():
             self.operation = "set"
-        elif (not argument.isdigit()):
-            self.operation ='error'
+        elif not argument.isdigit():
+            self.operation = "error"
         else:
             raise RuntimeError
 
@@ -269,7 +269,7 @@ class Economy(commands.Cog):
                     currency=currency,
                     user=to.display_name,
                 )
-            elif creds.operation =='set':
+            elif creds.operation == "set":
                 await bank.set_balance(to, creds.sum)
                 msg = _("{author} set {user}'s account balance to {num} {currency}.").format(
                     author=author.display_name,
@@ -277,8 +277,10 @@ class Economy(commands.Cog):
                     currency=currency,
                     user=to.display_name,
                 )
-            elif creds.operation =='error':
-                msg="Please enter a valid argument!"
+            elif creds.operation == "error":
+                msg = (
+                    "Can not set bank balance to a non-integer value.Try giving a valid argument!"
+                )
         except (ValueError, errors.BalanceTooHigh) as e:
             await ctx.send(str(e))
         else:
--- a/redbot/cogs/economy/economy.py
+++ b/redbot/cogs/economy/economy.py
@@ -278,9 +278,7 @@ class Economy(commands.Cog):
                     user=to.display_name,
                 )
             elif creds.operation == "error":
-                msg = (
-                    "Can not set bank balance to a non-integer value.Try using a valid argument such as...(+1. 1. -1)"
-                )
+                msg = "Can not set bank balance to a non-integer value.Try using a valid argument such as...(+1. 1. -1)"
         except (ValueError, errors.BalanceTooHigh) as e:
             await ctx.send(str(e))
         else:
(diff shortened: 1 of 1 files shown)


NeuroAssassin (changes requested):
Hey there, glad that you took this PR!  This is not a formal review of any sort, but can you run black on the project (`make reformat` in root folder) so that they style check will pass?

Drapersniper:
One thing worth mentioning is errors should be explicit, "Enter a **valid** argument" leaves me wondering what is a valid argument and how should I as a brand new user know what arguments are valid?

rijusougata13:
> One thing worth mentioning is errors should be explicit, "Enter a **valid** argument" leaves me wondering what is a valid argument and how should I as a brand new user know what arguments are valid?

ok i am adding some more meaningful message for the user!

rijusougata13:
@NeuroAssassin  all tests passed!!..is it okay now?

redbot/cogs/economy/economy.py
```diff
@@ -266,14 +269,16 @@ async def _set(self, ctx: commands.Context, to: discord.Member, creds: SetParser
...
+            elif creds.operation == "set":
                 await bank.set_balance(to, creds.sum)
                 msg = _("{author} set {user}'s account balance to {num} {currency}.").format(
                     author=author.display_name,
                     num=humanize_number(creds.sum),
                     currency=currency,
                     user=to.display_name,
                 )
+            elif creds.operation == "error":
+                msg = "Can't set bank balance to a non-integer value.Try using a valid argument,for example: `+100`, `-100`, or `100`"
```
madebylydia: Sentences should have a space before `.` and `,`.

madebylydia (changes requested):

rijusougata13:
@Predeactor  PTAL

madebylydia:
> @Predeactor  PTAL

This is still not correct. Ideally, it would be:

"Can't see bank balance to a non-integer value. Try using a valid argument, for example: `+100`, `-100`, or `100`."

Also next time do not ping me for deleting your mention right after in testing and then DMing me just for verifying your string.

Jackenmen (changes requested):
Hi @rijusougata13, thanks for deciding to contribute. Before we can get this merged, I noticed a few things that can be improved here. This review might seem a bit lengthy, but the suggested changes are not too advanced and I just wanted to explain the motivation between all of these suggestions properly.


Firstly, I think the error messages could be clearer, possibly different for each of the errors.
For example, when an integer with a sign is used and the number is equal to zero, the message could be something like:
> Invalid value, the amount of currency to increase or decrease must be an integer different from zero.

And when the given value is not an integer at all, the message could be something like:
> Invalid value, the argument must be an integer, optionally preceded with a `+` or `-` sign.

This way, it would be clearer to the user what was the issue with the input they given.


Another thing that I've noticed is the lack of `_()` usage, which is required in our project for the strings to get translated if the translation is available. This is easy to miss when contributing for the first time as most projects don't have (or need) the strings to be translatable. This is however very easy to fix and just requires you to wrap user-faced strings in `_()`, e.g. `_("Invalid value!")` instead of `"Invalid value!"`


Lastly, while the approach you chose to handle showing the errors to the user is not wrong, I think it would make more sense to raise `commands.BadArgument` exception in the `__init__` method of `SetParser`. Exceptions are made to handle exceptional cases and invalid input is such a case. This keeps the logic of the command simpler and leaves the handling of the error to global error handlers that Red has.

So to show a full example of all these improvements implemented, here's a piece of code with the existing `RuntimeError` exceptions replaced with `commands.BadArgument` exceptions constructed with meaningful error messages:
```py
        if argument and argument[0] in allowed:
            if self.sum < 0:
                self.operation = "withdraw"
            elif self.sum > 0:
                self.operation = "deposit"
            else:
                raise commands.BadArgument(
                    _(
                        "Invalid value, the amount of currency to increase or decrease"
                        " must be an integer different from zero."
                    )
                )
            self.sum = abs(self.sum)
        elif argument.isdigit():
            self.operation = "set"
        else:
            raise commands.BadArgument(
                _(
                    "Invalid value, the argument must be an integer,"
                    " optionally preceded with a `+` or `-` sign."
                )
            )
```
With this code, there's no need to touch the logic of the command at all and everything will be handled by Red's global error handler.

Jackenmen (approved):
I've updated the PR with the changes requested in my review and this is now ready to be merged.

Thanks for the contribution, @rijusougata13!

PR #4807, opened 2021-02-12
Update `[p]economyset slottime&paydaytime` to use `TimedeltaConverter` and not accept negative integers

felineturtle67:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

Paydaytime and Slottime command seconds param now only takes a positive integer, but accepts 0
Paydaytime and Slottime seconds param has a converter so `1d` would be 86400 seconds, along with seconds, minutes, etc.

### DIFF
--- a/redbot/cogs/economy/economy.py
+++ b/redbot/cogs/economy/economy.py
@@ -12,6 +12,7 @@ from redbot.cogs.bank import is_owner_if_bank_global
 from redbot.cogs.mod.converters import RawUserIds
 from redbot.core import Config, bank, commands, errors, checks
 from redbot.core.commands import BadArgument, Converter
+from redbot.core.commands.converter import TimedeltaConverter
 from redbot.core.bot import Red
 from redbot.core.i18n import Translator, cog_i18n
 from redbot.core.utils import AsyncIter
@@ -115,27 +116,6 @@ class SetParser:
         else:
             raise RuntimeError
             
-class TimeConverter(Converter):
-    async def convert(self, ctx: commands.Context, time: str) -> int:
-        conversions = {"s": 1, "m": 60, "h": 3600, "d": 86400, "w": 604800, "mo": 604800*30 }
-        
-        if str(time[-1]) not in conversions:
-            if not str(time).isdigit():
-                raise BadArgument(f"{time} was not able to be converted to a time.")
-            return int(time) 
-        
-        multiplier = conversions[str(time[-1])]
-        
-        time = time[:-1]
-        if not str(time).isdigit():
-            raise BadArgument(f"{time} was not able to be converted to a time.")
-        
-        if int(time) * multiplier < 0:
-            raise BadArgument("This is not a positive integer")
-
-        return int(time) * multiplier
-
-
 @cog_i18n(_)
 class Economy(commands.Cog):
     """Get rich and have fun with imaginary currency!"""
@@ -925,7 +905,7 @@ class Economy(commands.Cog):
         )
 
     @economyset.command()
-    async def slottime(self, ctx: commands.Context, seconds: TimeConverter):
+    async def slottime(self, ctx: commands.Context, seconds: TimedeltaConverter):
         """Set the cooldown for the slot machine.
 
         Example:
@@ -935,6 +915,7 @@ class Economy(commands.Cog):
 
         - `<seconds>` The new number of seconds to wait in between uses of the slot machine. Default is 5.
         """
+        seconds = seconds.total_seconds()
         guild = ctx.guild
         if await bank.is_global():
             await self.config.SLOT_TIME.set(seconds)
@@ -943,7 +924,7 @@ class Economy(commands.Cog):
         await ctx.send(_("Cooldown is now {num} seconds.").format(num=seconds))
 
     @economyset.command()
-    async def paydaytime(self, ctx: commands.Context, seconds: TimeConverter):
+    async def paydaytime(self, ctx: commands.Context, seconds: TimedeltaConverter):
         """Set the cooldown for the payday command.
 
         Example:
@@ -953,6 +934,7 @@ class Economy(commands.Cog):
 
         - `<seconds>` The new number of seconds to wait in between uses of payday. Default is 300.
         """
+        seconds = seconds.total_seconds()
         guild = ctx.guild
         if await bank.is_global():
             await self.config.PAYDAY_TIME.set(seconds)
--- a/redbot/cogs/economy/economy.py
+++ b/redbot/cogs/economy/economy.py
@@ -11,7 +11,6 @@ import discord
 from redbot.cogs.bank import is_owner_if_bank_global
 from redbot.cogs.mod.converters import RawUserIds
 from redbot.core import Config, bank, commands, errors, checks
-from redbot.core.commands import BadArgument, Converter
 from redbot.core.commands.converter import TimedeltaConverter
 from redbot.core.bot import Red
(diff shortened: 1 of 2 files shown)


flaree:
There's already a timedelta converter you can make use of rather than making a new one

felineturtle67:
but that returns something like 0:00:50 right

flaree:
Take a look at how slow mode if I remember takes use of it. You can convert a timedelta to seconds

felineturtle67:
oh its a timedelta object? alr

redbot/cogs/economy/economy.py
```diff
@@ -11,6 +11,8 @@
 from redbot.cogs.bank import is_owner_if_bank_global
 from redbot.cogs.mod.converters import RawUserIds
 from redbot.core import Config, bank, commands, errors, checks
+from redbot.core.commands import BadArgument, Converter
```
Jackenmen: Now that you removed `TimeConverter`, these imports are no longer used and can be removed.

redbot/cogs/economy/economy.py
```diff
@@ -113,8 +115,7 @@ def __init__(self, argument):
             self.operation = "set"
         else:
             raise RuntimeError
-
-
```
Jackenmen: You removed a line in here and added whitespace to the other line which makes this PR fail the style check. You need to ensure that there are two empty lines between the class definitions. Or you can run Black to reformat this for you as explained in [our contributing guidelines](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/CONTRIBUTING.md)

redbot/cogs/economy/economy.py
```diff
@@ -914,6 +915,7 @@ async def slottime(self, ctx: commands.Context, seconds: int):
 
         - `<seconds>` The new number of seconds to wait in between uses of the slot machine. Default is 5.
```
Jackenmen: The example above and the description of the `<seconds>` parameter here should be updated to reflect the type of input that the used converter accepts.

redbot/cogs/economy/economy.py
```diff
@@ -932,6 +934,7 @@ async def paydaytime(self, ctx: commands.Context, seconds: int):
 
         - `<seconds>` The new number of seconds to wait in between uses of payday. Default is 300.
```
Jackenmen: The example above and the description of the `<seconds>` parameter here should be updated to reflect the type of input that the used converter accepts.

Jackenmen (changes requested):
Thanks for making the fix, I have a few review comments that will need to be addressed before this can be approved and merged.

redbot/cogs/economy/economy.py
felineturtle67: finished

redbot/cogs/economy/economy.py
```diff
@@ -113,8 +114,9 @@ def __init__(self, argument):
             self.operation = "set"
         else:
             raise RuntimeError
-
-
+            
+            
+
```
Jackenmen: You're still failing style here. There should be exactly 2 lines between class definitions and there can be no characters in the line, and currently, you have trailing spaces in here.

redbot/cogs/economy/economy.py
```diff
@@ -904,16 +906,19 @@ async def slotmax(self, ctx: commands.Context, bid: positive_int):
         )
 
     @economyset.command()
-    async def slottime(self, ctx: commands.Context, seconds: int):
+    async def slottime(self, ctx: commands.Context, seconds: TimedeltaConverter):
```
Jackenmen: Now that this accepts numbers with units, it would make more sense if this argument were named `duration` (or `time`, `cooldown`, whatever you think makes sense here).

Remember to also update the docstring with the new name and update this code:
```py
seconds = seconds.total_seconds()
```
to be this instead (or whatever name you choose for argument):
```py
seconds = duration.total_seconds()
```

redbot/cogs/economy/economy.py
```diff
@@ -922,16 +927,19 @@ async def slottime(self, ctx: commands.Context, seconds: int):
         await ctx.send(_("Cooldown is now {num} seconds.").format(num=seconds))
 
     @economyset.command()
-    async def paydaytime(self, ctx: commands.Context, seconds: int):
+    async def paydaytime(self, ctx: commands.Context, seconds: TimedeltaConverter):
```
Jackenmen: Now that this accepts numbers with units, it would make more sense if this argument were named `duration` (or `time`, `cooldown`, whatever you think makes sense here).

Remember to also update the docstring with the new name and update this code:
```py
seconds = seconds.total_seconds()
```
to be this instead (or whatever name you choose for argument):
```py
seconds = duration.total_seconds()
```

redbot/cogs/economy/economy.py
Jackenmen: It probably makes sense to set default unit to seconds here:
```suggestion
    async def slottime(
        self, ctx: commands.Context, seconds: TimedeltaConverter(default_unit="seconds")
    ):
```

redbot/cogs/economy/economy.py
Jackenmen: It probably makes sense to set default unit to seconds here:
```suggestion
    async def paydaytime(
        self, ctx: commands.Context, seconds: TimedeltaConverter(default_unit="seconds")
    ):
```

redbot/cogs/economy/economy.py
```diff
@@ -922,16 +927,19 @@ async def slottime(self, ctx: commands.Context, seconds: int):
...
         """Set the cooldown for the payday command.
 
         Example:
             - `[p]economyset paydaytime 86400`
+            - `[p]economyset paydaytime 1d`
 
         **Arguments**
 
         - `<seconds>` The new number of seconds to wait in between uses of payday. Default is 300.
+        - It is also a timedeltaconverter so `1d` is converted to 1 day
```
Jackenmen: This isn't really clear to the user, I would suggest sth like this:
```suggestion
        - `<seconds>` The new duration to wait in between uses of payday. Default is 5 minutes.
        Accepts: seconds, minutes, hours, days, weeks
```

redbot/cogs/economy/economy.py
```diff
@@ -904,16 +906,19 @@ async def slotmax(self, ctx: commands.Context, bid: positive_int):
...
         """Set the cooldown for the slot machine.
 
         Example:
             - `[p]economyset slottime 10`
+            - `[p]economyset slottime 10m`
 
         **Arguments**
 
         - `<seconds>` The new number of seconds to wait in between uses of the slot machine. Default is 5.
+        - It is also a timedeltaconverter so `1d` is converted to 1 day
```
Jackenmen: This isn't really clear to the user, I would suggest sth like this:
```suggestion
        - `<seconds>` The new duration to wait in between uses of the slot machine. Default is 5 seconds.
        Accepts: seconds, minutes, hours, days, weeks
```

Jackenmen (changes requested):
There are few more things that need to be changed before merging.

After you deal with the review comments for the docstrings, you should also update the `docs/cog_guides/economy.rst` file with the updated contents of docstrings.

Jackenmen:
I'm assuming you'll be working on it in full later (or are working on it now), can you send a comment in here once you finish, @Andeeeee? It would make it easier for me to know when this is ready to be reviewed again :)

felineturtle67:
k done

Jackenmen (approved):
Looks good to me, thanks for contributing.

PR #4820, opened 2021-02-14
[Warnings/Modlog] Stop changing reason removing custom info in Warnings cog 

### LINKED ISSUE #4775: [Warnings] Unwarn command gets removed upon updating case
# Other bugs

Updating a warning case when the unwarn command usage in the case is removed. Also removing the warn ID which can only be obtained with `[p]warnings <user>`

Unwarn command stays after updating

#### What actually happened?

The unwarn command usage gets removed

1. Warn a user
2. Update the case
3. Observe

Vexed01:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Add a way for custom permanent information to be added to modlog cases. This enhancement allows for the bugfix of warnings removing some information on the case (points and unwarn command). Closes #4775


Vexed01:
I HATE GIT

Jackenmen:
Quite cool - allows us to split the information provided by the bot and the information provided by the user.

I wonder if there are any other cases in core/core cogs where this could be used instead/in addition to reason.
For example, would it make sense to replace the usage of `get_audit_reason()` with a split between reason and extra_info?

PR #4660, opened 2020-12-06
Replace Selfrole add/remove with a toggle

TheWyn:
Merged Selfrole add/remove commands into a single toggle.

### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

### Description of the changes

Merged Selfrole add/remove commands into a single toggle.

PR #4826, opened 2021-02-15
Make `[p]selfrole` without subcommand "toggle" the selfrole

TheWyn:
Replaces #4660 

Merged Selfrole add/remove commands into a single toggle.

### Type

- [ ] Bugfix
- [X] Enhancement
- [x] New feature

Merged Selfrole add/remove commands into a single toggle.


Kreusada (changes requested):
In the selfrole header, the quotation marks need to be inline with the 'selfrole' text in order to pass the docs check

(Sorry if this review looked weird, i'm on mobile)

Jackenmen (approved):
Looks good to me, thanks for contributing and the fast responses yesterday :)

PR #4828, opened 2021-02-16
make blacklist/whitelist verify its an actual user

felineturtle67:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

for blacklist add and remove allow discord.User typehints as well as fetching the user (blocklist add) and verifying the user exists when adding to the blocklist.


Drapersniper:
People should be able to disallow anyone from using the bot even if they no longer share a server with it, your change makes that no longer possible therefore it is a bad idea.

PR #4830, opened 2021-02-16
Fix the tracking of nicks that were set just before nick reset

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This PR fixes tracking of nicknames that were set just before nick reset (i.e. XYZ not getting saved when the user has nickname XYZ and then the nickname is reset rather than changed to something else) and stops adding `None`s to the list when the nickname wasn't set *before* the nickname change.

This issue was introduced by #4131.


Kowlin (approved):
Looks good

PR #4833, opened 2021-02-17
Change things up with Crowdin and release-related workflows

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

The new Crowdin workflows works as follows:
- New strings get uploaded to Crowdin on each push so that they're always up-to-date there
- Translations get downloaded and updated in the repository during the `prepare_release` workflow that should be manually dispatched by the release manager before the release

The new `prepare_release` workflow runs two parallel jobs:
- Translations download from Crowdin
- PR creation of stable bump (the version that workflow should bump to can be passed when dispatching the workflow BUT it should not be needed most of the time as by default the job will strip `.dev1` from our version which is generally how our stable version bump looks like)

The (previously named `publish_pypi`) `publish_release` workflow runs two jobs sequentially:
1. Build the wheels and publish to PyPI.
2. PR creation of dev bump which bumps micro version by 1 and adds the `.dev1` suffix (happens once PyPI release is done).
    - Note: this initially was done in the `prepare_release` workflow, but it is prone to merge conflicts and if we can automate it anyway, there's no benefit to making the PR with the bump *before* actually releasing first.

Improvements:
- All automatically created PRs get automatically closed and reopened by @Cog-CreatorsBot to trigger the CI checks

The workflows related to bumps have been tested, and Crowdin pretty much just reuses that code, so it *most likely* works too.


Jackenmen:
@Kowlin I've ended up working on the release-related workflows after all. This should be in a reviewable state now.

Kowlin (approved):
Looks about right! 👍

PR #4838, opened 2021-02-20
get_or_fetch_user: return None if user is not found

madebylydia:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Just to avoid the NotFound error, I think this is better, but feel free to discuss it in case it should be to devs to handle it.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -9,7 +9,7 @@ import contextlib
 import weakref
 import functools
 from collections import namedtuple
-from datetime import datetime_fetch
+from datetime import datetime
 from enum import IntEnum
 from importlib.machinery import ModuleSpec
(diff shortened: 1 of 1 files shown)


Twentysix26:
I have edited this PR's description / title since "fix" implied that the feature wasn't working as expected. The docstring and return type should be changed accordingly.

Drapersniper:
Sorry i missed this PR.

So the reason I have left is a a raisable util is that `fetch_user()` does raise if you were to call directly.

I do believe mirrying that behaviour allows this to be a better drop in preplacement for `fetch_user` i see why you done this PR, as its far more covinient, that being said it changes the return type which is no longer  shadowing the signature of `fetch_user`.

PR #4839, opened 2021-02-20
Add user/role names to `[p](local)allow/blocklist list`

Kreusada:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Feature Request: Allow for the bot to get the user in each allowlist/blocklist in the format:

```yaml
- user#1234 (1234567890)
```

I feel this feature would help these lists to be more expansive in the information they provide, because sometimes, especially on a large redbot, there will be lots of users on the blocklist/allowlist, and to sift through the IDs may not be so convenient, but also I can't think of a reason against why this couldn't be implemented.

If a user could not be found, I check against a variable getting the user and change the variable content to `Unknown or Deleted User` if that were to be the case.

I've tested these changes on my fork of Red and they appear to work just fine. I attempted to provide an image below. Please let me know if there are any changes required.

Example Image

~~Edit: 'Unknown or Deleted User' has been edited to have squares brackets around it inside the string.~~

Further edits:

A few more additional edits were made to the original UI to make it look slightly neater, seeing as usernames can vary drastically in length, but IDs cannot. The layout has been changed to look something like this:

```yaml
- 123456789012345 (user#1234)
- 543210987654321 (Unknown or Deleted User)
```

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -2977,7 +2977,7 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         for obj in curr_list:
             get_user_or_role = self.bot.get_user(obj) or ctx.guild.get_role(obj)
             if not get_user_or_role:
-                get_user_or_role = _("Unknown or Deleted User")
+                get_user_or_role = _("Unknown or Deleted User/Role")
             msg += "\n\t- {} ({})".format(obj, get_user_or_role)
 
         for page in pagify(msg):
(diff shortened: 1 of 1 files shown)


Drapersniper:
Rather than editing the translation string why don't you add the [] to the format line  that way the string won't need to be retranslated

Kreusada:
The latest commit on this PR changed the formatting slightly (thanks siu for the suggestion) IDs should come first given that they're pretty much the same length in comparison to usernames which could range in length and make the blocklist/whitelist look messy. This also removes the `[]` from when there is an Unknown or Deleted User to match the improved format.

Image

Jackenmen (commented):
Very cool change, it seems to work as intended.

@Kreusada would it be possible to make this show the role names as well? That would be awesome :)

Kreusada:
Yes of course, the changes have been made in the latest two commit

Jackenmen (approved):
Looks good to me, thanks for the changes.

PR #4844, opened 2021-02-23
Work around `actions/labeler`'s upstream issue with label sync setting

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Works around https://github.com/actions/labeler/issues/104 which causes `sync-labels: false` (which is the default) to not work as actual `false`.


Jackenmen:
Admin-merged as this issue causes more work for triagers and this does not affect any source code.

PR #4848, opened 2021-02-24
fix filter error

ltzmax:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

this fixes an `on_message` error that had no attribute to `id`.


ltzmax:
Well this was the wrong way to do it on.

Issue #4847, opened 2021-02-24
[Mutes] Mute error shows incorrect text


Dav-Git:
When muting a user fails, the message for unmuting is shown instead.

![image](https://user-images.githubusercontent.com/57032623/108998410-ca88c680-7698-11eb-9f2c-f641376b6d37.jpeg)

Jackenmen:
#4850 was meant to fix this, but looking at the full message here, it appears that the "That user cannot be muted" reason still needs fixing so I'm keeping this open.

PR #4850, opened 2021-02-26
[Mutes] Wrong error message when muting fails

elijabesu:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Just a minor text change in `parse_issues`, it should fix issue #4847


Kowlin (approved):
Thanks, made some minor alteration due the fact that this is used for both muting and unmuting.

PR #4865, opened 2021-02-28
Use `brew install --cask` instead of `brew cask install`

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

`brew cask install` has been removed in Brew 3.0.0: https://github.com/Homebrew/brew/pull/10397

You can see the error when running our current install instructions here:
https://github.com/jack1142/Red-Install-Tests/runs/1998480687?check_suite_focus=true

@aikaterna because this affects Mac, I would appreciate it if you could verify this works


aikaterna (approved):
Any user with an old Homebrew install will be forced into using the updated syntax as the program automatically updates when any Homebrew command is run, so that's good.

PR #4871, opened 2021-03-03
[Alias] Strip extra whitespace from fake command messages

### LINKED ISSUE #4766: Aliases for some commands which have optional arguments fail to operate normally
#### Command name

Aliases for _some_ commands which require optional arguments.

Alias, but also the cog from which the command of the alias is from. Currently, it seems to be the case especially with `Downloader`.

For the aliases for `pipinstall` and `load` to return the help menu, as no arguments were given.

#### What actually happened?

For `load`, the redbot would start typing but there would be no result.
For `pipinstall`, the redbot would start typing and then send "Libraries installed", although no arguments were given.

These are two examples I've discovered, but there may be more.

Create an alias for `pipinstall`, and/or `load`.
Use these aliases without any arguments.

Flame442:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

Discord messages are unable to have trailing whitespace when created via the client. Alias previously would leave trailing whitespace in fake command messages when the alias had no arguments passed to it. This whitespace broke certain commands that take an arbitrary, but non-zero, number of arguments, such as `[p]load`. This PR removes that trailing space.

Closes #4766

### DIFF
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -185,7 +185,7 @@ class Alias(commands.Cog):
         # noinspection PyDunderSlots
         new_message.content = "{}{} {}".format(
             prefix, command, " ".join(args[trackform.max + 1 :])
-        )
+        ).strip()
         await self.bot.process_commands(new_message)
 
     async def paginate_alias_list(


Jackenmen (approved):
I can't think of any *new* issues this could cause and it fixes what it says it fixes, so looks good to me.

PR #4728, opened 2021-01-13
Update d.py to 1.6.0, aiohttp to 3.7.3, Red-Lavalink to 0.7.2

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature
- [x] Dependency update

### Description of the changes
- Update d.py to 1.6.0, aiohttp to 3.7.3, Red-Lavalink to 0.7.2 (the last one because of the dependency conflict with aiohttp 3.7+)
- Remove suppression of deprecation warnings that are no longer relevant with the update of aiohttp
- Change things that require changing after bumping d.py 1.6
    - in this update, it's only some max concurrency things in `Command.prepare()`
- Remove two random `AttributeError` exception catching that doesn't make sense at least since d.py 1.1...

After this PR gets merged, #3529 should be updated appropriately.

PR #4876, opened 2021-03-06
Fix `AttributeError` in `Context.delete_messages()`

kingslayer268:
### Type

- [x] Bugfix

delete_messages() isn't an attribute of DMChannels which causes an error when sending interactives in DM channels

Regression was introduced in #4728


Jackenmen (approved):
Looks good to me, thanks so much for contributing a fix!

PR #4883, opened 2021-03-13
[Streams] Handle Twitch ratelimits

### LINKED ISSUE #4808: Handling of Twitch ratelimits to stream alerts

Streams

Currently streams cog can be ratelimited by Twitch API and it ignores it.
It needs some handling of 429s, to avoid continue sending requests to the API when the host is currently ratelimited.

PredaaA:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

This PR handle Twitch ratelimits, which closes #4808. Credits goes to @TrustyJAID [for the ratelimits handling logic](https://github.com/TrustyJAID/Trusty-cogs/blob/master/twitch/twitch_api.py).
Possible API or client errors are now handled too, but they won't be retried, they return a `None` and an empty dict, which finally will raise `APIError`.

### DIFF
--- a/redbot/cogs/streams/streamtypes.py
+++ b/redbot/cogs/streams/streamtypes.py
@@ -293,9 +293,9 @@ class TwitchStream(Stream):
         https://github.com/tsifrer/python-twitch-client/blob/master/twitch/helix/base.py
         https://github.com/TrustyJAID/Trusty-cogs/blob/master/twitch/twitch_api.py
         """
+        current_time = int(time.time())
+        self._rate_limit_resets = {x for x in self._rate_limit_resets if x > current_time}
         if self._rate_limit_remaining == 0:
-            current_time = int(time.time())
-            self._rate_limit_resets = set(x for x in self._rate_limit_resets if x > current_time)
 
             if self._rate_limit_resets:
                 reset_time = next(iter(self._rate_limit_resets))
--- a/redbot/cogs/streams/streamtypes.py
+++ b/redbot/cogs/streams/streamtypes.py
@@ -298,22 +298,20 @@ class TwitchStream(Stream):
             self._rate_limit_resets = set(x for x in self._rate_limit_resets if x > current_time)
 
             if len(self._rate_limit_resets) > 0:
-                reset_time = list(self._rate_limit_resets)[0]
+                reset_time = next(self._rate_limit_resets)
                 # Calculate wait time and add 0.1s to the wait time to allow Twitch to reset
                 # their counter
                 wait_time = reset_time - current_time + 0.1
                 await asyncio.sleep(wait_time)
 
-    async def get_data(self, url: str, params: dict = {}) -> Tuple[int, dict]:
+    async def get_data(self, url: str, params: dict = {}) -> Tuple[Optional[int], dict]:
         header = {"Client-ID": str(self._client_id)}
         if self._bearer is not None:
             header["Authorization"] = f"Bearer {self._bearer}"
         await self.wait_for_rate_limit_reset()
         async with aiohttp.ClientSession() as session:
             try:
-                async with session.get(
-                    url, headers=header, params=params, timeout=aiohttp.ClientTimeout(total=None)
-                ) as resp:
+                async with session.get(url, headers=header, params=params, timeout=60) as resp:
                     remaining = resp.headers.get("Ratelimit-Remaining")
                     if remaining:
                         self._rate_limit_remaining = int(remaining)
@@ -325,13 +323,14 @@ class TwitchStream(Stream):
                         log.info(
                             "Ratelimited. Trying again at %s.", datetime.fromtimestamp(int(reset))
                         )
+                        resp.release()
                         return await self.get_data(url)
 
                     if resp.status != 200:
-                        return None, {}
+                        return resp.status, {}
 
                     return resp.status, await resp.json(encoding="utf-8")
-            except aiohttp.ClientConnectionError:
+            except (aiohttp.ClientConnectionError, asyncio.TimeoutError):
                 return None, {}
 
     async def is_online(self):
--- a/redbot/cogs/streams/streamtypes.py
+++ b/redbot/cogs/streams/streamtypes.py
@@ -1,6 +1,8 @@
+import asyncio
 import contextlib
 import json
 import logging
+import time
 from dateutil.parser import parse as parse_time
 from random import choice
 from string import ascii_letters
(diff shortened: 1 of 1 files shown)


TrustyJAID (commented):
I want to leave a comment here stating that this change has been working well for me for a couple weeks now. I can't approve it though since most of these changes were in part written by myself.

redbot/cogs/streams/streamtypes.py
```diff
@@ -280,92 +282,101 @@ def __init__(self, **kwargs):
...
-            if user_profile_data:
-                profile_image_url = user_profile_data["data"][0]["profile_image_url"]
-                data["profile_image_url"] = profile_image_url
-                data["view_count"] = user_profile_data["data"][0]["view_count"]
-                data["login"] = user_profile_data["data"][0]["login"]
-
-            is_rerun = False
-            return self.make_embed(data), is_rerun
-        elif r.status == 400:
+                    url, headers=header, params=params, timeout=aiohttp.ClientTimeout(total=None)
```
Jackenmen: Why no timeout?

redbot/cogs/streams/streamtypes.py
```diff
@@ -280,92 +282,101 @@ def __init__(self, **kwargs):
...
+                        self._rate_limit_remaining = int(remaining)
+                    reset = resp.headers.get("Ratelimit-Reset")
+                    if reset:
+                        self._rate_limit_resets.add(int(reset))
+
+                    if resp.status == 429:
+                        log.info(
+                            "Ratelimited. Trying again at %s.", datetime.fromtimestamp(int(reset))
+                        )
+                        return await self.get_data(url)
```
Jackenmen, line 327: Probably wouldn't hurt to release the response object since we're doing all of this inside the context manager and therefore blocking the underlying connection from returning back to the pool.
```suggestion
                        resp.release()
                        return await self.get_data(url)
```

redbot/cogs/streams/streamtypes.py
```diff
@@ -280,22 +282,67 @@ def __init__(self, **kwargs):
...
+        From python-twitch-client and adaptated to asyncio from Trusty-cogs:
+        https://github.com/tsifrer/python-twitch-client/blob/master/twitch/helix/base.py
+        https://github.com/TrustyJAID/Trusty-cogs/blob/master/twitch/twitch_api.py
+        """
+        if self._rate_limit_remaining == 0:
+            current_time = int(time.time())
+            self._rate_limit_resets = set(x for x in self._rate_limit_resets if x > current_time)
+
+            if len(self._rate_limit_resets) > 0:
+                reset_time = list(self._rate_limit_resets)[0]
```
Jackenmen:
```suggestion
                reset_time = next(self._rate_limit_resets)
```

redbot/cogs/streams/streamtypes.py
```diff
@@ -280,22 +282,67 @@ def __init__(self, **kwargs):
...
+                        self._rate_limit_resets.add(int(reset))
+
+                    if resp.status == 429:
+                        log.info(
+                            "Ratelimited. Trying again at %s.", datetime.fromtimestamp(int(reset))
+                        )
+                        return await self.get_data(url)
+
+                    if resp.status != 200:
+                        return None, {}
```
Jackenmen: We still need to return status code here so that the handling of error codes works.
```suggestion
                        return resp.status, {}
```

redbot/cogs/streams/streamtypes.py
```diff
@@ -280,22 +282,67 @@ def __init__(self, **kwargs):
...
+            self._rate_limit_resets = set(x for x in self._rate_limit_resets if x > current_time)
+
+            if len(self._rate_limit_resets) > 0:
+                reset_time = list(self._rate_limit_resets)[0]
+                # Calculate wait time and add 0.1s to the wait time to allow Twitch to reset
+                # their counter
+                wait_time = reset_time - current_time + 0.1
+                await asyncio.sleep(wait_time)
+
+    async def get_data(self, url: str, params: dict = {}) -> Tuple[int, dict]:
```
Jackenmen: The status code might be `None` when `ClientConnectionError` is raised.
```suggestion
    async def get_data(self, url: str, params: dict = {}) -> Tuple[Optional[int], dict]:
```

Jackenmen (changes requested):
Thanks for adding this and splitting the PR in multiple :) I  have a few review comments that should be addressed before merging.

redbot/cogs/streams/streamtypes.py
```diff
@@ -280,22 +282,66 @@ def __init__(self, **kwargs):
...
+                            "Ratelimited. Trying again at %s.", datetime.fromtimestamp(int(reset))
+                        )
+                        resp.release()
+                        return await self.get_data(url)
+
+                    if resp.status != 200:
+                        return resp.status, {}
+
+                    return resp.status, await resp.json(encoding="utf-8")
+            except (aiohttp.ClientConnectionError, asyncio.TimeoutError):
```
Jackenmen: It would probably be good to log these.
```suggestion
            except (aiohttp.ClientConnectionError, asyncio.TimeoutError) as exc:
                log.warning("Connection error occurred when fetching Twitch stream", exc_info=exc)
```

redbot/cogs/streams/streamtypes.py
```diff
@@ -280,22 +282,67 @@ def __init__(self, **kwargs):
...
+
+        From python-twitch-client and adaptated to asyncio from Trusty-cogs:
+        https://github.com/tsifrer/python-twitch-client/blob/master/twitch/helix/base.py
+        https://github.com/TrustyJAID/Trusty-cogs/blob/master/twitch/twitch_api.py
+        """
+        if self._rate_limit_remaining == 0:
+            current_time = int(time.time())
+            self._rate_limit_resets = set(x for x in self._rate_limit_resets if x > current_time)
+
+            if len(self._rate_limit_resets) > 0:
```
Jackenmen: Simplification
```suggestion
            if self._rate_limit_resets:
```

redbot/cogs/streams/streamtypes.py
```diff
@@ -280,22 +282,67 @@ def __init__(self, **kwargs):
...
+        From python-twitch-client and adaptated to asyncio from Trusty-cogs:
+        https://github.com/tsifrer/python-twitch-client/blob/master/twitch/helix/base.py
+        https://github.com/TrustyJAID/Trusty-cogs/blob/master/twitch/twitch_api.py
+        """
+        if self._rate_limit_remaining == 0:
+            current_time = int(time.time())
+            self._rate_limit_resets = set(x for x in self._rate_limit_resets if x > current_time)
+
+            if self._rate_limit_resets:
+                reset_time = next(self._rate_limit_resets)
```
Jackenmen: turns out this needs to be wrapped in `iter()` :)
```suggestion
                reset_time = next(iter(self._rate_limit_resets))
```

redbot/cogs/streams/streamtypes.py
```diff
@@ -280,22 +282,67 @@ def __init__(self, **kwargs):
...
+    async def wait_for_rate_limit_reset(self) -> None:
+        """Check rate limits in response header and ensure we're following them.
+
+        From python-twitch-client and adaptated to asyncio from Trusty-cogs:
+        https://github.com/tsifrer/python-twitch-client/blob/master/twitch/helix/base.py
+        https://github.com/TrustyJAID/Trusty-cogs/blob/master/twitch/twitch_api.py
+        """
+        if self._rate_limit_remaining == 0:
+            current_time = int(time.time())
+            self._rate_limit_resets = set(x for x in self._rate_limit_resets if x > current_time)
```
Jackenmen: We need to regularly clean the `self._rate_limit_resets` set as otherwise it can grow indefinitely.
This would be one way of doing it:
```suggestion
        current_time = int(time.time())
        self._rate_limit_resets = {x for x in self._rate_limit_resets if x > current_time}
        if self._rate_limit_remaining == 0:
```

Jackenmen (changes requested):

Jackenmen (approved):
Looks good to me now, thanks for PRing!

PR #4897, opened 2021-03-18
Update ambiguous oid parsing to detect candidates on Git 2.31+

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

On Git 2.31, the returned error with candidate hints has been changed (https://github.com/git/git/commit/1ffa26c4614720f8bb6c04703c7b65b34bd93fe3) and this PR updates Downloader to be able to parse the updated message.

To be clear here, this is not something that can raise an error on Git 2.31+ without this PR as this logic was written so that it just doesn't show candidates if it can't parse the message from Git.

### DIFF
--- a/tests/cogs/downloader/test_git.py
+++ b/tests/cogs/downloader/test_git.py
@@ -5,6 +5,7 @@ import pytest
 
 from redbot.cogs.downloader.repo_manager import ProcessFormatter, Repo
 from redbot.pytest.downloader import (
+    GIT_VERSION,
     cloned_git_repo,
     git_repo,
     git_repo_with_remote,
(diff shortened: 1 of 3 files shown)


tests/cogs/downloader/test_git.py
```diff
@@ -326,16 +330,65 @@ async def test_git_get_full_sha1_from_ambiguous_commits(git_repo):
...
+        "hint:   c6f0e5e commit 2019-10-24 - Commit ambiguous with tag.\n"
+        "fatal: Needed a single revision"
+    )
+
+
+@pytest.mark.skipif(
+    GIT_VERSION >= (2, 31), reason="This is test for output from Git older than 2.31."
+)
+@pytest.mark.asyncio
+async def test_git_get_full_sha1_from_ambiguous_commits(git_repo):
```
Xiretza: Doesn't this need a different name?

Jackenmen: Yes, you're right. I just updated name of this test, thanks!

PR #4898, opened 2021-03-18
[Core] Improve responses to `[p]set (server)prefix` commands

Kreusada:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

* Prefix and serverprefix commands to alter their responses based on the number of arguments (e.g. "Prefix set", "Prefixes set").
* Resetting server prefixes to use `server` instead of `guild` in the response's wording, to unify with other aspects of the command such as the docstring and the command's name, and alias.
* Setting server prefixes to indicate that they are server prefixes in the response. Previously, both prefix commands responded with "Prefix set.".

Side note: This has been tested


Jackenmen (approved):
Looks good to me, thanks!

PR #4904, opened 2021-03-20
[Warnings] fix raw namedtuple output

### LINKED ISSUE #4900: [Warnings] Raw namedtuple output in [p]warnings

3.4.7

### Cog name

Warnings

### Command name

warnings

Have the member name properly show

### What actually happened?

Because `user` is treated as a member object, if it is a named tuple (when a user id is used for a member no longer in the server) it prints out that named tuple instead of the member name.

`Warnings for Member(id=00000000000, guild=<Guild id=00000000 name='Server' shard_id=0 chunked=True member_count=0>)`

https://github.com/Cog-Creators/Red-DiscordBot/blob/d26335b01b892e346e52cfd3bfcb86706647a76c/redbot/cogs/warnings/warnings.py#L529
https://github.com/Cog-Creators/Red-DiscordBot/blob/d26335b01b892e346e52cfd3bfcb86706647a76c/redbot/cogs/warnings/warnings.py#L555

1. Warn a user
2. Have that user leave the server
3. Run `[p]warnings <uid>` on that user

Obi-Wan3:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Updated the `[p]warnings` command to display the user's ID instead of the raw namedtuple if the member has left the server.

resolves #4900

### DIFF
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -552,7 +552,9 @@ class Warnings(commands.Cog):
                     )
                 await ctx.send_interactive(
                     pagify(msg, shorten_by=58),
-                    box_lang=_("Warnings for {user}").format(user=user),
+                    box_lang=_("Warnings for {user}").format(
+                        user=user if isinstance(user, discord.Member) else user.id
+                    ),
                 )
 
     @commands.command()


Jackenmen (approved):

PR #4906, opened 2021-03-21
[Mutes] Add guild ID to mutes data

PredaaA:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

The reason of this change is because the `bot.get_channel` in `_handle_channel_unmutes` was too much intensive on bot runtime, even more since this runs every 30s.

It includes a converter for the data that runs on `initialize`, once bot fired red_ready.

### DIFF
--- a/redbot/cogs/mutes/mutes.py
+++ b/redbot/cogs/mutes/mutes.py
@@ -81,7 +81,7 @@ class Mutes(VoiceMutes, commands.Cog, metaclass=CompositeMetaClass):
             "dm": False,
             "show_mod": False,
         }
-        self.config.register_global(force_role_mutes=True, version=0)
+        self.config.register_global(force_role_mutes=True, schema_version=0)
         # Tbh I would rather force everyone to use role mutes.
         # I also honestly think everyone would agree they're the
         # way to go. If for whatever reason someone wants to
--- a/redbot/cogs/mutes/mutes.py
+++ b/redbot/cogs/mutes/mutes.py
@@ -12,7 +12,7 @@ from .voicemutes import VoiceMutes
 
 from redbot.core.bot import Red
 from redbot.core import commands, checks, i18n, modlog, Config
-from redbot.core.utils import bounded_gather
+from redbot.core.utils import AsyncIter, bounded_gather
 from redbot.core.utils.chat_formatting import bold, humanize_timedelta, humanize_list, pagify
 from redbot.core.utils.mod import get_audit_reason
 from redbot.core.utils.menus import start_adding_reactions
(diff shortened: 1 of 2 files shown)


Kowlin (dismissed):
Code looks good in theory, haven't tested it.

Jackenmen (dismissed):
Just to block Kowlin's review (because I don't have plans for the full review, at least not right now), this won't act properly on bot startup (or at least not efficiently) as `initialize()` is awaited in `setup()` which happens before the bot is ready and its internal cache is filled.
And I've also mentioned to Preda it might be better to use integer for config schema version for the ease of use in future, although that's not as important.

PredaaA:
> Just to block Kowlin's review (because I don't have plans for the full review, at least not right now), this won't act properly on bot startup (or at least not efficiently) as `initialize()` is awaited in `setup()` which happens before the bot is ready and its internal cache is filled.
> And I've also mentioned to Preda it might be better to use integer for config schema version for the ease of use in future, although that's not as important.

PR got updated to match requested changes.
And also at the same time I can confirm that it fixes performance issues on bot.

redbot/cogs/mutes/mutes.py
```diff
@@ -139,10 +144,38 @@ async def initialize(self):
...
+                        mute_data["guild"] = channel.guild.id
+                        async with self.config.channel_from_id(
+                            channel_id
+                        ).muted_users() as muted_users:
+                            muted_users[str(user_id)] = mute_data
+                except (discord.NotFound, discord.Forbidden):
+                    await self.config.channel_from_id(channel_id).clear()
+
+            schema_version += 1
+            await self.config.version.set(schema_version)
```
TrustyJAID:
```suggestion
            await self.config.schema_version.set(schema_version)
```

redbot/cogs/mutes/mutes.py
```diff
@@ -139,10 +144,38 @@ async def initialize(self):
...
+
+        if schema_version == 0:
+            start = datetime.now()
+            log.info("Config conversion to schema_version 1 started.")
+            all_channels = await self.config.all_channels()
+            async for channel_id, channel_data in AsyncIter(all_channels.items()):
+                try:
+                    if (channel := self.bot.get_channel(channel_id)) is None:
+                        channel = await self.bot.fetch_channel(channel_id)
+                    for user_id, mute_data in channel_data["muted_users"].items():
```
TrustyJAID:
```suggestion
                    async with self.config.channel_from_id(channel_id) as muted_users:
                        for user_id, mute_data in muted_users.items():
                            mute_data["guild"] = channel.guild.id
                            
```

TrustyJAID (changes requested):
One issue needs to be fixed and another slightly better way to handle the conversion imo.

PredaaA:
Updated the PR with the requested changes.

TrustyJAID (approved):
This looks good to me. Thanks for making this addition!

PR #4907, opened 2021-03-21
[Mod] Only loop through guilds in config when checking tempbans

PredaaA:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Looping through all bot guilds was taking on bot performances way too much as it was almost directly looking to config.

### DIFF
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -243,36 +243,41 @@ class KickBanMixin(MixinMeta):
 
                     guild_tempbans = guild_data["current_tempbans"]
                     if guild_tempbans:
-                        for uid in guild_tempbans:
-                            unban_time = datetime.fromtimestamp(
-                                await self.config.member_from_ids(guild.id, uid).banned_until(),
-                                timezone.utc,
-                            )
-                            if datetime.now(timezone.utc) > unban_time:  # Time to unban the user
-                                try:
-                                    await guild.unban(
-                                        discord.Object(id=uid), reason=_("Tempban finished")
-                                    )
-                                except discord.NotFound:
-                                    # user is not banned anymore
-                                    guild_tempbans.remove(uid)
-                                except discord.HTTPException as e:
-                                    # 50013: Missing permissions error code or 403: Forbidden status
-                                    if e.code == 50013 or e.status == 403:
-                                        log.info(
-                                            f"Failed to unban ({uid}) user from "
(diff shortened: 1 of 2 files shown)


redbot/cogs/mod/kickban.py
```diff
@@ -230,41 +230,49 @@ async def ban_user(
...
+                    if not (guild := self.bot.get_guild(guild_id)):
+                        continue
+                    if not guild.me.guild_permissions.ban_members:
+                        continue
+                    if await self.bot.cog_disabled_in_guild(self, guild):
+                        continue
+
+                    guild_tempbans = guild_data["current_tempbans"]
+                    if guild_tempbans:
+                        for uid in guild_tempbans:
```
Jackenmen: Copying is needed so that `guild_tempbans.remove(uid)` works in the loop:
```suggestion
                        for uid in guild_tempbans.copy():
```

redbot/cogs/mod/kickban.py
```diff
@@ -230,41 +230,49 @@ async def ban_user(
...
+                                        break  # skip the rest of this guild
+                                    log.info(f"Failed to unban member: error code: {e.code}")
+                                else:
+                                    # user unbanned successfully
+                                    guild_tempbans.remove(uid)
+                        await self.config.guild(guild).current_tempbans.set(guild_tempbans)
+            except Exception:
+                log.exception("Something went wrong in check_tempban_expirations:")
+            finally:
+                await asyncio.sleep(60)
```
Jackenmen: This doesn't really need to be in finally so let's keep it simple:
```suggestion
            await asyncio.sleep(60)
```

redbot/cogs/mod/kickban.py
Jackenmen: It would be good to use either of
```
async with self.config.guild(guild).current_tempbans.get_lock():
    ...

async with self.config.guild(guild).current_tempbans():
    ...
```
to avoid a situation where we replace some tempban accidentally which doesn't seem *that* unlikely when you consider that this task runs every minute.

Jackenmen (changes requested):
This is a great improvement, I have just a few review comments before we can get this merged.

PredaaA:
Updated the PR with requested changes.

Jackenmen:
The indentation levels on this method started to get a bit ridiculous so I went ahead and split this into more methods :)

Jackenmen (approved):
Looks good to me, thanks for working on speeding up Red.

PR #4912, opened 2021-03-23
Add `deprecated-removed` directive to Sphinx

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Added a small extension for Sphinx inspired by CPython's directive called `deprecated-removed`.

This directive accepts 3 arguments - version number where the deprecation will happen (e.g. `3.4.8`), the minimum amount of days that must pass since the date of the release of that version (e.g. `30`), and the (technically optional but in practice it should always be used) deprecation "note".

Here's an example call to it:
```py
.. deprecated-removed:: 3.4.8 30
    Use ``guild`` parameter instead.
```
And this results in one of these two outputs, depending on whether the given version was already released:
image
or not:
image

The code is based on Sphinx's built-in [`deprecated` directive](https://github.com/sphinx-doc/sphinx/blob/0949735210abaa05b6448e531984f159403053f4/sphinx/domains/changeset.py).

### DIFF
--- a/docs/_ext/deprecated_removed.py
+++ b/docs/_ext/deprecated_removed.py
@@ -1,0 +1,129 @@
+"""
+A Sphinx extension adding a ``deprecated-removed`` directive that works
+similarly to CPython's directive with the same name.
+
+The key difference is that instead of passing the version of planned removal,
+the writer must provide the minimum amount of days that must pass
+since the date of the release it was deprecated in.
+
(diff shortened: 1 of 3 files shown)


Jackenmen:
Updated the PR to add a `deprecated_removed()` function to our internal utils which can be used to issue a similar warning to the one available in the Sphinx extension.

Kowlin (approved):

PR #4913, opened 2021-03-23
Add user param to menu()

laggron42:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

A user can be provided when using menus instead of having to rely on the hardcoded `ctx.author` for defining who should be allowed to interact. If nothing is provided, falls back to that default `ctx.author`


Jackenmen:
The way this is currently done is breaking as arguments to `menu()` can be passed positionally

Drapersniper:
> The way this is currently done is breaking as arguments to `menu()` can be passed positionally

A simple fix would be adding the user argument as the last argument in the fingerprint which should make it non breaking in positional usages

laggron42:
Actually that's a pretty annoying problem which I didn't think of. Yeah, passing the user as last argument wouldn't break anything, but that would then make the docs invalid

> Additionally, they must take all of the parameters of this function, in addition to a string representing the emoji reacted with. This parameter should be the last one, and none of the parameters in the handling functions are optional

Is this worth a breaking change? Can't think of a way to follow that without breaking other installations

laggron42:
Based on Jack's comments, I modified the way it's done. `user` is a keyword-only argument with a default set to `None`. We check if `user` is `None` to determine whether the parameter should be passed to the controls or not. That way, this implementation is not a breaking change.

I'm aware of the new d.py menus function, didn't really look into it when first writing this PR. If you prefer to close this in favor of their menus, that's ok, I'm probably going to switch to their menus too anyway. Just wanted to make that PR working at least.

Jackenmen (approved):
I think this is a useful change that shouldn't affect anything badly with the current implementation where we only call controls differently if the argument is passed.

Since the `menu()` API is ancient (aside from the recently added `set usebuttons` functionality and related fixes) I added provisional notes expiring in 2 months, let's see if we run into any issues here. Thanks for making the PR!

PR #4914, opened 2021-03-23
allowed_by_whitelist_blacklist() - add guild param, deprecate guild_id

### LINKED ISSUE #4905: `bot.allowed_by_whitelist_blacklist` doesn't work when passed with `guild_id` and `who_id`

3.4.7

I tried to run this code:
```py
await bot.allowed_by_whitelist_blacklist(who_id=133801473317404673, guild_id=guild.id)
```

It to return a bool.

### What actually happened?

It raised an exception:
```py
  File "/home/ubuntu/red-data/cogs/CogManager/cogs/diagnoser/diagnoser.py", line 165, in _check_whitelist_blacklist_issues
    is_role_related = not await self.bot.allowed_by_whitelist_blacklist(
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/core/bot.py", line 597, in allowed_by_whitelist_blacklist
    if guild.owner_id == who.id:
AttributeError: 'Object' object has no attribute 'owner_id'
```

Running this in an eval (in a guild text channel) should be enough:
```py
await bot.allowed_by_whitelist_blacklist(who_id=133801473317404673, guild_id=guild.id)
```

This is *a bit* problematic, guild ID doesn't give us enough data to check whether the user should be permitted to do an action as we need a guild's owner id for that.

On the bright note though, this was broken for 17 months (sic!) so we can (mostly) safely change the signature here (without the need of doing a 3.x.0 release) to require `guild` rather than `guild_id` as this parameter configuration has **never** worked.
But even with that said, it would not hurt to leave the `guild_id` param (with deprecation warning and it raising an AttributeError if it can't save itself by using `bot.get_guild`) until the next 3.x.0 release just so that we don't outright break someone's usage of the method in the cases where they weren't affected.

Jackenmen:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

This PR adds a new `guild` param to `bot.allowed_by_whitelist_blacklist()` and deprecates `guild_id`. The code tries its best to get the guild object when `guild_id` is passed, but if it can't manage to do this, it raises an `AttributeError` similarly to what was (unintentionally) raised before.

Fixes #4905

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -592,7 +592,7 @@ class RedBase(
                     "`guild_id` parameter",
                     "3.4.8",
                     30,
-                    "Use `guild_id` parameter instead.",
+                    "Use `guild` parameter instead.",
                     stacklevel=2,
                 )
                 if guild:
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -551,11 +551,13 @@ class RedBase(
         guild : Optional[discord.Guild]
             When used in conjunction with a provided value for ``who_id``, checks
             the lists for the corresponding guild as well.
+            This is ignored when ``who`` is passed.
         guild_id : Optional[int]
             When used in conjunction with a provided value for ``who_id``, checks
             the lists for the corresponding guild as well. This should not be used
             as it has unfixable bug that can cause it to raise an exception when
             the guild with the given ID couldn't have been found.
+            This is ignored when ``who`` is passed.
 
             .. deprecated-removed:: 3.4.8 30
                 Use ``guild`` parameter instead.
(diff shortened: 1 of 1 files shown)


redbot/core/bot.py
```diff
@@ -568,15 +581,24 @@ async def allowed_by_whitelist_blacklist(
...
                 raise TypeError("Must provide a value for either `who` or `who_id`")
             mocked = True
             who = discord.Object(id=who_id)
             if guild_id:
-                guild = discord.Object(id=guild_id)
+                deprecated_removed(
+                    "`guild_id` parameter",
+                    "3.4.8",
+                    30,
+                    "Use `guild_id` parameter instead.",
```
TrustyJAID: should this be "use `guild` paremeter instead."?
```suggestion
                    "Use `guild` parameter instead.",
```

Jackenmen: Yes, it should, thanks!

TrustyJAID (approved):
Looks good to me!

PR #4915, opened 2021-03-26
 a small typo

6days9weeks:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

A little too much small typo which I see every time when making servers and setting up the mutes cog. Don’t know if a small change like this is accepted but nonetheless.


Kowlin (approved):
Thanks for the fix.

PR #4917, opened 2021-03-27
Fix no dispatch on shared API token removal

### LINKED ISSUE #4916: Dispatch listener when deleting services

Shared API Keys

I have a cog that was previously fetching the shared API keys when running a command, I fixed this by making a little cache, so on, I made a listener in my cog to detect changes if the API key of the service I'm using is being touched, so if the API key is edited, my cog will then call again the shared API.

The problem here is that if the user delete the service, my cog will NOT get any updates, so it'll still use the cache I made, that is a problem since I want my cog to know if the key is updated, or deleted.

My proposal is to either add this to the actual `on_red_api_tokens_update` or create a new listener. (`on_red_api_service_delete`)

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #4916

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1230,6 +1230,7 @@ class RedBase(
         async with self._config.custom(SHARED_API_TOKENS, service_name).all() as group:
             for name in token_names:
                 group.pop(name, None)
+        self.dispatch("red_api_tokens_update", service_name, MappingProxyType(group))
 
     async def remove_shared_api_services(self, *service_names: str):
         """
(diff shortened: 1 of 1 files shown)


Jackenmen:
It would be appreciated if you could test whether this works and leave a comment here, @Predeactor

madebylydia (approved):
Tested and seems good for me, thank for the fast PR :+1:

TrustyJAID (approved):
This looks good to me!

PR #4918, opened 2021-03-27
Feat/schema update

issy:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

In both of the schemas provided in this repo I've added a `additionalProperties: false` rule at the base level. This is useful for anyone wanting to run a json validator against their cogs repo, as it will alert you of unrecognised keys (those that do not conform to the schema).
This shouldn't be a breaking change, and is in fact a bugfix for anyone currently attempting to run json validation against their cogs repo.

### DIFF
--- a/schema/red_cog.schema.json
+++ b/schema/red_cog.schema.json
@@ -1,98 +1,99 @@
 {
-  "$id": "https://raw.githubusercontent.com/Cog-Creators/Red-DiscordBot/V3/develop/schema/red_cog.schema.json",
-  "$schema": "http://json-schema.org/draft-07/schema#",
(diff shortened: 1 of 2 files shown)


Jackenmen:
When the schemas were first added, this was done on purpose - there are keys that are not currently a part of schema but are widely used like `name` or `permissions` keys. I didn't want this schema to be of no use to people who use these additional keys, but standardizing these (especially the `name` key) requires a more significant effort due to extra considerations we need to take like cog packages that contain multiple cogs (or no cogs).

issy:
I understand, thank you for the explanation. In this case I suppose the schema is better off as it was before

PR #4926, opened 2021-03-30
Pre-fetch app owners and fail early on no owners

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

There's really no reason to allow users to run the bot without having an owner set and this is a point of confusion for users that are trying to set up Red using a team application.

Marked as Breaking Change as this will affect anyone who runs Red without having any owner set for some strange reason...

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -28,7 +28,7 @@ from redbot import _early_init, __version__
 _early_init()
 
 import redbot.logging
-from redbot.core.bot import Red, ExitCodes
+from redbot.core.bot import Red, ExitCodes, _NoOwnerSet
 from redbot.core.cli import interactive_config, confirm, parse_cli_flags
 from redbot.setup import get_data_dir, get_name, save_config
 from redbot.core import data_manager, drivers
@@ -401,6 +401,24 @@ async def run_bot(red: Red, cli_flags: Namespace) -> None:
             "You can find out how to enable Privileged Intents with this guide:\n"
             "https://docs.discord.red/en/stable/bot_application_guide.html#enabling-privileged-intents",
             style="red",
+        )
+        sys.exit(1)
+    except _NoOwnerSet:
+        print(
+            "Bot doesn't have any owner set!\n"
+            "This can happen when your bot's application is owned by team"
+            " as team members are NOT owners by default.\n\n"
+            "Remember:\n"
+            "ONLY the person who is hosting Red should be owner."
+            " This has SERIOUS security implications."
+            " The owner can access any data that is present on the host system.\n"
+            "With that out of the way, depending on who you want to be considered as owner,"
+            " you can:\n"
+            "a) pass --team-members-are-owners when launching Red"
+            " - in this case Red will treat all members of the bot application's team as owners\n"
+            "b) set owner manually with redbot --edit <instance_name>\n"
+            "c) pass owner ID(s) when launching Red with --owner"
+            " (and --co-owner if you need more than one) flag\n"
         )
         sys.exit(1)
 
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -416,7 +416,7 @@ async def run_bot(red: Red, cli_flags: Namespace) -> None:
             " you can:\n"
             "a) pass --team-members-are-owners when launching Red"
             " - in this case Red will treat all members of the bot application's team as owners\n"
-            "b) set owner manually with redbot --edit <instance_name>\n"
+            f"b) set owner manually with `redbot --edit {cli_flags.instance_name}`\n"
             "c) pass owner ID(s) when launching Red with --owner"
             " (and --co-owner if you need more than one) flag\n"
         )
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -384,7 +384,7 @@ async def run_bot(red: Red, cli_flags: Namespace) -> None:
         await red.http.close()
         sys.exit(0)
     try:
-        await red.start(token, bot=True, cli_flags=cli_flags)
+        await red.start(token, bot=True)
     except discord.LoginFailure:
         log.critical("This token doesn't seem to be valid.")
         db_token = await red._config.token()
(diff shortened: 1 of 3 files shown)


redbot/__main__.py
```diff
@@ -401,6 +401,24 @@ async def run_bot(red: Red, cli_flags: Namespace) -> None:
...
+            " as team members are NOT owners by default.\n\n"
+            "Remember:\n"
+            "ONLY the person who is hosting Red should be owner."
+            " This has SERIOUS security implications."
+            " The owner can access any data that is present on the host system.\n"
+            "With that out of the way, depending on who you want to be considered as owner,"
+            " you can:\n"
+            "a) pass --team-members-are-owners when launching Red"
+            " - in this case Red will treat all members of the bot application's team as owners\n"
+            "b) set owner manually with redbot --edit <instance_name>\n"
```
fixator10: Probably better would be enclose `redbot --edit <instance_name>` in quotes

Jackenmen: Done.

redbot/core/events.py
```diff
@@ -70,19 +70,7 @@ async def on_ready():
...
-                bot.owner_ids.update(m.id for m in app_info.team.members)
-        elif bot._owner_id_overwrite is None:
-            bot.owner_ids.add(app_info.owner.id)
-        bot._app_owners_fetched = True
-
-        try:
-            invite_url = discord.utils.oauth_url(app_info.id)
-        except:
-            invite_url = "Could not fetch invite url"
+        invite_url = discord.utils.oauth_url(bot._app_info.id)
```
Flame442, line 73: Should this have a check `if bot._app_info is None`, like `is_owner` has? I don't see a case where that could be possible, so IDK why `is_owner` has it, but if there _is_ a reason, it should also be included here.

redbot/__main__.py
```diff
@@ -403,6 +403,24 @@ async def run_bot(red: Red, cli_flags: Namespace) -> None:
...
+            " as team members are NOT owners by default.\n\n"
+            "Remember:\n"
+            "ONLY the person who is hosting Red should be owner."
+            " This has SERIOUS security implications."
+            " The owner can access any data that is present on the host system.\n"
+            "With that out of the way, depending on who you want to be considered as owner,"
+            " you can:\n"
+            "a) pass --team-members-are-owners when launching Red"
+            " - in this case Red will treat all members of the bot application's team as owners\n"
+            f"b) set owner manually with `redbot --edit {cli_flags.instance_name}`\n"
```
Flame442, line 419: Having ticks here but on none of the other things is a little weird, not sure if they should be here.

redbot/__main__.py
```diff
@@ -403,6 +403,24 @@ async def run_bot(red: Red, cli_flags: Namespace) -> None:
...
+            "c) pass owner ID(s) when launching Red with --owner"
+            " (and --co-owner if you need more than one) flag\n"
+        )
+        sys.exit(1)
```
Flame442, line 423: A MASSIVE traceback covers up this message, maybe add an `input()` or some other method of *hey look at this here* before exiting?
image

Flame442 (changes requested):
Functionally works, but not very user friendly.

redbot/__main__.py
Jackenmen: I can't reproduce this issue.
image

redbot/__main__.py
Jackenmen: I used backticks because they're already used for indicating commands in other places:
https://github.com/Cog-Creators/Red-DiscordBot/blob/a8f35f762c09195f5b02d5ffbdfd3a093a80dafd/redbot/__main__.py#L57
https://github.com/Cog-Creators/Red-DiscordBot/blob/a8f35f762c09195f5b02d5ffbdfd3a093a80dafd/redbot/setup.py#L192

redbot/__main__.py
Flame442: Seems to only happen if downloader is in the cogs the bot tries to load (notice the exception in the first message said it was an unhandled error in a downloader task)
image
I don't know if that makes it ok, since most cases this warning is trying to catch would happen _before_ downloader is ever loaded, but that's how to reproduce it at least.

redbot/core/events.py
Jackenmen: I was thinking that this could be needed if `is_owner()` is run before app info is fetched (in `on_ready` it is definitely going to be already fetched). Looking at it now, I don't think that's really possible to do anyway - `pre_flight()` is ran *before* we log in so the owner wouldn't be able to be fetched then anyway *and* everything after that is done after we already have the app_info fetched.

I'll update this to not check for it since it's indeed unnecessary, thanks.

redbot/__main__.py
Jackenmen: We *could* change the order of things so that we log in to the REST API and pre-fetch owners *before* `pre_flight()`. This would require me to leave the `_app_info is None` check mentioned in the other comment but it would probably make sense to do since there's no reason to go *that far* into the startup if we're going to have to exit early.
This would (not purposefully) *also* allow people to use some of the REST API stuff *before* the bot is ready since we would already have the token set but that shouldn't cause any problems and might even make some people happy even though we wouldn't otherwise be making this change.

The issue with Downloader's traceback should probably also be reported separately as I don't think CancelledError should be allowed to propagate from the task's done callback.

Jackenmen: Note: this will also require splitting some of the early stuff from `pre_flight` into a separate place.

Jackenmen:
I have split `pre_flight` into `_pre_login` and `_pre_connect` so that we can do cog load after we log in while still doing some of the things before the login (and therefore before the owner pre-fetching).
That also allowed me to drop `_app_info is None` check as well.

Let me know if there are any other changes needed.

redbot/__main__.py
Flame442: image
yooooooooooo!

Flame442 (approved):
Working good now!

PR #4930, opened 2021-04-02
[Docs] Use pascal casing for example cog in cog creation guide

Kreusada:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This area of the docs is referenced by a lot by users trying to create their own cogs - I do think the example cog should use the camel case recommendation provided in the cog creators guide (see below)

image

### DIFF
--- a/docs/guide_cog_creation.rst
+++ b/docs/guide_cog_creation.rst
@@ -83,7 +83,7 @@ In that file, place the following code:
 
     from redbot.core import commands
 
-    class Mycog(commands.Cog):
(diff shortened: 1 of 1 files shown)


Rapptz:
Doesn't matter much but that's called `PascalCase`. This is `camelCase`.

fixator10:
PEP8 names it `CapWords`
https://www.python.org/dev/peps/pep-0008/#class-names

Jackenmen (approved):
Thanks!

PR #4934, opened 2021-04-03
[Mutes] Fix data conversion to shema_version 1 and add a note about it

PredaaA:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

This fixes data conversion that was missing value in context manager. And it also adds a note to avoid potential user confusion about the cog taking a while to be ready if it needs data conversion.
The conversion has also been moved to its own method.


TrustyJAID (dismissed):
This looks good to me, good catch on the missing attribute from my suggestion. In practice I don't see channel mutes used nearly as frequently as server mutes so I am not concerned with time to convert the old settings.

TrustyJAID (approved):

PR #4936, opened 2021-04-04
Make customcoms convert-parameters a core util

TrustyJAID:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

This adds the utility of converting ctx arguments into strings for use by other cogs where this was previously only available in customcommands. This also adds `{p}` for the bots prefix in the channel and `{pp}` for all the bots prefixes for the channel.

This has not been tested using Context but it has been tested using message object.

### DIFF
--- a/redbot/core/utils/chat_formatting.py
+++ b/redbot/core/utils/chat_formatting.py
@@ -1,16 +1,31 @@
+from __future__ import annotations
 import datetime
 import itertools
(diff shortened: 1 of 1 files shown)


flaree:
Is there a way for this to be able to accept more objects dynamically as to avoid just duplicating the code?

PR #4938, opened 2021-04-05
[Streams] Minimize Twitch API calls

PredaaA:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

### Description of the changes
The `is_online` method has been rewrote, it removes two uneeded API calls, first was `/helix/users` because it was called twice, and second was `/helix/games` to get game name, which is already existing in `/helix/streams` that we call too.

This PR depends off #4883.

PR #4939, opened 2021-04-05
[Streams] Remove stream alert if stream not found

PredaaA:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

It removes a stream from `self.streams` and there streams config value if it is not found.

This PR depends off #4938.

### DIFF
--- a/redbot/cogs/streams/streams.py
+++ b/redbot/cogs/streams/streams.py
@@ -778,6 +778,10 @@ class Streams(commands.Cog):
 
                     stream.messages.clear()
                     await self.save_streams()
+                except StreamNotFound:
+                    self.streams.remove(stream)
+                    await self.save_streams()
+                    continue
                 else:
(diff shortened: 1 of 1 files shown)


redbot/cogs/streams/streams.py
```diff
@@ -750,6 +750,10 @@ async def check_streams(self):
                                 await message.delete()
                     stream._messages_cache.clear()
                     await self.save_streams()
+                except StreamNotFound:
+                    self.streams.remove(stream)
```
Jackenmen: If we want to do this, we need to iterate on a copy of `self.streams` so as not to change what we iterate on during the iteration.

Jackenmen (changes requested):

Jackenmen (approved):
Looks simple enough, thanks.

PR #4955, opened 2021-04-06
[Context] Fixup docstrings in `react_quietly` function

Kreusada:
### Type

- [ ] Bugfix
- [x] Enhancement
- [x] New feature

~~Add a ``cross()`` function in context which is analogous to ``tick()``. Also, a small docstring fix with the ``react_quietly()`` function.
The cross function could be debatable due to the fact that an 'error' should respond with meaningful information, there are a few times where I've wanted to use this, though, and imo it would be better to have it, than to not - up for discussion.~~

Re-opened just to fix the docstring in `react_quietly`.

### DIFF
--- a/redbot/core/commands/context.py
+++ b/redbot/core/commands/context.py
@@ -18,7 +18,6 @@ if TYPE_CHECKING:
     from ..bot import Red
 
 TICK = "\N{WHITE HEAVY CHECK MARK}"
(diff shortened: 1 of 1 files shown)


Kreusada:
Re-opening this as as a different PR as there was still something in here which could be merged.

Drapersniper (approved):
choo choo the approval train is here!

PR #4958, opened 2021-04-06
Stop reloading unrelated modules

### LINKED ISSUE #4956: Cog packages starting with a strict prefix of `redbot` crash the drivers on load

3.4.9 / 3.4.10.dev1

### Cog name

CogManagerUI

### Command name

load

For a private cog package to be loaded, named 'red'.

### What actually happened?

The package failed to load, with the traceback provided below:
```py
Exception during import of package
  File "/home/ubuntu/redenv/lib/python3.8/site-packages/redbot/core/core_commands.py", line 156, in _load
    spec = await bot._cog_mgr.find_cog(name)
  File "/home/ubuntu/redenv/lib/python3.8/site-packages/redbot/core/cog_manager.py", line 285, in find_cog
    return await self._find_ext_cog(name)
  File "/home/ubuntu/redenv/lib/python3.8/site-packages/redbot/core/cog_manager.py", line 221, in _find_ext_cog
    real_paths = list(map(str, [await self.install_path()] + await self.user_defined_paths()))
  File "/home/ubuntu/redenv/lib/python3.8/site-packages/redbot/core/cog_manager.py", line 72, in install_path
    return Path(await self.config.install_path()).resolve()
  File "/home/ubuntu/redenv/lib/python3.8/site-packages/redbot/core/config.py", line 176, in _get
    ret = await self.driver.get(self.identifier_data)
  File "/home/ubuntu/redenv/lib/python3.8/site-packages/redbot/core/drivers/json.py", line 138, in get
    partial = partial[i]
TypeError: 'NoneType' object is not subscriptable
```

This occured on the following bots:

Bot 1:
Version: 3.4.10.dev1
Py version: 3.8.5
Dpy version: 1.7.0
Storage type: JSON
System arch: aarch64
OS: Linux

Bot 2:
Version: 3.4.9
Py version: 3.8.5
Dpy version: 1.7.1
Storage type: JSON
System arch: AMD64
OS: Windows

I haven't looked into the side-effects of this, but it could be damaging to the backend. Perhaps a simple fix for this would be to check if the cog is named 'red' when loading, and return if it is - im happy to PR something for that. Obviously, a sensible workaround would be to just rename the package, but I still believe this should be handled.

Edit: This red cog does not use config. It doesn't matter whether they use config or not.

1. Create a simple cog, with the package name as 'red', and the cog class as 'Red'. This can be installed from github, or loaded from a path, it makes no difference.
2. Load the cog, and get ready for your bot to error out on almost every command that evolves around config. For example, the basic `[p]ping` command will still function, because it does not use config. `[p]info` uses config to set custominfo, so this errors, same with help, or any customizable command.
3. Simply restart the bot for these errors to be pruned, I suggest then completely removing the cog from your bot.

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Fixes #4956 
We'll get rid of this whole thing in 3.5 but let's maintain this for a little bit longer...

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -240,7 +240,11 @@ class CoreLogic:
         for m in modules:
             maybe_reload(m)
 
-        children = {name: lib for name, lib in sys.modules.items() if name.startswith(module_name)}
+        children = {
+            name: lib
+            for name, lib in sys.modules.items()
+            if name == module_name or name.startswith(f"{module_name}.")
+        }
         for child_name, lib in children.items():
             importlib._bootstrap._exec(lib.__spec__, lib)
 


Drapersniper (approved):
I mean it works, i couldnt get it to mess up.

PR #4961, opened 2021-04-06
[Audio] Check if player.current is not None before resume

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

There was a missing check to know if player.current is not None before resume the player. I checked and looks like it was the only one missing.


redbot/cogs/audio/core/events/lavalink.py
```diff
@@ -64,8 +64,8 @@ async def lavalink_event_handler(
             event_channel_id = extra.get("channelID")
             _error_code = extra.get("code")
             if _error_code in [1000] or not guild:
-                if _error_code == 1000:
-                    await player.resume(player.current, start=player.position, replace=False)
+                if _error_code == 1000 and current_track is not None and player.is_playing:
+                    await player.resume(current_track, start=player.position, replace=False)
```
Drapersniper: Can you just in the case of a potential edge case ... change the `current_track` in here to `player.current` on both here.

In the off change theres an issue while its resuming it should resume the current song not a potentially  old/skipped song.

Drapersniper (approved):

aikaterna (approved):

PR #4962, opened 2021-04-07
embedset - fix perm check for subcommands, add missing return

bobloy:
…return to `embedset user`

### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

Corrects two small bugs in the Core cog.


Jackenmen (approved):
Works great, thanks for providing us with a fix!

PR #4964, opened 2021-04-08
Include more files in distributions

Jackenmen:
### Type

- [x] Bugfix
- [x] Enhancement
- [ ] New feature

The main `LICENSE` file was already included in dist-info, now it's also included in the root of the source dist.
In addition to that, now we *actually* include `discord-ext-menus.LICENSE` file as part of the wheel and source distribution.
I also added `tests` folder and `tox.ini` to sdist so that people downloading sdist can do a full tox run from the sdist, similarly to what we do when building.

I've ensured that the only difference between the files that were included before and the files that are included now is the inclusion of LICENSE files (+ tests folder and tox.ini file in case of sdist).

Since I've also updated our build system to use PEP 517 isolated builds and the new `build` package from PyPA, I've also ensured that the `python -m build` command works properly to avoid issue during release.

### DIFF
--- a/.github/workflows/publish_release.yml
+++ b/.github/workflows/publish_release.yml
@@ -19,13 +19,13 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install setuptools wheel twine
+          pip install build twine
       - name: Build and publish
         env:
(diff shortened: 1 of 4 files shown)


Kowlin (approved):
Sounds good, merge whenever.

PR #4965, opened 2021-04-08
Mutes API

TrustyJAID:
This introduces the mutes cog rewrite as a full API that can be utilized by cog creators. This is a draft PR until I am finished with the changes.

### DIFF
--- a/redbot/core/errors.py
+++ b/redbot/core/errors.py
@@ -91,6 +91,7 @@ class CannotSetSubfield(StoredTypeError):
 
 class MuteError(RedError):
     """Base error class for Mute-related errors."""
+
     def __init__(self, **kwargs):
         self.member = kwargs.get("member")
         self.reason = kwargs.get("reason")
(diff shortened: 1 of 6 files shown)


redbot/core/errors.py
```diff
@@ -87,3 +87,84 @@ class CannotSetSubfield(StoredTypeError):
...
+        self.reason = kwargs.get("reason")
+        self.guild = kwargs.get("guild")
+        self.channel = kwargs.get("channel")
+        self.channels = kwargs.get("channels")
+
+
+class MuteRoleMissingError(MuteError):
+    """Raised if role mutes are set and the role is missing."""
+
+    def __init__(self, **kwargs):
```
DevilXD, line 106: If all of those inherit from `MuteError`, then an `__init__` method like this is not necessary here, as it's signature is the same, and all it does is pass the kwargs along. I'd remove those, and leave only the doc string as the body of each exception. Unless, of course, you plan on expanding those in the future, and they'll be doing something more than just passing kwargs along.

TrustyJAID:
Closing this PR in favor of utilizing discord timeouts as an api method instead.

PR #4966, opened 2021-04-08
Fix a bunch of typos in commands in Core

Fabian-Evolved:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

Recent commit "Core Cog Guide" had a few grammatical mistakes, and some identation mistakes, i've fixed those in this pull request.


Fabian-Evolved:
I just noticed i forgot to put a title on the pull request, I hope that is not an issue.

If it is, let me know, and i'll open a new one as soon as possible (read: when i wake up later)

redbot/core/core_commands.py
```diff
@@ -3969,7 +3969,7 @@ async def localblocklist_remove(
         self, ctx: commands.Context, *users_or_roles: Union[discord.Member, discord.Role, int]
     ):
         """
-        Removes user or role from blocklist.
+        Removes user or role from local blocklist.
```
Drapersniper, line 3972: IIRC this was done on purpose as @Kowlin  wanted all mentions to the allowlist and denylist at a local level to be just "blocklist" and "allowlist" and references to the global one to be "global allowlist" and "global blocklist"

Dav-Git: Check above in line 388, there it is also referenced as "local" so consistency is definitely an issue.

Fabian-Evolved: > IIRC this was done on purpose as @Kowlin wanted all mentions to the allowlist and denylist at a local level to be just "blocklist" and "allowlist" and references to the global one to be "global allowlist" and "global blocklist"

Alright i will get rid of that, do i also remove the one from line 388? and i'll push that too.

Kowlin: Feel free and just leave it @Fabian-Evolved, that's something for me to do when I make a gigantic pass over all the wording later in the year.

What Draper wasn't aware of is that I'd do that stuff later on.

Flame442 (approved):
Thanks for the fixes!

PR #4967, opened 2021-04-08
[Streams] Check required permissions in alerts

PredaaA:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

This is to avoid 403s because of missing permissions in stream alerts messages.


Hbutlercapone (approved):

Jackenmen (commented):
see below

Jackenmen (dismissed):
I am not so sure whether we should be removing all the streams set for the channel once the bot no longer has permission to send messages in it. It seems to me like someone could be bitten by it if they weren't doing changes to permissions and temporarily removed bot's permissions in the process.

Perhaps just checking for the perms *without* removing the streams would be enough?

PredaaA:
> I am not so sure whether we should be removing all the streams set for the channel once the bot no longer has permission to send messages in it. It seems to me like someone could be bitten by it if they weren't doing changes to permissions and temporarily removed bot's permissions in the process.
> 
> Perhaps just checking for the perms _without_ removing the streams would be enough?

Well, it's been over a year but I've just applied your suggestion. That actually makes more sense that way.

palmtree5 (approved):
Looks good to me, the requested changes were implemented

PR #4968, opened 2021-04-08
[Streams] Improve config calls in stream alerts

PredaaA:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

It now does one config call for each channel, instead of each value from each channel.


Jackenmen (approved):
LGTM, I've done small code style changes with what I found to be a bit more readable.

PR #4972, opened 2021-04-08
[Help] Allow prefix formatting in taglines

### LINKED ISSUE #4669: [Help] Tagline formatting
# Feature request

- [x] Cog
- [ ] Command
- [ ] API functionality

Be able to use `[p]` in `[p]helpset tagline` to specify the context's prefix. Currently it will only use format on the default tagline. 

Nice to have: also the ability to reference the current command object for `command.signature` or `command.qualified_name`

Default tagline can use the context prefix: 
image

Custom tagline can't use context prefix:
image

Kreusada:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Allow the use of ``[p]`` to be replaced with the invoked prefix inside taglines. Fixes #4669.


Jackenmen:
> Contributes towards #4669.

Any reason why it only contributes towards that issue but doesn't fix it? I'm not sure what I am missing here that this PR doesn't fix.

Kreusada:
> > Contributes towards #4669.
> 
> Any reason why it only contributes towards that issue but doesn't fix it? I'm not sure what I am missing here that this PR doesn't fix.

Reason why I said "contributes" is because the idea of referencing the command object was also mentioned in the issue, of which I did not include in this PR. But I guess that was just a side-note, so yes this PR will fix the main outline of the issue.

Jackenmen:
Thanks for clarifying, that does make sense indeed. I'll update this PR's description to say it fixes #4669 then and comment on the issue with more information.

Flame442 (approved):

PR #4975, opened 2021-04-10
Fix logging and handle NotFound exception in Case.edit()

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

The logging call was missing the format fields used in the log message.
I also added handling for `discord.NotFound` as there's no reason to continue trying to edit the modlog message in the future if we are being told that it no longer exists.


Drapersniper (approved):
Approving, ahead of response in AC - that being said would like to hear the feedback on it before merging.

PR #4977, opened 2021-04-10
Stop fetching the messages when getting the cases in Modlog

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

This gives us an enormous performance improvement (in the case of my test bot that had 500 cases, it went down from 7:33 minutes to ~0.06s) with the very small cost of API users now having to *manually* fetch the message *if* they actually need its content for some reason.
From looking at various cog repositories, it does not seem to be something that people do at all, which is not really surprising and fetching the message with `PartialMessage` object is a matter of calling `partial_message.fetch()`. Also, `Case.message` will continue to be `discord.Message` for object returned from `modlog.create_case()` as I didn't see much reason to require the API users to fetch the message object that we already got from the API. I did not, however, make it update the `Case.message` with full `discord.Message` when doing any edits on the `Case` object as I think it's just an unnecessary complication.

Depends on #4975 (needed because this is more likely to occur with PartialMessage objects) and #4979 (this PR should update the documentation added by #4979 though it doesn't yet as I want to wait with that until #4979 gets merged)

### DIFF
--- a/redbot/core/modlog.py
+++ b/redbot/core/modlog.py
@@ -286,8 +286,11 @@ class Case:
     modified_at: Optional[float]
         The UNIX time of the last change to the case.
         `None` if the case was never edited.
-    message: Optional[discord.Message]
+    message: Optional[Union[discord.PartialMessage, discord.Message]]
         The message created by Modlog for this case.
+        Instance of `discord.Message` *if* the Case object was returned from
+        `modlog.create_case()`, otherwise `discord.PartialMessage`.
+
         `None` if we know that the message no longer exists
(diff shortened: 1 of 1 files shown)


Dav-Git (approved):
LGTM

SharkyTheKing:
Looks good.

Local hosted bot with ~500 modlog cases took less than a second to fetch all cases for a nice eval command to show which mod moderated the most.

VPS hosted bot with ~300 nearly 400 cases took 300 seconds to fully preview same information.

Works perfectly.

Kowlin (approved):

PR #4979, opened 2021-04-10
Add documentation for Case class in Modlog

Jackenmen:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature


Drapersniper (approved):
EWWWWH so much doc, we need a doc reduction PR ...

PR #4991, opened 2021-04-15
Remove PM2 guide

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] New feature
- [x] Removal

There's apparently no way to specify exit codes
for which the process should not be restarted.

This just shows how poor of a process manager PM2 is
so we really should not continue to recommend it in the documentation...

Ref:
https://stackoverflow.com/questions/43614017/how-to-only-restart-node-process-with-pm2-if-exit-code-0

---

Change my mind :P


Apoo711:
This'll remove auto restart for mac users, as pm2 works but we don't have systemd.

aikaterna:
Mac users use a plist file.

Drapersniper:
Consider this my approval here, dont wanna do an official one so the big green button doesnt show my and im physically forced to press it just yet...

Kowlin:
Can't find the damn approve button on mobile...

PR #4992, opened 2021-04-15
Allow mute commands to update duration of the mute

### LINKED ISSUE #4834: Allow mute commands to update duration of the mute

Mutes

Currently, there's no clear way of changing the duration of an existing mute that doesn't involve unmuting and muting with a new duration. It does seem that it already works with `[p]mute` when using role-based mutes, but it seems to be work incidentally.
I suggest that all mute commands get an enhancement that would allow the commands to update the duration of an existing mute rather than fail with "That user is already muted in this channel." or a similar error message.

davidhu99:
### Type

- [ ] Bugfix
- [x] Enhancement
- [ ] New feature

Previously, when the author wanted to mute an already muted user, if muted user already inside muted list (if statement on line 1666), "That user is already muted in this channel." would be sent to the text channel and the user's current mute duration would not get updated. 

Currently, when author mutes an already muted user, there will be an update to the mute duration. One case we accounted for is if the author changes the mute duration and happens to hit it after the mute task has been created. In this case, there may be an extra unmute task in the _self._unmute_tasks__ cache, which will do nothing. We accounted for this by checking if an unmute task is in the unmute_tasks cache, and cancelling it if there is. 

Closes #4834


bobloy:
Is it correct to allow the mute command to *shorten* the duration of the mute too?

davidhu99:
> Is it correct to allow the mute command to _shorten_ the duration of the mute too?

Good point, in my opinion it should not. However, in the original issue the author said to "update" which would be both shorten and lengthen. Let me know whether to only shorten!

TrustyJAID:
@bobloy Yes the mute command should allow shortening the duration of a mute as well as extending it, it is the moderators perogative to choose when to mute until and maybe they want to lessen the time to unmute.

Something that wasn't really discussed but should be addressed for consistency is the mute function should also check if there is an existing server unmute task when a user is server muted and cancel the task the same way you did for channel mutes.

davidhu99:
> @bobloy Yes the mute command should allow shortening the duration of a mute as well as extending it, it is the moderators perogative to choose when to mute until and maybe they want to lessen the time to unmute.
> 
> Something that wasn't really discussed but should be addressed for consistency is the mute function should also check if there is an existing server unmute task when a user is server muted and cancel the task the same way you did for channel mutes.

I believe I did that with my if condition on line 1668. Let me know if that's not the correct way to do it

TrustyJAID:
You handled it for channel mutes but role-based server mutes still suffer from the same issue.

davidhu99:
> You handled it for channel mutes but role-based server mutes still suffer from the same issue.

Added handling for that in my most recent commit. Please check it out!

davidhu99:
> You handled it for channel mutes but role-based server mutes still suffer from the same issue.

I see, please check out my last commit for the solution to that :)

redbot/cogs/mutes/mutes.py
```diff
@@ -1532,6 +1532,15 @@ async def mute_user(
             if guild.id not in self._server_mutes:
                 self._server_mutes[guild.id] = {}
 
+            # This is here in the event that the author tries to update the mute duration of user.
+            # we must check if there is an automatic unmute task in the self._server_mutes cache, for the specific guild.id and user.id.
+            # if there is, then we cancel the automatic unmute task.
+            if user.id in self._server_mutes[guild.id]:
+                task_name = f"server-unmute-{guild.id}-{user.id}"
+                print(self._unmute_tasks)
```
TrustyJAID: This file has logging setup. So if you really want to keep this here I recommend changing to `log.debug` rather than print or remove this entirely. It might be worth considering doing a `log.info(f"{user} has had their mute time updated in {guild}")` but I don't think that's entirely necessary.

davidhu99: > This file has logging setup. So if you really want to keep this here I recommend changing to `log.debug` rather than print or remove this entirely. It might be worth considering doing a `log.info(f"{user} has had their mute time updated in {guild}")` but I don't think that's entirely necessary.

Oh sorry that was just a sanity check for me, I'll delete it.

PR #4998, opened 2021-04-19
[Economy] Make top parameter optional in leaderboard

madebylydia:
### Type

- [ ] Bugfix
- [X] Enhancement
- [ ] New feature

The docstring says the top argument can be omited to show the global leaderboard, however, doing so will say `Converting to "int" failed for parameter "top".`
I prefered to fix this by making top an optional parameter.

### DIFF
--- a/redbot/cogs/economy/economy.py
+++ b/redbot/cogs/economy/economy.py
@@ -4,7 +4,7 @@ import random
 from collections import defaultdict, deque, namedtuple
 from enum import Enum
 from math import ceil
-from typing import cast, Iterable, Union, Literal
+from typing import cast, Iterable, Union, Literal, Optional
 
 import discord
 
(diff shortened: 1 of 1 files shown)


Jackenmen:
Examples quite clearly present that the command is meant to be run with both parameters passed if you want to use the second parameter, so I'm not really sure this should be done. It is also not possible to use the bool argument using `0/1` since it will then be consumed by `top` parameter.

madebylydia:
> Examples quite clearly present that the command is meant to be run with both parameters passed if you want to use the second parameter, so I'm not really sure this should be done.

Shame on me, I've been confused with another bot's docstring, so that would need to be edited then, since I consider it as a good idea anyway. 😅

> It is also not possible to use the bool argument using `0/1` since it will then be consumed by `top` parameter.

I haven't thought about it.
As a solution, I would propose to switch `top` and `show_global` arguments. A boolean makes more sense to me to be placed at first in this context, which would allow more simplicity to a lambda user.
I don't know what to think about something else otherwise.

Jackenmen:
> As a solution, I would propose to switch `top` and `show_global` arguments. A boolean makes more sense to me to be placed at first in this context, which would allow more simplicity to a lambda user.

That's assuming that `show_global` is more often used than `top`, is that really true?

Jackenmen:
There was a bit of discussion in #advanced-testing on this, @Predeactor will throw together a list of possible solutions later and then org members can give their feedback on a preferred solution.

madebylydia:
Closing this, Draper will be handling this in another PR.

PR #5001, opened 2021-04-19
[API] add ability to check for disabled command

### LINKED ISSUE #4130: [API] add ability to check for disabled command
# Feature request

- [ ] Cog
- [ ] Command
- [X] API functionality

Add ability to check if command is disabled. E.g. via command property `is_disabled_globally` and func `is_disabled(ctx)` or smth like that.

oyie1:
### Type of Feature Request

- [ ] Cog

- [ ] Command

- [x] API Functionality 

Previously, there was already a way to check whether the command is disabled using the command.enabled attribute, however, we needed to implement a way to check whether a command is disabled in a specific guild.

We made a method that checks whether a command is disabled in a specific guild in the Command class with the name is_disabled_in_guild(). This functions accepts a discord.Guild object as a positional-only argument and will return true if the command is disabled in the guild and returns false if the command is enabled in the guild.

Closes [#4130 ](https://github.com/Cog-Creators/Red-DiscordBot/issues/4130)

### DIFF
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -626,9 +626,7 @@ class Command(CogCommandMixin, DPYCommand):
         """
         disabler = get_command_disabler(guild)
         # If disabler is in self.checks, then command disabled; else the command is enabled
-        if disabler in self.checks:
-            return True
-        return False
+        return disabler in self.checks
 
     def allow_for(self, model_id: Union[int, str], guild_id: int) -> None:
         super().allow_for(model_id, guild_id=guild_id)
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -622,7 +622,6 @@ class Command(CogCommandMixin, DPYCommand):
         -------
         bool
             ``True`` if the command is disabled.
-            ``False`` if the command isn't disabled.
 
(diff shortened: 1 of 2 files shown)


redbot/core/commands/commands.py
```diff
@@ -610,6 +610,27 @@ def enable_in(self, guild: discord.Guild) -> bool:
...
+        guild : discord.Guild
+            The guild to enable the command in.
+
+        Returns
+        -------
+        bool
+            ``True`` if the command is disabled.
+            ``False`` if the command isn't disabled.
+
+        """
```
Twentysix26:
```suggestion
        -------
        bool
            ``True`` if the command is disabled.

        """
```
False is a bit redundant

redbot/core/commands/commands.py
```diff
@@ -610,6 +610,27 @@ def enable_in(self, guild: discord.Guild) -> bool:
...
+        disabler = get_command_disabler(guild)
+        # If disabler is in self.checks, then command disabled; else the command is enabled
+        if disabler in self.checks:
+            return True
+        return False
```
Twentysix26:
```suggestion
        return disabler in self.checks
```

PR #5002, opened 2021-04-20
[Economy] nanooptimizations

fixator10:
Prioritize variables in checks over function calls


Drapersniper (approved):
Thanks for the picooptimizations :awesome:

PR #5004, opened 2021-04-20
Send the DM to a user before applying the warn action.

### LINKED ISSUE #4713: the check for warning points should be after a DM is attempted
# Feature request

- [ ] Cog
- [X] Command
- [ ] API functionality

I wasn't really sure if this counted for a bug report or not, but since its not really a bug but just some weird placement I've put it under here.

as it stands, the check that handles users getting hit by a warnaction occurs **before** Red attempts to DM the warn to them.
[Line 434 here](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/warnings/warnings.py#L434]) should probably be realistically after [this block right under it.](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/warnings/warnings.py#L435-L516)

The common use case I see here is softban as a warnaction - it'll send the softban, say its for excessive warnings, but not send the actual warning.

_Thank you Deihim007 in the main server for confirming it wasn't me going crazy._

Danstr5544:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New Feature

### Description of Changes
When a warn action is enforced, it can sometimes prevent the actual warning being sent via DM too - the most reliable example I can get to trigger is setting a softban as a warn action. This PR aims to prevent that simply Moving the warn action handling to after the DM itself is sent. Thats It, really.

This technically fixes #4713 But considering the way I worded that rivals Topical English Football Controversy in how stupid it is, I'm half tempted to call the association accidental

 

-- Dan

### DIFF
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -430,13 +430,7 @@ class Warnings(commands.Cog):
                 "mod": ctx.author.id,
             }
         }
-        async with member_settings.warnings() as user_warnings:
-            user_warnings.update(warning_to_add)
-        current_point_count += reason_type["points"]
-        await member_settings.total_points.set(current_point_count)
-
-        await warning_points_add_check(self.config, ctx, user, current_point_count)
-        dm = guild_settings["toggle_dm"]
+		dm = guild_settings["toggle_dm"]
         showmod = guild_settings["show_mod"]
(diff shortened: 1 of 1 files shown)


Drapersniper (approved):
LGTM, thanks for improving warnings!

PR #5014, opened 2021-04-26
[General] Provide the urban command with embed color

Kreusada:


Drapersniper (approved):
Thanks for improving Red's UI!

PR #5015, opened 2021-04-27
[General] Fix usage in choose help

TrustyJAID:
Also improve doc-strings to explain usage.

### DIFF
--- a/docs/cog_guides/general.rst
+++ b/docs/cog_guides/general.rst
@@ -74,9 +74,9 @@ Options are separated by spaces.
 
 **Arguments**
 
-* ``<first>``: The first mandatory option to randomly choose from.
-* ``<second>``: The second mandatory option to randomly choose from.
-* ``[others...]``: Any remaining optional options to choose from.
+* ``<first>``: The first mandatory option.
+* ``<second>``: The second mandatory option.
+* ``[others...]``: Any remaining optional options.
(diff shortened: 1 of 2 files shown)


Jackenmen (commented):
I think the usage syntax should continue to represent that the command requires *at least* 2 choices to pick from.
One of the alternatives (suggested by Danny) would be:
`<first> <second> [others...]`

TrustyJAID:
While I don't disagree I believe this is the better option for a few reasons. First of all the requirement of 2 is already described in the help string and the only place the usage appears is in the help command. Second, in terms of educating users on what the usage actually means this is provides a good stepping stone that more than one option is available and there is only one type of argument. Or in other words this teaches that things like `<arg...>` means more than one option is possible while the help string aids this command by stating that at least 2 are required and it errors on fewer than 2 anyways.

Drapersniper (dismissed):
LGTM, I made a typo fix and fixed the line break in addition to updating the General docs with your new docstrings. Thanks for contributing!

Drapersniper (dismissed):
Jack Approves my own changes.

Drapersniper (approved):
THIS IS IT!

Jackenmen:
If this ever comes back again, I'm personally going to make this @Flame442's problem 😃

PR #5017, opened 2021-04-27
Added owner check for audiostats

benno1237:


redbot/cogs/audio/core/commands/miscellaneous.py
```diff
@@ -40,6 +40,7 @@ async def command_sing(self, ctx: commands.Context):
 
     @commands.command(name="audiostats")
     @commands.guild_only()
+    @checks.is_owner()
```
Drapersniper:
```suggestion
    @commands.is_owner()
```
Use the check from `commands` namespace, thanks

Drapersniper (approved):

aikaterna (approved):

PR #5018, opened 2021-04-30
[Audio] Disallow all users from modifying Playlist 42069.

Drapersniper:
As the title states, users should not manually edit this playlist, specially delete it, even though it comes back on next cog reload/bot restart it may leave the bot in a unsupported state as the presence of this playlist is mandatory, if autoplay is expected to work.

### DIFF
--- a/redbot/cogs/audio/core/abc.py
+++ b/redbot/cogs/audio/core/abc.py
@@ -322,7 +322,13 @@ class MixinMeta(ABC):
 
     @abstractmethod
     async def can_manage_playlist(
-        self, scope: str, playlist: "Playlist", ctx: commands.Context, user, guild
+        self,
+        scope: str,
+        playlist: "Playlist",
+        ctx: commands.Context,
+        user,
+        guild,
+        bypass: bool = False,
     ) -> bool:
         raise NotImplementedError()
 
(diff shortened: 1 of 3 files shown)


Jackenmen (changes requested):
It seems unnecessary to pass `bypass=False` everywhere considering it's the default.

Drapersniper:
> It seems unnecessary to pass `bypass=False` everywhere considering it's the default.

You've missed https://github.com/Cog-Creators/Red-DiscordBot/pull/5018/files#diff-bd87e54068100b18833ef1210198b0be47138aa7f667fca5876412c625399e88R1694


and Im dumb...

Jackenmen (changes requested):
I see that you changed `playlist.id` to `getattr(playlist, "id", 0)`. From what I can see `Playlist` object will always have the `id` attr, which suggests that the `playlist` variable might not be a `Playlist` object. If that can indeed happen, then the type hints should be updated appropriately.


Also this change with `getattr()` doesn't really seem that related to this PR's real purpose which is fixing the stuff with playlist 42069...

Drapersniper:
It can be a SimpleNamesace when mocked.

Jackenmen (approved):
The error message is IMO not ideal but it's the simplest way to fix this so I think it's good enough until proven otherwise.
image

Drapersniper:
I've filed #5052 to track it.

PR #5019, opened 2021-05-02
[Downloader] Fix `[p]cog uninstall` examples

Obi-Wan3:
Previously, the examples in the help command for `[p]cog uninstall` showed that the repository name was to be inputted, inconsistent with the actual parameters for the command. They have been fixed so users will not be confused when following the examples.

### DIFF
--- a/docs/cog_guides/downloader.rst
+++ b/docs/cog_guides/downloader.rst
@@ -235,8 +235,8 @@ You may only uninstall cogs which were previously installed
 by Downloader.
 
 Examples:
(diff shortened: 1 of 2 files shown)


Jackenmen (changes requested):
Same change needs to be done in `docs/cog_guides/downloader.rst` and then this is good to go, thanks!

Obi-Wan3:
@jack1142 Fixed the cog guide as well 👍

Jackenmen (approved):
Thanks!

PR #5022, opened 2021-05-03
Prefer home directory even for system users if it exists

Jackenmen:
Useful especially when you create system users with a home directory (`useradd -rm`) so that you don't have to use the shared directory for configuration. Might also benefit people on systems like CentOS that I believe start assigning the non-system UIDs from 500 rather than 1000.

This is written in a way that prefers the home directory (if it exists) *unless* the shared directory was already used before by Red (which can be easily detected by checking whether `config.json` file exists.

### DIFF
--- a/redbot/core/data_manager.py
+++ b/redbot/core/data_manager.py
@@ -36,13 +36,21 @@ basic_config_default: Dict[str, Any] = {
     "CORE_PATH_APPEND": "core",
 }
 
-config_dir = None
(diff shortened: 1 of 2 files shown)


Dav-Git (dismissed):
LGTM

madebylydia (commented):
LGTM!

TrustyJAID (approved):
This appears to be working as intended.

PR #5024, opened 2021-05-03
[Downloader] Check author attribute as a fallback for `[p]findcog` author(s)

Kreusada:
Check if the cog class has an `__author__` attribute as a fallback to when a cog is not installed via downloader.
```py
author_attr = getattr(cog.__class__, "__author__", None)
if author_attr:
    if isinstance(author_attr, (list, tuple)):
        made_by = humanize_list(author_attr)
    elif isinstance(author_attr, str):
        made_by = author_attr
```
IMO it's worth having in there, especially for private cog owners who would still like to display their name in the `[p]findcog` command.

This **has** been tested, can confirm that it works™️. If the attribute is not a list, tuple or str, or if the attribute does not exist, it is ignored and falls back to "Unknown".

From the following cog, `[p]findcog` will show the author's in a list, humanized.

```py

class Test(commands.Cog):

    __author__ = ["Kreusada", "Red"]

    @commands.command()
    async def test(self, ctx):
        await ctx.send("Hello world!")
```

list
The same applies if the attribute was a string (see files changed). 
Now, if we change this attribute to an unsupported type, it will still show Unknown.

```py
__author__ = 123
```

unsupported

It's only going to show if the type is a tuple, string or list. Anyhow, I'm not sure why i'm explaining this so much 😄

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -1724,14 +1724,12 @@ class Downloader(commands.Cog):
                 made_by = _("Unknown")
                 repo_url = _("None - this cog wasn't installed via downloader")
                 cog_name = cog.__class__.__name__
-
             author_attr = getattr(cog.__class__, "__author__", None)
             if author_attr:
                 if isinstance(author_attr, (list, tuple)):
                     made_by = humanize_list(author_attr)
                 elif isinstance(author_attr, str):
                     made_by = author_attr
-                    
         else:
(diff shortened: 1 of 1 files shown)


Drapersniper:
What if like a significante quantity of modules, the author is added at a module level?

Flame442:
`[p]findcog` is a command from the downloader cog, so whether or not the command should support cogs not installed through it should be discussed first.

Kreusada:
Yeah ok good point, thanks

PR #5029, opened 2021-05-07
Fix a typo in Track Stuck error

stellariumm:
When it came to Track Stuck error I noticed a small typo in word unexpected which I went and wanted to fix in lavalink.py


Drapersniper (approved):
Thanks for the fix!

aikaterna (approved):

PR #5041, opened 2021-05-14
Use ctx.send_interactive for SyntaxErrors in Dev cog

Jackenmen:
# Bugfix request

Fixes an issue where the SyntaxErrors (and any exceptions in case of `[p]debug`) were not sent interactively causing the command to error when the content is over 2k chars.


Drapersniper (approved):
LGTM Thanks jako

PR #5011, opened 2021-04-23
Audio Update - RLL 0.8.1

Drapersniper:
Control RLL logs verbosity with `--debug` flag
Bump RLL dep to 0.8.1
Make certain audio logs more verbose to improve debuggability 
LOOOADS of cleanup
just internal stuff, 
ensure the error fixed in #5042 can never happen
Improve logging audio.

PR #5042, opened 2021-05-14
[Audio] Fix error when player does not have connection time saved

Ryan5453:
Some RLL players won't have the connection times saved, so instead of erroring the command, it'll now just show "Unknown" for how long the player has been active.


Drapersniper:
Hmmm I would like to know why they have no connect time instead of monkey patching it here. 

I'll have a look into it today.

Drapersniper:
I've updated #5011 to ensure this never happens.

Drapersniper:
This will be made obsolete by #5046,

I've added you as cocontributor there, I will close this one so it isn't accidently merged, thanks for spotting this one and making a PR to fix it!

PR #5043, opened 2021-05-15
Add per channel command disable

PredaaA:
This PR introduce the ability to disable commands, but also in channels.
~~(Not ready for review yet)~~


PredaaA:
Actually... it *is* ready for review, but I need to do some more testing.

PR #5045, opened 2021-05-17
Cleanup `[p]volume` and fix infinite retrying in autoplay

Drapersniper:
- cleanup of [p]volume
- a fix for infinite retrying

What to test:
 - `[p]volume` shows the correct percentage when being set
 - `[p]volume` allows volumes >= 0
 - `[p]volume` correctly persist on a new song / cog reload

### DIFF
--- a/redbot/cogs/audio/core/commands/controller.py
+++ b/redbot/cogs/audio/core/commands/controller.py
@@ -703,7 +703,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
     @commands.guild_only()
     @commands.bot_has_permissions(embed_links=True)
     async def command_volume(self, ctx: commands.Context, vol: int = None):
-        """Set the volume in percentages greater or equal to 0%."""
+        """Set the volume, 1% - 150%."""
         dj_enabled = self._dj_status_cache.setdefault(
             ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
         )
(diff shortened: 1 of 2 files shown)


redbot/cogs/audio/core/commands/controller.py
```diff
@@ -733,25 +733,15 @@ async def command_volume(self, ctx: commands.Context, vol: int = None):
...
-                player.store("channel", ctx.channel.id)
-                player.store("guild", ctx.guild.id)
-        else:
-            await self.config.guild(ctx.guild).volume.set(vol)
-            if self._player_check(ctx):
-                player = lavalink.get_player(ctx.guild.id)
-                await player.set_volume(vol)
-                player.store("channel", ctx.channel.id)
-                player.store("guild", ctx.guild.id)
+        vol = max(0, vol)
```
Jackenmen: I suppose that volume over 150 should still be disallowed here?
```suggestion
        vol = max(0, min(vol, 150))
```

Jackenmen (changes requested):

redbot/cogs/audio/core/commands/controller.py
Drapersniper: This is an old restriction from when this was the cap. 

The volume can go above 150% and has been able to for a very long time, this just removed the artificial restriction.

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -703,14 +703,14 @@ async def command_summon(self, ctx: commands.Context):
     @commands.guild_only()
     @commands.bot_has_permissions(embed_links=True)
     async def command_volume(self, ctx: commands.Context, vol: int = None):
-        """Set the volume, 1% - 150%."""
+        """Set the volume in percentages greater or equal to 0%."""
```
Drapersniper:
```suggestion
        """Set the volume, 1% - 150%."""
```

Jackenmen (approved):
Looks good

PR #5046, opened 2021-05-17
Fix crash in [p]audiostats by using the new connected_at attr

Drapersniper:
This PR depends on bump from Part 1 + the new `Player.guild` attribute

- [player.store] channel -> notify_channel
- [player.store] removed guild
- [player.store] removed connect
- changes in [p]audiostats

Co-authored-by: alec <50505980+aleclol@users.noreply.github.com>

What to test:
 - `[p]audiostats` shows the correct join time.
 - `[p]audiostats` doesn't crash
 - `[p]audiostats` show the correct join time for a player that that wasnt created by audio (sfx cog/eval)

### DIFF
--- a/redbot/cogs/audio/core/commands/miscellaneous.py
+++ b/redbot/cogs/audio/core/commands/miscellaneous.py
@@ -53,7 +53,7 @@ class MiscellaneousCommands(MixinMeta, metaclass=CompositeMetaClass):
                 int(
                     (datetime.datetime.now(datetime.timezone.utc) - p.connected_at).total_seconds()
                 )
-            )
+            ) or "0s"
             try:
                 if not p.current:
                     raise AttributeError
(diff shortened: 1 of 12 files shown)


aikaterna:
Making a note here to refer to the changes for the notify channel player.store key that is present in #4960 so it doesn't get forgotten about.

redbot/cogs/audio/core/commands/miscellaneous.py
```diff
@@ -45,21 +45,22 @@ async def command_sing(self, ctx: commands.Context):
...
         msg = ""
-        async for p in AsyncIter(lavalink.all_players()):
-            connect_start = p.fetch("connect")
+        async for p in AsyncIter(lavalink.all_connected_players()):
             connect_dur = self.get_time_string(
-                int((datetime.datetime.utcnow() - connect_start).total_seconds())
+                int(
+                    (datetime.datetime.now(datetime.timezone.utc) - p.connected_at).total_seconds()
+                )
             )
```
Jackenmen, line 61: Perhaps we could put a default here in case not a full second passed yet?
```suggestion
            ) or "0s"
```
image

Optional cause it's not something that this PR causes.

Drapersniper: :aha: how fast did you run the command lmfao

Jackenmen (dismissed):
> how fast did you run the command lmfao

*fast*

Jackenmen (approved):
Seems like I approved *too fast* as well...

PR #5048, opened 2021-05-17
[Docs] Update alias userdoc to reflect updated help menu layout

Kreusada:
Update alias userdoc for updates from #3040 - aliases are no longer represented in the old way (`[colour|color]`), so the getting started guide should be updated accordingly


Twentysix26 (approved):
Ty

PR #5051, opened 2021-05-19
Fix resume for auto play which was broken in RLL 0.8.0 update.

Drapersniper:
Autoresume is currently partially broken, but it wasn't caught due to `[p]audioset restart` being borked.

to test this fix to the following:

- `[p]autoplay` 
- give it 10s 
- `[p]audioset restart`

Expected behaviour is that it should restore the player.

Things to keep note of:
1- running `[p]disconnect` 
2- running `[p]stop` 
3- kicking the bot from vc.

The 3 points about should not resume when you run `[p]reload audio` or `[p]audioset restart` 

1. Moving the bot to another VC.
2. Running `[p]pause`
3. Running `[p]summon` 

These 3 should resume when you run `[p]reload audio`  or `[p]audioset restart`

### DIFF
--- a/redbot/cogs/audio/core/events/lavalink.py
+++ b/redbot/cogs/audio/core/events/lavalink.py
@@ -54,7 +54,6 @@ class LavalinkEvents(MixinMeta, metaclass=CompositeMetaClass):
         disconnect = guild_data["disconnect"]
         if event_type == lavalink.LavalinkEvents.FORCED_DISCONNECT:
             self.bot.dispatch("red_audio_audio_disconnect", guild)
-            await self.config.guild_from_id(guild_id=guild_id).currently_auto_playing_in.set([])
             self._ll_guild_updates.discard(guild.id)
             return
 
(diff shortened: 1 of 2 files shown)


Drapersniper:
Added a few more things that should be kept in consideration for when testing this PR (In case anyone wants to help test it)

PredaaA (approved):
LGTM thank you for this fix!

rtemelcea:
Disconnects triggered by empty disconnect timeout don't terminate correctly Lavalink players. When the discord bot leaves the channel, Lavalink receives the following: "stop" command, then "destroy" command, then "play" command with the following attributes {"op": "play", "guildId": "(guildid)", "track": "(trackid)", "noReplace": false, "startTime": "0", "pause": false}. Also the bot send the message "Auto Play Started".

This causes the following:
-When [p]restart is issued, after the bot starts, connects to the last voice channel joined and sends a message in the last text channel with "Auto Play started.", then leaves. (This is consistent behavior with Auto Play enabled after the queue ends and empty disconnect timer)
-When [p]audioset restart, after the bot starts, connects to the last voice channel joined and sends a message in the last text channel with "Auto Play started.", then leaves. (This is consistent behavior with Auto Play enabled after the queue ends and empty disconnect timer)

On the bright side, I can restart Lavalink without the bot randomly doing the behavior above :) .

aikaterna:
Can you open a new issue instead of tacking this onto something that's closed? You can reference this if you feel like you need to link to it.

rtemelcea:
My bad. I had the impression that this is still open.

PR #5056, opened 2021-05-19
Bump discord.py to 1.7.2

Jackenmen:
d.py 1.7.2 contains just a bunch of bug fixes, none of which concern the commands extension.


Drapersniper (approved):
V4 When?

PR #5063, opened 2021-05-19
Ensure nothing initializes colorama when it isn't needed

Jackenmen:
# Bugfix request

Before this change, `apsw` and `tqdm` called `colorama.init()` causing the colors to break when using Windows Terminal.

#### Anything we need to know about this fix?
The reviewer should ensure that the colors (and terminal output in general) work properly both on Windows Terminal and on the Command Prompt (cmd.exe).
Might also be worth checking that the tqdm progress bar works properly on both as well which you can easily check by removing the Audio's jar and loading Audio after.

### DIFF
--- a/redbot/__init__.py
+++ b/redbot/__init__.py
@@ -191,6 +191,29 @@ def _update_event_loop_policy():
             _asyncio.set_event_loop_policy(_uvloop.EventLoopPolicy())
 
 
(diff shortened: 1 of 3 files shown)


Flame442 (approved):
Confirmed working on CMD and Windows Terminal

PR #5064, opened 2021-05-19
Use `rich.progress` instead of `tqdm`

Jackenmen:
# Enhancement request

#### Describe the enhancement

- Use `rich.progress` instead of `tqdm`
- Remove `tqdm` from deps

I tested it (yes, including the `redbot-setup convert`)!

- [ ] Yes
- [x] No


Kowlin (approved):
Looks great, thanks 👀

PR #5065, opened 2021-05-20
Text tells you what to put in info.json.

Apoo711:
Pull request adds paragraph about what you should add in info.json, also adds optional next to info.json.

### DIFF
--- a/docs/guide_cog_creation.rst
+++ b/docs/guide_cog_creation.rst
@@ -105,8 +105,11 @@ Open :code:`__init__.py`. In that file, place the following:
     def setup(bot):
         bot.add_cog(MyCog(bot))
         
-Finally, open :code:`info_json`(If you made it). In that file, place the following:
+Finally, open :code:`info_json` (If you made it). In that file, place the following:
         
+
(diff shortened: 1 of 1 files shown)


Apoo711:
Updated so code block actually works.

Drapersniper:
The keys for the info.json and information about it can be found on this section of the docs.
https://docs.discord.red/en/stable/guide_publish_cogs.html#publishing-cogs-for-red-v3

I do not believe this PR is necessary, as the info.json is only needed if you plan to public your cog.

Apoo711:
But then it shouldn't be mentioned in creating cogs, shouldn't it just be mentioned in https://docs.discord.red/en/stable/guide_publish_cogs.html#publishing-cogs-for-red-v3.

Drapersniper:
I disagree with you; perhaps it should link to the info.json section in our publishing docs sure, but it should still be mentioned as creating the info.json although not necessary for non published cogs its a good thing to get in the habbit of doing to for the benefit of users.

Apoo711:
Ahh I understand!

PR #5066, opened 2021-05-20
Fix the bug with Twitch display name being set as Twitch login

### LINKED ISSUE #5050: [Streams] twitch channel no longer exists and removed

3.4.9

### Cog name

Streams

### Command name

streamalert twitch

After i added the stream, it should announce everytime a stream is up

### What actually happened?

the stream was recognized as no longer exists and removed from list 

1. [p]streamalert twitch salmoninthebox
2. wait and check logs i suppose

### Anything else?

i checked all other streamalerts from other streams in other servers and they seem to be fine

Jackenmen:
# Bugfix request

Fixes #5050

Adds `{stream.display_name}` replacement field that can be used in stream message to have it be replaced with streamer's display name.
On Twitch, this can be different from streamer's login - either having different casing OR completely different when the streamer uses the localized display name (available for Chinese, Japanese, and Korean streamers).
On all other platforms this is the same as `{stream}`, at least for now.

### DIFF
--- a/docs/cog_guides/streams.rst
+++ b/docs/cog_guides/streams.rst
@@ -200,7 +200,10 @@ Use ``{mention}`` in the message to insert the selected mentions.
 
 Use ``{stream}`` in the message to insert the channel or user name.
 
-For example: ``[p]streamset message mention {mention}, {stream} is live!``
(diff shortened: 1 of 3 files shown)


Drapersniper (approved):
Finally got around to this one.

PR #5068, opened 2021-05-20
Reduce buffer set by JDA-NAS

Drapersniper:
Lets update all we need to update, so we don't have to require users to update now and then later


aikaterna:
The "jdanas" key is not ready to be put in Red yet. I'll accept the 300 ms timing but not the other key currently.

PR #5074, opened 2021-05-24
Update logger names to make them more intuitive

Drapersniper:
subjectively improve the logger names to make them more explicit.

I was going through red and notice that we are extremely inconsistent on how we name our loggers, we should improve the consistency here.

I took the liberty to tweak a somewhat used standard, `red.(core.cogs|core).<cog_name.camelcase()>[and a further breakdown if desidered here]

This allows us to control the level of all core cogs, all 3rd part cogs (if they adopt this schema) separately as well as making a lot clearer where something is getting logged from. for example `red.bank` is this the api or the cog? `red.core.bank` and `red.core.cogs.bank` make that uncertainty non existant.

### DIFF
--- a/redbot/setup.py
+++ b/redbot/setup.py
@@ -21,7 +21,7 @@ from redbot.core.utils._internal_utils import safe_delete, create_backup as red_
 from redbot.core import config, data_manager, drivers
 from redbot.core.drivers import BackendType, IdentifierData
 
-conversion_log = logging.getLogger("red.converter")
+conversion_log = logging.getLogger("red.driver.converter")
 
 config_dir = None
 appdir = appdirs.AppDirs("Red-DiscordBot")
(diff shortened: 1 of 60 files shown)


redbot/setup.py
```diff
@@ -21,7 +21,7 @@
 from redbot.core import config, data_manager, drivers
 from redbot.core.drivers import BackendType, IdentifierData
 
-conversion_log = logging.getLogger("red.converter")
+conversion_log = logging.getLogger("red.driver.converter")
```
```suggestion
conversion_log = logging.getLogger("red.core.driver.converter")
```

Jackenmen:
If we want to switch, we should IMO go for the use of `__name__` or a part of it for the specific modules.

Drapersniper:
name would work, should i make the replacement for that? the problem is that would cause RLL logs to become uncoupled.

PR #5078, opened 2021-05-25
Lower RTD build time

Jackenmen:
# Enhancement request

#### Describe the enhancement

For now this just remove PDF format from the build but I might also look into other ways of improving the speed of the build.


Drapersniper (approved):
After discussing change with Jack on discord LGTM

PR #5079, opened 2021-05-25
Update gitignore with more of Windows, Mac, Sublime, and PyCharm

Drapersniper:
Since i have these untracked locally sharing... I know jack is going to bitch.

1 - I do work on Windows and Mac (Every so often)
2 - I Use Sublime and Pycharm
3 - We use VirtualEnv


.gitignore
```diff
@@ -143,3 +143,166 @@ ENV/
...
+# Virtualenv
+# http://iamzed.com/2009/05/07/a-primer-on-virtualenv/
+[Bb]in
+[Ii]nclude
+[Ll]ib
+[Ll]ib64
+[Ll]ocal
+[Ss]cripts
+pyvenv.cfg
+pip-selfcheck.json
```
Jackenmen: virtualenv is fine but why aren't you creating the virtualenv in `venv` or `.venv` directory?

.gitignore
```diff
@@ -143,3 +143,166 @@ ENV/
...
+*.cab
+*.msi
+*.msix
+*.msm
+*.msp
+
+# Windows shortcuts
+*.lnk
+
+### JetBrains template
```
fixator10: This is already handled by line 15 (`.idea`)

.gitignore
```diff
@@ -143,3 +143,166 @@ ENV/
...
+
+# Cursive Clojure plugin
+
+# Crashlytics plugin (for Android Studio and IntelliJ)
+
+# Editor-based Rest Client
+.idea/httpRequests
+
+# Android studio 3.1+ serialized cache file
+.idea/caches/build_file_checksums.ser
```
Jackenmen: we already ignore the whole `.idea` directory

Jackenmen (changes requested):

.gitignore
Drapersniper: I have since, pycharms default is weird though.

image

Jackenmen: This is the default I see when creating the project
image

.gitignore
```diff
@@ -143,3 +143,97 @@ ENV/
...
+
+# Windows shortcuts
+*.lnk
+
+### JetBrains template
+# Covers JetBrains IDEs: IntelliJ, RubyMine, PhpStorm, AppCode, PyCharm, CLion, Android Studio, WebStorm and Rider
+# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+
+# CMake
+cmake-build-*/
```
Jackenmen: The comments don't really refer to anything since JetBrains config is gone and we don't need cmake ignore either.

Jackenmen (changes requested):
Everything except for the one review comment below seems fine

Jackenmen (approved):

PR #5080, opened 2021-05-25
Docstring Fix

Lifeismana:
Some errors i came across while translating.
In case i come across other errors while translating, it might be better to not merge this pr yet.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -2116,7 +2116,7 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         """
         Adds an admin role for this guild.
 
-        Admins have all the same access and Mods, plus additional admin level commands like:
+        Admins have the same access as Mods, plus additional admin level commands like:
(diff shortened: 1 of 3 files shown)


redbot/core/core_commands.py
```diff
@@ -2116,7 +2116,7 @@ async def addadminrole(self, ctx: commands.Context, *, role: discord.Role):
         """
         Adds an admin role for this guild.
 
-        Admins have all the same access and Mods, plus additional admin level commands like:
+        Admins have the same access as Mods, plus additional admin level commands like:
```
Jackenmen, line 2119: This one needs to also be updated in `docs/cog_guides/core.rst` file

Jackenmen (changes requested):

redbot/core/core_commands.py
Lifeismana: Done

Jackenmen (approved):
Thanks for the contribution!

PR #5082, opened 2021-05-26
[Downloader] Allow removal of multiple repos

### LINKED ISSUE #4765: Add possiblity to remove multiple repos
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality

Currently you cannot remove multiple repos, now you gotta remove one by one which is much of a case annoying. 
Maybe try to add something so you can remove more than one at once like you can with cogs?

Vexed01:
This will make removing repos similar to removing cogs.

I've chosen to use the same format for multi repo removal as in the `cog uninstall` command for continuity.

Closes #4765

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -605,7 +605,7 @@ class Downloader(commands.Cog):
                 if len(repos) > 1
                 else _("Successfully deleted the repo: ")
             )
-            + humanize_list([i.name for i in set(repos)])
+            + humanize_list([inline(i.name) for i in set(repos)])
         )
 
     @repo.command(name="list")
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -583,23 +583,30 @@ class Downloader(commands.Cog):
             if repo.install_msg:
                 await ctx.send(repo.install_msg.replace("[p]", ctx.clean_prefix))
 
-    @repo.command(name="delete", aliases=["remove", "del"])
-    async def _repo_del(self, ctx: commands.Context, repo: Repo) -> None:
+    @repo.command(name="delete", aliases=["remove", "del"], require_var_positional=True)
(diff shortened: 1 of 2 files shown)


redbot/cogs/downloader/downloader.py
```diff
@@ -583,23 +583,30 @@ async def _repo_add(
...
-            _("The repo `{repo.name}` has been deleted successfully.").format(repo=repo)
-        )
+        if len(repos) == 1:
+            await ctx.send(
+                _("The repo `{repo.name}` has been deleted successfully.").format(repo=repo)
+            )
+        else:
+            await ctx.send(
+                _("Successfully deleted repos: ") + humanize_list([i.name for i in set(repos)])
+            )
```
Jackenmen: I'm not sure whether I managed to be 100% consistent but I tried my best to have both the singular and plural forms look similar so that people can parse through it faster knowing the consistent look the message has.
Here's what I would suggest to use instead:
```suggestion
        await ctx.send(
            (
                _("Successfully deleted repos: ")
                if len(repos) > 1
                else _("Successfully deleted the repo: ")
            ) + humanize_list([i.name for i in set(repos)])
        )
```

Jackenmen (changes requested):

redbot/cogs/downloader/downloader.py
```diff
@@ -583,22 +583,29 @@ async def _repo_add(
...
+            await self._repo_manager.delete_repo(repo.name)
 
         await ctx.send(
-            _("The repo `{repo.name}` has been deleted successfully.").format(repo=repo)
+            (
+                _("Successfully deleted repos: ")
+                if len(repos) > 1
+                else _("Successfully deleted the repo: ")
+            )
+            + humanize_list([i.name for i in set(repos)])
```
Jackenmen:
```suggestion
            + humanize_list([inline(i.name) for i in set(repos)])
```

Jackenmen (changes requested):
One more thing and this is ready to go

Jackenmen (approved):
Looks great now, thanks 👍

PR #5083, opened 2021-05-27
Update Dev's code block regex to support `python` language

benno1237:
"python" as syntax in codeblocks wasn`t recognized. "py" however was. this PR fixes that

### DIFF
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -28,7 +28,7 @@ https://github.com/Rapptz/RoboDanny/blob/master/cogs/repl.py
 
 _ = Translator("Dev", __file__)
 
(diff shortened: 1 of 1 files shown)


redbot/core/dev_commands.py
```diff
@@ -28,7 +28,7 @@
 
 _ = Translator("Dev", __file__)
 
-START_CODE_BLOCK_RE = re.compile(r"^((```py)(?=\s)|(```))")
+START_CODE_BLOCK_RE = re.compile(r"^((```py)(?=\s)|(```python)(?=\s)|(```))")
```
Jackenmen: This can be simplified:
```suggestion
START_CODE_BLOCK_RE = re.compile(r"^((```py(thon)?)(?=\s)|(```))")
```

Jackenmen (changes requested):

Jackenmen (approved):
Whoopee, it works!

PR #5107, opened 2021-05-31
[Dev] Use `tick()` when debugs are successful

Kreusada:
Add `ctx.tick()` to `[p]debug` when the debug does not raise an exception. 
There are a few debugs you can run, such as `[p]debug "\t"`, and `[p]debug ""`, which do not respond with anything - leaving questions in the air as to whether something isn't working.

It will also be nice to have this tick, seeing as eval has it too. Although this was a very small change, it has still been tested, and it works.


Kowlin (approved):
Plain and simple, thanks!

PR #5109, opened 2021-06-01
[Downloader] Suppress NotFound errors in `[p]cog update` command

Kreusada:
To reproduce this issue:

1. Run `[p]cog update` and wait for the bot to successfully update the cogs.
2. Delete the message that says "Would you like to reload the updated cogs?". This is the message tied to the predicate event.
3. Delete this message within 30 seconds. Then the missing message will try to be deleted, raising a `discord.NotFound` exception.


Jackenmen (approved):
Thanks

PR #5114, opened 2021-06-02
[Core] Add toggle for the applications.commands invite scope

laggron42:
Adds the `[p]inviteset commandscope` toggle for enabling or disabling the commands scope, given with the URL returned by `[p]invite` .

I feel like the wording is wrong for the docstring, feel free to suggest something better.

----

Requested by Zephyrkul on the server.


Kowlin (dismissed):
Thank you very much.

Kowlin (approved):

PR #5116, opened 2021-06-03
[Mutes] Update new channels

### LINKED ISSUE #5111: Muted role for every new text channel
### Type of feature request

Cog, Other

in my opinion, the bot should add the Muted role with its permissions denied for each new channel created. In such a way as not to waste time for the admin of the server to add the role for each individual channel.

### Anything else?

_No response_

laggron42:
This adds a toggle under `[p]muteset autoupdate` for updating the permissions of new channels for the mute role.

Resolves #5111

### DIFF
--- a/docs/cog_guides/mutes.rst
+++ b/docs/cog_guides/mutes.rst
@@ -119,6 +119,36 @@ muteset
 
 Mute settings.
 
+.. _mutes-command-muteset-autoupdate:
+
+""""""""""""""""""
+muteset autoupdate
+""""""""""""""""""
+
+.. note:: |admin-lock|
+
+**Syntax**
+
+.. code-block:: none
+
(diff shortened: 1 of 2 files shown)


TrustyJAID:
I don't know if this is appropriate to handle considering channels are often auto-synced with channel categories when created and this can potentially break that sync.

Jackenmen:
Perhaps we could only handle it if it's not already in sync with the category?

TrustyJAID:
The thing is if it's not synced then the server mod who created the channel will potentially not want this messing with things and it should be fully controlled by the mod creating the channel. This setting could end up with a lot of confusion but checking for sync first as long as the cache is up-to-date in time for the bot to know about when the channel is created should suffice.

PR #5117, opened 2021-06-03
[Core] Limit server prefix length

### LINKED ISSUE #5091: Limit Server Prefix length
### Type of feature request

Other

Impose a length limit on server prefix (and possible raise a warning on a global prefix)

### Anything else?

_No response_

laggron42:
This sets a hard limit of 20 characters for server prefixes. Any prefix longer than that will cancel the command.

Global prefixes can bypass this limit, although this will raise a warning and ask for confirm.

Resolves #5091

### DIFF
--- a/docs/cog_guides/core.rst
+++ b/docs/cog_guides/core.rst
@@ -3174,6 +3174,8 @@ Sets Red's server prefix(es).
 
     This is not additive. It will replace all current server prefixes.
 
+    You cannot have a prefix with more than 20 characters.
+
 **Examples:**
     - ``[p]set serverprefix !``
     - ``[p]set serverprefix "! "`` - Quotes are needed to use spaces in prefixes.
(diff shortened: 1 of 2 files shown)


phenom4n4n:
With this implementation the limit can be ignored when using 3rd party cogs with the `Red.set_prefixes` method. A runtime error could be raised by the prefix cache manager itself when setting guild prefixes if the prefix length is above 20 which would prevent any way to bypass this limit.

Kreusada:
It may also be nice to restrict or at least warn users if they decide to set their prefix as nothing (`[p]set prefix/serverprefix ""`). Reason being is that users who are normally talking can invoke commands without intention which is something they didn't opt into, and could be dangerous dependent on which command they unintentionally interacted with. Maybe this should be implemented with the same theme as this PR, disabled completely for global but then warned upon for servers. Because some bot owners may still enjoy using no prefix in a private server, so this option should still remain partially available to them.

Kowlin (approved):
This is stupid that this has to be a thing... But here we are... Thanks Laggron.

PR #5120, opened 2021-06-03
[Warnings] Warn mod if 0 or less points are given

### LINKED ISSUE #5119: `[p]warn` accepts non-positive int points values

3.4.9

### Cog name

Warnings

### Command name

warn

An error when trying to run `[p]warn` with a negative or 0 number of points for a custom reason.

### What actually happened?

It blindly accepts the passed points.

1. Enable custom reasons
2. `[p]warn <user> -1 test`
3. `[p]warnings <user>`

### Anything else?

_No response_

Just-Jojo:
This adds a warning message for when `allowcustomreasons` is enabled and a moderator applies 0 or less points

Fixes #5119


Flame442 (approved):
Thanks for making that fix!

Crossedfall:
Paaain. We actively used 0 point warnings. This also was not in the CL

PR #5121, opened 2021-06-06
Update deps, allow Python 3.9, drop Fedora 32

### LINKED ISSUE #4655: Arch Linux instructions no longer work as Arch ships with Python 3.9
Arch Linux uses bleeding edge technology so its python package already uses Python 3.9 that is currently not compatible with Red.
We could tell people to install the `python38` package from AUR (https://aur.archlinux.org/packages/python38) however, I don't think it's our job to tell Arch users *how* they can do it since Arch is meant to be a system for knowledgeable users and aside from the manual, recommended way, there are also plenty of tools that make it a bit easier for the user but are not endorsed by Arch Linux.

palmtree5:
Fixes #4655 

I've tested on Ubuntu and things seem to be working fine with commands in Discord, as well as the console output via `rich` (which is one of the updated dependencies). Some docs dependencies also updated; building the docs succeeded, albeit with a couple warnings about multiple targets for 'any' cross-reference. Tests pass as well, though with multiple instances of the following:

```py
DeprecationWarning: The explicit passing of coroutine objects to asyncio.wait() is deprecated since Python 3.8, and scheduled for removal in Python 3.11.
```

Not sure if we want to update docs here to point people towards 3.9 or not but that can be discussed. Obviously at some point before this gets merged, I will ensure we remove the mentions of 3.9 not being supported.

I am leaving `black` on version `20.8b1` for the time being as attempting to run a stylediff with black updated led to it wanting to change 17 files (as well as per discussion with jack).

Our `python_requires` has been updated to `>=3.8.1,<3.10` which will allow people to install Red on Python 3.9 without use of a pip flag.

Things still needing testing:

- [x] Windows
- [x] Install guides (@jack1142 should also update the tests at https://github.com/jack1142/Red-Install-Tests)


Jackenmen:
I suggest reusing my work for CI, tests, tox, and docs that's available here:
https://github.com/Cog-Creators/Red-DiscordBot/compare/V3/develop...jack1142:V3/py39

Jackenmen:
According to https://github.com/sphinx-doc/sphinx/commit/93cf1a57d916a1ff96c8e8a0356d0256e40489ac, I think that we should probably update our `extlinks` dictionary in `docs/conf.py`

palmtree5:
Will look after work today

Hbutlercapone (approved):

Jackenmen (approved):
This should be good to go now. I'll merge it a bit later, probably after I'll look through some of the remaining PRs.

PR #5125, opened 2021-06-09
Fix Filter - None issue

Stonedestroyer:
This can happen in cases where guild is None and it tries to compare a None object.


Flame442 (approved):
Thanks for the fix!

PR #5137, opened 2021-06-15
Update `[p]addpath` to use consume rest on the path argument

npc203:
[p]addpath doesn't take in spaces. This pr just does that by consuming rest so that noobs like me don't fail in adding path.


Jackenmen (approved):
Looks good to me!

PR #5156, opened 2021-06-24
Refined debuginfo

### LINKED ISSUE #4997: Add more information to `[p]debuginfo`

Core

Add information about things like:
- instance name
- more system information, i.e. RAM
- ??? (@Kowlin seems to have more ideas than me so he can add them here)

### Anything else?

I would highly suggest dropping the embed version entirely while doing these changes - the embed is getting big and it's not that easy to read already and it will only get worse. Also, we might eventually run out of fields as you can only put 20...

Kowlin:

Rewritten debug info to be pure embedless and just frankly... more useful!

Fixes #4997


Jackenmen (commented):
The changes look good to me. I've asked in #help-desk whether there's anything else that could be potentially useful to include here but if nothing is brought up, this PR will be good to go.

// I've done a bunch of code formatting changes, hope you don't mind 🙃

Jackenmen (commented):
Markdown code block breaks when a character like `_` is used.
As an example, it could appear in architecture name, user name, instance name, owner name, and any paths:
image

Considering that characters cannot be escaped in code blocks, I'm not sure if we should use any language in the code block.

Jackenmen:
Alternative 1 (no box lang):
image

Alternative 2 (box lang only for headers):
image

Jackenmen:
Alternative 3 (same as Alternative 2 but with more vertical spacing between sections):
image

It's good to keep in mind that these are likely to just be copy-pasted straight into #support so it needs to play properly with regular copy-paste.

Jackenmen (approved):
Looks awesome!

PR #5159, opened 2021-06-25
Rename RedBase to Red, remove the old Red

Jackenmen:
The separation of `RedBase` and `Red` is no longer needed (historically, this was used to aid with user bots usage) and so this PR joins them.

### DIFF
--- a/docs/changelog_3_3_0.rst
+++ b/docs/changelog_3_3_0.rst
@@ -253,7 +253,7 @@ Developer changelog
 Core Bot
 ********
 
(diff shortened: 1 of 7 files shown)


leetfin (dismissed):
"Heck, I would do it if it weren't for the the review requirement for all PRs" - Jack

Doesn't look like it'll explode to me!

Dav-Git (approved):
Looks like he didn't break anything that's not meant to be broken.

Flame442 (approved):
The docs seem to be working fine, and the bot launches.

PR #5163, opened 2021-06-28
[Mutes] Fix NotFound error when trying to delete message

Kreusada:

Fixes an issue when `discord.NotFound` is raised when the query is not found.


Jackenmen (approved):
Thanks

PR #5167, opened 2021-06-29
Improve release correctness and safety by using GH Environments

Jackenmen:
This PR does 2 things to improve release correctness and safety:
1. Make the release workflow require `Release` environment *before* pushing the release to PyPI:
2. A small summary of the release to be able to easily check that what is getting released is what we want to be released
    - This also adds a check that automatically compares the tag name to the version that is getting released to minimize the possibility of human error (though it does not rule out all the cases which is why the human review of the summary is needed)

Here's the current configuration of the `Release` environment:
image
It would probably be good to move the `PYPI_TOKEN` repository secret to this environment. I wasn't sure whether we should choose just admin core devs or all core devs but I chose admin core devs per the principle of least privileges. The reviewer can be the person that made the release so this is not a bottleneck and seems to be in line with our policies.

#### Test cases
- Correct release
https://github.com/jack1142/Red-DiscordBot/runs/2947260520?check_suite_focus=true
Summary: https://github.com/jack1142/Red-DiscordBot/actions/runs/984437295

- Tag name not corresponding to release version (`3.4.13` vs `3.5.0.dev1`):
https://github.com/jack1142/Red-DiscordBot/runs/2947181460?check_suite_focus=true
Summary: https://github.com/jack1142/Red-DiscordBot/actions/runs/984407534

- Branch the tag was based off being incorrect (`V3/improve_release_safety_test` vs `V3/improve_release_safety_test2`):
https://github.com/jack1142/Red-DiscordBot/runs/2947246411?check_suite_focus=true
Summary: https://github.com/jack1142/Red-DiscordBot/actions/runs/984431767

#### Example of deployment review
image

### DIFF
--- a/.github/workflows/publish_release.yml
+++ b/.github/workflows/publish_release.yml
@@ -5,8 +5,58 @@ on:
       - "*"
 
 jobs:
-  release_to_pypi:
+  release_information:
     if: github.repository == 'Cog-Creators/Red-DiscordBot'
+    name: GO HERE BEFORE APPROVING
+    runs-on: ubuntu-latest
+    steps:
+      # Checkout repository and install Python
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+
+      # Get version to release
+      - name: Get version to release
+        id: version_to_release
+        run: |
+          python .github/workflows/scripts/bump_version.py
+        env:
+          PYTHONPATH: ${{ github.workspace }}:${{ env.PYTHONPATH }}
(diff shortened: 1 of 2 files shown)


Jackenmen:
> Here's the current configuration of the Release environment:

As you can see on the screenshot above, the admin core devs team is not actually listed in the environment protection rules. For the time being, I set it to the core devs team as I'm not entirely sure why GitHub is not allowing me to set the admin core devs team as a required reviewer. Might have something to do with the fact that the team is hidden, but the UI does not give any indication of that.

Kowlin (approved):

PR #5172, opened 2021-07-03
[Trivia] Handle potential Discord errors in session

PredaaA:
This PR handle NotFound or Forbidden errors in Trivia's session wait_for_answer method.

### DIFF
--- a/redbot/cogs/trivia/session.py
+++ b/redbot/cogs/trivia/session.py
@@ -114,6 +114,8 @@ class TriviaSession:
             fut.result()
         except asyncio.CancelledError:
             pass
+        except (discord.NotFound, discord.Forbidden):
+            self.stop()
         except Exception as exc:
             LOG.error("A trivia session has encountered an error.\n", exc_info=exc)
             asyncio.create_task(
--- a/redbot/cogs/trivia/session.py
+++ b/redbot/cogs/trivia/session.py
@@ -1,5 +1,6 @@
 """Module to manage trivia sessions."""
 import asyncio
+import contextlib
 import time
 import random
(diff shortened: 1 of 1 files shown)


redbot/cogs/trivia/session.py
```diff
@@ -207,7 +208,8 @@ async def wait_for_answer(self, answers, delay: float, timeout: float):
             )
         except asyncio.TimeoutError:
             if time.time() - self._last_response >= timeout:
-                await self.ctx.send(_("Guys...? Well, I guess I'll stop then."))
+                with contextlib.suppress(discord.HTTPException):
+                    await self.ctx.send(_("Guys...? Well, I guess I'll stop then."))
```
Jackenmen: Could you send the traceback that you are getting here that isn't a Forbidden nor NotFound?

PredaaA: Yeah I don't know why I have set HTTPException here, just changed.

Jackenmen (commented):
see below

Jackenmen (changes requested):
Hmm, there are a lot of places where we send messages and might possibly run into such issue, perhaps it would be better to instead just add this to `_error_handler()` method?

PredaaA:
> Hmm, there are a lot of places where we send messages and might possibly run into such issue, perhaps it would be better to instead just add this to `_error_handler()` method?

Good point, thanks for pointing it out! That's something I didn't saw.
I've added those to `_error_handler` method on latest commit.

Jackenmen (approved):
Seems like this should just work 😏

PR #5173, opened 2021-07-03
[Mod] Check if guild is unavailable in tempban expirations

PredaaA:
Fixes `guild.me` being None errors when the guild is unavailable.
```py
AttributeError: 'NoneType' object has no attribute 'guild_permissions'
  File "mod/kickban.py", line 235, in tempban_expirations_task
    await self._check_tempban_expirations()
  File "mod/kickban.py", line 246, in _check_tempban_expirations
    if not guild.me.guild_permissions.ban_members:
```

### DIFF
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -243,7 +243,7 @@ class KickBanMixin(MixinMeta):
         async for guild_id, guild_data in AsyncIter(guilds_data.items(), steps=100):
             if not (guild := self.bot.get_guild(guild_id)):
                 continue
-            if not guild.me or not guild.me.guild_permissions.ban_members:
+            if guild.unavailable or not guild.me.guild_permissions.ban_members:
                 continue
             if await self.bot.cog_disabled_in_guild(self, guild):
                 continue
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -243,7 +243,7 @@ class KickBanMixin(MixinMeta):
         async for guild_id, guild_data in AsyncIter(guilds_data.items(), steps=100):
             if not (guild := self.bot.get_guild(guild_id)):
                 continue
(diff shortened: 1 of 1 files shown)


Jackenmen:
How often do you run into this exception? Would it be possible for you to test if this would be fixed if we were checking for `guild.unavailable` instead?

PredaaA:
> How often do you run into this exception? Would it be possible for you to test if this would be fixed if we were checking for `guild.unavailable` instead?

That's something I see at every boot. And sure I can try, but I think that `guild.unavailable` is False when booting up.

Jackenmen:
We might want to wait until ready too I guess...

I would appreciate it if you could test whether testing for `guild.unavailable` is enough to solve this.

Rapptz:
> That's something I see at every boot. And sure I can try, but I think that guild.unavailable is False when booting up.

No. It's `True`.

The gateway originally sends guilds with just an `id` and `unavailable: true` and later streams them in via `GUILD_CREATE`. This `GUILD_CREATE` will always have the self member (populating `Guild.me`) barring Discord bugs. Checking for `None` for `guild.me` is just a bandaid of another problem.

PredaaA:
> > That's something I see at every boot. And sure I can try, but I think that guild.unavailable is False when booting up.
> 
> No. It's `True`.
> 
> The gateway originally sends guilds with just an `id` and `unavailable: true` and later streams them in via `GUILD_CREATE`. This `GUILD_CREATE` will always have the self member (populating `Guild.me`) barring Discord bugs. Checking for `None` for `guild.me` is just a bandaid of another problem.

Thank you very much on the clarification about this, I always like to know how things are *actually* working.

I will then change to check `guild.unavailable` instead.

PredaaA:
Can confirm that checking for `guild.unavailable` fix this issue. I just updated the PR.

Jackenmen (approved):
LGTM

PR #5179, opened 2021-07-04
Update economy.py

kpopdev:


Flame442:
Hi there! We're not currently interested in adding this alias to the global command namespace. If you want this alias on your bot, the core `alias` cog can be used to add custom aliases to existing commands without modifying the code.
`[p]load alias`
`[p]alias add lb leaderboard`
If you want the alias to be applied to all servers, you can instead use
`[p]alias global add lb leaderboard`
Hopefully that helps you out.

PR #5184, opened 2021-07-06
Make no intents warning stand out

Vexed01:
Use Rich's console print to print the lack of privileged intents warning in red text to make it noticeable. This also has the effect of making the link cyan as well, making it even more noticeable and hopefully helping people click/copy-paste it
This stops the error from 'blending in'.

Top one is old, bottom is new.
image

Shamelessly stolen from #support, this is what the whole process looks like otherwise:
image

@kowlin might be interested in this

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -18,6 +18,7 @@ from pathlib import Path
 from typing import NoReturn
 
 import discord
+import rich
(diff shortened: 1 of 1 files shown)


Kowlin (approved):
Thank you.

PR #5185, opened 2021-07-08
[Downloader] Fix various formatting issues in cog update notices

Kreusada:

Another small formatting fix alike #5183. I have checked the notice strings to check for formatting issues and this appears to be the last one that needs changing.


Jackenmen (changes requested):
It appears there's the same problem in line 1444 and 1489

Kreusada:
Apologies, fixed

Jackenmen (approved):

PR #5186, opened 2021-07-13
[Audio] UX improvements to `[p]summon` command

ltzmax:
This pr adds an ✅ when you do `[p]summon`.

summon

Edit: It also now send a message when the bot and the command author are already in the same channel, before it was silently returning.


PredaaA (changes requested):
While we are at improving UX on this command, it could be nice to return a message in the case that the bot and the author are already in the same channel, currently it just returns silently. It's at line 670 (I can't select it on the review).

ltzmax:
> While we are at improving UX on this command, it could be nice to return a message in the case that the bot and the author are already in the same channel, currently it just returns silently. It's at line 670 (I can't select it on the review).

Done 👍

PredaaA (approved):
Thank you for this PR!

PR #5191, opened 2021-07-18
[Cleanup] Handle NotFound error of prompt in check_100_plus

PredaaA:


redbot/cogs/cleanup/cleanup.py
```diff
@@ -58,11 +59,9 @@ async def check_100_plus(ctx: commands.Context, number: int) -> bool:
         response = await ctx.bot.wait_for("message", check=MessagePredicate.same_context(ctx))
 
         if response.content.lower().startswith("y"):
-            await prompt.delete()
-            try:
+            with contextlib.suppress(discord.HTTPException):
+                await prompt.delete()
                 await response.delete()
```
Jackenmen: - the error in the first delete should not result in not doing the second delete 
- it should be enough to only suppress `NotFound` here as it's our own message
```suggestion
            with contextlib.suppress(discord.NotFound):
                await prompt.delete()
            with contextlib.suppress(discord.HTTPException):
                await response.delete()
```

Jackenmen (changes requested):

Jackenmen (approved):
LGTM

PR #5194, opened 2021-07-22
Make the indexing argument optional in leaderboard command

Kreusada:

Adds an Optional typehint to the "top" argument in the leaderboard command, allowing you to pass the global argument without having to provide the indexing argument.


bobloy:
This doesn't work if the user wishes to use `1` or `0` as their boolean values.

Kreusada:
> This doesn't work if the user wishes to use `1` or `0` as their boolean values.

You're right, this probably won't work then. Thanks

PR #5196, opened 2021-07-25
[Downloader] Fix a potential permission error when installing a package with pip

Lifeismana:
When installing a package with pip with the target option as we are doing here, some package are installing files in the cwd (which is the cwd of the redbot cmd), to make things clean (files being put in unintended folder) and make sure we have file perms on the folder, we should define the cwd on one of our folder, defining it on the lib folder works.

Example of a package that does this is `greenlet`, when putting it as a req on a cog, on cog install, it will create a file `include/site/python3.8/greenlet/greenlet.h` in the cwd of the redbot cmd. 
(Note: in that example, when the target option is not used, the file gets put in the venv folder, so the lib folder seems like the best place to put that file)


Lifeismana:
can't reproduce that issue anymore, so i'm gonna quietly close that pr
(this was probably resolved either by pip or by the lib, my bet is that pip resolved that)

PR #5198, opened 2021-07-26
Add test for presence of upper and lower bound on Python version

Jackenmen:
# Enhancement request

#### Describe the enhancement

Due to constant issues in support with Red being installed on a Python version that was not supported by any Red version, it is important that we have both an upper and lower bound set.
This PR adds a test that ensures that this is always the case. This test might have false-negatives if we were to use `~=` or `==` with a wildcard in the future, and not use the `>`/`>=` and `<`/`<=` comparators. It seems rather unlikely that these would be used so if it does happen, the test can just be updated to work with it.

- [ ] Yes
- [x] No

##### No upper-bound
https://github.com/Cog-Creators/Red-DiscordBot/runs/3162462606

##### No lower-bound
https://github.com/Cog-Creators/Red-DiscordBot/runs/3162482116

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,7 +33,7 @@ license_files =
 
 [options]
 packages = find_namespace:
-python_requires = <3.9
+python_requires = >=3.8.1,<3.9
 include_package_data = True
 install_requires =
     aiohttp==3.7.3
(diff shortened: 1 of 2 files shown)


hDmtP (commented):
LGTM

Kowlin (approved):

PR #5199, opened 2021-07-27
[Core] add `ctx.tick()` to `[p]invite`

Arman0334:

This adds a simple tick reaction on the `[p]invite` command. Previously it just sent the invite url to the author and returned.


Jackenmen (approved):
Thanks for the PR, looks good to me.

PR #5206, opened 2021-08-11
[Core] Add more APIs for allowlists and blocklists

### LINKED ISSUE #5171: further API support for the blacklist manager.
### Type of feature request

API functionality

Hello, I wanted to request API functionality for Red's private blacklist manager.

Either that or some public method(s) to allow cog creators access to manually add and/or remove users from the bot's blacklist. This could be useful if someone is intrigued to change the functionality of the inbuilt blacklist command (like me and jojo). This would allow us to access the blacklist and modify it as per our use case and leisure.

### Anything else?

_No response_

Just-Jojo:
Adds a supported way to interact with Red's blacklist cache. Before you had to use private attributes to access it :D
Closes #5171

### DIFF
--- a/redbot/core/_diagnoser.py
+++ b/redbot/core/_diagnoser.py
@@ -166,7 +166,7 @@ class DetailedGlobalCallOnceChecksMixin(IssueDiagnoserBase):
         )
 
     async def _get_detailed_global_whitelist_blacklist_result(self, label: str) -> CheckResult:
-        global_whitelist = await self.bot._whiteblacklist_cache.get_whitelist()
+        global_whitelist = await self.bot.get_whitelist()
         if global_whitelist:
             return CheckResult(
                 False,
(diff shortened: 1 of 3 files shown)


Jackenmen:
Hello, thanks for making this PR.

I've done a bunch of changes while reviewing:
- changed the signatures of the add/remove methods to accept iterable of users/roles instead of accepting an arbitrary amount of positional arguments (`*args`) as it seemed to be a more ergonomic choice based on the usage in core (with the usage in core, we usually already have an iterable that we can pass straight to the API rather than having to unpack it so that's why I think it's more ergonomic)
- updated all usage of the private cache to instead use the new public APIs
    - rebased on top of V3/develop in order to update the usage in `[p]diagnoseissues`
- done a bunch of updates to docstrings to hopefully make them a bit more clear

I'll be doing some testing soon and then I'll approve unless I see some issues.

Jackenmen (approved):
Looks good to me, thanks again for the PR!

PR #5210, opened 2021-08-13
[Docs] Fix capitalization from "python" to "Python"

fredster33:


Twentysix26 (changes requested):
Hi, I see the same occurrences in other guides: mac auto-restart, virtual environments, systemd to name a few. Mind taking care of those too?

PR #5215, opened 2021-08-15
Document parameters in remaining chat_formatting functions

Kreusada:

Adds parameters to the remaining chat formatting functions without parameters, so that they appear in the docs *with* parameters. I noticed some inconsistency between various functions and whether they had parameters in their docstrings, regardless of the number of arguments the function takes.

For example, `inline` and `quote` functions include parameters inside the docstring, despite only having one argument in the function's signature. The functions contained in this PR also only have one parameter each, therefore it would be nice for consistency measures to provide these functions with parameters in their docstrings too.

TLDR; adds parameters to function docstrings if they don't have parameters in their docstrings.


Jackenmen (approved):
Thanks

PR #5217, opened 2021-08-15
Add Debian 11 to install guide and clarify "Buster" as Debian 10

### LINKED ISSUE #5213: Add install instructions for Debian 11
https://www.debian.org/News/2021/20210814

Unlike Debian 10, Debian 11 should not require pyenv which means many build requirements won't need to be specified either. It will probably make most sense to base the instructions on Ubuntu 20.04, minus the Git ppa. From what I can tell, Debian 11 only supports Python 3.9 but this is fine - we will add support for it before merging the docs. 

This is up for grabs, I suggest testing whether the installation works without issues but it will be checked during review anyway.

TwinDragon:

Quick update to the installation instructions to add Debian 11 "Bullseye" instructions. As per #5213, the instructions for Ubuntu 20.04 worked fine, Red installed without problems and no Git errors when adding a repo and some cogs.

There's also a quick update to the heading for Debian Buster to better clarify that it's referring to Debian 10 and then Raspberry Pi OS being based on it.

Fixes #5213

### DIFF
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -124,23 +124,6 @@ with apt:
 
 .. prompt:: bash
 
-    sudo apt -y install python3 python3-dev python3-venv python3-pip git openjdk-11-jre-headless build-essential nano
-
-Continue by `creating-venv-linux`.
-
----
-
-.. _install-debian11:
-
-~~~~~~~~~~~~~~~~~~~~
-Debian 11 Bullseye
-~~~~~~~~~~~~~~~~~~~~
-
-Debian 11 "Bullseye" has all required packages available in official repositories. Install them
-with apt:
-
-.. prompt:: bash
-
     sudo apt update
     sudo apt -y install python3 python3-dev python3-venv python3-pip git openjdk-11-jre-headless build-essential nano
 
--- a/docs/install_linux_mac.rst
+++ b/docs/install_linux_mac.rst
@@ -116,6 +116,23 @@ Debian/Raspbian Buster. This guide will tell you how. First, run the following c
 
 Complete the rest of the installation by `installing Python 3.8 with pyenv <install-python-pyenv>`.
 
+---
+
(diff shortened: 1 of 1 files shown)


ghost:
The python version included with bullseye is 3.9, but if i recall correctly that only causes problems in Windows. Still to be aware though.

Jackenmen:
@BreezeQS as mentioned in the linked issue:
> From what I can tell, Debian 11 only supports Python 3.9 but this is fine - we will add support for it before merging the docs.

docs/install_linux_mac.rst
```diff
@@ -111,6 +111,23 @@ Debian/Raspbian Buster. This guide will tell you how. First, run the following c
...
+~~~~~~~~~~~~~~~~~~~~
+Debian 11 Bullseye
+~~~~~~~~~~~~~~~~~~~~
+
+Debian 11 "Bullseye" has all required packages available in official repositories. Install them
+with apt:
+
+.. prompt:: bash
+
+    sudo apt -y install python3 python3-dev python3-venv python3-pip git openjdk-11-jre-headless build-essential nano
```
Jackenmen, line 128: Better to add `sudo apt update` in case repositories are outdated (this was a reason of failure for CI build [here](https://github.com/jack1142/Red-Install-Tests/runs/3526740476?check_suite_focus=true)):
```suggestion
    sudo apt update
    sudo apt -y install python3 python3-dev python3-venv python3-pip git openjdk-11-jre-headless build-essential nano
```

Jackenmen (changes requested):

docs/install_linux_mac.rst
TwinDragon: Good point. I just auto-pilot when I do apt stuff and just do `sudo apt update` when I do package updates. Line added in!

Jackenmen (approved):
Looks good to me, thanks!

PR #5218, opened 2021-08-16
Add toggleable notification of deleted messages count in Cleanup

### LINKED ISSUE #4732: [Cleanup] let cleanup show in channel how many messages it deleted for a short while
# Feature request

- [ ] Cog
- [ ] Command
- [ ] API functionality
- [x] Cog UX improvement

EDIT: I'm suggesting this feature as a toggle. 

Currently, messages deleted through cleanup commands gets deleted silently and how many messages it deleted gets logged to console/bot logs. However, in many instances it would be useful if bot displayed how many messages it actually deleted in said channel for a short while (5 seconds or less) before auto-vanishing, so as to inform the person who executed said cleanup command. Below GIF can demonstrate what I'm trying to say:

unn2zSDxKu

This could be helpful in many instances, where, people who are managing a discord server as trial mod or helpers with just manage messages perms and don't have access to neither audit logs nor bot's console logs , OR to mods and admins who don't have to keep checking audit logs after using cleanup commands to see how many messages bot actually deleted in said channel, etc etc. It saves time, even by little amount and also makes the cleanup process transparent for them by showing deleted messages count. I also personally believe it improves user experience too in context of using cleanup cog. Just my 26 cents.

flaree:

Addition of a toggleable notification allowing to see how many messages have been deleted. Defaults to on

Addition of cleanupset commands, locked to admin or manage message until further discussion.

closes #4732

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -217,7 +217,7 @@ class Cleanup(commands.Cog):
         log.info(reason)
 
         await mass_purge(to_delete, channel)
-        await self.send_optional_notification(len(to_delete), channel, True)
+        await self.send_optional_notification(len(to_delete), channel, subtract_invoking=True)
 
     @cleanup.command()
     @commands.guild_only()
@@ -288,7 +288,7 @@ class Cleanup(commands.Cog):
         log.info(reason)
 
         await mass_purge(to_delete, channel)
-        await self.send_optional_notification(len(to_delete), channel, True)
+        await self.send_optional_notification(len(to_delete), channel, subtract_invoking=True)
 
     @cleanup.command()
     @commands.guild_only()
@@ -397,7 +397,7 @@ class Cleanup(commands.Cog):
         log.info(reason)
 
         await mass_purge(to_delete, channel)
-        await self.send_optional_notification(len(to_delete), channel, True)
+        await self.send_optional_notification(len(to_delete), channel, subtract_invoking=True)
 
     @cleanup.command()
     @commands.guild_only()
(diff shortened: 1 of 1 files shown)


redbot/cogs/cleanup/cleanup.py
```diff
@@ -32,6 +32,8 @@ class Cleanup(commands.Cog):
     def __init__(self, bot: Red):
         super().__init__()
         self.bot = bot
+        self.config = Config.get_conf(self, 8927348724)
```
Jackenmen:
```suggestion
        self.config = Config.get_conf(self, 8927348724, force_registration=True)
```

Jackenmen (changes requested):
Works fine, need to add `force_registration=True` when getting the Config object.

I am wondering about one possible source of confusion though.
When I run `[p]cleanup messages 2`, I'll get "3 messages were deleted." message. While this is *technically* true, I feel like including the command message as part of the count is a bit counter-intuitive and may cause some unnecessary worry.
What do you think about it?

Small note on the possible implementation: in some cases (i.e. command message in `[p]cleanup self`), it is not enough to just unconditionally subtract 1.

flaree:
> Works fine, need to add `force_registration=True` when getting the Config object.
> 
> I am wondering about one possible source of confusion though.
> When I run `[p]cleanup messages 2`, I'll get "3 messages were deleted." message. While this is *technically* true, I feel like including the command message as part of the count is a bit counter-intuitive and may cause some unnecessary worry.
> What do you think about it?
> 
> Small note on the possible implementation: in some cases (i.e. command message in `[p]cleanup self`), it is not enough to just unconditionally subtract 1.

~~I actually didn't even know the invoking message got deleted.~~

Yeah, I agree it can be confusing in the cases where you provide an amount.

flaree:
Haven't went through each command properly, just what I quickly thought needed the invoking message not included.

Jackenmen (changes requested):
- `[p]cleanup bot` should have one message subtracted too
- `[p]cleanup self` should use `can_mass_purge` variable to determine whether one message should be subtracted from the count

Jackenmen (approved):
Looks good to me, thanks for the PR!

PR #5225, opened 2021-08-25
[Docs] Add Oracle Cloud $300 credits to hosting guide

Vexed01:
Oracle Cloud offers $300 of free credits, valid for 30 days, so this has been added to where the same is mentioned for GCP.

Let's not be Googleist :)


Twentysix26 (approved):
Ty

PR #5226, opened 2021-08-25
Remove semicolon as a valid separator...

### LINKED ISSUE #5222: [Core set api] Bot ignore the api if token contain ";"

3.4.12

### Cog name

Core

### Command name

set api travitia api_key tok;en

I expect the bot to successfully add the api token.

### What actually happened?

The bot just ignores it, and paste help.
image

For the example, my token is an unfortunate API token, it contain a comma ";" inside. For obvious security reason I will not paste it here so I will call it `tok;en` to represent the unusual token.

- [p]set api travitia api_key tok;en
- [p]set api travitia api_key,tok;en
- [p]set api travitia api_key;tok;en
- [p]set api travitia api_key "tok;en"

I tried all the commands above, no luck, still the same results.

### Anything else?

_No response_

Kowlin:
*sigh*... I don't like this change...

Closes #5222

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -96,7 +96,7 @@ log = logging.getLogger("red")
 
 _ = i18n.Translator("Core", __file__)
 
-TokenConverter = commands.get_dict_converter(delims=[" ", ",", ";"])
+TokenConverter = commands.get_dict_converter(delims=[" ", ","])
 
 MAX_PREFIX_LENGTH = 20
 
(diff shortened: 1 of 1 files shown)


Jackenmen:
I think we should go ahead with this one. Ideally we would also remove the comma but that one is a lot more likely to be used in some old cog's instructions since it used to be the only way to specify these. Let's just leave it like that and reconsider for a future breaking version if anyone starts needing a comma in a token (which I'm kind of surprised hasn't happened *before* semicolon...)

redbot/core/core_commands.py
```diff
@@ -2970,7 +2970,7 @@ async def api(self, ctx: commands.Context, service: str, *, tokens: TokenConvert
 
         **Arguments:**
             - `<service>` - The service you're adding tokens to.
-            - `<tokens>` - Pairs of token keys and values. The key and value should be separated by one of ` `, `,`, or `;`.
+            - `<tokens>` - Pairs of token keys and values. The key and value should be separated by one of ` `, or `,`.
```
Jackenmen, line 2973: This needs to be updated in the relevant cog guide as well + the merge conflict needs to be resolved.

Jackenmen (changes requested):

Kowlin:
Not really needed after all these years it seems.

PR #5227, opened 2021-08-25
Respects quotes in set api

### LINKED ISSUE #5222: [Core set api] Bot ignore the api if token contain ";"

bobloy:

Uses a regex pattern that respects quotes and uses findall instead of split.

image

Closes #5222 

Alternative to #5226

### DIFF
--- a/redbot/core/commands/converter.py
+++ b/redbot/core/commands/converter.py
@@ -192,11 +192,13 @@ else:
         def __init__(self, *expected_keys: str, delims: Optional[List[str]] = None):
             self.expected_keys = expected_keys
             self.delims = delims or [" "]
-            self.pattern = re.compile(r"|".join(re.escape(d) for d in self.delims))
+            self.pattern = re.compile(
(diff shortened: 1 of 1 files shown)


bobloy:
Previous version included the quotes in the token, now strips quotes.
image

Note: Only works with double quotes.

Jackenmen:
Superseded by #6762

PR #5234, opened 2021-08-30
Update Mac OS instructions to work properly on Apple Silicon

Jackenmen:
# Bugfix request

Uses `brew --prefix` which will always give the correct prefix.


aikaterna:
I was able to create a venv with no issues when using this new prefix on an M1 mac. It was able to find the python exec with no problems.

aikaterna (approved):

PR #5240, opened 2021-09-01
[Mod] names command: default to cmd issuer

Apoo711:
If a user is not specified in `[p]name` then it will get the author's name.

### DIFF
--- a/redbot/cogs/mod/names.py
+++ b/redbot/cogs/mod/names.py
@@ -294,11 +294,9 @@ class ModInfo(MixinMeta):
     @commands.command()
     async def names(self, ctx: commands.Context, *, member: discord.Member):
         """Show previous names and nicknames of a member."""
-        if user is None:
-            user = ctx.author
-        else:
-            user = member
-        names, nicks = await self.get_names_and_nicks(user)
+        if member is None:
+            member = ctx.author
+        names, nicks = await self.get_names_and_nicks(member)
         msg = ""
         if names:
             msg += _("**Past 20 names**:")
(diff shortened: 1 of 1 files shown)


redbot/cogs/mod/names.py
```diff
@@ -294,7 +294,11 @@ async def userinfo(self, ctx, *, member: discord.Member = None):
     @commands.command()
     async def names(self, ctx: commands.Context, *, member: discord.Member):
         """Show previous names and nicknames of a member."""
-        names, nicks = await self.get_names_and_nicks(member)
+        if user is None:
+            user = ctx.author
+        else:
+            user = member
+        names, nicks = await self.get_names_and_nicks(user)
```
Kreusada:
```suggestion
        if member is None:
            member = ctx.author
        names, nicks = await self.get_names_and_nicks(member)
```

redbot/cogs/mod/names.py
```diff
@@ -294,7 +294,11 @@ async def userinfo(self, ctx, *, member: discord.Member = None):
     @commands.command()
     async def names(self, ctx: commands.Context, *, member: discord.Member):
```
Kreusada:
```suggestion
    async def names(self, ctx: commands.Context, *, member: discord.Member = None):
```

Kreusada (changes requested):
There are a few changes here that are needed - you hadn't defined "user", nor had you added the None default argument to the member parameter. I've made some suggestions for you to take a look at.

Apoo711:
Thanks Kreusada!

Twentysix26:
I don't think a mod tool should have the command issuer as a default. It's certainly something I wouldn't expect from a UX standpoint

Kowlin:
I'm gonna agree with Twentysix here, and go ahead and close this PR.

PR #5243, opened 2021-09-03
Add `[p]diagnoseissues` command

### LINKED ISSUE #4717: Add white/blacklist check to canrun command
# Feature request

- [X] Cog
- [X] Command
- [ ] API functionality

Change "canrun" functionality to check allow/ignorelists before reporting result. Perhaps include what specific setting prevents their use of the command.

Example Command: [p]permissions canrun  BobTheBuilder warn
Example Result: "That user has the permissions to run the command, however, is prevented by localblocklist."

When trying to debug issues around the a users ability to run a command with "canrun", it only reports if the users has the permissions. This led to an situation where the user had the correct permissions so "canrun" reported as "allowed", however, this did not account for them being excluded by the allowlist/ignorelist. This resulted in the user not being able to run the command, despite "canrun" claiming they could.

Jackenmen:
Adds a new command that should make it a lot easier to diagnose issues with various command checks.

The commit history from before today can be found at:
https://github.com/jack1142/DiagnoserCog

The diagnostic message is made of two parts, the explanation of the issue and a detailed report:
image
It is important that the first part of the message explains what's wrong as clearly as possible so I suppose that other than the code review itself, you should focus especially on making sure that those are clear enough for what they're explaining.

What should you test? Ideally, everything... I don't know, there are a lot of conditions here and many of them have their own diagnostic message so it's hard to say which parts here are the most likely to have issues.

Fixes #4717

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3696,7 +3696,17 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         *,
         command_name: str,
     ) -> None:
-        """Diagnose issues with command checks with ease!"""
+        """
+        Diagnose issues with the command checks with ease!
+
+        **Example:**
+            - `[p]diagnoseissues #general @Slime ban` - Diagnose why @Slime can't use `[p]ban` in #general channel.
+
+        **Arguments:**
+            - `<channel>` - The channel that the command should be tested for.
+            - `<member>` - The member that should be considered as the command caller.
+            - `<command_name>` - The name of the command to test.
+        """
         command = self.bot.get_command(command_name)
         if command is None:
             await ctx.send("Command not found!")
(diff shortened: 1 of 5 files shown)


redbot/core/core_commands.py
```diff
@@ -3680,6 +3681,57 @@ def _datasize(num: int):
...
+    # cause after all it could be quite useful to guild owners!
+    # Truth to be told, that would require us to make some part of this
+    # more end-user friendly rather than just bot owner friendly - terms like
+    # 'global call once checks' are not of any use to someone who isn't bot owner.
+    @commands.is_owner()
+    @commands.command()
+    async def diagnoseissues(
+        self,
+        ctx: commands.Context,
+        channel: discord.TextChannel,
```
Kowlin: Perhaps an optional for the channel? Also may be worth considering having the ability to query on channels globally on the bot, not sure how intensive that would be at scale though.

Jackenmen: If by globally you mean being able to check channels from a different server than you're running the command in, then this is possible when running the command from DMs. I guess I should put that info in docstring.

Jackenmen: Optional for channel sounds fine, I didnt do it to favour explicitness I think but if it improves usability then it doesn't really matter.

Kowlin (approved):

PR #5245, opened 2021-09-06
[Docs] Update Cleanup docs with new cleanupset command group

Kreusada:
Updates the cleanup cog guide to include the new cleanupset group.


Jackenmen (changes requested):
This should update the command's docstring appropriately too (with the more helpful description of what the feature does)

Jackenmen (approved):
LGTM, thanks!

PR #5313, opened 2021-09-12
Only initialize and teardown driver if actually needed

### LINKED ISSUE #5312: Removing an instance with nonexistant PostgreSQL database fails

3.4.13

Delete an instance with a nonexistent PostgreSQL database.

The instance to be deleted.

### What actually happened?

A really long error message, and the instance did not get deleted.

1. Run redbot-setup <name>
2. Choose PostgreSQL
3. Input settings to a server that doesn't exist
4. Try to delete the instance

### Anything else?

[Discord support message where issue was found.](https://discord.com/channels/133049272517001216/387398816317440000/886422097692332052)

leetfin:
Check if delete_data is True before initializing or tearing down driver.
Closes #5312.

### DIFF
--- a/redbot/setup.py
+++ b/redbot/setup.py
@@ -264,7 +264,8 @@ async def remove_instance(
 
     backend = get_current_backend(instance)
     driver_cls = drivers.get_driver_class(backend)
-    await driver_cls.initialize(**data_manager.storage_details())
+    if delete_data is True:
+        await driver_cls.initialize(**data_manager.storage_details())
     try:
         if delete_data is True:
             await driver_cls.delete_all_data(interactive=interactive, drop_db=drop_db)
@@ -280,7 +281,8 @@ async def remove_instance(
 
         save_config(instance, {}, remove=True)
     finally:
-        await driver_cls.teardown()
+        if delete_data is True:
+            await driver_cls.teardown()
     print("The instance {} has been removed\n".format(instance))
 
 
--- a/redbot/setup.py
+++ b/redbot/setup.py
@@ -251,21 +251,16 @@ async def remove_instance(
     backend = get_current_backend(instance)
 
     if interactive is True and delete_data is None:
+        msg = "Would you like to delete this instance's data?"
         if backend != BackendType.JSON:
-            delete_data = click.confirm(
-                "Would you like to delete this instance's data?"
-                " The database server must be running for this to work.",
-                default=False,
-            )
-        else:
-            delete_data = click.confirm(
-                "Would you like to delete this instance's data?", default=False
-            )
+            msg += " The database server must be running for this to work."
+        delete_data = click.confirm(msg, default=False)
 
     if interactive is True and _create_backup is None:
-        _create_backup = click.confirm(
-            "Would you like to make a backup of the data for this instance?", default=False
-        )
+        msg = "Would you like to make a backup of the data for this instance?"
+        if backend != BackendType.JSON:
+            msg += " The database server must be running for this to work."
+        _create_backup = click.confirm(msg, default=False)
 
     if _create_backup is True:
         await create_backup(instance)
--- a/redbot/setup.py
+++ b/redbot/setup.py
@@ -248,11 +248,17 @@ async def remove_instance(
     remove_datapath: Optional[bool] = None,
 ):
     data_manager.load_basic_configuration(instance)
+    backend = get_current_backend(instance)
 
     if interactive is True and delete_data is None:
-        delete_data = click.confirm(
-            "Would you like to delete this instance's data? If you have not created the database, please select No.", default=False
-        )
+        if backend != BackendType.JSON:
+            delete_data = click.confirm(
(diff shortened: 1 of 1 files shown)


madebylydia:
I think it would be a great idea to let the user know that he should select "No" in case he didn't created the database, otherwise it could be confusing of why it throw an error and doesn't try again with the "No" option. 🤷

redbot/setup.py
```diff
@@ -251,7 +251,7 @@ async def remove_instance(
 
     if interactive is True and delete_data is None:
         delete_data = click.confirm(
-            "Would you like to delete this instance's data?", default=False
+            "Would you like to delete this instance's data? If you have not created the database, please select No.", default=False
```
Jackenmen: This should only be shown when a non-JSON backend is used. Also, you should just say that the removal of instance's data requires the database server to be running as what you currently put is not the only case this can happen.

redbot/setup.py
```diff
@@ -264,7 +264,8 @@ async def remove_instance(
 
     backend = get_current_backend(instance)
     driver_cls = drivers.get_driver_class(backend)
-    await driver_cls.initialize(**data_manager.storage_details())
+    if delete_data is True:
+        await driver_cls.initialize(**data_manager.storage_details())
```
Jackenmen: It would be better if all of this were under a single `if` statement along with `try/finally` clause:
```py
if delete_data is True:
    try:
        await driver_cls.initialize(**data_manager.storage_details())
        await driver_cls.delete_all_data(interactive=interactive, drop_db=drop_db)
    finally:
        await driver_cls.teardown()
```

Jackenmen (changes requested):

redbot/setup.py
leetfin: done

redbot/setup.py
leetfin: Not sure how to check which backend is in use, would I get it from this?
Line 176 `default_dirs["STORAGE_TYPE"] = storage_type.value`

Jackenmen: There's a `backend = get_current_backend(instance)` line a bit below, just move it a bit higher and then you can check it with `if backend != BackendType.JSON`

leetfin: I think I did that right - it's still failing style though. Not sure why

redbot/setup.py
```diff
@@ -248,11 +248,17 @@ async def remove_instance(
...
     data_manager.load_basic_configuration(instance)
+    backend = get_current_backend(instance)
 
     if interactive is True and delete_data is None:
-        delete_data = click.confirm(
-            "Would you like to delete this instance's data?", default=False
-        )
+        if backend != BackendType.JSON:
+            delete_data = click.confirm(
+                "Would you like to delete this instance's data? The database server must be running for this to work. If you do not have a database server, select No.", default=False
```
Jackenmen: There is no need to mention not having the database server, it should be clear enough with just the first part and otherwise, it is also quite long.
```suggestion
                "Would you like to delete this instance's data?"
                " The database server must be running for this to work.",
                default=False,
```
Also, this code suggestion fixes issues with the black style.

redbot/setup.py
```diff
@@ -262,13 +268,13 @@ async def remove_instance(
...
-        if delete_data is True:
+    driver_cls = drivers.get_driver_class(backend)  
+    if delete_data is True:
+        try:
+            await driver_cls.initialize(**data_manager.storage_details())
             await driver_cls.delete_all_data(interactive=interactive, drop_db=drop_db)
-
+        finally:
+            await driver_cls.teardown()
         if interactive is True and remove_datapath is None:
```
Jackenmen: This and the following lines should not be part of the `if delete_date is True` block

redbot/setup.py
```diff
@@ -262,13 +268,13 @@ async def remove_instance(
     if _create_backup is True:
         await create_backup(instance)
 
-    backend = get_current_backend(instance)
-    driver_cls = drivers.get_driver_class(backend)
-    await driver_cls.initialize(**data_manager.storage_details())
-    try:
-        if delete_data is True:
+    driver_cls = drivers.get_driver_class(backend)
```
Jackenmen: The code can't have trailing whitespace:
```suggestion
    driver_cls = drivers.get_driver_class(backend)
```

Jackenmen (changes requested):
Note: the issues with the Black style should generally be fixed with the Black auto-formatter per our contributing guidelines but in this case, it should be enough to just accept my code suggestions.

redbot/setup.py
L33Tech: Hopefully done!

leetfin:
I think everything is fixed - sorry about all the issues. I probably shouldn't have tried to do this.

Jackenmen:
> I think everything is fixed - sorry about all the issues. I probably shouldn't have tried to do this.

Don't worry, it's your first PR, it can't be perfect :)
It seems all good to me now, I'll test it out later and merge if I don't see any issues with this approach, thanks!

Jackenmen (approved):
Looks good to me, thanks for the PR!

PR #5317, opened 2021-09-14
Fix Case's modified_at attr and its format in message content

Kowlin:
It appears that ``modified_at`` is in fact NOT an int in all cases... Its sometimes a float!


Jackenmen (approved):
LGTM

PR #5323, opened 2021-09-15
Use discord native timestamp in [p]uptime

ltzmax:

Since it wasn't already added, i've added the new discord's timestamp on `since` only to show current timezone.

5a2rJqX2i

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -561,7 +561,11 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         delta = datetime.datetime.utcnow() - self.bot.uptime
         uptime = self.bot.uptime.replace(tzinfo=datetime.timezone.utc)
         uptime_str = humanize_timedelta(timedelta=delta) or _("Less than one second.")
-        await ctx.send(_(f"Been up for: **{uptime_str}** (since <t:{int(uptime.timestamp())}:f>)"))
+        await ctx.send(
+            _(f"Been up for: **{time_quantity}** (since {timestamp})").format(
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -558,14 +558,10 @@ async def info(self, ctx: commands.Context):
...
         delta = datetime.datetime.utcnow() - self.bot.uptime
-        uptime_str = humanize_timedelta(timedelta=delta) or _("Less than one second")
-        await ctx.send(
-            _("Been up for: **{time_quantity}** (since {timestamp} UTC)").format(
-                time_quantity=uptime_str, timestamp=since
-            )
-        )
+        uptime = self.bot.uptime.replace(tzinfo=datetime.timezone.utc)
+        uptime_str = humanize_timedelta(timedelta=delta) or _("Less than one second.")
+        await ctx.send(_(f"Been up for: **{uptime_str}** (since <t:{int(uptime.timestamp())}:f>)"))
```
Jackenmen: f-strings do not work with i18n, use str.format() instead:
```suggestion
        await ctx.send(
            _(f"Been up for: **{time_quantity}** (since {timestamp})").format(
                time_quantity=uptime_str, timestamp=f"<t:{int(uptime.timestamp())}:f>"
            )
        )
```

Jackenmen (changes requested):

redbot/core/core_commands.py
ltzmax: Didn't know that... fixed it.

Jackenmen (approved):
Looks good to me, thanks!

PR #5325, opened 2021-09-18
[Bank] Dispatch events for account events

### LINKED ISSUE #5141: [Bank] Add event dispatch on account events

Bank's API

It would be nice if cogs can take advantage of listeners for the Bank cog.
Cogs could know when an user's account get changed, pruned, spend money, and so on...

### Anything else?

_No response_

yamikaitou:
Adds some event dispatches to Bank

red_bank_set_global
red_bank_set_balance
red_bank_transfer_credits
red_bank_wipe
red_bank_prune_accounts

Resolves #5141


redbot/core/bank.py
```diff
@@ -1085,3 +1120,41 @@ async def wrapped(*args, **kwargs):
...
+        recipient_old_balance: int,
+        recipient_new_balance: int,
+    ):
+        self.recipient_id = recipient_id
+        self.guild_id = guild_id
+        self.recipient_old_balance = recipient_old_balance
+        self.recipient_new_balance = recipient_new_balance
+
+
+class BankPruneInformation:
```
vertyco: The type annotation for `pruned_users` should be `Union[List[int], Dict[str, List[int]]]` instead of `Union(List[int], Dict[str, List[int]])`

Zephyrkul:
Would it be possible it to make these payload classes into [NamedTuple](https://docs.python.org/3/library/typing.html#typing.NamedTuple)s instead? That should be a free performance improvement since these classes don't need to be modified after they're created.

docs/framework_events.rst
```diff
@@ -4,6 +4,29 @@
...
+Bank
+^^^^
+
+.. py:method:: red_bank_set_global(global_state: bool)
+
+    Dispatched when the Bank gets toggled between Global and Local. :code:`global_state` will be True if the Bank is being set to Global or False if the bank is being set to Local
+
+.. py:method:: red_bank_set_balance(payload: redbot.core.bank.BankSetBalanceInformation)
+
+    Dispatched when a user has their balance changed.
```
Flame442, line 16: The documented events that dispatch custom objects should link to those objects in some way. It might make sense to use a similar format as the autogenerated API descriptions for methods, which have a "parameters" section describing each of the parameters in detail.

redbot/core/bank.py
```diff
@@ -1085,3 +1120,24 @@ async def wrapped(*args, **kwargs):
             return coro_or_command
 
     return deco
+
+
+class BankTransferInformation(NamedTuple):
```
Flame442, line 1125: These objects should be added to `__all__`, since they are part of the public API. They also need to be documented, which may or may not happen automatically after doing so.

redbot/core/bank.py
```diff
@@ -524,6 +544,14 @@ async def bank_prune(bot: Red, guild: discord.Guild = None, user_id: int = None)
 
     """
 
+    print(f"{guild} | {user_id}")
```
Flame442, line 547: Leftover print?

redbot/core/bank.py
```diff
@@ -345,6 +351,11 @@ async def set_balance(member: Union[discord.Member, discord.User], amount: int)
     if await group.name() == "":
         await group.name.set(member.display_name)
 
+    payload = BankSetBalanceInformation(
+        member.id, (0 if _cache_is_global else guild.id), old_balance, amount
```
Flame442, line 355: It's more typical for `guild` like attributes to be `Optional` and use `None` as the value to represent non-guild actions (IE, `message.guild`). If it is consistently possible to provide the full user/guild object rather than an ID, I also think that would be preferred.

docs/framework_events.rst
```diff
@@ -4,6 +4,29 @@
...
+
+.. py:method:: red_bank_transfer_credits(payload: redbot.core.bank.BankTransferInformation)
+
+    Dispatched when credits gets transfered from one user to another.
+
+.. py:method:: red_bank_wipe(guild_id: int)
+    
+    Dispatched when the Bank gets wiped. :code:`guild_id` will be the ID of the Guild that was wiped, 0 if all Guilds were wiped (Bank is Local), or -1 if all Users were wiped (Bank is Global)
```
Flame442, line 24: Having both 0 and -1 as special values seems a bit smelly. Is there a meaningful difference for the CC to account for that makes differentiating between wiping guilds or users something that needs to happen? If so, I think it would be neater to just have a second argument, whether or not the bank was global at the time of wiping.

I would say the CC can just check if the bank is global or not to get the same value, but that only applies for `[p]bankset reset`. `[p]bankset toggleglobal` also implicitly results in this being called, since it clears the bank during the process of swapping scope. Trying to test for it at that time would likely result in an incorrect value due to race conditions.

redbot/core/bank.py
```diff
@@ -563,6 +591,10 @@ async def bank_prune(bot: Red, guild: discord.Guild = None, user_id: int = None)
             if user_id in bank_data:
                 del bank_data[user_id]
 
+    payload = BankPruneInformation(scope, tmp)
```
Flame442, line 594: `tmp` is only defined when `user_id is None`. This errors if a particular user is pruned.
(I do think the conditionally defined variables in this function are smelly, making this sort of error very easy to make)

Flame442 (changes requested):
Initial design feedback.

redbot/core/bank.py
```diff
@@ -524,6 +544,14 @@ async def bank_prune(bot: Red, guild: discord.Guild = None, user_id: int = None)
 
     """
 
+    print(f"{guild} | {user_id}")
+    if guild is None and user_id is None:
+        scope = 1  # prune global
+    if guild is not None and user_id is None:
+        scope = 2  # prune server
+    else:
+        scope = 3  # prune user
```
Flame442, line 553: Forgot to hit submit on the comment I was writing here....

Doing it like this loses the ability to know what guild was being pruned. Unless there is some reason to differentiate between specifically pruning a member and a single member happening to be the only invalid member from a bulk prune, the signature of this event object would likely be better as:
`guild: Optional[discord.Guild], pruned_user_ids: List[int]`

Jackenmen:
It would seem that this got superseded by #6339 - we'll definitely make sure that you are listed as one of the authors before the merge.

PR #5326, opened 2021-09-21
[modlog] Fix broken timestamp

anclint01:

Converted timestamp from a float to an int as floats do not work with Discord's timestamp

notworktimestamp

yesworktimestamp


Jackenmen:
This has already been fixed by #5317

PR #5344, opened 2021-09-25
[Config] Add type-hints to _ValueContextManager methods

Tobotimus:
Without these type-hints, I've found that PyCharm misidentifies the returned type, causing a number of false positives in static type checking.

### Before these changes:
image

### After these changes:
image


Jackenmen (approved):
PyCharm isn't particularly good at type checking but in this case, the return types are good to have there anyway so thanks!

PR #5345, opened 2021-09-25
[Mute] Check for the user's role list before unmuting

### LINKED ISSUE #5161: Mute command should check for the user's role list before muting/unmuting

Mutes Cog

I think red should check for the user's role list whilst muting / unmuting a user and respond with User#1234 is already muted/unmuted in this server. Because when we run mute / unmute it will keep saying muted / unmuted the user all thou they would have been unmuted or muted.

### Anything else?

That's all

veni-vidi-code:
Closes #5161

Adds a check if the member was muted

There was a corresponding MUTE_UNMUTE_ISSUE before, but it was only checked for channelwise muting

### DIFF
--- a/redbot/cogs/mutes/mutes.py
+++ b/redbot/cogs/mutes/mutes.py
@@ -1602,15 +1602,18 @@ class Mutes(VoiceMutes, commands.Cog, metaclass=CompositeMetaClass):
             if not role:
                 ret["reason"] = _(MUTE_UNMUTE_ISSUES["role_missing"])
                 return ret
-
-            if guild.id in self._server_mutes:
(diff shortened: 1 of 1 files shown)


veni-vidi-code:
is it normal that it takes this long until a review for 8 lines @TrustyJAID

TrustyJAID:
> is it normal that it takes this long until a review for 8 lines @TrustyJAID

This is an open source project maintained by volunteers. People will get around to reviews when they have time and it's a necessary update. There has not been a planned release that would include this PR since you have made it so there's been no push to have it included yet. Also while I may be the code owner for the mutes cog I am hardly the only person who can and likely will review this PR.

PR #5351, opened 2021-09-30
[Docs] Removes LXC as an unsupported hosting platform

Crossedfall:
LXC is listed as an unsupported hosting platform when a vast majority of VPS rentals utilize that technology. 

While a seemingly pedantic change, broadly declaring that LXCs (and relating technologies) are unsupported would be overreaching; as I don't believe the intent is to require full VMs for Red.

### DIFF
--- a/docs/host-list.rst
+++ b/docs/host-list.rst
@@ -21,7 +21,7 @@ First, we would like to make something clear:
 .. warning::
     Due to their inability to handle Red's data structure and meet the
     conditions of being a supported platform; platforms such as Heroku, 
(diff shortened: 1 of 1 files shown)


Kowlin:
I'm inclined to agree since it's pretty much as bog standard as one can get with virtualization tools. And just as a sidenote, if someone has experience in an platfkrm they'd like to use, I'm not gonna stop the fr doing so. The problem mostly arises from people who *do not* understand the platforms and tools they are using.

Jackenmen (approved):
We will still have to ensure Support Squad knows what will be happening with this but I see no reason to not at least merge this, thanks!

PR #5352, opened 2021-10-01
[Modlog] Make `modlogset cases` docstring less vague

OofChair:

Mentioned in Red so thought it would be an easy fix :P


aikaterna:
What I was thinking when I mentioned this was that the argument of "action" in the function signature was vague, and that it didn't lend very many hints to the user to what "action" does or what they need to do. So my idea of this improvement was:  

`Syntax: [p]modlogset cases [action]`

Enable or disable case creation for a mod action.

An action can be enabling or disabling specific cases.
Example: `[p]modlogset cases kick enabled`

Or something similar to this that explains it better.

OofChair:
I updated the docstring, and I added in examples of what you can enable/disable. Should be good for review :)

Vexed01 (commented):
It may be worth adding the same examples of what actions are to the documentation at `docs/cog_guides/modlog.rst`.

OofChair:
> It may be worth adding the same examples of what actions are to the documentation at `docs/cog_guides/modlog.rst`.

Good idea, I just added that ☺

Jackenmen:
I resolved the conflicts so this can be reviewed again.

aikaterna:
Hi Oof, your docstring is still vague to me. Please refer to my earlier comment about a command example in the docstring.

aikaterna:
Thanks, the example added to the docstring was what I was looking for. Can you do the same to the docs too please, since in the docs there are examples of command usage for other various core commands?

aikaterna:
Please make the formatting on the docs' subtitle headers similar to the existing ones.

An existing example ("Examples" and "Arguments" are bolded and italicised):
Screen Shot 2022-02-24 at 6 18 36 PM

Your current implementation:
Screen Shot 2022-02-24 at 6 18 48 PM

aikaterna (approved):
Thank you for the help and your patience.

PR #5353, opened 2021-10-01
Add a tick on traceback command when it's sent to DMs

ltzmax:

This only add a ✅ when the traceback is sent to dm.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1456,7 +1456,6 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         """
         if not public:
             destination = ctx.author
-            await ctx.tick()
         else:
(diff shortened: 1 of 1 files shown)


Jackenmen (changes requested):
The tick should probably be sent after the messages are sent since it's a confirmation that the action has completed.

redbot/core/core_commands.py
```diff
@@ -1469,6 +1469,8 @@ async def traceback(self, ctx: commands.Context, public: bool = False):
                         "Either you blocked me or you disabled DMs in this server."
                     )
                     return
+        if not public:
+            await ctx.tick()
```
```suggestion
            if not public:
                await ctx.tick()
```

Jackenmen (changes requested):

Jackenmen (approved):
LGTM.

PR #5354, opened 2021-10-01
[Help] Optimisation on getting commands in Cog-help

npc203:

Now this PR might be nitpicky, but getting every command from the bot and filtering the cog commands when a cog help is requested is slower compared to using a default method (get_commands) provided by the cog.

This further saves a lot of overhead in `get_bot_help_mapping` which requests the `get_cog_help_mapping` with every loaded cog.  
In short, the bot used to get every command present in the bot again and again (and then filter it) for every cog loaded to display help. Now it's faster but simply getting the commands of the cog requested.

One possible downside (though I haven't tested it) is, This prevents monkey-patched/Commands created dynamically and the cog object injected into it, to not show up in the help. Tho such are forbidden methods to make commands I assume.


redbot/core/commands/help.py
```diff
@@ -274,7 +274,7 @@ async def send_help(
     async def get_cog_help_mapping(
         self, ctx: Context, obj: commands.Cog, help_settings: HelpSettings
     ):
-        iterator = filter(lambda c: c.parent is None and c.cog is obj, ctx.bot.commands)
+        iterator = obj.get_commands()
```
Flame442: Fails with `AttributeError: 'NoneType' object has no attribute 'get_commands'` when fetching the entire help output (`[p]help` with no arguments). 

```py
  File "c:\users\flame\venv\dev\lib\site-packages\discord\ext\commands\core.py", line 229, in wrapped
  File "c:\users\flame\venv\dev\lib\site-packages\redbot\core\commands\help.py", line 911, in red_help
    await ctx.bot.send_help_for(ctx, thing_to_get_help_for, from_help_command=True)
  File "c:\users\flame\venv\dev\lib\site-packages\redbot\core\bot.py", line 1272, in send_help_for
    return await self._help_formatter.send_help(
  File "c:\users\flame\venv\dev\lib\site-packages\redbot\core\commands\help.py", line 252, in send_help
    await self.format_bot_help(ctx, help_settings=help_settings)
  File "c:\users\flame\venv\dev\lib\site-packages\redbot\core\commands\help.py", line 631, in format_bot_help
    coms = await self.get_bot_help_mapping(ctx, help_settings=help_settings)
  File "c:\users\flame\venv\dev\lib\site-packages\redbot\core\commands\help.py", line 296, in get_bot_help_mapping
    cm = await self.get_cog_help_mapping(ctx, cog, help_settings=help_settings)
  File "c:\users\flame\venv\dev\lib\site-packages\redbot\core\commands\help.py", line 277, in get_cog_help_mapping
    iterator = obj.get_commands()
AttributeError: 'NoneType' object has no attribute 'get_commands'
```

Flame442 (dismissed):

redbot/core/commands/help.py
Flame442: This happens specifically because `None` is passed to this method as a "cog" object to catch any commands with no cog associated with them when listing all commands in the bot. I don't know if there's any special way to get a list of commands *not* associated with a cog, so this might need to use the existing check in this case.
https://github.com/Cog-Creators/Red-DiscordBot/blob/aa51fd9ad1463484039f1578ae71210a516000b8/redbot/core/commands/help.py#L295-L296

Flame442 (approved):

PR #5355, opened 2021-10-01
Handle os error in cogmanager with "friendly" message

### LINKED ISSUE #5138: `[p]addpath` raises `OSError`s for paths with invalid characters

3.4.11

### Cog name

CogManagerUI

### Command name

addpath

The command to send a friendly error message.

### What actually happened?

An exception was raised during the execution of the command.

1. Run `[p]addpath 'C:\'`
2. See error:
```py
Exception in command 'addpath'
  File "C:\Users\Jakub\OneDrive\Dokumenty\Red-DiscordBot\.venv\lib\site-packages\discord\ext\commands\core.py", line 85, in wrapped
  File "C:\Users\Jakub\OneDrive\Dokumenty\Red-DiscordBot\redbot\core\cog_manager.py", line 350, in addpath
    if not path.is_dir():
  File "C:\Python38\lib\pathlib.py", line 1415, in is_dir
    return S_ISDIR(self.stat().st_mode)
  File "C:\Python38\lib\pathlib.py", line 1197, in stat
    return self._accessor.stat(self)
OSError: [WinError 123] The filename, directory name, or volume label syntax is incorrect: "'C:\\'"
```

### Anything else?

Tested on Windows (as in the NTFS file system), might also be reproducible with the use of `:` on Mac (as in the HFS/HFS+ file system)

Dav-Git:
Closes #5138

Seems like pathlib was not happy with special characters.

I have played around with the original issue a bit and it seems to me that the " and ' together with \ in the case described by @jack1142 are the issue.
image

Since \ is discord's escape character I also tried escaping the \ with another \ ("C:\\") which also produced the error.

### DIFF
--- a/redbot/core/cog_manager.py
+++ b/redbot/core/cog_manager.py
@@ -357,7 +357,7 @@ class CogManagerUI(commands.Cog):
             await ctx.send(
                 _(
                     "That path does not exist or does not point to a valid directory. \nCheck that your path does not contain any {quotation_mark}, {apostrophe} or other invalid characters."
-                ).format(quotation_mark='``"``', apostrophe="``'``")
+                ).format(quotation_mark='`"`', apostrophe="`'`")
             )
             return
         try:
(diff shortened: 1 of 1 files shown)


Jackenmen:
Issue #5138 has been fixed in Python 3.8.10 and 3.9.5 so this PR is no longer needed.

Kardoudia:
﻿stop

PR #5356, opened 2021-10-01
Address common issues with load path configuration in docs

Jackenmen:
- Add instructions on how to add pyenv to load path
    - I include support for zsh in pyenv load path instructions because why not
- Make instructions for Mac work when shell-specific profile (.bash_profile/.zprofile) file exists

See docs preview:
- Mac: https://red-discordbot--5356.org.readthedocs.build/en/5356/install_guides/mac.html
- pyenv instructions:
    - Python 3.9 (shown for CentOS but it's same for other OSes): https://red-discordbot--5356.org.readthedocs.build/en/5356/install_guides/centos-7.html#installing-python-with-pyenv
    - Python 3.8 (shown for Raspberry Pi OS): https://red-discordbot--5356.org.readthedocs.build/en/5356/install_guides/raspberry-pi-os-10.html#installing-python-with-pyenv

I know these are somewhat complicated but I expect users will just copy-paste though so it shouldn't particularly matter.

Could use some additional testing.

### DIFF
--- a/docs/install_guides/_includes/_install-pyenv-and-setup-path.rst
+++ b/docs/install_guides/_includes/_install-pyenv-and-setup-path.rst
@@ -16,4 +16,4 @@ you should run these commands:
     echo 'eval "$(pyenv init -)"' >> "$rcfile"
     echo 'eval "$(pyenv virtualenv-init -)"' >> "$rcfile"
 
-Then **close and reopen to your shell** and run the following command:
+Then **log out and log back in** and run the following command:
--- a/docs/install_guides/_includes/_install-pyenv-and-setup-path.rst
+++ b/docs/install_guides/_includes/_install-pyenv-and-setup-path.rst
@@ -1,0 +1,18 @@
+To install/update pyenv, run the following command:
+
+.. prompt:: bash
+
+    command -v pyenv && pyenv update || curl https://pyenv.run | bash
+
+After this command, you will see a warning about 'pyenv' not being in the load path. To address this,
+you should run these commands:
+
+.. prompt:: bash
+
+    profile=$([ -f ~/.bash_profile ] && echo ~/.bash_profile || echo ~/.profile)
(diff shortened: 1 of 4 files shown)


TwinDragon (dismissed):
Before I ran through the commands, I cleared out anything in both my `.bashrc` and `.profile` files for Debian 10 so I had a clean slate to work with and ran through each command as provided. They worked fine, no errors to report and pyenv was running just fine in my PATH.

I think this will help in the future so long as pyenv doesn't change anything again.

docs/install_guides/_includes/_install-pyenv-and-setup-path.rst
```diff
@@ -0,0 +1,19 @@
...
+.. prompt:: bash
+
+    profile=$([ -n "$ZSH_VERSION" ] && [ -f ~/.zprofile ] && echo ~/.zprofile || [ -f ~/.bash_profile ] && echo ~/.bash_profile || echo ~/.profile)
+    rcfile=$([ -n "$ZSH_VERSION" ] && echo ~/.zshrc || echo ~/.bashrc)
+    printf '%s\n%s\n%s\n' 'export PYENV_ROOT="$HOME/.pyenv"' 'export PATH="$PYENV_ROOT/bin:$PATH"' "$([ -f "$profile" ] && cat "$profile")" > "$profile"
+    echo 'eval "$(pyenv init --path)"' >> "$profile"
+    echo 'eval "$(pyenv init -)"' >> "$rcfile"
+    echo 'eval "$(pyenv virtualenv-init -)"' >> "$rcfile"
+
+Then **close and reopen to your shell** and run the following command:
```
Jackenmen: This should instead say to log out and re-login - in desktop environments with a terminal app, `~/.profile` only gets sourced on login.

aikaterna (approved):
Mac changes look good now that we got to work through this together. Thanks for the help Jack.

PR #5359, opened 2021-10-01
[Audio] Spotify post request fix

npc203:
Report by alec in discord. When the status of request to fetch token from spotify isn't 200, The content-type is not json.
This PR handles it ,makes sure the error is propagated and handled.

Contributes towards #6776

### DIFF
--- a/redbot/cogs/audio/apis/spotify.py
+++ b/redbot/cogs/audio/apis/spotify.py
@@ -126,7 +126,7 @@ class SpotifyWrapper:
             else "US"
         )
 
-    async def request_access_token(self) -> MutableMapping:
+    async def request_access_token(self) -> Optional[MutableMapping]:
         """Make a spotify call to get the auth token."""
         await self.get_token()
         payload = {"grant_type": "client_credentials"}
@@ -141,6 +141,7 @@ class SpotifyWrapper:
         token = await self.request_access_token()
         if token is None:
             log.debug("Requested a token from Spotify, did not end up getting one.")
+            return None
         try:
             token["expires_at"] = int(time.time()) + int(token["expires_in"])
         except KeyError:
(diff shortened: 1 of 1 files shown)


npc203:
While we are still here, the `self.get()` suffers from the same problem, it tries to parse json despite the response not necessarily being one (mostly when response status is 5xx). However fixing the get req needs lots of other changes and error handling pretty much all over the api interface, because it is used in many places unlike post (which just occurs once). Thus I didn't cramp it up into a single PR and let better people deal with it. but do let me know if you would like me to update this PR with a similar approach as done in post.

redbot/cogs/audio/apis/spotify.py
```diff
@@ -151,12 +152,18 @@ async def get_access_token(self) -> Optional[str]:
 
     async def post(
         self, url: str, payload: MutableMapping, headers: MutableMapping = None
-    ) -> MutableMapping:
+    ) -> Optional[MutableMapping]:
         """Make a POST call to spotify."""
         async with self.session.post(url, data=payload, headers=headers) as r:
-            data = await r.json(loads=json.loads)
-            if r.status != 200:
```
Jackenmen, line 158: We still should be checking for status code as some (if not most) non-200 responses contain valid JSON. Yes, with current usage this doesn't technically matter but we should be doing it nonetheless.
If we do this, we could probably even simplify `get_access_token()` and remove the `try-except` with `KeyError` from it and just directly assign instead.

Jackenmen (changes requested):

Jackenmen:
> While we are still here, the `self.get()` suffers from the same problem, it tries to parse json despite the response not necessarily being one (mostly when response status is 5xx). However fixing the get req needs lots of other changes and error handling pretty much all over the api interface, because it is used in many places unlike post (which just occurs once). Thus I didn't cramp it up into a single PR and let better people deal with it. but do let me know if you would like me to update this PR with a similar approach as done in post.

I suppose this could be added to this PR as well, it's part of the same issue.

Jackenmen:
It's been over 4 years since the last changes on this PR were made, and there are still unresolved review comments on this PR. It's time to just close this and let someone else pick it up. I made a tracking issue at #6776

PR #5361, opened 2021-10-02
[Mod] fix possible issue in userinfo if joined_at is None

fixator10:
The current code implies that `joined_at` can be `None` (likely, if target is in "lurker" mode). 
However, the current implementation will fail on `joined_at.replace` if `joined_at` is `None`.


Flame442 (approved):
Thanks!

PR #5362, opened 2021-10-02
Check if branch exists before updating

### LINKED ISSUE #3397: [Downloader] Repos set with tag as branch break `[p]cog update`
# Command bugs

#### Command name

`[p]cog update`

Downloader

Cog update to proceed without error or at least to error gracefully.

#### What actually happened?

Got unhandled exception:
```
Exception in command 'cog update'
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/discord/ext/commands/core.py", line 79, in wrapped
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/cogs/downloader/downloader.py", line 841, in _cog_update
    await self._cog_update_logic(ctx, cogs=cogs)
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/cogs/downloader/downloader.py", line 915, in _cog_update_logic
    cogs_to_check, check_failed = await self._get_cogs_to_check(repos=repos, cogs=cogs)
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/cogs/downloader/downloader.py", line 1159, in _get_cogs_to_check
    __, failed = await self._repo_manager.update_repos()
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/cogs/downloader/repo_manager.py", line 1175, in update_repos
    updated_repo, (old, new) = await self.update_repo(repo.name)
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/cogs/downloader/repo_manager.py", line 1142, in update_repo
    old, new = await repo.update()
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/cogs/downloader/repo_manager.py", line 805, in update
    await self.hard_reset()
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/cogs/downloader/repo_manager.py", line 785, in hard_reset
    raise errors.HardResetError(
redbot.cogs.downloader.errors.HardResetError: Some error occurred when trying to execute a hard reset on the repo at the following path: /home/ubuntu/red-local/cogs/RepoManager/repos/jack-rlstats31
```

1. Add repo with a tag as a branch, i.e:
`[p]repo add jackcogs-test https://github.com/jack1142/JackCogs rlstats-3.1`
2. Run `[p]cog update`

Jan200101:
Checks is the branch used for the repo exists using `git show-ref --quiet --verify` before trying to update

closes #3397

### DIFF
--- a/redbot/cogs/downloader/repo_manager.py
+++ b/redbot/cogs/downloader/repo_manager.py
@@ -126,6 +126,7 @@ class Repo(RepoJSONMixin):
         "git -C {path} log --diff-filter=D --pretty=format:%H -n 1 {descendant_rev}"
         " -- {module_name}/__init__.py"
     )
+    GIT_VERIFY_BRANCH = "git -C {path} show-ref --quiet --verify refs/heads/{branch}"
 
     PIP_INSTALL = "{python} -m pip install -U -t {target_dir} {reqs}"
 
@@ -246,6 +247,22 @@ class Repo(RepoJSONMixin):
 
         """
         return await self.latest_commit() == self.commit
+
+    async def is_branch(self) -> bool:
+        """
+        Check if the branch assosiated with the repo is a tag.
+
+        Returns
+        -------
+        bool
+            `True` if branch is a tag or `False` otherwise
+        """
+        git_command = ProcessFormatter().format(
+            self.GIT_VERIFY_BRANCH, branch=self.branch, path=self.folder_path
+        )
+        p = await self._run(git_command)
+
+        return p.returncode == 0
 
     async def _get_file_update_statuses(
         self, old_rev: str, new_rev: Optional[str] = None
@@ -815,6 +832,9 @@ class Repo(RepoJSONMixin):
         -------
         `UpdateError` - if git pull results with non-zero exit code
         """
+        if not await self.is_branch():
+            return self.commit, self.commit
+
         old_commit = await self.latest_commit()
 
         await self.hard_reset()
--- a/tests/cogs/downloader/test_downloader.py
+++ b/tests/cogs/downloader/test_downloader.py
@@ -324,6 +324,7 @@ async def test_update(mocker, repo):
     new_commit = "a0ccc2390883c85a361f5a90c72e1b07958939fa"
     m = _mock_run(mocker, repo, 0)
     _mock_setup_repo(mocker, repo, new_commit)
+    mocker.patch.object(repo, "is_branch", autospec=True, return_value=True)
     mocker.patch.object(repo, "latest_commit", autospec=True, return_value=old_commit)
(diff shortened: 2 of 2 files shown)


Jackenmen:
> check if HEAD is detached

Can we just check if the given revision (which is saved in Config and therefore doesn't depend on the current, possibly invalid state of the repository) is a branch (I assume that is probably possible to do with git)? I think that's the only thing that isn't constant (and is therefore the only thing that Downloader can update)? *Technically*, tags can be modified too but I don't think we need to support it right now, especially that we don't even properly support when people force push to the branch.

Jan200101:
> 
> 
> > check if HEAD is detached
> 
> Can we just check if the given revision (which is saved in Config and therefore doesn't depend on the current, possibly invalid state of the repository) is a branch (I assume that is probably possible to do with git)? I think that's the only thing that isn't constant (and is therefore the only thing that Downloader can update)?

This sounds like a good idea but I am seeing one edge case, if the user manually updates the repo (for whatever reason) the commit in the config is no longer the latest commit of the branch.

I am unsure if this is something that needs to be tackled to prevent PEBKAC or if modifying red data outside of red is unsupported

> _Technically_, tags can be modified too but I don't think we need to support it right now, especially that we don't even properly support when people force push to the branch.

.

Jan200101:
having had a bit to think about this, what about when a tag and branch are both on the same commit?
I think it would be better to generally look if the head is detached, which would leave branches entirely alone (unless there was an issue fast forwarding which would imply a force push or local change)

Jackenmen:
> having had a bit to think about this, what about when a tag and branch are both on the same commit?

It probably doesn't matter. When you add a repo, you pass a named reference (or omit it, in which case the default branch is used and saved) which is saved with Config:
```json
{
    "170708480":
    {
        "GLOBAL":
        {
            "repos":
            {
                "Fox-V3": "master",
                "aika": "v3",
                "preda": "master",
                "trusty": "master",
                "kenny": "v3-cogs",
                "jack-private": "v3",
                "draper": "master",
                "ax": "V3",
                "laggron": "v3",
                "PR4086": "master",
                "sinbad": "v3",
                "me-bounty": "main",
                "palmtree5": "redv3-rewrites",
                "weird": "master",
                "flame": "master"
            }
        }
    }
}
```
The current state of the repository doesn't really matter as we have the indication of whether we're dealing with a branch or something else based on this alone. We just need to detect whether the given name is a branch name which can be done by adding `refs/heads/` (if it's not already there) and then verifying with `git show-ref --verify`.

Jan200101:
right, ignore me forgetting to run black

I've changed the approach to the one you suggested: checking if the branch exists before updating.

This is implemented in a naive way and simply uses `refs/heads/{branch}` since I don't think the stored branch would ever be the full ref.

PR #5364, opened 2021-10-02
Do not include expected wait_for responses in translated strings

### LINKED ISSUE #5363: Russian translation issue

3.4.14

### Cog name

Core

### Command name

repo add

No translation for substring "I agree" or bot's reaction to the translated message, in Russian it is "Я согласен".

### What actually happened?

image

1. Set i18n settings to Russian language
2. Add any repo by command "[p] repo add"

### Anything else?

_No response_

Jackenmen:
This should remove all occurrences of translatable strings including the responses expected by us in `bot.wait_for()` even though we do not actually support translated variants. Also, expands `(y/n)` to full `(yes/no)` to avoid any confusion on what these mean.

Fixes #5363

### DIFF
--- a/redbot/cogs/downloader/checks.py
+++ b/redbot/cogs/downloader/checks.py
@@ -2,6 +2,7 @@ import asyncio
 
 from redbot.core import commands
 from redbot.core.i18n import Translator
+from redbot.core.utils.chat_formatting import bold
 from redbot.core.utils.predicates import MessagePredicate
 
 __all__ = ["do_install_agreement"]
@@ -13,7 +14,7 @@ REPO_INSTALL_MSG = _(
     "You're about to add a 3rd party repository. The creator of Red"
     " and its community have no responsibility for any potential "
     "damage that the content of 3rd party repositories might cause."
-    "\n\nBy typing '**I agree**' you declare that you have read and"
+    "\n\nBy typing '{confirmation_message}' you declare that you have read and"
     " fully understand the above message. This message won't be "
     "shown again until the next reboot.\n\nYou have **30** seconds"
     " to reply to this message."
@@ -26,7 +27,7 @@ async def do_install_agreement(ctx: commands.Context) -> bool:
     if downloader is None or downloader.already_agreed:
         return True
 
-    await ctx.send(T_(REPO_INSTALL_MSG))
+    await ctx.send(T_(REPO_INSTALL_MSG).format(confirmation_message=bold("I agree")))
 
     try:
         await ctx.bot.wait_for(
--- a/redbot/cogs/downloader/checks.py
+++ b/redbot/cogs/downloader/checks.py
@@ -7,19 +7,7 @@ from redbot.core.utils.predicates import MessagePredicate
 
 __all__ = ["do_install_agreement"]
 
-T_ = Translator("DownloaderChecks", __file__)
-
-_ = lambda s: s
-REPO_INSTALL_MSG = _(
-    "You're about to add a 3rd party repository. The creator of Red"
-    " and its community have no responsibility for any potential "
-    "damage that the content of 3rd party repositories might cause."
-    "\n\nBy typing '{confirmation_message}' you declare that you have read and"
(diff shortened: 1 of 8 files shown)


Jackenmen:
This should now include all occurrences of the strings including the response expected by us when we use `bot.wait_for()`.

redbot/cogs/mutes/mutes.py
```diff
@@ -1014,10 +1020,16 @@ async def _check_for_mute_role(self, ctx: commands.Context) -> bool:
...
                 msg += _(
-                    "Saying `yes` will continue "
+                    "Saying {response_1} will continue "
                     "the mute with overwrites and stop this message from appearing again, "
-                    "saying `no` will end the mute attempt."
+                    "saying {response_2} will end the mute attempt."
                 )
+            msg = msg.format(
+                command_1=inline(command_1),
+                command_2=inline(command_2),
```
Dav-Git, line 1029: Having this closer to where the variables were defined would make this code a lot more readable.

redbot/cogs/downloader/checks.py
```diff
@@ -26,7 +27,7 @@ async def do_install_agreement(ctx: commands.Context) -> bool:
     if downloader is None or downloader.already_agreed:
         return True
 
-    await ctx.send(T_(REPO_INSTALL_MSG))
+    await ctx.send(T_(REPO_INSTALL_MSG).format(confirmation_message=bold("I agree")))
```
Dav-Git: I see why you put the `.format()` here. If we're already moving the text that needs to be typed into the function, you'll probably want to use the same string here as in line 34.
```py
confirmation_message="I agree"
await ctx.send(T_(REPO_INSTALL_MESSAGE).fornmat(confirmation_message=bold(confirmation_message)))
# ...
await ctx.bot.wait_for(
	"message", check=MessagePredicate.lower_equal_to(confirmation_message.lower(), ctx), timeout=30
)
```
That way there's only one place to look for the correct string.

Dav-Git (changes requested):
There's a few things that would make the code better readable, but from a functional perspective everything looks fine.

redbot/cogs/mutes/mutes.py
Jackenmen: I'm not sure I agree - these variables are defined right above the definition of `msg` variable which makes it easy to see what are the contents of the replacement fields in that text.

redbot/cogs/downloader/checks.py
Jackenmen: Done 👍

Dav-Git (dismissed):

redbot/cogs/mutes/mutes.py
```diff
@@ -983,9 +989,9 @@ async def _check_for_mute_role(self, ctx: commands.Context) -> bool:
         command_2 = f"{ctx.clean_prefix}muteset makerole"
         msg = _(
             "This server does not have a mute role setup. "
-            " You can setup a mute role with `{command_1}` or"
-            "`{command_2}` if you just want a basic role created setup.\n\n"
-        ).format(command_1=command_1, command_2=command_2)
+            " You can setup a mute role with {command_1} or"
+            "{command_2} if you just want a basic role created setup.\n\n"
```
Flame442: There is no space between the previous line and this one (same behavior as the old code, but incorrect)

redbot/core/core_commands.py
```diff
@@ -2688,8 +2688,9 @@ async def prefix(self, ctx: commands.Context, *prefixes: str):
             await ctx.send(
                 _(
                     "Warning: A prefix is above the recommended length (20 characters).\n"
-                    "Do you want to continue? (y/n)"
+                    "Do you want to continue?"
                 )
+                + "(yes/no)"
```
Flame442: This one still needs a leading space

Flame442 (changes requested):
Mild formatting issues.

Flame442 (approved):

PR #5367, opened 2021-10-02
[Mod] Indicate successful run in `[p]voicekick`

the-krak3n:
This PR will send a message that the user has been successfully kicked from the vc when `[p]voicekick` is used to kick someone from a vc.
Before it used to kick the user from vc silently without any message or tick.
voicekick


Jackenmen (approved):
Looks good 👍

PR #5369, opened 2021-10-03
[Audio] Allow to skip even if the queue is empty

### LINKED ISSUE #5349: [Audio] Allow skipping with empty queue

Audio

Allow for skipping songs even if the queue is empty. It does not make any sense that you'd have to listen to the current song just because there is nothing coming up in the queue, and having to queue a 10 second clip just to be able to skip the current one is a very weird workaround.

The skip command should skip the current song, not "play next".

### Anything else?

_No response_

Wayrion:
Now skips even when queue is empty 

Fixes #5349


Jackenmen (commented):
It seems that this might still fail to skip if the `except` block in line 135 gets hit.

Jackenmen:
After looking at this again, it seems to me that this `except` block should only get hit *if* there is no track currently playing so this shouldn't need to be changed.

Edit: I don't know anymore...

redbot/cogs/audio/core/utilities/player.py
```diff
@@ -142,14 +142,16 @@ async def _skip_action(self, ctx: commands.Context, skip_to_track: int = None) -
...
-                    )
+                embed = discord.Embed(
+                    title=_("Track Skipped"),
+                    description=await self.get_track_description(
+                        player.current, self.local_folder_current_path
+                    ),
                 )
-            await self.send_embed_msg(ctx, embed=embed)
-            return
+                await self.send_embed_msg(ctx, embed=embed)
```
aikaterna: This indentation you added removes the embed send from the livestreaming if block above it.

aikaterna:
After skipping a livestream, it shows:
Screen Shot 2021-10-20 at 10 01 31 AM
Please change this to the standard skip response instead:
Screen Shot 2021-10-20 at 10 02 22 AM

Wayrion:
Sorry about that, I forgot to save file before commiting. Should be fixed now 😅

aikaterna:
Your last commit did not resolve what I mentioned in the previous comment.  
If you skip a livestream, it will say:

Then it will skip anyway. Please change the response to the standard skip response.

benno1237:
I still think you are overcomplicating things slightly. As far as i am aware the whole block of code starting from L132 to L152 can be removed and that's all it takes. Skip should ignore empty queues then. 
https://github.com/benno1237/Red-DiscordBot/commit/daec57666444c167f5c74136ea8268c8643422c5
It's your PR so I don't want to talk you into anything. I however feel like that's the easier and prettier solution. Does it have bugs? Hell i don't know. Don't rely on me

Drapersniper:
There is a far simpler implementation for this. 

- Just invoke the `[p]stop` command when the user skips ONLY if there is a song currently being played and there is no queue (Break up the existing OR logic)
Instead of 
```py
         if not player.current or (not player.queue and not autoplay):
            try:
                pos, dur = player.position, player.current.length
            except AttributeError:
                await self.send_embed_msg(ctx, title=_("There's nothing in the queue."))
                return
            time_remain = self.format_time(dur - pos)
            if player.current.is_stream:
                embed = discord.Embed(title=_("There's nothing in the queue."))
                embed.set_footer(
                    text=_("Currently livestreaming {track}").format(track=player.current.title)
                )
            else:
                embed = discord.Embed(title=_("There's nothing in the queue."))
                embed.set_footer(
                    text=_("{time} left on {track}").format(
                        time=time_remain, track=player.current.title
                    )
                )
            await self.send_embed_msg(ctx, embed=embed)
            return         
        elif autoplay and not player.queue:

```

You should change it to 

```py
        if not player.current: # No matter what you don't want to skip if nothing is playing.
            await self.send_embed_msg(ctx, title=_("There's nothing being played."))
            return
        elif not player.queue and not autoplay: # You want to stop the existing playback if there is nothing in the queue and autoplay is off and there is something currently playing, by invoking the `[p]stop` command you are ensuring all the handling is done properly as it is already done by the stop command.
        	await ctx.invoke(self.command_stop)
            return
        elif autoplay and not player.queue:
            ### current code
```

aikaterna:
There is a voteskip issue regarding this change.

pleb = non-mod, non-admin, non-bot owner, etc: use non-privileged accounts
Have 2 plebs in a voice channel with the bot.
Set the voteskip threshold to less than 50% (so that one pleb vote for skipping will cause a skip)
Put 1 song in the queue.
One pleb uses skip - the bot responds with "Vote threshold met."
The vote threshold is met, but the next response is "Unable To Stop Player/There are other people listening - vote to skip instead." and the bot does not stop playback.

Drapersniper:
The correct fix here imo would he checking if user is privilege if he is allow skipping of last song (stopping the player)

if the user is not privileged say something along the lines of  "unable to stop the player as you not a DJ"

aikaterna:
Part of the fun here is the stopping point/response is coming from invoking stop on the player and the stop function's attached checks are getting hit (https://github.com/Cog-Creators/Red-DiscordBot/blob/6acff9961256b7ae3b5b994866e6dd873ffeca2e/redbot/cogs/audio/core/commands/controller.py#L581). 

I tested this with DJ roles involved and everything seemed to work as intended but I gave it a rather superficial pass.

PR #5372, opened 2021-10-04
Fix descriptions for examples of ID-based command invocations in Mod

WreckRox:

The example command in [kickban.py#L385](https://github.com/Cog-Creators/Red-DiscordBot/blob/a0a433b13d698250ba6689fb07889ea507127c68/redbot/cogs/mod/kickban.py#L385) stated 
`[p]ban 428675506947227648 7 Continued to spam after told to stop.
This will ban Twentysix and it will delete 7 days worth of messages.`

The description is inconsistent with the example command. Changed the example command description to 
`This will ban the User ID 428675506947227648 and it will delete 7 days worth of messages.`

### DIFF
--- a/docs/cog_guides/mod.rst
+++ b/docs/cog_guides/mod.rst
@@ -92,7 +92,7 @@ Kick a user.
 **Example Usage**
 
 * ``[p]kick 428675506947227648 wanted to be kicked.``
-    This will kick user with ID 428675506947227648 from the server.
+    This will kick the user with ID 428675506947227648 from the server.
 * ``[p]kick @Twentysix wanted to be kicked.``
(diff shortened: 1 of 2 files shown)


Jackenmen (changes requested):
Thanks for the PR! This should also be fixed in tempban and kick command.
Also, it should probably be "user with ID ..." rather than "user ID ..." as this does not sound that great IMO.

Kreusada (changes requested):
These changes should also be made present in the documentation for the Mod cog (`docs/cog_guides/mod.rst`).

Jackenmen (approved):
Looks good to me

PR #5375, opened 2021-10-07
[Help] Send menus to dm if maxpages is 0

### LINKED ISSUE #5093: Allowing Help to be DMable when Menu is toggled.
### Type of feature request

Other

As stated in the Title when using the menu system which is enabled via [p]helpset usemenus 

It now uses help in the chat you used it, I think a good feature would be being able to toggle this so people dont flood the chat with help commands if they want to look at the command list!

Thank you in advance!

### Anything else?

_No response_

Dav-Git:
~~Blocked by #6282~~
Closes #5093

This PR will change help to send help menus to DMs if maxpages is 0.
Inconsistent with the "non-menu" mode the menus would always send in the invocation context even if maxpages was 0.

Currently i am a little stuck on the menus in the DM. For a reason i have yet to figure out the menu_ctx context does not play nicely with the menu.
While all messages and reactions can be sent/added, I can not flip the pages in the menu.


Dav-Git:
I can‘t seem to find the issue here. The menu works fine with the old implementation when invoked in dms but for some reason my context for m doesn‘t want to work with it.

Flame442 (dismissed):
`[p]helpset maxpages`'s help currently states:
> Note: This setting does not apply to menu help.

This needs to be changed to explain the new behavior specific to menus as a part of this PR.

Flame442:
This PR has conflicts due to the addition of button support to help's menu options. Looking at the `SimpleMenu` class, it requires a `ctx` object to send the menu in, which won't exist for DMs. It does not support the `message=` kwarg that the older `menu` supported, to put a menu onto an existing message. Continuing to include this change would require a different approach to support button menus.

@TrustyJAID, you wrote the PR to add buttons to help, do you have any insight as to the best way to approach this?

TrustyJAID:
I think the best path forward would be adding a new option on SimpleMenu called `start_dm` which accepts a user parameter instead. This way it's a new feature for the menu and would allow this to happen. Looking at the SimpleMenu ctx is only used to extract the author and provide a messageable which can be simply replaced with the author in the new method.

Flame442:
Blocked until #6282 is resolved and this PR can use the new method.

Flame442:
> I can‘t seem to find the issue here. The menu works fine with the old implementation when invoked in dms but for some reason my context for m doesn‘t want to work with it.

The issue appears to be that generating `ctx` from a message sent by the _bot_ sets `ctx.author` equal to the bot, rather than the invoker of `help`. This means the menu is listening for reactions from the bot, rather than from the user.
https://github.com/Cog-Creators/Red-DiscordBot/blob/409ece427f9b8014cf0f28ff03a019cab323107e/redbot/core/utils/menus.py#L200
It should be possible to monkeypatch the author attribute to fix this issue.

Flame442 (approved):

PR #5380, opened 2021-10-11
Update changelog_3_4_0.rst

RehanPlayz:


Jackenmen:
The jar links in the changelogs correspond to the Lavalink release that the release ships with, not the latest release so this does not need changing.

PR #5386, opened 2021-10-12
[Modlog]Raise error instead of failing silently on invalid arguments

### LINKED ISSUE #3346: modlog.create_case() should raise exception not fail silently when passing invalid action_type
Current behaviour is to `return None` when an invalid `action_type` is passed.

```py  
 # redbot/core/modlog.py
case_type = await get_casetype(action_type, guild)
    if case_type is None:
        return
```
[Quick link](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/core/modlog.py#L750)

Dav-Git:
Closes #3346 

Raise ValueError when an invalid casetype is passed to modlog.create_case()

Shouldn't an error also be raised when the bot is passed as the user?
Imo that would then be consistent behaviour within the method.


Dav-Git:
Someone who knows how to fix pytest... pls help (and teach me how to do it)

redbot/core/modlog.py
```diff
@@ -940,16 +940,24 @@ async def create_case(
...
     if case_type is None:
-        return
+        raise ValueError("That is not a valid action type.")
 
     if not await case_type.is_enabled():
         return
 
-    if user == bot.user:
-        return
+    user_id = user if type(user) == int else user.id
```
Jackenmen: We could get away with checking against `type()` if it weren't a user-provided argument but since it isn't, we should use `isinstance` here instead:
```suggestion
    user_id = user if isinstance(user, int) else user.id
```

Jackenmen (dismissed):

redbot/core/modlog.py
```diff
@@ -1013,18 +1013,23 @@ async def create_case(
...
+        If the action type is not a valid action type.
+    RuntimeError
+        If user is the bot itself
     TypeError
         If ``channel`` is of type `discord.PartialMessageable`.
     """
     case_type = await get_casetype(action_type, guild)
     if case_type is None:
-        return
+        raise ValueError("That is not a valid action type.")
```
Flame442: Instead of "that", put the action type that was entered.
```suggestion
        raise ValueError(f"{action_type} is not a valid action type.")
```

redbot/core/modlog.py
```diff
@@ -1013,18 +1013,23 @@ async def create_case(
 
     Raises
     ------
+    ValueError
+        If the action type is not a valid action type.
+    RuntimeError
+        If user is the bot itself
```
Flame442:
```suggestion
        If user is the bot itself.
```

Flame442 (changes requested):

Flame442 (approved):
Thanks!

PR #5387, opened 2021-10-13
[Admin] Add `[p]selfroleset clear` command

Kreusada:

This PR adds `[p]selfroleset clear` which can be used to clear the guild's list of available selfroles.

### DIFF
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -7,6 +7,7 @@ import discord
 from redbot.core import Config, checks, commands
 from redbot.core.i18n import Translator, cog_i18n
 from redbot.core.utils.chat_formatting import box
+from redbot.core.utils.predicates import MessagePredicate
 from .announcer import Announcer
 from .converters import SelfRole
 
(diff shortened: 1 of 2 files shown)


redbot/cogs/admin/admin.py
```diff
@@ -509,6 +510,40 @@ async def selfroleset_remove(self, ctx: commands.Context, *roles: SelfRole):
...
+            return
+        if pred.result:
+            for role in current_selfroles:
+                role = ctx.guild.get_role(role)
+                if role is None:
+                    continue
+                if not self.pass_user_hierarchy_check(ctx, role):
+                    await ctx.send(
+                        _(
+                            "I cannot let you remove {role.name} from being a selfrole because that role is higher than or equal to your highest role in the Discord hierarchy."
```
Vexed01: It might be worth mentioning the whole clear operation was aborted as it is currently ambiguous as to whether this aborted the operation completely or cleared everything except the mentioned role. This sentence is a little long now so feel free ignore the suggestion and reword if you can think of something better than me.
```suggestion
                            "I cannot clear selfroles because I cannot let you remove {role.name} from being a selfrole because that role is higher than or equal to your highest role in the Discord hierarchy."
```

PredaaA (approved):
Looks good. Thanks!

PR #5388, opened 2021-10-13
[Mod] Account for roles in mention spam

### LINKED ISSUE #4471: [Mod] Mention spam accounting for roles
# Feature request

After finalizing https://github.com/Cog-Creators/Red-DiscordBot/pull/4359 (allowing for duplicate mentions). A small discussion was brought up (I forgot with whom), to possibly make something that would / could account for role mentions as well. 

- [ ] Cog
- [x] Command
- [ ] API functionality

This could either be it's own count system (separate from total user mentions. So user mentions could be set to 7, while role mentions could be set to 3), or it would take both user mentions and role mentions into the same context. (each would account for 7 in this example). 

I wanted to make this issue to gather feedback, I am more than willing to work on this myself, though, I want to make 100% sure that this is something worth working on, getting proper feedback on how to set this up and how it should action.

The reason why I believe this should be something for Core Mod is because while it can 100% be used as a third party cog, I figured this would give us and any user who uses core mod better advantages and control over antimentionspam we offer.

Vexed01:
Add up both user and role mentions when checking for mention spam, fixes #4471

The discussion on the original issue did not come to a conclusion on which way to implement this. For both end-user and code simplicity (and my personal preference), I have chosen to not add separate trigger values to config and their associated new commands, instead just totalling up role + user mentions. I would be happy to expand on this and implement one of the variations in the original issue if this is what is preferred, such as separate customisation for user mentions/role mentions.


Kowlin (approved):
Changes work like a charm, thank you.

PR #5390, opened 2021-10-13
Document setting overrides in Trivia list format

### LINKED ISSUE #3977: [Docs] Document CONFIG key for custom Trivia overrides
After looking at #3582 and Trivia, custom overrides are allowed for Trivia files. We should add documentation regarding how to format Trivia files to include these overrides.

Vexed01:
Document the CONFIG key in the Trivia List Creation Guide, fixes #3977

Doc build at https://red-discordbot--5390.org.readthedocs.build/en/5390/guide_trivia_list_creation.html#optional-custom-overrides

Also fix the spelling of medallist :)


Jackenmen:
I'm not sure something like "bool" or "int" tells a regular user (as in, not a developer) much and that's who we target with this document.

Vexed01:
That's certainly a very valid point that I hadn't considered, thanks for pointing it out.

Jackenmen:
Most of these accept any number with or without fractional component but there's also `max_score` which only accepts integers. Do you think this could be improved in some way?

Vexed01:
I've added decimals to the defaults which accept them and clarified "whole number" for `max_score`.

Would it be worth a new issue to actually validate these, as at the moment anything would be accepted?

Jackenmen:
> Would it be worth a new issue to actually validate these, as at the moment anything would be accepted?

They are validated, see #4571/#4659.

As for the change, I'm not entirely sure "whole number" is clearer than "integer" would have been. I'm not a native speaker so someone else might want to chime in here.

Vexed01:
As far as I'm aware, integers can be positive or negative while whole numbers are always positive - and a negative number for max_score (well, as well as 0) wouldn't be good.

I'm not saying that "whole number" is better as I would hope someone advanced enough to touch this wouldn't put in a negative for max score. Though, "integer" is better than "int". Basically idk either and a 3rd opinion would be useful, but from a technical standpoint I'm pretty sure "whole number" is more accurate for what's wanted, and I think it would look odd if "integer" was surrounded by "number" for the floats.

Flame442:
"positive number" / "positive integer" would be my votes, unless it supports zero, in which case "whole number" is best. For the floats, perhaps "number with decimal" or something like that, to avoid the oddity of using just "number" for floats?

Jackenmen:
The valid values for each of these are:
- max_score: positive integer (>0)
- delay: positive number greater than or equal to 4 (>=4.0)
- timeout: positive number (>0.0)
- payout_multiplier: non-negative number (>=0.0)

Additional notes:
- for `timeout`, equivalent `[p]triviaset` command also imposes an additional requirement that this value is higher than `delay`, we might want to restrict in the same way for YAML files
- for all of the numeric fields, we need to be stricter with what we accept per the requirements above as currently we only restrict to int/float

Jackenmen:
#5947 is now merged, this PR still needs updates to the used wording. It's been settled in #5947 so it's just a matter of making it consistent with what we use in error messages there.

docs/guide_trivia_list_creation.rst
```diff
@@ -110,16 +110,67 @@ As you've added more questions, your file should look something like this:
...
+
+So, your final file might look something like this:
+
+.. code-block:: yaml
+
+    AUTHOR: Red
+    CONFIG:
+        bot_plays: false
+        use_spoilers: true
+        delay: 20
```
Jackenmen, line 158: I suggest adding a couple more numeric examples to show what can be passed (hopefully better shows the difference between an integer and a number):
```suggestion
        delay: 20
        payout_multiplier: 0.5
        max_score: 20
```

redbot/cogs/trivia/data/lists/sports.yaml
```diff
@@ -31,7 +31,7 @@ How many times has a Super Bowl went into overtime?:
 How many world records did swimmer Mark Spitz set when he won seven gold medals at the 1972 Olympics?:
 - 7
 - seven
-How much money is a US Olympic gold medalist awarded?:
+How much money is a US Olympic gold medallist awarded?:
 - $25,000
```
Jackenmen: If you want to correct answers in this PR, you should update answers here as well.

docs/guide_trivia_list_creation.rst
```diff
@@ -110,16 +110,67 @@ As you've added more questions, your file should look something like this:
     Who won the Premier League in 2015?:
     - Chelsea
     - chelsea f.c.
-    How much money is a US Olympic gold medalist awarded?:
-    - $25,000
-    - 25,000
-    - 25k
-    - 25000
-    - $25000
+    How much money is a US Olympic gold medallist awarded?:
```
Jackenmen: `medalist` is a correct US spelling:
https://dictionary.cambridge.org/us/dictionary/english/medalist

redbot/cogs/trivia/data/lists/sports.yaml
```diff
@@ -31,7 +31,7 @@ How many times has a Super Bowl went into overtime?:
 How many world records did swimmer Mark Spitz set when he won seven gold medals at the 1972 Olympics?:
 - 7
 - seven
-How much money is a US Olympic gold medalist awarded?:
+How much money is a US Olympic gold medallist awarded?:
```
Jackenmen: `medalist` is a correct US spelling:
https://dictionary.cambridge.org/us/dictionary/english/medalist

Jackenmen (changes requested):

Jackenmen (approved):
Looks good to me, thanks for writing this new section.

PR #5391, opened 2021-10-14
Add plural forms to the responses of `[p]leave` command

the-krak3n:
This PR will improve the response of the `[p]leave` command. Before the response was a little bit messed up but now it will give a better response(idk how to word it).
Before
leave0

After: 
deku

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -1600,10 +1600,7 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
             return await ctx.send(_("You need to specify at least one server ID."))
 
         leaving_local_guild = not guilds
-        s = "s" if len(guilds) > 1 else ""
         number = len(guilds)
-        t = "these" if len(guilds) > 1 else "this" 
-        th = "those" if len(guilds) > 1 else "that" 
 
         if leaving_local_guild:
             guilds = (ctx.guild,)
@@ -1611,9 +1608,16 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
                 _("You haven't passed any server ID. Do you want me to leave this server?")
                 + " (y/n)"
             )
+        if len(guilds) > 1:
+            msg = (
+                _(f"Are you sure you want me to leave these servers?")
+                + " (y/n):\n"
+                + "\n".join(f"- {guild.name} (`{guild.id}`)" for guild in guilds)
+            )
+
         else:
             msg = (
-                _(f"Are you sure you want me to leave {t} server{s}?")
+                _("Are you sure you want me to leave this server")   
                 + " (y/n):\n"
                 + "\n".join(f"- {guild.name} (`{guild.id}`)" for guild in guilds)
             )
@@ -1637,10 +1641,14 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         else:
             if pred.result is True:
                 if leaving_local_guild is True:
-                    await ctx.send(_("Alright. Bye :wave:"))
+                    await ctx.send(_("Alright. Bye :wave:")
+                    )
+                if len(guilds) > 1:
+                    await ctx.send(_(f"Alright. Leaving {number} servers...")
+                    )   
                 else:
                     await ctx.send(
-                        _(f"Alright. Leaving {number} server{s}...")
+                        _(f"Alright. Leaving {number} server...")
                     )
                 for guild in guilds:
                     log.debug("Leaving guild '%s' (%s)", guild.name, guild.id)
@@ -1649,7 +1657,10 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
                 if leaving_local_guild is True:
                     await ctx.send(_("Alright, I'll stay then. :)"))
                 else:
-                    await ctx.send(_(f"Alright, I'm not leaving {th} server{s}."))
+                    if len(guilds) > 1:
+                        await ctx.send(_("Alright, I'm not leaving those servers."))
+                    else:
+                        await ctx.send(_("Alright, I'm not leaving that server."))  
 
     @commands.command()
     @checks.is_owner()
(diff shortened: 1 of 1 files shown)


Jackenmen (changes requested):
This does not work well with non-English locales.
You should instead have entirely separate strings for singular and plural forms.
Here's an example of what I mean (
```py
if len(guilds) > 1:
    _("Are you sure you want me to leave these servers?")
else:
    _("Are you sure you want me to leave this server?")
```

the-krak3n:
so is this alright?

redbot/core/core_commands.py
```diff
@@ -1600,16 +1600,24 @@ async def leave(self, ctx: commands.Context, *servers: GuildConverter):
...
+                + "\n".join(f"- {guild.name} (`{guild.id}`)" for guild in guilds)
+            )
+
         else:
             msg = (
-                _("Are you sure you want me to leave these servers?")
+                _("Are you sure you want me to leave this server")
                 + " (y/n):\n"
                 + "\n".join(f"- {guild.name} (`{guild.id}`)" for guild in guilds)
             )
```
Jackenmen: This changes the logic of this code - this should all be inside an `else` block of the `if leaving_local_guild`. As it is right now, it completely disregards that the bot will be leaving the *current* server.

redbot/core/core_commands.py
```diff
@@ -1634,18 +1642,21 @@ async def leave(self, ctx: commands.Context, *servers: GuildConverter):
             if pred.result is True:
                 if leaving_local_guild is True:
                     await ctx.send(_("Alright. Bye :wave:"))
+                if number > 1:
+                    await ctx.send(_(f"Alright. Leaving {number} servers..."))
                 else:
-                    await ctx.send(
-                        _("Alright. Leaving {number} servers...").format(number=len(guilds))
-                    )
+                    await ctx.send(_(f"Alright. Leaving {number} server..."))
```
Jackenmen: This changes the logic of this code - this should all be inside an `else` block of the `if leaving_local_guild is True`. As it is right now, it results in two messages getting sent when leaving the current server.

redbot/core/core_commands.py
```diff
@@ -1600,16 +1600,24 @@ async def leave(self, ctx: commands.Context, *servers: GuildConverter):
...
+            msg = (
+                _(f"Are you sure you want me to leave these servers?")
+                + " (y/n):\n"
+                + "\n".join(f"- {guild.name} (`{guild.id}`)" for guild in guilds)
+            )
+
         else:
             msg = (
-                _("Are you sure you want me to leave these servers?")
+                _("Are you sure you want me to leave this server")
```
Jackenmen: Missing question mark.
```suggestion
                _("Are you sure you want me to leave this server?")
```

redbot/core/core_commands.py
Jackenmen: No need to use a replacement field in here. Also, f-strings cannot be used for translatable strings.
```suggestion
                    await ctx.send(_("Alright. Leaving one server..."))
```

redbot/core/core_commands.py
```diff
@@ -1634,18 +1642,21 @@ async def leave(self, ctx: commands.Context, *servers: GuildConverter):
             if pred.result is True:
                 if leaving_local_guild is True:
                     await ctx.send(_("Alright. Bye :wave:"))
+                if number > 1:
+                    await ctx.send(_(f"Alright. Leaving {number} servers..."))
```
Jackenmen: F-strings cannot be used for translatable strings.
```suggestion
                    await ctx.send(
                        _("Alright. Leaving {number} servers...").format(number=number)
                    )
```

redbot/core/core_commands.py
```diff
@@ -1600,16 +1600,24 @@ async def leave(self, ctx: commands.Context, *servers: GuildConverter):
...
+                + " (y/n):\n"
+                + "\n".join(f"- {guild.name} (`{guild.id}`)" for guild in guilds)
+            )
+
         else:
             msg = (
-                _("Are you sure you want me to leave these servers?")
+                _("Are you sure you want me to leave this server")
                 + " (y/n):\n"
                 + "\n".join(f"- {guild.name} (`{guild.id}`)" for guild in guilds)
```
Jackenmen: This can be simplified considering we're only dealing with a single guild:
```suggestion
                + f"- {guilds[0].name} (`{guilds[0].id}`)"
```

Jackenmen (changes requested):

redbot/core/core_commands.py
the-krak3n: oh my bad , also i just tried to fix it like this but it didnt work for me 
```python
       if leaving_local_guild:
            guilds = (ctx.guild,)
            msg = (
                _("You haven't passed any server ID. Do you want me to leave this server?")
                + " (y/n)"
            )
        else:
            if number > 1:
                msg = (
                    _(f"Are you sure you want me to leave these servers?")
                    + " (y/n):\n"
                    + "\n".join(f"- {guild.name} (`{guild.id}`)" for guild in guilds)
                )
            else:                
                msg = (
                    _("Are you sure you want me to leave this server?")
                    + " (y/n):\n"
                    + f"- {guilds[0].name} (`{guilds[0].id}`)"
                )
```

Jackenmen: What exactly didn't work?

the-krak3n: The expected outcome from the bot when the command is run without passing any guild id was "You haven't passed any server ID. Do you want me to leave this server?" but it sent "Are you sure you want me to leave this server?". Well should i commit the changes?

Jackenmen: Yeah, if you commit the changes, I can try to see what's wrong.

the-krak3n: Well nvm its working well 👍 I just forgot to update my core path of bot so was facing some issues sorry for that.

Jackenmen (approved):
LGTM!

PR #5393, opened 2021-10-15
Prevent unexpected timedelta matches

### LINKED ISSUE #5385: Relative/time delta parsing regex can have unexpected outcomes

3.5.0.dev1

Test some of new code, which uses the `parse_relativedelta` function, which calls the culprit [`_parse_and_match` function](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/core/commands/converter.py#L74), with "1 decade" as this is not an allowed or programmed unit.

Nothing to be parsed and None to be returned (or in a converter a friendly error message), because `decade` is not in the regex or an allowed unit.

### What actually happened?

A decade was parsed as a day. Slight difference /s

1. Use any command with time converter (eg remindme, tempban, mute) or the actual function in an eval and pass "1 decade"
2. See how it is parsed as "1 day"

### Anything else?

The regex used in `_parse_and_match` will allow for any word beginning with y, m, w, d, s to be parsed as the respective non-short time, such as "decade" being caught under "d" for day in the example below.

This would also affect the timedelta and telativedelta converters, which use this function.

This does not just affect the part of the regex for days, but decade is just what I thought of - the fix for this shouldn't just single out "decade".

Example using the function directly:
image

Flame442:
Requires groups to end with whitespace or eof before accepting them as valid.

Fixes #5385.

### DIFF
--- a/redbot/core/commands/converter.py
+++ b/redbot/core/commands/converter.py
@@ -58,13 +58,13 @@ ID_REGEX = re.compile(r"([0-9]{15,20})")
 # https://github.com/mikeshardmind/SinbadCogs/blob/816f3bc2ba860243f75112904b82009a8a9e1f99/scheduler/time_utils.py#L9-L19
 TIME_RE_STRING = r"\s?".join(
     [
-        r"((?P<years>\d+?)\s?(years?|y)($|\s))?",
-        r"((?P<months>\d+?)\s?(months?|mo)($|\s))?",
-        r"((?P<weeks>\d+?)\s?(weeks?|w)($|\s))?",
-        r"((?P<days>\d+?)\s?(days?|d)($|\s))?",
-        r"((?P<hours>\d+?)\s?(hours?|hrs|hr?)($|\s))?",
-        r"((?P<minutes>\d+?)\s?(minutes?|mins?|m(?!o))($|\s))?",  # prevent matching "months"
-        r"((?P<seconds>\d+?)\s?(seconds?|secs?|s)($|\s))?",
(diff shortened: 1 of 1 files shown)


Vexed01 (changes requested):
This will break the matching of single letters when there are multiple together, for example `1d6h` would only parse the 6h part.

This format is used in the docstring examples of quite a few commands that use the timedelta converter in core (such as [tempban](https://github.com/Cog-Creators/Red-DiscordBot/blob/ad92110f0e3b79bff99267ed51ec1cd4a00deb27/redbot/cogs/mod/kickban.py#L599)), and I would also imagine so in 3rd party cogs.

Flame442:
That commit should fix the issue you were having (and seems to make the regex more efficient according to regex101 :P)

Vexed01 (dismissed):
LGTM

Jackenmen:
Hmm, would there be any harm in doing a fullmatch instead of a match in `_parse_and_match()`?
This would additionally prevent us from matching something like "1 year 2 quarters" as 1 year even though technically that wasn't the user's entire input.

It would be a tiny change in behavior for `parse_timedelta`/`parse_relativedelta` so we would have to delay this to 3.5.0 but I'm more wondering if I'm missing some **big** issue that could be really problematic for someone's usage if we were to change this?

Jackenmen (dismissed):
Functionally, this is sound. Depending on the response to my comment, this PR might still need changes though.

Jackenmen (approved):
LGTM!

PR #5394, opened 2021-10-15
[Audio] Fix UnboundLocalError in edge case

### LINKED ISSUE #5378: playlist list causing UnboundLocalError

3.4.14

### Cog name

Audio

### Command name

playlist list

I first created an empty playlist, and then called `playlist list`, where I expected Red to show me the empty playlist.

### What actually happened?

Red confirmed that it created the empty playlist, but reported that no playlists were found.

1. `[p] playlist create playlistName`
2. `[p] playlist list`

### Anything else?

I found an error in my log files, an `UnboundLocalError` as referenced in the title.
image
Additionally, here is the output of `[p] debuginfo`.
image

Flame442:
https://github.com/Cog-Creators/Red-DiscordBot/blob/334cd4fa2a2d89a0df87d8fa72714edb665db2a3/redbot/cogs/audio/apis/playlist_wrapper.py#L107-L111
This exception handling will always lead to an `UnboundLocalError` in `row_result`, as it is not defined when an exception occurs. Though there appears to be no way an exception should be raised under normal conditions, this PR nevertheless fixes that error.

https://github.com/Cog-Creators/Red-DiscordBot/blob/334cd4fa2a2d89a0df87d8fa72714edb665db2a3/redbot/cogs/audio/apis/playlist_wrapper.py#L139-L143
The one location that does not lead to an `UnboundLocalError` handles it by returning the base case for when no playlists are found.

https://github.com/Cog-Creators/Red-DiscordBot/blob/334cd4fa2a2d89a0df87d8fa72714edb665db2a3/redbot/cogs/audio/apis/playlist_wrapper.py#L111-L114
For `fetch`, the base case is `None`, as `fetchone` will return either a row or `None`, and the `None` case simply returns the `None`.

https://github.com/Cog-Creators/Red-DiscordBot/blob/334cd4fa2a2d89a0df87d8fa72714edb665db2a3/redbot/cogs/audio/apis/playlist_wrapper.py#L196-L198
For `fetch_all_converter`, the base case is `[]`, as that is the default value of `output`, and it is returned when no rows are found.

These base cases were used as a return value for the fetch exception handling.
Fixes #5378.

### DIFF
--- a/redbot/cogs/audio/apis/playlist_wrapper.py
+++ b/redbot/cogs/audio/apis/playlist_wrapper.py
@@ -84,7 +84,7 @@ class PlaylistWrapper:
             table = 2
         return table
 
-    async def fetch(self, scope: str, playlist_id: int, scope_id: int) -> PlaylistFetchResult:
+    async def fetch(self, scope: str, playlist_id: int, scope_id: int) -> Optional[PlaylistFetchResult]:
         """Fetch a single playlist."""
         scope_type = self.get_scope_type(scope)
 
--- a/redbot/cogs/audio/apis/playlist_wrapper.py
+++ b/redbot/cogs/audio/apis/playlist_wrapper.py
@@ -107,7 +107,8 @@ class PlaylistWrapper:
                 try:
                     row_result = future.result()
                 except Exception as exc:
-                    debug_exc_log(log, exc, "Failed to completed playlist fetch from database")
+                    debug_exc_log(log, exc, "Failed to complete playlist fetch from database")
+                    return None
             row = row_result.fetchone()
             if row:
                 row = PlaylistFetchResult(*row)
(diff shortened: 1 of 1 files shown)


aikaterna:
https://github.com/Cog-Creators/Red-DiscordBot/blob/de53d15cf8d64778979862946529537701c45f9d/redbot/cogs/audio/apis/playlist_interface.py#L388 seems to be the only usage of .fetch(). Returning None there is handled, but could I ask you to add Optional to the PlaylistFetchResult typehint on https://github.com/Cog-Creators/Red-DiscordBot/blob/aa8e11189c352efeb50bbd655c2d85ac857757f7/redbot/cogs/audio/apis/playlist_wrapper.py#L87?

Flame442:
All set

aikaterna (approved):

Issue #5244, opened 2021-09-06
Use new RelativedeltaConverter/parse_relativedelta in core cogs


Jackenmen:
### What component of Red (cog, command, API) would you like to see improvements on?

Mod, Mutes

### Describe the enhancement you're suggesting.

After looking through existing commands, the commands that could benefit from this are `[p]tempban` (and by extension, also `[p]modset defaultduration`) and the mute commands in Mutes cog. Both of these however are not just a simple find&replace.

Change to `[p]tempban` itself is simple, however we should also support setting a relativedelta as the default duration in `[p]modset defaultduration`. I suppose that the easiest acceptable way of doing it would be to have a custom converter that accepts a string, uses `parse_relativedelta` on it, and returns the original string which we can then save to Config. There would also be a need for a config migration that would convert the existing values (integers) to the new format. An alternative solution for storing would be to use a serialized relativedelta - dictionary with the proper kwargs for relativedelta.

As for mute commands, the above applies but the regex for Mutes's cogs custom MuteTime converter is a bit different to allow time interval both at the beginning and the end of the mute reason. I'm guessing that we will just use relativedelta here directly rather than the converter.

### Anything else?

_Originally posted by @jack1142 in https://github.com/Cog-Creators/Red-DiscordBot/pull/5000#pullrequestreview-746704199_

PR #5402, opened 2021-10-16
[Mutes/Core] RelativeDeltaWithReason

Flame442:
This PR takes the converter from the mutes cog which converts a string into a timedelta and a string reason and adds it to `redbot.core.commands.converter`. Also implements a version of this converter which produces a relativedelta and a string reason, as well as functions which can be used to just do the conversion process on an arbitrary string. Swaps `mutes` to use `redbot.core.commands.converter` converters, including these.

Specific implementation details are subject to change, including typehints and method names. Some things I really just did not understand are omitted. This PR is untested beyond ensuring it works under the most expected of situations.

This PR serves to satisfy the `mutes` part of #5244. This PR implements the changes to the regex present in #5393 in new regex strings, however does not modify the existing strings in an attempt to not conflict with that PR.


flaree:
Can you solve the 2 conflicts when you get a chance?

Flame442:
Conflicts have been resolved

PR #5403, opened 2021-10-17
Update JDK URL and add Python version upper-bound in Windows guide

yamikaitou:

Update the Java download URL for manual dependencies with Adopt's move to Temurin


Jackenmen (approved):
Looks good to me, thanks

PR #5404, opened 2021-10-17
[Mutes] Fix vars not getting formatted due to previous PR

Flame442:
#5364 introduced a bug where variables would not get formatted in some cases. This fixes that issue.

~~sorry for not catching it in the review~~


redbot/cogs/mutes/mutes.py
```diff
@@ -1025,8 +1028,6 @@ async def _check_for_mute_role(self, ctx: commands.Context) -> bool:
                     "saying {response_2} will end the mute attempt."
                 )
```
Jackenmen, line 1032: Could you move the format call from the line below here? This way we will avoid calling .format() for the second time on a string that already was formatted once.
Due to our past experiences, I would go as far as calling it an anti-pattern 😅

Jackenmen (approved):
Thanks for fixing my mistakes :)

PR #5407, opened 2021-10-17
Stop muted users from creating threads

### LINKED ISSUE #5371: [Mutes] Deny thread creation

Muting

Also deny the ability to create threads. This can be easily abused as the default role made can still create threads.

### Anything else?

_No response_

Jan200101:
closes #5371

Does what it says on the tin

Permission bits for creating public and private threads are stored privately in the class with a DEP-WARN comment saying to remove these when time comes.

The logic is quite simply and just bitewise or's all create thread permission, inverts it and then does a bitwise end on the permission value.

### DIFF
--- a/redbot/cogs/mutes/mutes.py
+++ b/redbot/cogs/mutes/mutes.py
@@ -110,6 +110,10 @@ class Mutes(VoiceMutes, commands.Cog, metaclass=CompositeMetaClass):
 
         self._init_task = self.bot.loop.create_task(self._initialize())
 
+        # DEP-WARN: Remove when updating to Discord.py 2.0
+        self._create_public_threads: int = 1 << 35
+        self._create_private_threads: int = 1 << 36
+
     async def red_delete_data_for_user(
         self,
         *,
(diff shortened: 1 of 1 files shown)


redbot/cogs/mutes/mutes.py
```diff
@@ -1664,6 +1671,9 @@ async def channel_mute_user(
         new_overs: dict = {}
         move_channel = False
         new_overs.update(send_messages=False, add_reactions=False, speak=False)
+        new_overs.value &= ~(
+            self._create_public_threads ^ self._create_private_threads
+        )  # DEP-WARN Remove when updating to Discord.py 2.0
```
Jackenmen, line 1676: This won't work with a dictionary. Looking at how `PermissionOverwrite` works, I don't think there's a way to make this work before d.py 2.0. If that's the case, this PR can be closed and we'll just wait and properly implement it after we switch.

Jan200101: agreed

The only way I can see to do this completely is to patch multiple parts of discord.py, which is not the best option

PR #5389, opened 2021-10-13
Audio API

benno1237:
### (Semi) Functional API for Red's audio cog

this is proof of concept PR and is far from being ready to review yet. I am however more than happy about any comments/suggestions/contributions. It's quite a lot of code so expect some things to break please. I will thoroughly test this in the near future and hopefully eliminate each and every issue.  
Since it is just a draft PR for now, I also didn't really care about code style tools like isort and black.

Yes, it adds 10k+ lines. Atm, there is heaps of duplicated code in the API and core audio. My idea was however to keep the initial PR as "small" as somehow possible. Thus core audio only has the changes which are needed in order for it to work (hopefully?). IF an interest exists however, I will go ahead and rewrite the majority of core audio to use the functionality provided by this API.

Semi good docs of what it includes can be seen here: https://red-discordbot-audio.readthedocs.io/en/latest/framework_audio.html

Please, ignore the messy commit history.

PR #5408, opened 2021-10-17
Audio config optimization

benno1237:
Sums up config calls in core audio by using .all()
Gets rid of guild_from_id wherever possible

Some config things I noticed while writing #5389. Currently, audio uses .all occasionally but sometimes also does multiple single config calls in a row. Sometimes even both combined in a single command. This PR fixes that

### DIFF
--- a/redbot/cogs/audio/core/commands/controller.py
+++ b/redbot/cogs/audio/core/commands/controller.py
@@ -34,9 +34,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
             return await self.send_embed_msg(ctx, title=_("Nothing playing."))
         else:
             guild_data = await self.config.guild(ctx.guild).all()
-            dj_enabled = self._dj_status_cache.setdefault(
-                ctx.guild.id, guild_data["dj_enabled"]
-            )
+            dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
             vote_enabled = guild_data["vote_enabled"]
             player = lavalink.get_player(ctx.guild.id)
             can_skip = await self._can_instaskip(ctx, ctx.author)
@@ -74,9 +72,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
                 await self.config.custom("EQUALIZER", ctx.guild.id).eq_bands.set(eq.bands)
             await player.stop()
             await player.disconnect()
-            await self.config.guild(ctx.guild).currently_auto_playing_in.set(
-                []
-            )
+            await self.config.guild(ctx.guild).currently_auto_playing_in.set([])
             self._ll_guild_updates.discard(ctx.guild.id)
             await self.api_interface.persistent_queue_api.drop(ctx.guild.id)
 
@@ -147,9 +143,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
 
         player.store("np_message", message)
 
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, guild_data["dj_enabled"]
-        )
+        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
         vote_enabled = guild_data["vote_enabled"]
         if (
             (dj_enabled or vote_enabled)
@@ -244,9 +238,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
         guild_data = await self.config.guild(ctx.guild).all()
         if not self._player_check(ctx):
             return await self.send_embed_msg(ctx, title=_("Nothing playing."))
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, guild_data["dj_enabled"]
-        )
+        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
         vote_enabled = guild_data["vote_enabled"]
         is_alone = await self.is_requester_alone(ctx)
         is_requester = await self.is_requester(ctx, ctx.author)
@@ -309,9 +301,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
         Accepts seconds or a value formatted like 00:00:00 (`hh:mm:ss`) or 00:00 (`mm:ss`).
         """
         guild_data = await self.config.guild(ctx.guild).all()
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, guild_data["dj_enabled"]
-        )
+        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
         vote_enabled = guild_data["vote_enabled"]
         is_alone = await self.is_requester_alone(ctx)
         is_requester = await self.is_requester(ctx, ctx.author)
@@ -393,9 +383,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
         """Toggle shuffle."""
         guild_data = await self.config.guild(ctx.guild).all()
         if ctx.invoked_subcommand is None:
-            dj_enabled = self._dj_status_cache.setdefault(
-                ctx.guild.id, guild_data["dj_enabled"]
-            )
+            dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
             can_skip = await self._can_instaskip(ctx, ctx.author)
             if dj_enabled and not can_skip:
                 return await self.send_embed_msg(
@@ -438,9 +426,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
         over `[p]shuffle`.
         """
         guild_data = await self.config.guild(ctx.guild).all()
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, guild_data["dj_enabled"]
-        )
+        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
         can_skip = await self._can_instaskip(ctx, ctx.author)
         if dj_enabled and not can_skip:
             return await self.send_embed_msg(
@@ -491,9 +477,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
         if not player.current:
             return await self.send_embed_msg(ctx, title=_("Nothing playing."))
         guild_data = await self.config.guild(ctx.guild).all()
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, guild_data["dj_enabled"]
-        )
+        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
         vote_enabled = guild_data["vote_enabled"]
         is_alone = await self.is_requester_alone(ctx)
         is_requester = await self.is_requester(ctx, ctx.author)
@@ -570,9 +554,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
     async def command_stop(self, ctx: commands.Context):
         """Stop playback and clear the queue."""
         guild_data = await self.config.guild(ctx.guild).all()
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, guild_data["dj_enabled"]
-        )
+        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
         vote_enabled = guild_data["vote_enabled"]
         if not self._player_check(ctx):
             return await self.send_embed_msg(ctx, title=_("Nothing playing."))
@@ -614,9 +596,7 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
             player.store("requester", None)
             player.store("autoplay_notified", False)
             await player.stop()
-            await self.config.guild(ctx.guild).currently_auto_playing_in.set(
-                []
-            )
+            await self.config.guild(ctx.guild).currently_auto_playing_in.set([])
             await self.send_embed_msg(ctx, title=_("Stopping..."))
             await self.api_interface.persistent_queue_api.drop(ctx.guild.id)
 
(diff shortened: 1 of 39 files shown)


redbot/cogs/audio/core/commands/controller.py
```diff
@@ -33,10 +33,14 @@ async def command_disconnect(self, ctx: commands.Context):
...
-            )
-            vote_enabled = await self.config.guild(ctx.guild).vote_enabled()
+            guild_data = await self.config.guild(ctx.guild).all()
+            vote_enabled = guild_data["vote_enabled"]
+            if ctx.guild.id in self._dj_status_cache:
+                dj_enabled = self._dj_status_cache[ctx.guild.id]
+            else:
+                dj_enabled = self._dj_status_cahce.setdefault(
+                    ctx.guild.id, guild_data["dj_enabled"]
+                )
```
Jackenmen, line 43: These are unnecessary changes to the code logic, let's just keep it as similar to the original code as possible since this PR is meant for config optimizations, not moving code around (especially when the reason for this moving is unknown).
```suggestion
            guild_data = await self.config.guild(ctx.guild).all()
            dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
            vote_enabled = guild_data["vote_enabled"]
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -242,9 +248,12 @@ async def command_prev(self, ctx: commands.Context):
...
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-        )
+        if ctx.guild.id in self._dj_status_cache:
+            dj_enabled = self._dj_status_cache[ctx.guild.id]
+        else:
+            dj_enabled = self._dj_status_cache.setdefault(
+                ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
+            )
         vote_enabled = await self.config.guild(ctx.guild).vote_enabled()
```
Jackenmen, line 257: I'm pretty sure you wanted to switch to call to `all()` here:
```suggestion
        guild_data = await self.config.guild(ctx.guild).all()
        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
        vote_enabled = guild_data["vote_enabled"]
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -306,9 +315,12 @@ async def command_seek(self, ctx: commands.Context, seconds: Union[int, str]):
...
```
Jackenmen, line 324: I'm pretty sure you wanted to switch to call to `all()` here:
```suggestion
        guild_data = await self.config.guild(ctx.guild).all()
        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
        vote_enabled = guild_data["vote_enabled"]
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -389,9 +401,12 @@ async def command_seek(self, ctx: commands.Context, seconds: Union[int, str]):
...
         if ctx.invoked_subcommand is None:
-            dj_enabled = self._dj_status_cache.setdefault(
-                ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-            )
+            if ctx.guild.id in self._dj_status_cache:
+                dj_enabled = self._dj_status_cache[ctx.guild.id]
+            else:
+                dj_enabled = self._dj_status_cache.setdefault(
+                    ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
+                )
```
Jackenmen, line 409: If we already know that the ID isn't in cache, we don't need to use `setdefault()`:
```suggestion
                dj_enabled = self._dj_status_cache[ctx.guild.id] = (
                    await self.config.guild(ctx.guild).dj_enabled()
                )
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -433,9 +448,12 @@ async def command_shuffle_bumpped(self, ctx: commands.Context):
...
         """
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-        )
+        if ctx.guild.id in self._dj_status_cache:
+            dj_enabled = self._dj_status_cache[ctx.guild.id]
+        else:
+            dj_enabled = self._dj_status_cache.setdefault(
+                ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
+            )
```
Jackenmen, line 456: If we already know that the ID isn't in cache, we don't need to use `setdefault()`:
```suggestion
            dj_enabled = self._dj_status_cache[ctx.guild.id] = (
                await self.config.guild(ctx.guild).dj_enabled()
            )
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -146,10 +150,12 @@ async def command_now(self, ctx: commands.Context):
...
 
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-        )
-        vote_enabled = await self.config.guild(ctx.guild).vote_enabled()
+        if ctx.guild.id in self._dj_status_cache:
+            dj_enabled = self._dj_status_cache[ctx.guild.id]
+        else:
+            dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
+
```
Jackenmen, line 157: These are unnecessary changes to the code logic, let's just keep it as similar to the original code as possible since this PR is meant for config optimizations, not moving code around (especially when the reason for this moving is unknown).
```suggestion
        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -485,10 +503,12 @@ async def command_skip(self, ctx: commands.Context, skip_to_track: int = None):
...
             return await self.send_embed_msg(ctx, title=_("Nothing playing."))
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-        )
-        vote_enabled = await self.config.guild(ctx.guild).vote_enabled()
+        guild_data = await self.config.guild(ctx.guild).all()
+        if ctx.guild.id in self._dj_status_cache:
+            dj_enabled = self._dj_status_cache[ctx.guild.id]
+        else:
+            dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
```
Jackenmen, line 510: These are unnecessary changes to the code logic, let's just keep it as similar to the original code as possible since this PR is meant for config optimizations, not moving code around (especially when the reason for this moving is unknown).
```suggestion
        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -563,9 +583,12 @@ async def command_skip(self, ctx: commands.Context, skip_to_track: int = None):
...
         """Stop playback and clear the queue."""
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-        )
+        if ctx.guild.id in self._dj_status_cache:
+            dj_enabled = self._dj_status_cache[ctx.guild.id]
+        else:
+            dj_enabled = self._dj_status_cache.setdefault(
+                ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
+            )
```
Jackenmen, line 591: If we already know that the ID isn't in cache, we don't need to use `setdefault()`:
```suggestion
            dj_enabled = self._dj_status_cache[ctx.guild.id] = (
                await self.config.guild(ctx.guild).dj_enabled()
            )
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -619,10 +642,12 @@ async def command_stop(self, ctx: commands.Context):
...
         """Summon the bot to a voice channel."""
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-        )
-        vote_enabled = await self.config.guild(ctx.guild).vote_enabled()
+        guild_data = await self.config.guild(ctx.guild).all()
+        if ctx.guild.id in self._dj_status_cache:
+            dj_enabled = self._dj_status_cache[ctx.guild.id]
+        else:
+            dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
```
Jackenmen, line 649: These are unnecessary changes to the code logic, let's just keep it as similar to the original code as possible since this PR is meant for config optimizations, not moving code around (especially when the reason for this moving is unknown).
```suggestion
        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -697,14 +722,16 @@ async def command_summon(self, ctx: commands.Context):
...
     async def command_volume(self, ctx: commands.Context, vol: int = None):
         """Set the volume, 1% - 150%."""
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-        )
+        guild_data = await self.config.guild(ctx.guild).all()
+        if ctx.guild.id in self._dj_status_cache:
+            dj_enabled = self._dj_status_cache[ctx.guild.id]
+        else:
+            dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
```
Jackenmen, line 729: These are unnecessary changes to the code logic, let's just keep it as similar to the original code as possible since this PR is meant for config optimizations, not moving code around (especially when the reason for this moving is unknown).
```suggestion
        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -697,14 +722,16 @@ async def command_summon(self, ctx: commands.Context):
...
         can_skip = await self._can_instaskip(ctx, ctx.author)
-        max_volume = await self.config.guild(ctx.guild).max_volume()
+        max_volume = guild_data["maxvolume"]
```
Jackenmen, line 731: typo
```suggestion
        max_volume = guild_data["max_volume"]
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -744,9 +771,12 @@ async def command_volume(self, ctx: commands.Context, vol: int = None):
...
         """Toggle repeat."""
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-        )
+        if ctx.guild.id in self._dj_status_cache:
+            dj_enabled = self._dj_status_cache[ctx.guild.id]
+        else:
+            dj_enabled = self._dj_status_cache.setdefault(
+                ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
+            )
```
Jackenmen, line 779: If we already know that the ID isn't in cache, we don't need to use `setdefault()`:
```suggestion
            dj_enabled = self._dj_status_cache[ctx.guild.id] = (
                await self.config.guild(ctx.guild).dj_enabled()
            )
```

redbot/cogs/audio/core/commands/playlists.py
```diff
@@ -1471,9 +1471,12 @@ async def command_playlist_start(
...
         scope, author, guild, specified_user = scope_data
-        dj_enabled = self._dj_status_cache.setdefault(
-            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-        )
+
+        guild_data = await self.config.guild(ctx.guild).all()
+        if ctx.guild.id in self._dj_status_cache:
+            dj_enabled = self._dj_status_cache[ctx.guild.id]
+        else:
+            dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
```
Jackenmen, line 1479: These are unnecessary changes to the code logic, let's just keep it as similar to the original code as possible since this PR is meant for config optimizations, not moving code around (especially when the reason for this moving is unknown).
```suggestion
        dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, guild_data["dj_enabled"])
```

Jackenmen (changes requested):

redbot/cogs/audio/core/commands/controller.py
aikaterna: > These are unnecessary changes to the code logic, let's just keep it as similar to the original code as possible since this PR is meant for config optimizations, not moving code around (especially when the reason for this moving is unknown).

More information about why those changes were made can be found at: https://discord.com/channels/133049272517001216/171665724262055936/902639810294988881

For clarity I will repeat some of what I said in the convo at that link when I partially reviewed these changes in that discussion without commenting on this issue:

___
Instead of 
```
dj_enabled = self._dj_status_cache.setdefault(
            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
        )
```
Change it to:
```
if ctx.guild.id in self._dj_status_cache:
    dj_enabled = self._dj_status_cache[ctx.guild.id]
else:
    dj_enabled = self._dj_status_cache.setdefault(ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled())
```

2 - Keep guild_from_id as is as it serves as a demo example for newcomers and that's what core cogs are supposed to do in my opinion
3 - Be considerate of .all() calls as they can be expensive, especially on larger bots
3.1 - Sometimes 2 config calls is cheaper than an .all() so if a command has early exits that's likely why .all() wasnt used on it... as making an expensive .all() call would be a waste of resources and potentially slow the bot down (like a user spamming [p]repeat when DJ mode is enabled)
___

Jack, if you want changes like the example here done in another PR that's understandable and I don't need to fight against it, this comment is for providing clarity on why changes like this were added to this PR.

redbot/cogs/audio/core/commands/controller.py
aikaterna: > I'm pretty sure you wanted to switch to call to `all()` here:

While I was talking with @Drapersniper about these changes and setdefault, he said this to me:

The idea behind dj_enabled here is to cache is in an attribute
and NEVER call config for it if its in the cache
his "improvement" here calls .all()
which means its always calling the dj_attribute
while i say this 
```
dj_enabled = self._dj_status_cache.setdefault(
            ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
        )
```
this right now is broken in audio
because of the way setdefault works
so its not "broken" but its for behaving as intended
the idea here is self._dj_status_cache.setdefault will only call await self.config.guild(ctx.guild).dj_enabled() if the cache for this guild is empty
but thats not how setdefault works unfortunately
which means await self.config.guild(ctx.guild).dj_enabled() ALWAYS gets called

Jackenmen:
It's been over 4 years since the last changes on this PR were made, and there are still unresolved review comments on this PR. It's time to just close this.

PR #5412, opened 2021-10-20
Fix CodeQL workflow

RasmusWL:
Hi :wave: I'm working on the CodeQL analysis of Python, and have looked into your recent problems with the codeql-action, which this PR should solve :blush: 

The new pip 21.3 (released 2021-10-11) uses in-tree builds when `pip install .` is used (see [release notes](https://pip.pypa.io/en/stable/news/#v21-3) and [implementation PR](https://github.com/pypa/pip/pull/10495)). This means CodeQL will see all files twice (once in the real location that is part of this repository, and once in the copy in the `build/` dir), which trips up the analysis.

When an editable install is used instead with `pip install -e .`, there is no `build/` dir, so the analysis will work again.

/cc @jack1142

### DIFF
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -27,7 +27,7 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install -U pip setuptools wheel
-        python -m pip install .[all]
+        python -m pip install -e .[all]
         # Set the `CODEQL-PYTHON` environment variable to the Python executable
         # that includes the dependencies
         echo "CODEQL_PYTHON=$(which python)" >> $GITHUB_ENV


HeyLaurel:
Sorry, was out Monday and Tuesday, I just emailed the user, feel free to close if needed and I can re-open if there are more issues

Jackenmen:
Looks like it works just fine now, thanks!

> This means CodeQL will see all files twice (once in the real location that is part of this repository, and once in the copy in the `build/` dir), which trips up the analysis.

Is there some tracking issue for this?

Jackenmen (approved):

RasmusWL:
Thanks @jack1142 

> Is there some tracking issue for this?

Not currently no. Do you see any benefit of doing `pip install .` instead of `pip install -e .`? -- my plan was to warn people if they did `pip install .` and recommending that they do `pip install -e .` instead, but if there are significant benefits of doing `pip install .` that I am not aware of, I might have to reconsider that approach.

Jackenmen:
> my plan was to warn people if they did `pip install .` and recommending that they do `pip install -e .` instead, but if there are significant benefits of doing `pip install .` that I am not aware of, I might have to reconsider that approach.

If this would also be part of Code scanning documentation as well, I think it's a fine solution. Currently, the documentation only shows how to do this when one has a `requirements.txt` file but it is common for the package developers to only set the dependencies in the build system configuration which is where `pip install -e .` would come handy:
https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning#analyzing-python-dependencies

> Do you see any benefit of doing `pip install .` instead of `pip install -e .`?

I automatically go for `pip install` without `-e` for anything done in CI because that puts it closer to the way the user would have it when installing the package from PyPI. It doesn't really matter for CodeQL analysis though as it really just needs all the dependencies installed so that it can include them in analysis and `pip install -e .` is more than enough for that. Probably also faster by a few seconds 😄

RasmusWL:
Great, thanks for clearing that up :+1: Adding it to the documentation is definitely on my list of things to do, so hopefully it'll get updated soon :wink:

RasmusWL:
Docs PR opened now: https://github.com/github/docs/pull/11329

PR #5416, opened 2021-10-20
Fix issues with loading config.json when it doesn't exist

### LINKED ISSUE #5415: Trying to run a red with no instance fails on fresh install

3.4.14

Trying to run an red without an instance since I'm lazy and wanted to test something and call it a day.

Start the red normally without an instance

### What actually happened?

Getting this if i do `redbot --no-instance --token TOKEN --prefix ~ --owner MY ID` 

```py
Warning: The data will be placed in a temporary folder and removed on next system reboot.
Unexpected exception (<class 'FileNotFoundError'>): 
  File "/Users/kato/Downloads/red/redenv/lib/python3.9/site-packages/redbot/__main__.py", line 503, in main
    data_manager.create_temp_config()
  File "/Users/kato/Downloads/red/redenv/lib/python3.9/site-packages/redbot/core/data_manager.py", line 64, in create_temp_config
    with config_file.open("r", encoding="utf-8") as fs:
  File "/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 1242, in open
    return io.open(self, mode, buffering, encoding, errors, newline,
  File "/usr/local/Cellar/python@3.9/3.9.6/Frameworks/Python.framework/Versions/3.9/lib/python3.9/pathlib.py", line 1110, in _opener
    return self._accessor.open(self, flags, mode)
FileNotFoundError: [Errno 2] No such file or directory: '/Users/kato/Library/Application Support/Red-DiscordBot/config.json'
```

1. Install red in a fresh system (cause didn't happen on my main vps)
2. Do not setup an instance and go to step 3.
3. Start red with `--no-instance` flag

### Anything else?

_No response_

Vexed01:
Move `load_existing_config` to `data_manager.py` and add a docstring, and utilise it in `create_temp_config`, so that FileNotFoundError is handled gracefully.

Primarily fixes #5415, tested by changing the path Red looks for the config file (at L46, `data_managager.py`) to a non-existent file.

### DIFF
--- a/redbot/core/data_manager.py
+++ b/redbot/core/data_manager.py
@@ -61,8 +61,11 @@ def create_temp_config():
     default_dirs["STORAGE_TYPE"] = "JSON"
     default_dirs["STORAGE_DETAILS"] = {}
 
-    with config_file.open("r", encoding="utf-8") as fs:
-        config = json.load(fs)
+    try:
+        with config_file.open("r", encoding="utf-8") as fs:
+            config = json.load(fs)
+    except FileNotFoundError:
+        config = {}
 
     config[name] = default_dirs
 
--- a/redbot/core/data_manager.py
+++ b/redbot/core/data_manager.py
@@ -46,6 +46,21 @@ if not config_dir:
 config_file = config_dir / "config.json"
 
 
+def load_existing_config():
+    """Get the contents of the config file, or an empty dictionary if it does not exist.
+
+    Returns
+    -------
+    dict
+        The config data.
+    """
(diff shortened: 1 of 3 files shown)


Jackenmen (commented):
How about we instead move `load_existing_config()` from `redbot/setup.py` to `redbot/core/data_manager.py` and reuse it here?

Jackenmen (approved):
Looks good!

PR #5417, opened 2021-10-20
Document more of Red's events

### LINKED ISSUE #3787: [Docs] Document events
# Feature request

- [X] Documentation

Currently only RPC events are documented here https://red-discordbot.readthedocs.io/en/stable/framework_events.html .
Other custom events might be useful for cog devs.
Obviously we don't need to document the events that are in dpy already but the new ones would be nice.

List that draper made:
- ``on_message_without_command``
- ``on_red_api_tokens_update``
- ``on_filter_message_delete``
- ``on_trivia_end``
- ``on_modlog_case_edit``
- ``on_modlog_case_create``
- all the audio ones

benno1237:

Documenting the custom events which I thought are worth being documented
The change in docs/conf.py is needed to allow proper documentation of the audio events

Potentially closes #3787

### DIFF
--- a/docs/framework_events.rst
+++ b/docs/framework_events.rst
@@ -4,6 +4,95 @@
 Custom Events
 =============
 
+Audio
+^^^^^
+
+.. py:method:: Red.on_red_audio_audio_disconnect(guild)
+
+    Dispatched when a player disconnects from a channel
+
+    :param discord.Guild guild: The guild in which the player disconnected
+
+.. py:method:: Red.on_red_audio_queue_end(guild, prev_track, prev_requester)
+
+    Dispatched when the queue ends
+
+    :param discord.Guild guild: The guild in which the queue ended
+    :param lavalink.Track prev_track: The last track
+    :param discord.Member prev_requester: The track's requester
+
+.. py:method:: Red.on_red_audio_skip_track(guild, prev_track, prev_requester)
+
+    Dispatched when a track is skipped
+
+    :param discord.Guild guild: The guild in which the track was skipped
+    :param lavalink.Track prev_track: The track that was skipped
+    :param discord.Member prev_requester: The track's requester
+
+.. py:method:: Red.on_red_audio_track_auto_play(guild, track, requester, player)
+
+    Dispatched when a track is auto-played
+
+    :param discord.Guild guild: The guild in which the track was auto-played
+    :param lavalink.Track track: The track that was started
+    :param discord.Member requester: The track's requester: :py:meth:`discord.Guild.me` in this case
+    :param lavalink.Player player: The player associated to this event
+
+.. py:method:: Red.on_red_audio_track_enqueue(guild, track, requester)
+
+    Dispatched when a track is enqueued
+
+    :param discord.Guild guild: The guild in which the track was enqueued
+    :param lavalink.Track track: The track that was enqueued
+    :param discord.Member requester: The track's requester
+
+.. py:method:: Red.on_red_audio_track_end(guild, track, requester)
+
+    Dispatched when a playing track ends
+
+    :param discord.Guild guild: The guild in which the track ended
+    :param lavalink.Track track: The track that ended
+    :param discord.Member requester: The track's requester
+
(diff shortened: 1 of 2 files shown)


docs/framework_events.rst
```diff
@@ -4,6 +4,95 @@
...
+.. py:method:: Red.on_red_audio_unload(cog)
+
+    Dispatched when the audio cog starts it's unloading task
+
+    :param cog: Audio cog instance: similar to :py:func:`redbot.core.bot.RedBase.get_cog()`
+
+Filter
+^^^^^^
+
+.. py:method:: Red.filter_message_delete(message, hits)
```
Jackenmen:
```suggestion
.. py:method:: Red.on_filter_message_delete(message, hits)
```

docs/framework_events.rst
```diff
@@ -4,6 +4,95 @@
...
+
+    :param discord.Guild guild: The guild in which the track started
+    :param lavalink.Track track: The track that was started
+    :param discord.Member requester: The track's requester
+
+.. py:method:: Red.on_red_audio_unload(cog)
+
+    Dispatched when the audio cog starts it's unloading task
+
+    :param cog: Audio cog instance: similar to :py:func:`redbot.core.bot.RedBase.get_cog()`
```
Jackenmen:
```suggestion
    :param cog: Audio cog instance: similar to :py:func:`redbot.core.bot.Red.get_cog()`
```

docs/framework_events.rst
```diff
@@ -4,6 +4,95 @@
...
+    :param Set[str] hits: Words which got detected in the message
+
+Modlog
+^^^^^^
+
+.. py:method:: Red.on_modlog_case_create(case)
+
+    Dispatched when a new modlog case is created
+
+    :param redbot.core.modlog.Case case: The modlog case associated to this event
```
Jackenmen:
```suggestion
    :param redbot.core.modlog.Case case: The modlog case associated with this event.
```

docs/framework_events.rst
```diff
@@ -4,6 +4,95 @@
...
+
+.. py:method:: Red.on_modlog_case_edit(case)
+
+    Dispatched when a modlog case is edited
+
+    :param redbot.core.modlog.Case case: The modlog case associated to this event
```
Jackenmen:
```suggestion
    :param redbot.core.modlog.Case case: The modlog case associated with this event.
```

docs/framework_events.rst
Jackenmen: @aikaterna do we actually support these?

Jackenmen (changes requested):
All of the method and parameters descriptions should have a full stop (`.`) at the end.

docs/framework_events.rst
benno1237: Thanks for having a look Jack. Sorry for not answering for such a long time. Christmas holidays.

Regarding your question for aika: I cannot really tell you whether the audio events are supported or not, they seem to work just fine tho. I did some basic testing with all of them and they seem to be dispatched reliably.

aikaterna: They have been in audio since 2019* or so I believe but they were never documented (obviously). They have been there for people to make use of and I don't know of any issues anyone has ever had with them. I would say they are supported imo.

edit: 2018 -> 2019

docs/framework_events.rst
```diff
@@ -4,6 +4,95 @@
...
+
+    :param discord.Guild guild: The guild in which the track started
+    :param lavalink.Track track: The track that was started
+    :param discord.Member requester: The track's requester
+
+.. py:method:: Red.on_red_audio_unload(cog)
+
+    Dispatched when the audio cog starts it's unloading task
+
+    :param cog: Audio cog instance: similar to :py:func:`redbot.core.bot.Red.get_cog()`
```
Jackenmen, line 69: I suggest removing this as the Audio class is not a public API.
```suggestion
```

Jackenmen (changes requested):
In addition to the review comment I posted, you still need to do one change from the previous review - there are still no full stops (`.`) at the end of all the method and parameter descriptions.

PR #5420, opened 2021-10-24
Detect git authentication failure

### LINKED ISSUE #2749: [Downloader] Show more specific message, when repo can't be found
Currently when using command for nonexistent repo:
```
[p]repo add oof https://github.com/non-existent-user/non-existent-repo
```
Bot responds with "Something went wrong during the cloning process."
I think the message should be more specific than that.

Jan200101:
closes #2749

the logic implemented has been tested against https clones and ssh clones using the following addition to .gitconfig
```
[url "ssh://git@github.com/"]
    insteadOf = https://github.com/
```

I've decided to go against reporting it as a missing repository because we cannot know if the repo is truly missing or simply behind authentication.

Some more environment variables are set or removed when running a command to ensure git doesn't do anything smart with appropriate comments and links to the related code in git.

Right now this bundles `ERROR: Repository not found.` with other permission errors (since it also returns `fatal: Could not read from remote repository.`) but that can be changed if desired.

For reference, here is a list of git errors and approriate regexes
https://github.com/desktop/dugite/blob/master/lib/errors.ts#L64


Jan200101:
Perhaps adding tests for this would be worthwhile.

One way I could think of is cloning from a user that cannot exist like Ghost: `https://github.com/ghost/non-existant-repo`

Flame442 (approved):

PR #5421, opened 2021-10-24
Make `emptydisconnect` disconnect when all vc members are bots

Parnassius:
If two Red instances, both with `audioset emptydisconnect` enabled and `audioset dc` disabled, are for whatever reason in the same voice chat they keep each other from disconnecting. Now the bot disconnects when every user in the voice chat is a bot.

### DIFF
--- a/redbot/cogs/audio/core/tasks/player.py
+++ b/redbot/cogs/audio/core/tasks/player.py
@@ -28,7 +28,7 @@ class PlayerTasks(MixinMeta, metaclass=CompositeMetaClass):
                 if await self.bot.cog_disabled_in_guild(self, server):
                     continue
 
-                if [self.bot.user] == p.channel.members:
+                if not any(m for m in p.channel.members if not m.bot):
                     stop_times.setdefault(server.id, time.time())
                     pause_times.setdefault(server.id, time.time())
                 else:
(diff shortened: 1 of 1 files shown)


Kowlin:
Nice catch, thanks for the PR.

redbot/cogs/audio/core/tasks/player.py
```diff
@@ -28,7 +28,7 @@ async def player_automated_timer(self) -> None:
                 if await self.bot.cog_disabled_in_guild(self, server):
                     continue
 
-                if [self.bot.user] == p.channel.members:
+                if not any(m for m in p.channel.members if not m.bot):
```
Jackenmen: I think this is easier to comprehend when reading the code:
```suggestion
                if p.channel.members and all(m.bot for m in p.channel.members):
```
It also fixes an issue with an empty sequence triggering this `if` (though it's probably a case that wouldn't be hit anyway and even if so, it would just be ignored later on).

Jackenmen (changes requested):

redbot/cogs/audio/core/tasks/player.py
Parnassius: Fair point, I've committed the change

Jackenmen (approved):
LGTM, thank you!

PR #5422, opened 2021-10-25
isolate direction of username within modlog

Jan200101:
Came up on the Discord as something that could possibly be implemented

Before this PR:
image

After this PR:
image

### DIFF
--- a/redbot/core/modlog.py
+++ b/redbot/core/modlog.py
@@ -464,10 +464,19 @@ class Case:
                 translated = _("Unknown or Deleted User")
                 user = f"[{translated}] ({self.user})"
             else:
-                user = f"{self.last_known_username} ({self.user})"
+                name = self.last_known_username[:-5]
+                discriminator = self.last_known_username[-4:]
+                user = (
+                    f"\N{FIRST STRONG ISOLATE}{name}"
+                    f"\N{POP DIRECTIONAL ISOLATE}#{discriminator} ({self.user})"
+                )
         else:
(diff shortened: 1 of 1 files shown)


Jan200101:
Force pushed first time to correct the styling checks

Force pushed the second time after I was asked if the second LEFT-TO-RIGHT was really needed, which I had some hours ago concluded it was but came to realize it really isn't

mathstrains21 (dismissed):

redbot/core/modlog.py
```diff
@@ -466,8 +466,13 @@ async def message_content(self, embed: bool = True):
             else:
                 user = f"{self.last_known_username} ({self.user})"
```
Jackenmen: This needs to be updated as well.

Jackenmen (changes requested):
This doesn't work in embeds:
image
image

From my testing, putting LTR override at the beginning of the text instead works just fine so I think this should be a good solution:
image

It *could* make sense to also add this LTR override for the moderator though, it is way less likely to be an issue for a moderator's name.

Jackenmen:
Sorry, I did not realize my screenshot contained an example with two LTR overrides rather than just one at the beginning (which works fine as well). I just removed the second LTR override myself in the commit I just pushed.

Jan200101:
image
Talked about this on Discord

Isolating the name should keep its direction intact while allowing the format of `name#discrim (id)` to stay intact

Jackenmen (approved):
Works great now, thanks for the PR!

cat-xmas

PR #5424, opened 2021-10-26
Make invite URL part of public API

### LINKED ISSUE #5152: Expose `[p]inviteset` Functionality (for Cogs)
### Type of feature request

API functionality

Currently, there is no supported method for third-party cogs to utilize the bot's settings for `[p]inviteset` in order to generate invite links with the same permissions and scopes as `[p]invite`. It would be great if either the settings/config info or the [link-generation function](https://github.com/Cog-Creators/Red-DiscordBot/blob/52f96e043bb43de9d5bb12d6c9dfdb34fb1cbf4d/redbot/core/core_commands.py#L372) were exposed for cogs to utilize.

### Anything else?

An example where this would be useful is in my [BotAccess](https://github.com/Obi-Wan3/OB13-Cogs/blob/400d75b932b0a1a508ed63943fdff082ee2b42a7/botaccess/botaccess.py#L195) cog, where server supporters are granted the ability to invite the bot to a limited number of their own servers. However, without this endpoint, either the generated invite link would be the "default" or `[p]inviteset` would need to be re-implemented.

Stonedestroyer:
Adds invite as a public API, refactored command to use the public API instead as there's no difference from private and public for that. 

Resolves #5152

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1310,7 +1310,6 @@ class Red(
         Generates the invite URL for the bot.
 
         Does not check if invites are public.
-        To check if invites are public, use `is_invite_public()`.
         To check if invites are public, use `Bot.is_invite_public()`.
 
         Returns
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1318,13 +1318,12 @@ class Red(
         str
             Invite URL.
         """
-        app_info = self._app_info
         data = await self._config.all()
         commands_scope = data["invite_commands_scope"]
         scopes = ("bot", "applications.commands") if commands_scope else None
         perms_int = data["invite_perm"]
         permissions = discord.Permissions(perms_int)
-        return discord.utils.oauth_url(app_info.id, permissions, scopes=scopes)
+        return discord.utils.oauth_url(self._app_info.id, permissions, scopes=scopes)
 
     async def is_invite_url_public(self) -> bool:
         """
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1318,7 +1318,7 @@ class Red(
         str
             Invite URL.
         """
-        app_info = await self._app_info
+        app_info = self._app_info
         data = await self._config.all()
         commands_scope = data["invite_commands_scope"]
         scopes = ("bot", "applications.commands") if commands_scope else None
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1309,6 +1309,8 @@ class Red(
         """
         Generates the invite URL for the bot.
 
+        Does not check if invites are public.
+
         Returns
         -------
         str
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1309,8 +1309,9 @@ class Red(
         """
         Generates the invite URL for the bot.
 
-        Does not check if invites are public.
-        To check if invites are public, use `Red.is_invite_url_public()`.
+        Does not check if the invite URL is configured to be public
+        with ``[p]inviteset public``. To check if invites are public,
+        use `Red.is_invite_url_public()`.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


Stonedestroyer:
I am not sure if we should make invite link return null if the invite is private or we should ignore it, as it's a method for cog creator we may wanna offer that as an option as it's meaningless to get invite if it's private.

redbot/core/core_commands.py
```diff
@@ -1487,7 +1470,7 @@ async def invite(self, ctx):
             - `[p]invite`
         """
         try:
-            await ctx.author.send(await self._invite_url())
+            await ctx.author.send(await self.bot.get_invite_url)
```
psykzz:
```suggestion
            await ctx.author.send(await self.bot.get_invite_url())
```

psykzz: I don't see this registered as a property - so unless im missing something this still needs to be a function call.

redbot/core/bot.py
```diff
@@ -1305,6 +1305,25 @@ async def is_owner(self, user: Union[discord.User, discord.Member]) -> bool:
         """
         return user.id in self.owner_ids
 
+    async def get_invite_url(self) -> str:
+        """
+        Generates the invite URL for the bot.
+
+        Does not check if invites are public.
```
Jackenmen: It would seem there's no public API to know if the invite is public and it seems like an important part of this functionality. Could you add a method for it as well?

redbot/core/bot.py
```diff
@@ -1305,6 +1305,25 @@ async def is_owner(self, user: Union[discord.User, discord.Member]) -> bool:
...
+
+        Returns
+        -------
+        str
+            Invite URL.
+        """
+        app_info = await self.bot.application_info()
+        data = await self.bot._config.all()
```
Jackenmen: We're already *in* the bot object.
```suggestion
        data = await self._config.all()
```

redbot/core/bot.py
```diff
@@ -1305,6 +1305,25 @@ async def is_owner(self, user: Union[discord.User, discord.Member]) -> bool:
...
+        Generates the invite URL for the bot.
+
+        Does not check if invites are public.
+
+        Returns
+        -------
+        str
+            Invite URL.
+        """
+        app_info = await self.bot.application_info()
```
Jackenmen: For better performance, you could use `self._app_info` attribute instead here

redbot/core/core_commands.py
Jackenmen: This indeed needs to be changed.

redbot/core/core_commands.py
```diff
@@ -112,7 +112,7 @@ def __init__(self, bot: "Red"):
         self.bot.register_rpc_handler(self._name)
         self.bot.register_rpc_handler(self._prefixes)
         self.bot.register_rpc_handler(self._version_info)
-        self.bot.register_rpc_handler(self._invite_url)
+        self.bot.register_rpc_handler(bot.invite_url)
```
Jackenmen: This needs to stay as it was and the `_invite_url` method here should just call the bot's `get_invite_url()` method. Otherwise, the RPC handler won't get auto-unregistered (well, it will because it's Core and it never gets unloaded but still, I would like this to be done properly).

Jackenmen (changes requested):

redbot/core/core_commands.py
Stonedestroyer: Thank you

redbot/core/bot.py
Stonedestroyer: Added under is_invite_url_public

Stonedestroyer:
All requested changes should be done now. I have tested the two methods but have not been able to test RPC.

redbot/core/core_commands.py
```diff
@@ -381,13 +381,7 @@ async def _invite_url(self) -> str:
...
             Invite URL.
         """
-        app_info = await self.bot.application_info()
-        data = await self.bot._config.all()
-        commands_scope = data["invite_commands_scope"]
-        scopes = ("bot", "applications.commands") if commands_scope else None
-        perms_int = data["invite_perm"]
-        permissions = discord.Permissions(perms_int)
-        return discord.utils.oauth_url(app_info.id, permissions, scopes=scopes)
+        return self.bot.get_invite_url()
```
```suggestion
        return await self.bot.get_invite_url()
```

Jackenmen (approved):
LGTM, thanks for your contribution!

PR #5430, opened 2021-10-28
Prevent an IndexError from occuring

Kowlin:
If an page value is negative it doesn't properly loop around to the end.


leetfin (dismissed):

Jackenmen:
Do you remember in what command this was an issue?

Ryan5453:
queue (audio)

Flame442 (changes requested):
`next_page` L138 should also be modified to catch similar out of bounds cases on the top end.

_Or if you want to swap these two methods to use modular arithmetic instead, that would also be a valid way to fix this bug :)_

Flame442 (approved):
Thanks!

PR #5437, opened 2021-11-03
Add guarantees for supported OS versions

Jackenmen:
This PR is meant to codify what is already happening in the version guarantees.

Our OS support policy at the base level is as follows:
- we strive (but do not guarantee) to support *all* versions that have not yet reached their EOL date (but we MAY drop the support for the non-latest version sooner, as happened for Ubuntu 16.04 and versions of RHEL lower than 8.4)
- we guarantee to support the current latest stable version of each operating system that was supported by previous **micro releases** (as in, we will NOT drop the whole Ubuntu operating system in version 3.4.26 *but* we could drop it in version 3.5.0)
- we drop support for the operating system in the first **micro release** that happens after the EOL date (e.g. we regularly drop support for older Fedora versions as they reach their EOL dates every ~6 months)

Docs preview: https://red-discordbot--5437.org.readthedocs.build/en/5437/version_guarantees.html#end-user-guarantees

### DIFF
--- a/docs/version_guarantees.rst
+++ b/docs/version_guarantees.rst
@@ -10,9 +10,58 @@ While this is very similar to SemVer, we have our own set of guarantees.
 
 Major versions are for project wide rewrites and are not expected in the foreseeable future.
 
-==========
-Guarantees
-==========
+===================
+End-user Guarantees
+===================
+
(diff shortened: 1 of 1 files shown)


Kowlin (approved):
I don't disagree here. Looks good.

PR #5444, opened 2021-11-11
[Docs] Correct set api examples to be consistent for Spotify

yamikaitou:

Update the examples in `set api` to use "spotify" instead of "Spotify" as to stop confusing uninformed users when they are trying to set their Spotify Credentials for Audio usage


Vexed01 (commented):
The `audiodb` example should probably also be replaced with something else, maybe youtube or twitch?

yamikaitou:
> The `audiodb` example should probably also be replaced with something else, maybe youtube or twitch?

Updated it to youtube

Jackenmen (approved):
Looks good, thanks!

PR #5448, opened 2021-11-17
Add non-interactive mode to redbot-setup

### LINKED ISSUE #2396: Allow redbot-setup to be run in headless mode

Jackenmen:
Adds a bunch of CLI options for non-interactive usage of `redbot-setup`.
`--no-prompt` can be used to ensure that the setup is non-interactive. Note that at current time, PostgreSQL backend cannot be configured non-interactively which is an issue that is also present in `redbot-setup convert` command (though it's not as problematic there).

Fixes #2396

### DIFF
--- a/redbot/setup.py
+++ b/redbot/setup.py
@@ -64,8 +64,12 @@ def save_config(name, data, remove=False):
         json.dump(_config, fs, indent=4)
 
 
-def get_data_dir(instance_name: str):
+def get_data_dir(*, instance_name: str, data_path: Optional[Path], interactive: bool) -> str:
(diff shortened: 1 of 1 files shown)


yamikaitou:
Works fine in my testing

Kowlin (approved):
Works like a charm.

PR #5449, opened 2021-11-20
Remove voice region field in [p]serverinfo

PredaaA:
This PR removes the guild region field from serverinfo command, as `guild.region` attribute got deprecated a while back, but was still showing from the API. Now it is set to "deprecated" for all guilds (see attached screenshot).
https://github.com/discord/discord-api-docs/pull/3001

image


Jackenmen (approved):
It works!

We could probably consider adding something else to the short version of `[p]serverinfo` now that we no longer have this field in it.

PR #5450, opened 2021-11-23
[Cleanup] Add `clearchannel` command.

AAA3A-AAA3A:

Since deleting messages is limited to 14 days, here is the "[p]cleanup clearchannel" command to dupe a channel and delete the old one in order to remove all messages from it. The request must be approved by a "y" message like the other one and the "manage_channel" permission.
Thanks in advance.

### DIFF
--- a/redbot/cogs/cleanup/cleanup.py
+++ b/redbot/cogs/cleanup/cleanup.py
@@ -770,11 +770,11 @@ class Cleanup(commands.Cog):
         await new_channel.edit(position=channel_position, reason=f"Clear Channel requested by {ctx.author} ({ctx.author.id})")
 
         log.info(
-            "%s (%s) deleted %s spam messages in channel %s (%s).",
+            "%s (%s) deleted all messages in channel %s (%s).",
             ctx.author,
             ctx.author.id,
-            ctx.channel,
-            ctx.channel.id,
+            new.channel.name,
+            new.channel.id,
         )
         
         await ctx.author.send(f"All messages in channel #{old_channel.name} ({old_channel.id}) have been deleted! You can find the new channel, with the same permissions: #{new_channel.name} ({new_channel.id}).")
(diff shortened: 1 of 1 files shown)


i-am-zaidali (commented):
**From line 87-97**: Message predicate has a [yes_or_no method](https://docs.discord.red/en/stable/framework_utils.html?highlight=yes_or_no#redbot.core.utils.predicates.MessagePredicate.yes_or_no) that you could use to achieve this in a better, more compact way. 

**From line 772-778**: You have 5 percentage signs (%) for formatting, yet you only provide 4 values for it.

Twentysix26:
-1
This introduces a particularly destructive command (can potentially delete years worth of message history) for very little benefit (doing it through the UI from desktop and mobile is trivial) to users that maybe *shouldn't* be able to delete channels that easily (how each server defines "admin" varies greatly and misconfigurations can and do happen).
Finally, in my opinion servers that have this bizarre requirement to wipe entire channels on a regular basis should be using threads to begin with.

AAA3A-AAA3A:
> -1 This introduces a particularly destructive command (can potentially delete years worth of message history) for very little benefit (doing it through the UI from desktop and mobile is trivial) to users that maybe _shouldn't_ be able to delete channels that easily (how each server defines "admin" varies greatly and misconfigurations can and do happen). Finally, in my opinion servers that have this bizarre requirement to wipe entire channels on a regular basis should be using threads to begin with.

I coded it because a server owner needed it. Many bots have this feature. The redbot does not have access to the threads and therefore cannot handle moderation. There may also be spam channels that need to be cleared quickly. I understand what you mean. You are right.
Thanks for your answer!

Jackenmen:
Sorry, this is not functionality we are interested in adding to Core Red. I don't think it's really related to the Cleanup cog and it is way too niche of a use case to be suitable for Core.

If someone wants such a command, it would be better to make a 3rd-party cog for it.

PR #5452, opened 2021-11-25
Make embeds in help output consistent

Jackenmen:
This makes embeds in help consistent regardless of whether it's sent because `[p]help` was used (`[p]help set`, `[p]help set prefix`) or because the command is command group (`[p]set`) or is missing required argument (`[p]set prefix`).

This PR also adds a new keyword argument to `bot.embed_requested()` called `check_permissions`. When set to `True` (defaults to `False`), it will additionally check that the bot has permissions to send embeds in the given channel, same as `ctx.embed_requested()` does.


Kowlin (approved):
LGTM

PR #5453, opened 2021-11-25
Fix RPC cog load and unload by returning dicts

### LINKED ISSUE #5451: [RPC] Some built-in methods error

3.4.14

Use the built-in `CORE__LOAD` RPC method to load a cog via RPC.

The cog to load and some form of output through RPC confirming it succeeded (which looking at the code would be a tuple)

### What actually happened?

The RPC server tried to serialize a tuple (which was returned from `CoreLogic._load`) but it can't serialize it.

Bot error:
https://img.vexcodes.com/hUm2gu

RPC client error:
https://img.vexcodes.com/W8CtEJ

1. Start red with `--rpc`
2. Start up an RPC client
3. Call the `CORE__LOAD` function with any arguments
4. Observe RPC returning Internal Error and the bot logs saying serialization error.

### Anything else?

Reproduction code, using the RPC library Red uses:
```py
from aiohttp_json_rpc import JsonRpcClient

client = JsonRpcClient()
await client.connect("localhost", 6133)

try:
    await client.call("CORE__LOAD", ["customcom"])  # raises RpcInternalError
finally:
    await client.disconnect()
```

`CORE__RELOAD` is broken in the same way as `CORE__LOAD`

I would expect `CORE_UNLOAD` to also be broken but instead it does this:
https://img.vexcodes.com/X1OvRx

Vexed01:
Use a dictionary for returns from the CoreLogic package management functions, `_load`, `_reload`, `_unload`. This makes the returns JSON-serializable and therefore closes #5451 
This has been mostly tested and it appears to work. I have not tested shared libraries.

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -116,11 +116,7 @@ class CoreLogic:
         self.bot.register_rpc_handler(self._version_info)
         self.bot.register_rpc_handler(self._invite_url)
 
-    async def _load(
-        self, pkg_names: Iterable[str]
-    ) -> Tuple[
-        List[str], List[str], List[str], List[str], List[str], List[Tuple[str, str]], Set[str]
-    ]:
+    async def _load(self, pkg_names: Iterable[str]) -> Dict[str, Union[List[str], Dict[str, str]]]:
         """
         Loads packages by name.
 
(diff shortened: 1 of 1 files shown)


Jackenmen (changes requested):
- From what I can tell, the used keys should always be present so you should be accessing the `outcomes` dictionary using `outcomes[key]` syntax, not `outcomes.get(key)`.
- The existing documentation of the `_load` return value should be re-used to document the keys present in the returned dictionary. It would also be great if you added similar documentation for the return value of `_unload`

Jackenmen (approved):
Looking good, thanks!

PR #5464, opened 2021-12-11
Remove Bank cog from labeler configuration

Kreusada:
Removes the Bank cog from the labeler configuration seeing as these directories/files no longer exist.


Jackenmen (approved):
Thanks

PR #5466, opened 2021-12-12
Restrict `[p]cleanupset notify` to guild only

Kreusada:
Running `[p]cleanupset notify` in DMs raises an exception with Config seeing as a DM is not a guild. I don't feel that the hassle to add another setting per user for DM notifying is really necessary (imo), so just adding the `guild_only` decorator would fix this error.

image


Jackenmen (approved):
Thanks for the fix!

PR #5467, opened 2021-12-14
[Mutes] Catch failure to add or remove role

### LINKED ISSUE #5161: Mute command should check for the user's role list before muting/unmuting

Jan200101:
Closes #5161
Alternative of #5345

catches `discord.errors.HTTPException` thrown by add_roles and remove_roles and states the user is already (un)muted


redbot/cogs/mutes/mutes.py
```diff
@@ -1559,6 +1559,9 @@ async def mute_user(
             try:
                 await user.add_roles(role, reason=reason)
                 await self.config.guild(guild).muted_users.set(self._server_mutes[guild.id])
+            except discord.errors.HTTPException:
```
Jan200101: it might be worth adding the guild.muted_users.set here to ensure the user is part of the stored _server_mutes

redbot/cogs/mutes/mutes.py
```diff
@@ -1615,14 +1618,16 @@ async def unmute_user(
                 ret["reason"] = _(MUTE_UNMUTE_ISSUES["role_missing"])
                 return ret
 
-            if guild.id in self._server_mutes:
-                if user.id in self._server_mutes[guild.id]:
-                    del self._server_mutes[guild.id][user.id]
+            if guild.id in self._server_mutes and user.id in self._server_mutes[guild.id]:
```
Jan200101, line 1664: if statement changed to match the one in `mute_user`

redbot/cogs/mutes/mutes.py
```diff
@@ -1546,6 +1546,9 @@ async def mute_user(
             if not guild.me.guild_permissions.manage_roles or role >= guild.me.top_role:
                 ret["reason"] = _(MUTE_UNMUTE_ISSUES["permissions_issue_role"])
                 return ret
+            if role in user.roles:
```
TrustyJAID: As Jack stated in #5161 we should care more about `[p]unmute` not so much `[p]mute`. This would prevent updating a mute duration forcing the user to unmute and then re-apply the mute which is inconvenient.

Jan200101: you make a good point.

add_roles won't fail if the user already has the role and since the user is already muted the user gets what they expected.

It might be worthwhile to add a message for updating the mute duration and printing out already muted if none is given, but this is not the point of this PR

Jan200101: Here is something along the lines of what I mean (keeping the already existing text)
image

PR #5472, opened 2021-12-18
[3.4] Fix error in tempban for guilds with unused vanity feature

Jackenmen:
Workaround until we can use a version with the fix for https://github.com/Rapptz/discord.py/issues/7103


Kowlin (approved):
I'm gonna assume that this works as I have no guilds where I can manage vanity URLs.

PR #5475, opened 2021-12-20
Update host list with new locations for Hetzner and Contabo

ltzmax:

Hetzner recently got a location in the US.
Contabo has gotten Asia location so i added that as well.


Jackenmen (approved):
Thanks.

PR #5476, opened 2021-12-20
[Core] Changed prefix length to `25`. 

ltzmax:

This PR changes the length from `20` to `40` so people can set their prefixes to bot mention (`"<@!12345678901234567890> "` - total of 25 characters) or a rendered bot mention (`"@AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA#1234 "` - total of 39 characters)

### DIFF
--- a/docs/cog_guides/core.rst
+++ b/docs/cog_guides/core.rst
@@ -3651,7 +3651,7 @@ Sets Red's server prefix(es).
 
     This is not additive. It will replace all current server prefixes.
 
-    You cannot have a prefix with more than 40 characters.
+    You cannot have a prefix with more than 25 characters.
 
 **Examples:**
     - ``[p]set serverprefix !``
--- a/docs/cog_guides/core.rst
+++ b/docs/cog_guides/core.rst
@@ -3651,7 +3651,7 @@ Sets Red's server prefix(es).
 
     This is not additive. It will replace all current server prefixes.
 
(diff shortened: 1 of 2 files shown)


Jackenmen:
You need to update this in more places, see https://github.com/Cog-Creators/Red-DiscordBot/pull/5117 for referemce

ltzmax:
> You need to update this in more places, see #5117 for referemce

Thanks, forgot about that. Updated now.

leetfin (commented):
LGTM

Jackenmen (commented):
The change is fine but I don't think anyone would possibly use 40 characters long prefix. 25 has merit because it's the max length for mention + space but I really don't think 40 has any merit.

ltzmax:
> The change is fine but I don't think anyone would possibly use 40 characters long prefix. 25 has merit because it's the max length for mention + space but I really don't think 40 has any merit.

I'm all down to change it to 25 if you want that instead of 40. Both works though.

Jackenmen:
> I'm all down to change it to 25 if you want that instead of 40.

Yeah, I think that would be great.

ltzmax:
> > I'm all down to change it to 25 if you want that instead of 40.
> 
> 
> 
> Yeah, I think that would be great.

Alright I've changed it to 25 now.

Jackenmen (approved):
LGTM

PR #5477, opened 2021-12-22
Add necessary None checks to Core's usage of `Requires.privilege_level`

Ryan5453:
Related Issue: https://sentry.io/share/issue/bf896b87103549b2b9b113c496d2493f/

This happens when someone is trying to disable a command that has user permission requirements set but not a Red PrivilegeLevel. (PrivilegeLevel.NONE was fine though)


Jackenmen (approved):
Looks good to me, thanks for making a fix for this!

PR #5498, opened 2021-12-29
[Events] Rename guild remove listener

Just-Jojo:
Renames `on_guild_leave` to `on_guild_remove`

Edit: This edits a non-existent listener event to the correct event. ~~Kowlin


Kowlin (approved):
LGTM, Thanks.

PR #5499, opened 2021-12-30
Ensure Nitro users can't make CCs that are too long

Kowlin:

This PR solves an issue whereby users with Nitro could create custom commands that are over 2000 characters and therefor silently error out the bot when trying to call the newly created custom command.

Source: https://discord.com/channels/133049272517001216/387398816317440000/925955050621272146

### DIFF
--- a/redbot/cogs/customcom/customcom.py
+++ b/redbot/cogs/customcom/customcom.py
@@ -4,7 +4,7 @@ import random
 from datetime import datetime, timedelta
 from inspect import Parameter
 from collections import OrderedDict
-from typing import Iterable, List, Mapping, Tuple, Dict, Set, Literal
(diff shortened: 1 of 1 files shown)


Jackenmen (changes requested):
You need to check for this in ``[p]cc edit`` as well.
No response about a too-long response before adding all responses in ``[p]cc create random`` is not ideal either but acceptable if you don't want to improve it further (I changed a limit from 2000 to 1000 in order to test this so ignore that on the screenshot):
image

Kowlin:
I'll see if I can push those changes in tonight, and I was not aware that `edit` exists

Jackenmen (approved):
LGTM

PR #5516, opened 2021-12-31
Fix number of cogs in Getting started doc after Bank cog removal

Jackenmen:

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -134,7 +134,7 @@ Red is built with cogs, a fancy term for plugins. They are
 modules that add functionality to Red. They contain
 commands to use.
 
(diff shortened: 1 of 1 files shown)


Flame442 (approved):
This is fine, but I remember seeing some discussion about removing the number entirely and using a general phrase like "many" or "over x". Would that be preferred over this approach?

Jackenmen:
That was in regard to the number of steps needed to upgrade the bot (#5556) where that number doesn't really have any meaning. I would say that in this case, the number of cogs Red is bundled with is a bit more meaningful, and presents (advertises?) that we already bundle quite a few cogs with Red itself.

I'm not very opposed to this idea and I suppose "over x" would work just as well but I do think that this situation is not *that* similar to the one in #5556.

Issue #5443, opened 2021-11-10
Use CamelCase in the [p]findcog command.


Honkertonken:
### Type of feature request

Cog, Command

### Description of the feature you're suggesting

In the `[p]findcog` command, all the cogs installed using downloader don't use CamelCase under the Cog Name field.

Flame442:
Currently, the `Cog Name` field uses the `name` info.json key if the cog is installed via downloader, and the class name otherwise. It should probably be made consistent, and making it the class name would make it easier to find help for a cog from a command. Unless there is some reason to show the `name` field that I'm not thinking of, this should be an acceptable change.

madebylydia:
Instead of replacing the cog name (#5519), putting the class's name between parentheses next to the cog's name could do the trick?

Jackenmen:
I don't agree with this change, I consider package name (the name that is currently shown for Downloader-installed cogs in `[p]findcog`) to be more useful - *every* command in Downloader, as well as `[p]load/unload/reload` commands requires *this* name as an argument, not the class name. If anything, we should instead try to figure out a package name for the other cases, not use the class name for Downloader-installed cogs.
I'm not sure that such a package name could always be fully accurate but it's IMO more useful to have it this way.
Another option would be to just put both cog package name and class name into `[p]findcog`. The output of the command is getting somewhat long so perhaps doing it in parentheses as @Predeactor said would be preferable over adding another field.

Honkertonken:
fair, so something like this should do the trick. https://rick-roll.click-if-you-da.re/5jYP8g0Za.png , made the changes required.

Jackenmen:
Resolved by #5772

PR #5427, opened 2021-10-26
Modernize syntax in chat formatting module and add `success()`

madebylydia:
### Description of the changes

[As discussed in advanced-coding](https://ptb.discord.com/channels/133049272517001216/718148684629540905/902620754112315474), this PR is made to add a `success` function to the others utils that add an emoji to a text (`error`, `warning`, `info`, etc...), which was missing to me.

In the same time, I have changed all functions using `.format` so they can use f-string, as the processing time is faster.

image
Credits to Zephyrkul#1089 for running tests.

For some reasons this PR is a ton of bullshitting commits coming out of nowhere so don't pay attention, I guess, and be sure to squash commit this one :v

PR #5519, opened 2021-12-31
Fix casing issue of cog name field in the findcog command.

Honkertonken:
This should fix casing issue of cog name field in the findcog command.

Solves #5443


redbot/cogs/downloader/downloader.py
```diff
@@ -1735,7 +1735,7 @@ async def findcog(self, ctx: commands.Context, command_name: str) -> None:
                     if cog_installable.repo is None
                     else cog_installable.repo.name
                 )
-                cog_name = cog_installable.name
+                cog_name = "{} ({})".format(cog_installable.name, cog.__class__.__name__)
```
madebylydia:
```suggestion
                cog_name = f"{cog_installable.name} ({cog.__class__.__name__})"
```
f-string are faster and more modern (See #5427 for tests result), they should be preferred. 😄

Honkertonken: welp in my testing i initially used f-strings, but a few days ago i was told that they are a bit slower & and mess up translation so hence i didn't use them. Thanks for the quick response and correction though.

PR #5525, opened 2021-12-31
[general] updated `[p]serverinfo 1` feature list.

ltzmax:
This PR updates the list of features in `[p]serverinfo 1` to the updated ones.


Kowlin:
I feel like most of these flags are better reflected in a server boosting level. As to not repeat information for no particular reason. Especially now that this is their current boosting page. 
image

ltzmax:
> I feel like most of these flags are better reflected in a server boosting level. As to not repeat information for no particular reason. Especially now that this is their current boosting page. 
> 
> image
> 
> 

I don't disagree about that. And i can do what's only in the server boost levels, if that's what wanted instead of the entire discord features from [here](https://discord.com/developers/docs/resources/guild#guild-object-guild-features).

PR #5532, opened 2022-01-05
[Mod] Custom ban/kick messages.

flaree:

Beginning work on custom ban/kick messages using the same func as customcom's transform args feature.

Leaving as draft for future discussion/changes/additions.


redbot/cogs/mod/kickban.py
```diff
@@ -254,6 +283,12 @@ async def _check_tempban_expirations(self) -> None:
                 if await self._check_guild_tempban_expirations(guild, guild_tempbans):
                     await self.config.guild(guild).current_tempbans.set(guild_tempbans)
 
+    def transform_message(self, message, objects):
+        results = re.findall(r"{([^}]+)\}", message)
```
Kowlin, line 280: This may work better as a compiled regex string?

flaree: Considering it was only ran on a mod action, I didn't think it was really necessary to do so but can change it otherwise. (it's also not compiled in custom com)

madebylydia:
@flaree Have you decided to drop this PR?

flaree:
> @flaree Have you decided to drop this PR?


I'm taking a step back with a lot of things and going to enjoy my final break mostly code free. As it's been here quite a bit I feel right to close it until I have the desire to return

madebylydia:
Do you mind if I make my own PR with the same purpose?
If you're okay with it, do you mind if I take some of the code you've written? If not, I'll completely write the new code from scratch on my own.

flaree:
The code used within this is already code within Red, I had just repurposed it for this PR. I did plan on revisiting this prior to 3.5 as I seen that was the milestone it was in. I guess I'll just leave it open until such time

PR #5533, opened 2022-01-07
Added Autostart_win.rst to docs folder

1tsMeMario:

### DIFF
--- a/docs/autostart_win.rst
+++ b/docs/autostart_win.rst
@@ -1,0 +1,109 @@
+.. launchd guide
+
+==============================
+Setting up auto-restart on Mac
+==============================
+
+-----------------------
+Creating the plist file
+-----------------------
+
+Start by activating your venv. Then run the following command:
+
+.. code-block:: none
+
+    which python
+
+Copy the output of that command.
+
+Now run :code:`sudo nano /Library/LaunchDaemons/red.plist`
+
+Paste the following and replace the following: 
+
+- all instances of :code:`username` with your Mac username 
+- :code:`path` with the path you copied earlier
+- :code:`instance-name` with your instance name:
+
+.. code-block:: none
+
+    <?xml version="1.0" encoding="UTF-8"?>
+    <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+    <plist version="1.0">
+        <dict>
+            <key>Label</key>
+            <string>red</string>
+            <key>ProgramArguments</key>
+            <array>
+                <string>path</string>
+                <string>-O</string>
+                <string>-m</string>
+                <string>redbot</string>
+                <string>instance-name</string>
+                <string>--no-prompt</string>
+            </array>
+            <key>RunAtLoad</key>
+            <true/>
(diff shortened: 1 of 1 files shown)


Jackenmen:
We already have 2 competing PRs (#3893, #4081) adding auto-restart guides for Windows so we're not interested in another competing PR when there's no benefit in using it over the existing solutions.

1tsMeMario:
how is there no benefit. i see comments regarding issues with certain windows versions in one and overall aspects of another being lacking. 
this solves both problems being very easy to setup and use as well as working on any windows version that supports startup apps

Jackenmen:
To give some reasoning:
This PR is closer to #3893 but unlike that PR, it requires an external closed-source tool which I wouldn't personally trust. There might be alternative tools that do the same thing but it's not clear what would be the benefit of it in comparison to #3893.

If we want to look from the perspective of #4081, that PR unlike this one, creates a Windows service that allows the system to nicely control the running process and allows the user to manage the service with the built-in service manager.

I also think that this guide is written in a way that allows too much room for the user to make mistakes, but this would be fixable so I'm just mentioning it.

1tsMeMario:
u can manage the service directly as you would running the service from the command line. the outside tool is opensource as well so the source code can be looked at. the benefits are that it only took a day to get both issues ive seen with the existing pr's solved. this quide is easy and straight forward. you copy and paste into a .bat file. anyone whos been selfhosting discord bots is sure to know what a .bat file is. then you right click and click on thing. and boom the main part done. now you just set the exe file to autostart. after that your done. it takes a min or two at most. and you can literally decompile the .exe file that gets created if youre concerned with it.

1tsMeMario:
ok problem solved since your against the bat to exe converter. ill re-edit the doc. you will no longer need to download anything to make this work

Jackenmen:
> u can manage the service directly as you would running the service from the command line

Not in the same capacity as with Windows's service manager.

> the outside tool is opensource as well so the source code can be looked at.

It's not. It's distributed for free but it is not open source. In fact, the authors of the program even offer a paid version that removes some licensing restrictions of the free version.

> the benefits are that it only took a day to get both issues ive seen with the existing pr's solved. 

There's no need for that condescending tone.

> this quide is easy and straight forward. you copy and paste into a .bat file. anyone whos been selfhosting discord bots is sure to know what a .bat file is. then you right click and click on thing. and boom the main part done. now you just set the exe file to autostart. after that your done. it takes a min or two at most. 

The simplicity of the solution does not mean it's a good solution. There are problems with your solution that I have not dived into because I frankly don't think it's worth my time convincing you otherwise considering your behavior.

Kowlin:
First of all, I'd like to thank you for sending in this PR. However I think that we are going to have to decline this.

I'm sorry, but I'm not really convinced by the program you've linked. You state that its open source software, but looking at the source you've linked it is not. The only thing it states is "Freeware", and that's hardly a classification of open source I'm afraid. And such I don't feel like this is a trust worthy source of what this program does, unlike #4081, where the source is audited by an external 3rd party (Chocolatey) before being published. Furthermore I also think that the language and writing style used in this PR is not up to standard. Where as it feels more as a blog post then proper documentation.

1tsMeMario:
i just reupdated it where it no longer needs any software

PR #5534, opened 2022-01-07
Updated Windows auto start: no longer need external software.

1tsMeMario:


Kowlin:
See https://github.com/Cog-Creators/Red-DiscordBot/pull/5533#issuecomment-1007440110.

PR #5535, opened 2022-01-08
[General] Removed 1 extra space 😅

Wayrion:
Thanks to MAX#1000 and GhOsT#0231 for pointing that out


Twentysix26:
It's not worth having people translate the string again over a space that bothers no one

PR #5540, opened 2022-01-12
Fix for using redbot name --edit for changing name

### LINKED ISSUE #5545: Cannot change bot's name using "redbot --edit"

3.4.16

I was trying to edit one of my instance using `redbot <Name> --edit`. In the process, I wanted to edit the bot's name too, but when answered "Yes", the command make you return to the terminal.
See the included GIF.

Code_htfcdRR4U3

Being able to change's the bot name.

### What actually happened?

The command exited.

1. Use `redbot <name> --edit`
2. Skip questions and say "yes" for changing the instance's name.

### Anything else?

The others questions work just fine, this questions seems to be the only issue in here.

aikaterna:
Currently if you use `redbot <bot name> --edit` and choose `y` for `Would you like to change the instance name?: [y/N]`, it returns the user back to the command prompt with no changes. By using the imported `get_name` function from setup.py, it is expecting a str to be passed with the `name = get_name()` line. By passing a name with a len of 0, that is when setup.py can prompt the user for a new name.

Fixes #5545

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -222,7 +222,7 @@ def _edit_instance_name(old_name, new_name, confirm_overwrite, no_prompt):
                 " run this command with --overwrite-existing-instance flag."
             )
     elif not no_prompt and confirm("Would you like to change the instance name?", default=False):
-        name = get_name()
+        name = get_name("")
         if name in _get_instance_names():
             print(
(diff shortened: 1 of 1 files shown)


madebylydia (commented):
This fix worked like a charm. 👍

Flame442 (approved):

PR #4127, opened 2020-07-29
Add variation selector-16 as appropriate to emojis across Red

Jackenmen:
### Type

- [x] Bugfix
- [ ] Enhancement
- [ ] New feature

### Description of the changes
Fixes #4106

Issue #4106, opened 2020-07-23
Look through emojis used in Red to see if some don't need variation selector


Jackenmen:
#4105 handles emojis in chat formatting, but all other places in Red still need to be checked. 

Info copied from Sinbad's PR:
  - Ensures consistent rendering across devices
  - See:  https://unicode.org/Public/emoji/12.1/emoji-variation-sequences.txt

flaree:
left/right arrow in DEFAULT_CONTROLS also needs the selector

PR #4105, opened 2020-07-23
[chat formatting] Add Variation Selector-16 as appropriate

mikeshardmind:
### Type

- [X] Bugfix
- [ ] Enhancement
- [ ] New feature

### Description of the changes
  - Ensures consistent rendering across devices
  - See:  https://unicode.org/Public/emoji/12.1/emoji-variation-sequences.txt

(Yes, one of these did not have the selector added)

PR #5544, opened 2022-01-16
[Diagnoser] Use VARIATION SELECTOR-16 for passed check mark

Kreusada:

Make passed check mark use VARIATION SELECTOR-16 to correspond with the failed emoji.

#### Before

before

#### After

after


Jackenmen:
If this is a problem, please make an issue for it to be triaged as the way we currently apply variation selectors is consistent (see #4105, #4106, #4127) and this breaks that consistency.

PR #5548, opened 2022-01-22
Emphasize lines with things to replace in autostart guides

Jackenmen:
Should make it easier to notice places you need to replace the vars in.


Flame442 (dismissed):
The emphasize is still pretty subtle, but it is technically working. Not sure if there's an option to change how that looks, but if there is one it should probably be changed to make it stand out more.
image

Flame442 (approved):

PR #5550, opened 2022-01-23
Update disable commands logic during cogs loading

PredaaA:
This PR changes the way disabled commands are handling when a cog is loaded.
Instead of checking at every command added (which is two config calls), it does it only at every cog added. However it still listens to command_add in case a command added is not part of a cog.

That remove a important config abuse basically. So this might get some bots starting significantly faster with this change.

Thanks to @phenom4n4n for pointing this out, and for the help!


leetfin (commented):
LGTMBIDKWILA

Flame442 (approved):

PR #5552, opened 2022-01-24
Use a SnowflakeList for Command disable instead of checks

PredaaA:
This avoids having many checks in memory for commands disable handling. Instead uses SnowflakeList from discord.utils.


phenom4n4n:
Just tested, works well.

redbot/core/commands/commands.py
```diff
@@ -461,10 +461,22 @@ async def can_run(
...
                 ctx.permission_state = original_state
 
+    def is_enabled(self, ctx) -> bool:
+        if not self.enabled:
+            return False
+        if ctx.guild:
+            if self._disabled_in.has(ctx.guild.id):
+                return False
+            if self._disabled_in.has(ctx.channel.id):
+                return False
```
flaree: You check for channel here but both the enable_in and disable_in only take in a guild obj

flaree (approved):

PR #5553, opened 2022-01-24
Add a way to include where a command is disabled in disabled message

PredaaA:
By using `{origin}` placeholder in the command `[p]command disabledmsg`, it will show where the command is disabled.
So `globally` for global, and `in this server` if it's in the current guild.

There's a cleaner way of doing it with #5552 than currently though.

### DIFF
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -481,10 +481,11 @@ class Command(CogCommandMixin, DPYCommand):
 
         if not self.enabled:
             disabled_message = await ctx.bot._config.disabled_command_msg()
-            disabled_message.replace("{command}", ctx.invoked_with).replace(
-                "{origin}", _("globally")
+            raise DisabledCommand(
+                disabled_message.replace("{command}", ctx.invoked_with).replace(
+                    "{origin}", _("globally")
+                )
             )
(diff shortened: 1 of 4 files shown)


madebylydia:
Is there a reason to keep `str.replace` here? It might be interesting to setup a `string.Template` here (While still keeping the same format of delimiters: `{}`)
This PR works as intended, tested on commit [54f7af5](https://github.com/Cog-Creators/Red-DiscordBot/pull/5553/commits/54f7af53513ee37e9f14781925bd85bd6949fa0e).

fixator10:
IMHO, better would be splitting disabled message in two messages. One for disabled globally, and one for locally disabled commands.

Since the current implementation can be confusing with i18n.

Jackenmen:
> IMHO, better would be splitting disabled message in two messages. One for disabled globally, and one for locally disabled commands.

I agree.


> Is there a reason to keep `str.replace` here? It might be interesting to setup a `string.Template` here (While still keeping the same format of delimiters: `{}`)

Simplicity.

Jackenmen:
Per the previous 2 comments, I created issue #6763. I'm closing this PR as I think the path forward should follow the approach proposed in #6763 instead + this has many conflicts now.

PR #5557, opened 2022-01-30
Ensure that registered Config defaults are serializable to JSON

Jackenmen:
Instead of pickle, let's use JSON here so that in addition to deep copying, we also verify that the value is JSON serializable. Might come with a very slight performance penalty but defaults are only registered on cog load so this does not matter.


Flame442 (approved):
Properly catches non-json serializables and deep copies mutables.

Issue #5558, opened 2022-01-30
Social Media Command


jliby:
### Type of feature request

Cog, Command

### Description of the feature you're suggesting

I am curious to know if I can add a cogg that will manage the social media for a particular server. I am thinking of a discord bot that will return the social media account of a specified user.

Kowlin:
Hey there, these issues are only for issues related to Red's code and its internal systems. If you require support regarding Red you're welcome to join [our Discord server](https://discord.gg/red). Also I suggest you check out the [cog index](https://index.discord.red/) to see if someone has already created something similar as you requested.

PR #5559, opened 2022-01-30
Adding a social media feature for the discord bot.

jliby:
Added a discord command called grab_social. You type #grab_social (user) and it will return the Instagram, LinkedIn, and more. If a user does not have any accounts they will have to add them to the social channel at the given server.

### DIFF
--- a/general.py
+++ b/general.py
@@ -1,0 +1,571 @@
+import datetime
+import time
+from enum import Enum
+from random import randint, choice
+from typing import Final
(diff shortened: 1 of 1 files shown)


Twentysix26:
Hello, as told in #5558 we are not accepting this feature: it looks like something that would be useful only in specific communities / server setups. If you're interested in expanding your own bot's functionalities you should familiarize yourself with how the cog system works to write your own.

PR #5462, opened 2021-12-09
Add explicit type checking for *_from_id(s) methods (#5459)

i-am-zaidali:
### Description of the changes

This PR adds type checking for the parameters of *_from_id(s) methods in config in order to ensure config files do not break incase parameters are not of type `int`.

If I've made any sort of mistakes, please correct me as this is my first PR and i would be happy to recieve constructive criticism

Fixes #5459

PR #5564, opened 2022-02-03
Add typechecking to config from_id methods

### LINKED ISSUE #5459: Config *_from_id(s) methods not typechecking.

3.4.14

Trying to access a guild's data from config using the `Config.guild_from_id()` method but accidentally passed in a guild object instead of its id.

An error or exception to raise that indicated that my passed value wasn't of type int.

### What actually happened?

The `Config.guild_from_id()` method took the guild object and casted it to a string which turned into the guild's name and upon bot restart, errors were raised about it not being of base 10 for int casting when using the `Config.all_guilds()` method in a cog's initialization

1. Use the `Config.guild_from_id()` method but instead of an integer id, pass a proper guild object.
2. Restart the bot and call the `Config.all_guilds()` method in a cog.

### Anything else?

I'll be happy to push a pr fix for this if needed

i-am-zaidali:
This PR is a redo for #5462

This PR adds type checking for the parameters of *_from_id(s) methods in config in order to ensure config files do not break in case parameters are not of type `int`.

If I've made any sort of mistakes, please correct me as this is my first PR and I would be happy to receive constructive criticism

Fixes #5459

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -956,7 +956,15 @@ class Config(metaclass=ConfigMeta):
         `Group <redbot.core.config.Group>`
             The guild's Group object.
 
+        Raises
+        ------
+        TypeError
+            If the given guild_id parameter is not of type int
         """
+        if not isinstance(guild_id, int):
+            raise TypeError(
+                f"`guild_id` should be of type `int`, not {guild_id.__class__.__name__}"
+            )
         return self._get_base_group(self.GUILD, str(guild_id))
 
     def guild(self, guild: discord.Guild) -> Group:
@@ -990,7 +998,15 @@ class Config(metaclass=ConfigMeta):
         `Group <redbot.core.config.Group>`
             The channel's Group object.
 
+        Raises
+        ------
+        TypeError
+            If the given channel_id parameter is not of type int
         """
+        if not isinstance(channel_id, int):
+            raise TypeError(
+                f"`channel_id` should be of type `int`, not {channel_id.__class__.__name__}"
+            )
         return self._get_base_group(self.CHANNEL, str(channel_id))
 
     def channel(self, channel: discord.abc.GuildChannel) -> Group:
@@ -1024,7 +1040,13 @@ class Config(metaclass=ConfigMeta):
         `Group <redbot.core.config.Group>`
             The role's Group object.
 
+        Raises
+        ------
+        TypeError
+            If the given role_id parameter is not of type int
         """
+        if not isinstance(role_id, int):
+            raise TypeError(f"`role_id` should be of type `int`, not {role_id.__class__.__name__}")
         return self._get_base_group(self.ROLE, str(role_id))
 
     def role(self, role: discord.Role) -> Group:
@@ -1056,7 +1078,13 @@ class Config(metaclass=ConfigMeta):
         `Group <redbot.core.config.Group>`
             The user's Group object.
 
+        Raises
+        ------
+        TypeError
+            If the given user_id parameter is not of type int
         """
+        if not isinstance(user_id, int):
+            raise TypeError(f"`user_id` should be of type `int`, not {user_id.__class__.__name__}")
         return self._get_base_group(self.USER, str(user_id))
 
     def user(self, user: discord.abc.User) -> Group:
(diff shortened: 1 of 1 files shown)


redbot/core/config.py
```diff
@@ -956,7 +956,15 @@ def guild_from_id(self, guild_id: int) -> Group:
         `Group <redbot.core.config.Group>`
             The guild's Group object.
 
+        Raises
+        ------
+        TypeError
+            If the given guild_id parameter is not of type int
         """
+        if not isinstance(guild_id, int):
```
Jackenmen:
```suggestion
        if type(guild_id) is not int:
```

redbot/core/config.py
```diff
@@ -990,7 +998,15 @@ def channel_from_id(self, channel_id: int) -> Group:
         `Group <redbot.core.config.Group>`
             The channel's Group object.
 
+        Raises
+        ------
+        TypeError
+            If the given channel_id parameter is not of type int
         """
+        if not isinstance(channel_id, int):
```
Jackenmen:
```suggestion
        if type(channel_id) is not int:
```

redbot/core/config.py
```diff
@@ -1024,7 +1040,13 @@ def role_from_id(self, role_id: int) -> Group:
         `Group <redbot.core.config.Group>`
             The role's Group object.
 
+        Raises
+        ------
+        TypeError
+            If the given role_id parameter is not of type int
         """
+        if not isinstance(role_id, int):
```
Jackenmen:
```suggestion
        if type(role_id) is not int:
```

redbot/core/config.py
```diff
@@ -1056,7 +1078,13 @@ def user_from_id(self, user_id: int) -> Group:
         `Group <redbot.core.config.Group>`
             The user's Group object.
 
+        Raises
+        ------
+        TypeError
+            If the given user_id parameter is not of type int
         """
+        if not isinstance(user_id, int):
```
Jackenmen:
```suggestion
        if type(user_id) is not int:
```

redbot/core/config.py
```diff
@@ -1090,7 +1118,21 @@ def member_from_ids(self, guild_id: int, member_id: int) -> Group:
         `Group <redbot.core.config.Group>`
             The member's Group object.
 
+        Raises
+        ------
+        TypeError
+            If the given guild_id or member_id parameter is not of type int
         """
+        if not isinstance(guild_id, int):
```
Jackenmen:
```suggestion
        if type(guild_id) is not int:
```

redbot/core/config.py
```diff
@@ -1090,7 +1118,21 @@ def member_from_ids(self, guild_id: int, member_id: int) -> Group:
...
+            raise TypeError(
+                f"`guild_id` should be of type `int`, not {guild_id.__class__.__name__}"
+            )
+
+        if not isinstance(member_id, int):
```
Jackenmen:
```suggestion
        if type(member_id) is not int:
```

Jackenmen (dismissed):
Considering that this is a performance-critical section *and* that there could be subclasses of `int` that we don't intend to accept (e.g. `bool`), checking for the exact `int` type is preferred here.

redbot/core/config.py
i-am-zaidali: lemme do that real quick. brb

redbot/core/config.py
```diff
@@ -956,7 +956,15 @@ def guild_from_id(self, guild_id: int) -> Group:
...
             The guild's Group object.
 
+        Raises
+        ------
+        TypeError
+            If the given guild_id parameter is not of type int
         """
+        if type(guild_id) is not int:
+            raise TypeError(
+                f"`guild_id` should be of type `int`, not {guild_id.__class__.__name__}"
```
Flame442:
```suggestion
                f"guild_id should be of type int, not {guild_id.__class__.__name__}"
```
The backticks should either be there for all "code" items or for none of them. I personally vote they should not be there.

Flame442 (changes requested):
Almost ready...

i-am-zaidali:
i committed from phone so checks failed. Gonna run them in the morning sorry for the delay

Flame442 (approved):
Thanks for the PR!

PR #5565, opened 2022-02-03
Add JSON Schema for Trivia Lists

### LINKED ISSUE #5301: Add JSON Schema for Trivia lists
### Type of feature request

Other

Similar to JSON Schema files we have for Downloader's repo and cog `info.json` files, we should have a JSON Schema for the Trivia lists.

### Anything else?

_No response_

Kreusada:

Closes #5301. May be altered pending comments on #5390, and #5897 if it gets merged.

Adds JSON schema for Trivia list author and configuration keys.


Jackenmen:
Once this is approved and merged, you or I should make a PR against https://github.com/SchemaStore/schemastore as well (see SchemaStore/schemastore#1226)

Kreusada:
> Once this is approved and merged, you or I should make a PR against https://github.com/SchemaStore/schemastore as well (see [SchemaStore/schemastore#1226](https://github.com/SchemaStore/schemastore/pull/1226))

I'd be happy doing this 👍 
Draft (pending): https://github.com/SchemaStore/schemastore/pull/2068

schema/trivia.schema.json
```diff
@@ -0,0 +1,54 @@
+{
+    "$id": "https://raw.githubusercontent.com/Cog-Creators/Red-DiscordBot/V3/develop/schema/trivia.schema.json",
+    "$schema": "http://json-schema.org/draft-07/schema#",
+    "title": "Red-DiscordBot Trivia metadata file",
```
Jackenmen:
```suggestion
    "title": "Red-DiscordBot Trivia List file",
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,54 @@
...
+    "type": "object",
+    "properties": {
+        "AUTHOR": {
+            "type": "string",
+            "description": "Author of the Trivia quiz."
+        },
+        "CONFIG": {
+            "type": "object",
+            "description": "The configuration for a trivia session.",
+            "required": [],
```
Jackenmen: This is the default value so it should not need to be specified:
```suggestion
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,54 @@
...
+                    "type": ["integer", "number"],
+                    "description": "Timeout per question in seconds, must be positive number greater than 0.0. Defaults to 120.0.",
+                    "default": 120.0
+                },
+                "use_spoilers": {
+                    "type": "boolean",
+                    "description": "Whether to use spoiler markdown when revealing question answers. Defaults to false.",
+                    "default": false
+                }
+            }
```
Jackenmen: The only valid keys are specified in the schema so we can disallow names not listed here:
```suggestion
            },
            "additionalProperties": false
```

Jackenmen (changes requested):
In addition to the review comments:
- the JSON file should be formatted with 2 space indentation to be consistent with existing schema files
- it would be great if the trivia cog ignored `$schema` key instead of treating it as a question
- schema should specify that all other keys should be lists of strings
    - this should be achievable with `additionalProperties`: https://json-schema.org/draft/2020-12/json-schema-core.html#section-10.3.2.3

schema/trivia.schema.json
```diff
@@ -0,0 +1,54 @@
...
+            "required": [],
+            "properties": {
+                "bot_plays": {
+                    "type": "boolean",
+                    "description": "Whether or not the bot gains points during the session. Defaults to true.",
+                    "default": true
+                },
+                "delay": {
+                    "type": ["integer", "number"],
+                    "description": "Answer time limit in seconds, must be positive number greater than 0.0. Defaults to 15.0.",
```
Jackenmen:
```suggestion
                    "description": "The maximum number of seconds permitted to answer a question, must be a positive number greater than or equal to 4.0. Defaults to 15.0.",
                    "minimum": 4.0,
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,54 @@
...
+                    "default": 15.0
+                },
+                "max_score": {
+                    "type": "integer",
+                    "description": "Number of points required in order win the trivia, must be positive integer. Defaults to 10.",
```
Jackenmen:
```suggestion
                    "description": "Number of points required in order to win the trivia, must be a positive integer. Defaults to 10.",
                    "exclusiveMinimum": 0,
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,54 @@
...
+                    "default": 0.0
+                },
+                "reveal_answer": {
+                    "type": "boolean",
+                    "description": "Whether or not to reveal the answer when the question times out. Defaults to true.",
+                    "default": true
+                },
+                "timeout": {
+                    "type": ["integer", "number"],
+                    "description": "Timeout per question in seconds, must be positive number greater than 0.0. Defaults to 120.0.",
```
Jackenmen:
```suggestion
                    "description": "Number of seconds that need to pass until trivia stops due to no response, must be a positive number greater than 0.0. Defaults to 120.0.",
                    "exclusiveMinimum": 0
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,54 @@
...
+                    "default": 15.0
+                },
+                "max_score": {
+                    "type": "integer",
+                    "description": "Number of points required in order win the trivia, must be positive integer. Defaults to 10.",
+                    "default": 10
+                },
+                "payout_multiplier": {
+                    "type": ["integer", "number"],
+                    "description": "The payout multiplier, must be positive number. Defaults to 0.0.",
```
Jackenmen:
```suggestion
                    "description": "The payout multiplier, must be a positive number or zero. Defaults to 0.0.",
                    "minimum": 0
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,54 @@
...
+                    "default": true
+                },
+                "timeout": {
+                    "type": ["integer", "number"],
+                    "description": "Timeout per question in seconds, must be positive number greater than 0.0. Defaults to 120.0.",
+                    "default": 120.0
+                },
+                "use_spoilers": {
+                    "type": "boolean",
+                    "description": "Whether to use spoiler markdown when revealing question answers. Defaults to false.",
```
Jackenmen:
```suggestion
                    "description": "Whether to hide the answers in spoilers when revealing the question's answers. Defaults to false.",
```

Jackenmen (changes requested):
I sent the first review a bit too soon - I forgot to review properties in `CONFIG` key...
Aside from review comments, you should probably also remove the default values from everything as they're not actually accurate since they depend on whatever the server set those values to. All of the keys simply have no default value, if they're not specified in the YAML file then they're simply not being overridden.

Kreusada:
> it would be great if the trivia cog ignored `$schema` key instead of treating it as a question

What do you mean by this?

Jackenmen:
JSON schema url can be specified with a top-level `$schema` key (to be explicit rather than rely on editor's auto-detection). However, Trivia cog will currently reject this as it expects all keys except AUTHOR and CONFIG to be questions.

Jackenmen:
This should probably just require adding this line to the `TRIVIA_LIST_SCHEMA` dictionary:
```
Optional("$schema"): And(str, error=_("{key} key must be a text value.")),
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,52 @@
...
+          "exclusiveMinimum": 0
+        },
+        "use_spoilers": {
+          "type": "boolean",
+          "description": "Whether to hide the answers in spoilers when revealing the question's answers. Defaults to false."
+        }
+      },
+      "additionalProperties": false
+    }
+  }
```
Jackenmen, line 60: This remains to be unaddressed:
> - schema should specify that all other keys should be lists of strings
>     - this should be achievable with `additionalProperties`: https://json-schema.org/draft/2020-12/json-schema-core.html#section-10.3.2.3

This should work:
```suggestion
  },
  "additionalProperties": {
    "type": "array",
    "items": {
      "type": "string"
    }
  }
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,52 @@
+{
+  "$id": "https://raw.githubusercontent.com/Cog-Creators/Red-DiscordBot/V3/develop/schema/trivia.schema.json",
+  "$schema": "http://json-schema.org/draft-07/schema#",
+  "title": "Red-DiscordBot Trivia List file",
+  "type": "object",
+  "properties": {
```
Jackenmen, line 6: I'm unsure this is required but some schema validators seem to reject `$schema` field as having an invalid type when `additionalProperties` is set to an array of strings so it's probably better to define it:
```suggestion
  "properties": {
    "$schema": {
      "type": "string",
      "format": "uri"
    },
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,52 @@
...
+    "AUTHOR": {
+      "type": "string",
+      "description": "Author of the Trivia quiz."
```
Jackenmen:
```suggestion
      "description": "Author of the Trivia list."
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,52 @@
...
+        },
+        "payout_multiplier": {
+          "type": ["integer", "number"],
+          "description": "The payout multiplier, must be a positive number or zero. Defaults to 0.0.",
+          "minimum": 0
+        },
+        "reveal_answer": {
+          "type": "boolean",
+          "description": "Whether or not to reveal the answer when the question times out. Defaults to true.",
+          "default": true
```
Jackenmen: Missed the removal of default in this one:
```suggestion
          "description": "Whether or not to reveal the answer when the question times out. Defaults to true."
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,52 @@
...
+    "CONFIG": {
+      "type": "object",
+      "description": "The configuration for a trivia session.",
+      "properties": {
+        "bot_plays": {
+          "type": "boolean",
+          "description": "Whether or not the bot gains points during the session. Defaults to true."
+        },
+        "delay": {
+          "type": ["integer", "number"],
```
Jackenmen: `number` already supports integers:
```suggestion
          "type": "number",
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,52 @@
...
+          "description": "The maximum number of seconds permitted to answer a question, must be a positive number greater than or equal to 4.0. Defaults to 15.0.",
+          "minimum": 4.0
+        },
+        "max_score": {
+          "type": "integer",
+          "description": "Number of points required in order to win the trivia, must be a positive integer. Defaults to 10.",
+          "exclusiveMinimum": 0
+        },
+        "payout_multiplier": {
+          "type": ["integer", "number"],
```
Jackenmen: `number` already supports integers:
```suggestion
          "type": "number",
```

schema/trivia.schema.json
```diff
@@ -0,0 +1,52 @@
...
+          "description": "The payout multiplier, must be a positive number or zero. Defaults to 0.0.",
+          "minimum": 0
+        },
+        "reveal_answer": {
+          "type": "boolean",
+          "description": "Whether or not to reveal the answer when the question times out. Defaults to true.",
+          "default": true
+        },
+        "timeout": {
+          "type": ["integer", "number"],
```
Jackenmen: `number` already supports integers:
```suggestion
          "type": "number",
```

Jackenmen (changes requested):
In addition to the review comments, it seems that the default values are still in the property descriptions, they should be removed from there as well.

Jackenmen:
> In addition to the review comments, it seems that the default values are still in the property descriptions, they should be removed from there as well.

Unsure if you know but above remains unaddressed.

It also seems that this PR isn't really blocked by the PR mentioned in the description, it would just need changes *if* the other one gets merged first. Am I missing something?

Kreusada:
> In addition to the review comments, it seems that the default values are still in the property descriptions, they should be removed from there as well. Unsure if you know but above remains unaddressed.

Ah I had missed this, I had addressed these changes now.

Jackenmen (approved):
LGTM, sorry for the delay. As mentioned before, this should now be added to schemastore repository.

Kreusada:
https://github.com/SchemaStore/schemastore/pull/2810

PR #5569, opened 2022-02-08
Update enqueued track message to distinguish album from playlist

Ryan5453:
Pretty small change, all it does is check for if it's an album or playlist (the Query object was already providing this data) and display the correct type instead of always defaulting to saying a playlist was enqueued.

Also, while doing this I noticed playlist_url was set to None at the top of the function, but it was never changed to be the playlist url later, so I 'fixed' (was that intentional?) that as well.

### DIFF
--- a/redbot/cogs/audio/core/utilities/player.py
+++ b/redbot/cogs/audio/core/utilities/player.py
@@ -383,6 +383,8 @@ class PlayerUtilities(MixinMeta, metaclass=CompositeMetaClass):
                 index = query.track_index
                 if query.start_time:
                     seek = query.start_time
+            if query.is_url:
+                playlist_url = query.uri
             try:
                 result, called_api = await self.api_interface.fetch_track(ctx, player, query)
             except TrackEnqueueError:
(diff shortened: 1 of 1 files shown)


Drapersniper (commented):
LGTM, yeah no the playlist_url was a bug

Drapersniper:
image


LGTM - I see you changed the album/playlist title from bold to normal - I'm not against it and rather like the contrast caused on the embed.

@aikaterna  if you are okay with the change in Bold -> Not bold then this can be merged

redbot/cogs/audio/core/utilities/player.py
```diff
@@ -491,11 +493,12 @@ async def _enqueue_tracks(
             playlist_name = escape(
                 playlist_data.name if playlist_data else _("No Title"), formatting=True
             )
+            title = _("Playlist Enqueued") if not query.is_album else _("Album Enqueued")
             embed = discord.Embed(
-                description=bold(f"[{playlist_name}]({playlist_url})")
+                description=f"[{playlist_name}]({playlist_url})"
```
Jackenmen:
```suggestion
                description=bold(f"[{playlist_name}]({playlist_url})")
```

Jackenmen (changes requested):

Jackenmen (approved):
Thanks for the PR!

PR #5570, opened 2022-02-11
Dispatch a cog unload event [on_cog_remove]

npc203:
This PR adds an extra event akin to it's counter-part (on_cog_add). It dispatches when a cog is (successfully) unloaded with the **cog object** as it's only argument.

Thanks in advance!


npc203:
Erm, just noticed, the other doc string line edit was done by black and not me

Flame442 (approved):

PR #5572, opened 2022-02-12
Improve topography trivia

Visne:
What prompted me to do this was that "Swaziland" was the only correct answer to the worldmap question, even though that's wrong. While I was there I cleaned up some other things. I tried to mostly add answers but removed some wrong ones as well:

- Using island names as country names (except Ireland, as it is extremely common)
- Inconsistencies
- Native language names of countries (Deutschland, eSwatini, Palaos Bandera)
- Names no longer in use (Swaziland)


Kreusada (commented):
I've just looked through these changes and double checked them, they all look good to me.

redbot/cogs/trivia/data/lists/worldflags.yaml
```diff
@@ -21,8 +23,8 @@ What country is represented by this flag? https://i.imgur.com/222q836.png:
 What country is represented by this flag? https://i.imgur.com/2Wxbjv5.png:
 - South Korea
 What country is represented by this flag? https://i.imgur.com/3H3Lfat.png:
-- Republic of the Philippines
```
Flame442, line 24: Not sure why you removed this option, when another official name `Republic of Maldives` was added. Can you explain your reasoning for this?

redbot/cogs/trivia/data/lists/worldflags.yaml
```diff
@@ -225,25 +235,25 @@ What country is represented by this flag? https://i.imgur.com/YaA07yt.png:
 - Belize
 What country is represented by this flag? https://i.imgur.com/Yvry0ap.png:
 - Central African Republic
+- CAR
+- C.A.R.
 What country is represented by this flag? https://i.imgur.com/ZZJK5He.png:
 - Republic of the Congo
-- Congo
 - Republic of Congo
```
Flame442, line 245: A previous answer for this country uses 
```
Democratic Republic of the Congo
Dem. Rep. of Congo
Democratic Republic of Congo
DRC
```
as the answers. Bare minimum, the same country should have the same answers across the board.

redbot/cogs/trivia/data/lists/worldflags.yaml
```diff
@@ -295,7 +304,6 @@ What country is represented by this flag? https://i.imgur.com/hbJ4J2A.png:
 - Nigeria
 What country is represented by this flag? https://i.imgur.com/hiNRJUm.png:
 - South Africa
-- Republic of South Africa
```
Flame442, line 307: Again, not sure why this was removed, when this is the official name.

redbot/cogs/trivia/data/lists/worldmap.yaml
```diff
@@ -357,7 +376,6 @@ What country is highlighted on this map? https://i.imgur.com/mM0RJG0.png:
 - Croatia
 What country is highlighted on this map? https://i.imgur.com/mUXpeOu.png:
 - Germany
```
Flame442, line 380: Officially `Federal Republic of Germany`, since those are being added

Flame442 (changes requested):

Visne:
@Flame442 I'm think I originally added the `Republic of Maldives` because I thought "Maldives" referred to a geoghrapic island, which is not the case. Therefore I will remove the change.

The reasoning is as follows:
1. Don't use an official country name, because this isn't really done for most of the answers.
2. However, if the informal name would be incorrect, use the official country name.

Example: people use "Ireland" to refer to the Republic of Ireland. However, Ireland is a geographic island, and the Republic of Ireland is only a part of that island.

redbot/cogs/trivia/data/lists/worldflags.yaml
Visne: Will be removed.

redbot/cogs/trivia/data/lists/worldflags.yaml
Visne: I'm not sure what you mean, the Democratic Republic of the Congo and the Republic of the Congo are two different countries.

Flame442 (approved):

PR #5576, opened 2022-02-19
Support all Messageables in `bot.embed_requested()`

Jackenmen:
`await User.create_dm()`, no more :)

This PR makes the following changes to the signature of `Red.embed_requested()`:
- `channel` parameter now accepts all kinds of `Messageable` objects
    - for `Context`, `Context.channel` is used and `command` is auto-filled with `Context.command`
    - for `GroupChannel`, only global settings are checked (bots can't send to group channels but it's part of the `Messageable` contract)
- `user` parameter has been removed
    - I know that if we instead just ignored the `user` param, we could avoid breakage but 3.5 seems like the right time to do this + there's a chance this results in more people using `command` so seems like a win to me :)
- `command` parameter has been made keyword-only
    - this makes it so that `bot.embed_requested(channel, user)` usage correctly starts raising errors
    - and it's a better choice for this parameter anyway

We *could* also do another breakage and have `check_permissions` default to True as it seems like a more sane default though I have not done this change for now. Opinions here would be very much welcome.

### DIFF
--- a/redbot/cogs/mutes/mutes.py
+++ b/redbot/cogs/mutes/mutes.py
@@ -543,8 +543,7 @@ class Mutes(VoiceMutes, commands.Cog, metaclass=CompositeMetaClass):
         if not reason:
             reason = _("No reason provided.")
 
-        # okay, this is some poor API to require PrivateChannel here...
-        if await self.bot.embed_requested(await user.create_dm(), user):
+        if await self.bot.embed_requested(user):
             em = discord.Embed(
                 title=title,
                 description=reason,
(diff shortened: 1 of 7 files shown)


TrustyJAID (commented):
I think we should set the default for`check_permissions` to `True`. It just makes more sense and can allow cog creators to more reliably know if they're even allowed to send embeds. Overall this is a fairly underutilized method that can absolutely be improved.

TrustyJAID (approved):
Looks good to me.

PR #5578, opened 2022-02-19
Avoid potential memory leak in Filter cog

Jackenmen:
Replaces a `(guild, channel)` key with `(guild.id, channel.id)` to ensure that the cog doesn't prevent the garbage collector from removing stale Guild and TextChannel objects.

### DIFF
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -348,10 +348,10 @@ class Filter(commands.Cog):
 
     def invalidate_cache(self, guild: discord.Guild, channel: discord.TextChannel = None):
         """ Invalidate a cached pattern"""
-        self.pattern_cache.pop((guild, channel), None)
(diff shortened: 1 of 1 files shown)


redbot/cogs/filter/filter.py
```diff
@@ -348,10 +348,10 @@ async def filter_names(self, ctx: commands.Context):
 
     def invalidate_cache(self, guild: discord.Guild, channel: discord.TextChannel = None):
         """ Invalidate a cached pattern"""
-        self.pattern_cache.pop((guild, channel), None)
+        self.pattern_cache.pop((guild.id, channel.id), None)
```
PredaaA: Will errors out if channel is None.

PredaaA (changes requested):

PredaaA (approved):
Thanks! Mem leak no

PR #5589, opened 2022-02-23
[Audio] Docstring changes, & hiding llset group command

### LINKED ISSUE #5571: [Audio] Discourage users from using [p]llset to fix unrelated issues

Audio

Currently, users are directing themselves towards `[p]llset` in order to fix issues with lavalink that should be fixed in other ways. This causes slower resolution times for support staff who need to start off by getting these changes reverted before the actual problem can be diagnosed and fixed.

`[p]llset`'s docstrings do not make it abundantly clear that the subcommands are intended for advanced usage only, not for normal setup of the bot. These should be modified to discourage changing the default values unless the user is absolutely sure they know what they are doing. The command may also need to be hidden, however this is an unideal solution as it makes it less obvious for users who would use these commands appropriately.

The "**Unable to play tracks** Connection to lavalink has not yet been established." response when lavalink has not finished starting up does not make it clear that the bot *is attempting to start and connect to* lavalink when lavalink is managed internally without looking at console. Thus, users think the connection was not established not because they need to *wait* for it to be established, but because they haven't properly set up the connection. This causes many to get lost in `[p]llset` trying to fix the issue. When lavalink is managed internally, this message should be expanded to make it clear that the bot is trying to start the server and that console will have information about its progress. When lavalink is managed externally, this message should be expanded to remind the user that they need to start up lavalink for the bot to connect to.

### Anything else?

_No response_

JeffJrShim:

Resolves #5571

### DIFF
--- a/redbot/cogs/audio/core/commands/llset.py
+++ b/redbot/cogs/audio/core/commands/llset.py
@@ -20,6 +20,7 @@ class LavalinkSetupCommands(MixinMeta, metaclass=CompositeMetaClass):
     @commands.bot_has_permissions(embed_links=True)
     async def command_llsetup(self, ctx: commands.Context):
         """Lavalink server configuration options.
+        
         These commands are intended for advanced usage only, not for the normal setup of the bot. Do not modify any of the settings here unless you know what you are doing.
         """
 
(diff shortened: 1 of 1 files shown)


Drapersniper:
Just hiding a command block is not a valid solution to https://github.com/Cog-Creators/Red-DiscordBot/issues/5571

Users should be discouraged to run commands and not just not told about their existence.

Hiding `llset` would be akin to hiding the bots `set` command/


A valid solution for it is better documentation on command docstrings and confirmation before action is taken to ensure this is what the user wants to actually do.

PR #5594, opened 2022-02-26
Support max_python_version key in info.json

### LINKED ISSUE #5590: [Downloader] max_python_version key in cog info.json

Downloader

Similar to the `min_python_version` key (#1452, #1455), it would be great to have a similar key for the maximum version for the (albeit rare) deprecations/removals between minor Python versions.

I am imagining this would mirror the behaviour of the `max_bot_version` key - only impacting installations and updates of the cog through Downloader.

### Anything else?

I would be happy to take this on if approved, and note to me/whoever does this: make sure to update the docs & info.json schema, as well as the code

plooshi:
Make cogs be able to have a maximum python version incase a new version has a change that breaks that cog. (#5590)


redbot/cogs/downloader/downloader.py
```diff
@@ -1490,10 +1499,17 @@ def _filter_incorrect_cogs(
         message = ""
         if outdated_python_version:
             message += (
-                _("\nThese cogs require higher python version than you have: ")
+                _("\nThese cogs require a higher python version than you have: ")
                 if len(outdated_python_version)
```
Vexed01:
```suggestion
                if len(outdated_python_version) > 1
```

redbot/cogs/downloader/downloader.py
```diff
@@ -1490,10 +1499,17 @@ def _filter_incorrect_cogs(
...
-                else _("\nThis cog requires higher python version than you have: ")
+                else _("\nThis cog requires a higher python version than you have: ")
             ) + humanize_list(outdated_python_version)
+        if newer_python_version:
+            message += (
+                _("\nThese cogs require a lower python version than you have: ")
+                if len(newer_python_version)
```
Vexed01:
```suggestion
                if len(newer_python_version) > 1
```

Vexed01 (commented):
I have tested this and can confirm it works as I described in #5590. In the process of testing I noticed the plural/singluar string selection doesn't actually function in the new code and in the min version code - see suggestions.

plooshi:
@Vexed01 i applied the patches you gave in the suggestions.

Drapersniper:
ensure_max_python_version_info

Can probably be simplified to 3,10 as micro versions shouldn't have any breaking changes.

PR #5435, opened 2021-11-02
[Core/Core Cogs] Prevent translation errors and use formatting utils

Dav-Git:
This PR proposes changes in accordance with #3919 to prevent errors in translation and uses the proper chat_formating utils wherever reasonable.

PR #5601, opened 2022-03-06
Fix formatting breaking with nested formatting utils in help

### LINKED ISSUE #5581: Formatting utils escapes if nested

dev

Just running help.

No escaping.

### What actually happened?

Text being escaped while it should not. Seems to be related off #5435.
image
image

Run any command that uses the formatters edited in the said PR.

### Anything else?

_No response_

Arman0334:

This PR is a simple fix to the broken formatting in the help command because of the nested formatters.

Fixes #5581

Before:
image

After:
image

Yes


Jackenmen (dismissed):

Jackenmen (approved):
I reverted a few changes to formatting that shouldn't really have been introduced in #5435. Time to approve again :)
hi mom

PR #5602, opened 2022-03-06
Improve changelog format

Jackenmen:

The current changelog format is not ideal so I tried to make a bunch of improvements to it.

A few alternative looks that have been considered for this change in changelog's format can be found in #4620.

The things I've done:
- The actual changelog files have been merged and moved into one called CHANGES.rst in the root directory
    - There now exists `docs/changelog.rst` file that only consists of an `include CHANGES.rst` statement
- Added changelogs for releases 3.0.1-3.1.9
- Made changelogs for all releases consistent
- Changed the format of the changelog:
    - What stays the same is that within each changelog there are sections (each is optional):
        - Read before updating
        - End-user changelog
        - Developer changelog
        - Documentation changes
    - However, now the changes are grouped by the type of change rather than its category, so e.g:
      image
      This is the list of change types (in the order they should appear in the changelog):
        - Security
        - Additions
        - Changes
        - Deprecations
        - Removals
        - Fixes
    - Additionally (with the exception of documentation changes), the change's category is still shown as a prefix for each entry (other than the few that apply to multiple categories):
      image
    - Category names have been reworked as follows:
        - Categories for each cog are now in a format "Cogs - CogName"
        - "Core Bot" category has now been further split into subcategories like "Core - Bot Commands" or "Core - Command-line Interfaces"
            - note that some entries may be just "Core" if they apply to multiple subcategories
        - The developer changelog now also contains categories

Things that still need to be considered:
- I was considering adding release highlights a long time ago so these could be added to changelogs (not necessarily all of them).
    - Release highlights would be a few changes that are most exciting in our opinion.
    - The format could either be a dedicated section *or* something like a ❤️ emoji before the changelog entries that we want to highlight.
- I should probably verify that making a redirect from changelog_3_4_0.html->changelog.html on RTD will keep the anchor link with it. If not, then oh well, we'll have to live with it I guess. But either way, we need to actually make the redirect once this PR is merged.

Yes


Flame442 (changes requested):
Is it possible to visually separate the releases more, maybe with a horizontal line between them? Having them all on one page and scrolling around, it's a bit hard to see where one begins and one ends at a glance.

I was also wondering about the old changelog links and if those would break, but it sounds like they would at least be able to redirect to the changelog page if not the specific release? If it will at least do that it's probably fine.

Jackenmen:
> Is it possible to visually separate the releases more, maybe with a horizontal line between them?

Sure, will do this in a moment.

> I was also wondering about the old changelog links and if those would break, but it sounds like they would at least be able to redirect to the changelog page if not the specific release?

Based on my previous testing, when browsers handle redirects, they keep the anchor link which means that as long as the anchor names stay the same (which they do), they should link to the changelog of a specific release (at least assuming that RTD doesn't do anything weird that would somehow strip the anchor link).

Flame442 (approved):

Jackenmen:
I've just added the redirects, they seem to be working as expected:
https://docs.discord.red/en/latest/changelog_3_4_0.html#redbot-3-4-11-2021-06-12

The changelog documents in the current stable documentation also work fine:
https://docs.discord.red/en/stable/changelog_3_4_0.html#redbot-3-4-11-2021-06-12

PR #5604, opened 2022-03-07
Add support for timeouts to Mutes cog

TrustyJAID:

Add timeouts support to the Mutes cog. This changes the default server mute behavior to utilize timeouts while maintaining role mutes if they have been previously setup. This adds a new `[p]timeout` command that exclusively uses discords built in timeout feature. This also allows the `[p]activemutes` command to list the currently muted and timed out users in the server. This also adds some models that make it easier for managing typed responses for internal muting methods.

Yes


BlizzardTheWolf:
Could we make it so that you can choose wether to use timeouts or mutes by using like `[p]muteset usetimeouts`, and having the command stay `[p]mute`, instead of having it be `[p]timeout`?

AAA3A-AAA3A:
> Could we make it so that you can choose wether to use timeouts or mutes by using like `[p]muteset usetimeouts`, and having the command stay `[p]mute`, instead of having it be `[p]timeout`?

I think we should have a `[p]timeout` command, but a setting could redirect `[p]mute` to `[p]timeout` if server's admins prefer this option to a real role.

BlizzardTheWolf:
Yeah that's a good idea!

> > Could we make it so that you can choose wether to use timeouts or mutes by using like `[p]muteset usetimeouts`, and having the command stay `[p]mute`, instead of having it be `[p]timeout`?
> 
> I think we should have a `[p]timeout` command, but a setting could redirect `[p]mute` to `[p]timeout` if server's admins prefer this option to a real role.

redbot/cogs/mutes/mutes.py
```diff
@@ -902,7 +905,7 @@ async def mute_role(self, ctx: commands.Context, *, role: discord.Role = None):
             await self.config.guild(ctx.guild).sent_instructions.set(False)
             # reset this to warn users next time they may have accidentally
             # removed the mute role
-            await ctx.send(_("Channel overwrites will be used for mutes instead."))
```
Flame442, line 905: If this PR is entirely replacing channel override mutes with timeout mutes, the docstring for this command and the output of `[p]muteset force_role_mutes` needs to be update to reflect that.

redbot/cogs/mutes/mutes.py
```diff
@@ -902,7 +905,7 @@ async def mute_role(self, ctx: commands.Context, *, role: discord.Role = None):
             await self.config.guild(ctx.guild).sent_instructions.set(False)
             # reset this to warn users next time they may have accidentally
             # removed the mute role
```
Flame442: It seems like this is no longer used, so it should probably be removed from here and the defaults.

redbot/cogs/mutes/mutes.py
```diff
@@ -1467,7 +1483,7 @@ async def unmute(
                         user, author, guild, _("Server unmute"), reason
                     )
                 else:
-                    issue_list.append(success)
+                    issue_list.append(response)
```
Flame442, line 1490: This appears to be appending a failing response twice. I believe the above check is because the failure message for moving users is successful with a reason, however it will also append an issue in cases where there was not a success, since they to have a reason set. They would then be appended again by this line.

redbot/cogs/mutes/voicemutes.py
```diff
@@ -222,7 +222,7 @@ async def unmute_voice(
                         user, author, guild, _("Voice unmute"), reason
                     )
                 else:
-                    issue_list.append((user, success["reason"]))
+                    issue_list.append((user, result.reason))
```
Flame442, line 225: Same issue with potentially double appending as before.

redbot/cogs/mutes/mutes.py
```diff
@@ -1560,43 +1576,31 @@ async def mute_user(
...
-            "user": user,
-        }
-        # TODO: This typing is ugly and should probably be an object on its own
-        # along with this entire method and some other refactorization
-        # v1.0.0 is meant to look ugly right :')
+        ret: MuteResponse = MuteResponse(success=False, reason=None, user=user)
+
         if permissions.administrator:
-            ret["reason"] = _(MUTE_UNMUTE_ISSUES["is_admin"])
+            ret.reason = _(MUTE_UNMUTE_ISSUES["is_admin"])
```
Flame442, line 1665: This isn't strictly an error with your PR, but there is no reason to be translating the strings both here and in the definition of `MUTE_UNMUTE_ISSUES`. This applies to this whole function, `unmute_user`, `channel_mute_user`, and `channel_unmute_user`.

Flame442 (changes requested):
Various developer-facing documentation should be updated to reflect an accurate description of the behavior under timeouts rather than channel override based mutes (lines 101, 315, 771).

I have not yet run the code, these are the issues I found from simply reviewing it.

Flame442:
@BlizzardTheWolf @AAA3A-AAA3A, the PR currently replaces the backup behavior for `[p]mute` when a mute role is not configured to use timeouts, rather than channel overrides. The `[p]timeout` command also exists for explicitly timing out a user, regardless of whether or not `[p]mute` is configured to use a role.

redbot/cogs/mutes/mutes.py
TrustyJAID: Actually this *is* necessary since the call to translate the string does not actually happen until here with the correct context. The initialized variables can be argued for not needing it and I can't remember the exact reason for why we left it this way and I can't find the discussion on it now. I just tested the change and can confirm that removing the translation `_()` from any of these cause the result to not correctly translate the message.

Jackenmen: `_()` calls in globals are so that redgettext can collect these strings but they're just passthrough and don't do anything (nor should they since the translations need to happen in a context where the locale is properly set).

`_()` calls in actual code are so that the strings actually get translated but those won't be collected by redgettext since they don't operate on string literals.

redbot/cogs/mutes/voicemutes.py
TrustyJAID: This, and the previous mention of a similar situation, are both intended. There are two lists we're dealing with here one for successful mutes we can largely ignore and another for ones with potential problems.

The only situation where we have successfully muted the user but still have a reason is when the user is currently in a voice channel and the bot was unable to move the user in the voice channel triggering discord to actually mute them. Therefore it should be notified to the command issuer that we were able to mute them but they may still be speaking in the voice channel and further action may need to be taken.

Flame442: I missed the if/else block...

redbot/cogs/mutes/mutes.py
```diff
@@ -831,7 +833,7 @@ async def force_role_mutes(self, ctx: commands.Context, true_or_false: bool):
         if true_or_false:
             await ctx.send(_("Okay I will enforce role mutes before muting users."))
         else:
-            await ctx.send(_("Okay I will allow channel overwrites for muting users."))
+            await ctx.send(_("Okay I will use timeouts for muting users."))
```
Flame442: I don't think this makes it use timeouts, it just allows timeouts to be used at all. I'm not sure if this setting is even needed, now that there's no API spam to need to opt in to.

Flame442 (dismissed):

Flame442 (dismissed):

redbot/cogs/mutes/mutes.py
```diff
@@ -1467,7 +1477,78 @@ async def unmute(
...
+        if not users:
+            return await ctx.send_help()
+        if ctx.me in users:
+            return await ctx.send(_("You cannot unmute me."))
+        if ctx.author in users:
+            return await ctx.send(_("You cannot unmute yourself."))
+        async with ctx.typing():
+            guild = ctx.guild
+            author = ctx.author
+            issue_list = []
```
Jackenmen, line 1532: Nitpick but `[p]unmute`/`[p]channelunmute` sets `audit_reason` generated from `reason` and passes that to `channel_unmute_user()` instead:
```suggestion
            audit_reason = get_audit_reason(author, reason, shorten=True)
            issue_list = []
```

redbot/cogs/mutes/mutes.py
```diff
@@ -1411,8 +1428,8 @@ async def channel_mute(
             )
         if issue_list:
             msg = _("The following users could not be muted\n")
-            for user, issue in issue_list:
-                msg += f"{user}: {issue}\n"
+            for issue in issue_list:
+                msg += f"{issue.user}: {issue.reason}\n"
```
Jackenmen:
```suggestion
                msg += f"- {issue.user}: {issue.reason}\n"
```

redbot/cogs/mutes/mutes.py
```diff
@@ -1549,8 +1643,8 @@ async def unmute_channel(
             )
         if issue_list:
             msg = _("The following users could not be unmuted\n")
-            for user, issue in issue_list:
-                msg += f"{user}: {issue}\n"
+            for issue in issue_list:
+                msg += f"{issue.user}: {issue.reason}\n"
```
Jackenmen:
```suggestion
                msg += f"- {issue.user}: {issue.reason}\n"
```

Jackenmen (approved):
meow?

PR #5613, opened 2022-03-10
Add a custom Logger class with both verbose and trace levels.

Drapersniper:

`logging.getLogger()` returns this custom class which has the 2 new logging levels.

~~Things to do~~  
- ~~Decide how to increase the verbosity of logs via CLI, I think adding `--trace` wouldn't be a good approach. a counter for `-v` would be the idea that's the industry standard for verbosity. (`-v` would enable debug, `-v -v ` or `-vv` would enable trace)~~

CLI now accepts `-v` and `--verbose` in addition to `--debug`
CLI is no longer a toggle but a counter  - Meaning `-v` will increase the verbosity of logs to DEBUG, and `-v -v -v` will increase it to TRACE

Yes

### DIFF
--- a/redbot/setup.py
+++ b/redbot/setup.py
@@ -17,7 +17,11 @@ import appdirs
 import click
 
 from redbot.core.cli import confirm
-from redbot.core.utils._internal_utils import safe_delete, create_backup as red_create_backup
+from redbot.core.utils._internal_utils import (
+    safe_delete,
+    create_backup as red_create_backup,
+    cli_level_to_log_level,
+)
 from redbot.core import config, data_manager, drivers
 from redbot.core.drivers import BackendType, IdentifierData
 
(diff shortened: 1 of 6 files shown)


Drapersniper:
Updated PR to remove dep due to its license - Added a custom Logger class instead

Drapersniper:
image
Latest commit regex tests

Drapersniper:
Merge first

redbot/setup.py
```diff
@@ -352,7 +356,7 @@ async def remove_instance_interaction() -> None:
 
 
 @click.group(invoke_without_command=True)
-@click.option("--debug", type=bool)
+@click.option("--debug", "-d", "--verbose", "-v", count=True)
```
```suggestion
@click.option("--debug", "--verbose", "-v", count=True)
```

Flame442 (dismissed):
image
Functions as expected. I have no opinions on the colors for each of the levels, however here is a screenshot of how it currently looks if anyone has a strong opinion that it should be changed.

Flame442 (approved):
image
image

PR #5615, opened 2022-03-12
Minor bugfixes resolving around modlog UI

Kowlin:

 * Fixed lack of new lines for `listcases`
     * regression introduced by #5435 
 * Fixed response text for case. It can only be used in the current server.

Yes

(To this day I'll never understand why we have ``listcases`` and ``casesfor``...)


Jackenmen (approved):
LGTM.

Looks like this is V3/dev-only regression, guess you could say I'm reviewing 3.5 PRs now /s

PR #5617, opened 2022-03-14
Pipinstall confirmation

### LINKED ISSUE #5614: Confirmation in the [p]pipinstall command.

Downloader Cog.

Having a confirmation on the ```pipinstall``` should be handy as to avoid typos and stuff. 
As mentioned in the docstring improper usage can potentially brick your bot. 

Something like this should do the trick.
image

### Anything else?

_No response_

Honkertonken:

Adds a simple reaction confirmation to the pipinstall command.

Resolves #5614.

Yes

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -1735,7 +1735,7 @@ class Downloader(commands.Cog):
                     if cog_installable.repo is None
                     else cog_installable.repo.name
                 )
-                cog_name = "{} ({})".format(cog_installable.name, cog.__class__.__name__)
+                cog_name = f"{cog_installable.name} ({cog.__class__.__name__})"
             elif cog.__module__.startswith("redbot."):  # core commands or core cog
                 made_by = "Cog Creators"
                 repo_url = "https://github.com/Cog-Creators/Red-DiscordBot"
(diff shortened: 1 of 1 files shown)


Honkertonken:
Apologies for the super messy commits

redbot/cogs/downloader/downloader.py
```diff
@@ -494,12 +494,23 @@ async def pipinstall(self, ctx: commands.Context, *deps: str) -> None:
 
         - `<deps...>` The package or packages you wish to install.
         """
+        confirmation_msg = await ctx.send(f"Are you sure you want to install `{(', '.join(deps))}`?")
```
Drapersniper: You are missing translation support for this string.

Red also has a `humanize_list` utility function which should be used when showing a list to end-users

Flame442:
We are not interested in this change at this time.

As `[p]pipinstall` is an owner only command, the owner could do similar things using `pip` in console, which also does not have protections. The kinds of mistakes a quick check like this would catch are typos, which would **not** break the bot as they would most likely be invalid libs. The only way to actually break the bot with this command would be to change the version of an existing lib that red depends on, however this change would not prevent that as that requires entering specific lib names.

In the future to avoid this, please only work on issues labeled as "Accepted". If the issue is still in "Needs Triage", feel free to say you are *interested* in working on the issue, but want to know if it's something we are interested in, that way we can let you know prior to you creating the PR.

Regardless, thank you for being interested in supporting the project.

PR #5624, opened 2022-03-18
Add and use Red-Commons library

Drapersniper:
yes


redbot/__init__.py
```diff
@@ -16,7 +16,7 @@
     Union as _Union,
 )
 
-from redbot._log import RedTraceLogger
+from red_commons.logging import maybe_update_logger_class
```
Jackenmen: You can't do this, this will cause Red to fail to build (because it's an external dependency).

Jackenmen (approved):

PR #5625, opened 2022-03-19
Make decorator combining in Requires more consistent

Jackenmen:

This changes the behavior of decorators in `redbot.core.commands.requires` to be more consistent.
Specifically, this code:
```py
@commands.bot_has_permissions(send_messages=True)
@commands.command()
async def somecom(ctx):
    ...


@commands.command()
@commands.bot_has_permissions(send_messages=True)
async def anothercom(ctx):
    ...


print(somecom.requires.bot_perms)
print(anothercom.requires.bot_perms)
```
will now give the same result for both prints which it did not before.

No


TrustyJAID (approved):
This looks to be working as intended.

PR #5626, opened 2022-03-20
Switch from low-level loop.create_task and asyncio.ensure_future

Jackenmen:

Replaces usage of `loop.create_task()` and `asyncio.ensure_future()` with `asyncio.create_task()`
There's only a single occurrence of `loop.create_task()` left in a place where it actually is needed.

Yes

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -819,7 +819,7 @@ class Permissions(commands.Cog):
                     cog_or_command.deny_to(model_id, guild_id=guild_id)
 
     def cog_unload(self) -> None:
-        self.bot.loop.create_task(self._unload_all_rules())
+        asyncio.create_task(self._unload_all_rules())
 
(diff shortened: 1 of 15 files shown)


redbot/cogs/permissions/permissions.py
```diff
@@ -819,7 +819,7 @@ def _load_rules_for(
                     cog_or_command.deny_to(model_id, guild_id=guild_id)
 
     def cog_unload(self) -> None:
-        self.bot.loop.create_task(self._unload_all_rules())
+        asyncio.create_task(self._unload_all_rules())
```
TrustyJAID, line 822: Do we want to just bulk replace with `asyncio.create_task` or utilize the new async `cog_unload` where applicable instead?

Jackenmen: This PR is independent of d.py 2.0 changes so no changes related to it should be done here.

TrustyJAID (approved):
Looks good to me!

PR #5627, opened 2022-03-20
Set Red._color before login rather than in on_ready

Jackenmen:

Does what it says on the tin, there's no reason for us to set this variable that late in the start-up process.

Yes

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -216,7 +216,6 @@ class Red(
 
         self._uptime = None
         self._checked_time_accuracy = None
-        self._color = discord.Embed.Empty  # This is needed or color ends up 0x000000
 
         self._main_dir = bot_dir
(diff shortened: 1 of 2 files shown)


redbot/core/bot.py
```diff
@@ -216,7 +216,6 @@ async def prefix_manager(bot, message) -> List[str]:
 
         self._uptime = None
         self._checked_time_accuracy = None
-        self._color = discord.Embed.Empty  # This is needed or color ends up 0x000000
```
TrustyJAID, line 219: Should we be assigning this attribute outside of `__init__`?

Jackenmen: No 3rd-party code executes before it's set in _pre_login() so it's fine. Same as with self.description

TrustyJAID (approved):
Looks good to me.

PR #5593, opened 2022-02-26
Audio changes

Drapersniper:
1 - Allow configuringJava's `-Xmx` flag (supersedes #5542)
2 - Replace the application.yml in favor of config storage.
3 - Allow toggling sources on and off
4 - Allow changing the managed node binding IP, password, and port.
5 - Allow customizing the framebuffer and NAS buffer
6 - Framework to modify the application YAML is baked in, all it takes to add more stuff (plugins for example) if following the structure currently present.
7 - Added captcha for bot owners for specific `llset` commands
8 - Reduce the number of config calls in `cog_before_invoke` by 4
9 - Unify all defaults into a single location to make changing them easier in the future
10 - Add `llset reset` which resets all changes made by `llset`
11 - Update docstring for all `llset` subcommands
12 - `llset host|password|port` are now only usable if you first enable an external Lavalink server.
13 - `llset info` update
14 - `llset yaml` added
15 - `audioset info` updated
16 - `-Xms64M` always set.
17 - `llset` commands will not be available depending on Node mode (i.e `llset config` will not be available if you run an external mode, `llset password` will not be available if you run a managed node)
18 - Too many fixes to mention all of them around the shutdown and reconnection logic
19 - `llset secured` added 

Depends on: 

~~https://github.com/Cog-Creators/Red-Lavalink/pull/118~~
https://github.com/Cog-Creators/Red-Lavalink/pull/117
https://github.com/Cog-Creators/Red-Lavalink/pull/116
https://github.com/Cog-Creators/Red-Lavalink/pull/115
https://github.com/Cog-Creators/Red-Lavalink/pull/114

To Test use: https://github.com/Cog-Creators/Red-Lavalink/pull/113

Before merging there are TODO's and dependencies changes required in the PR which are in place now just to facilitate testing

PR #5629, opened 2022-03-21
Add an option to auto-use default HTTP(S) port for Lavalink

Drapersniper:

Users are now allowed to set port to `-1` which will automatically set the connection ports to `80/443` depending on the security settings.

Sorry but you cannot test the 443 part of it as it is added on #5593

Yes


Drapersniper:
### What to test

Scenario 1:
`[p]llset host lavalink.oops.wtf`
`[p]llset port -1` 
`[p]llset secured` # Needs to be set to secured

It should connect successfully


Scenario 2:
`[p]llset host lavalink-coders.ml`
`[p]llset port -1` 
`[p]llset secured` # Needs to be set to unsecured

Scenario 3
Repeat Scenario 1 or 2 but have the incorrect `[p]llset secured` value
It should fail to connect

Scenario 4
Straight after scenario 3 change the secure state `[p]llset secure` to the correct value.

Scenario 5
`[p]llset port` docstring should tell users to set the value to `-1` to set it to "None"

### What to expect
1 - Scenario 1 user is able to connect successfully and play music
2 - Scenario 2 user is able to connect successfully and play music
3 - Scenario 3 connection will fail, it will attempt up to 10 times and then give up, Audio should say "Connection to Lavalink lost"
4 - Audio should tell the user to run `[p]audioset restart`, after running it user should be able to connect and play music
5 - Docstring is clear and directs the user on how to set the port to None

Jackenmen (changes requested):
I think it would have been clearer if, instead of "None", the "Port" field said something like "Default HTTP(S) port".
image

Jackenmen (approved):
Thanks!

PR #5632, opened 2022-03-22
add timestamps to all audio embeds

Drapersniper:
as the title states

Accidently closed #5620  and can't reopen due to force push

Yes


Drapersniper:
### What to test

1 - `[p]audioset notify`  and play a few tracks
2 - `[p]search`

### What to expect
1 - Timestamp to be present on all embeds (the 2 scenarios above cover the case where the embed already has a footer - if there's no weird formatting that means its looking good)

PredaaA (approved):

PR #5634, opened 2022-03-22
Add buttons to help

TrustyJAID:

This adds support for button menus in help and adds a SimpleMenu cog creators can import to simply paginate items with buttons similar to the existing menu in terms of available features.

Yes


redbot/core/utils/views.py
```diff
@@ -0,0 +1,174 @@
...
+            self.view.current_page = 0
+        elif self.direction == self.view.source.get_max_pages():
+            self.view.current_page = self.view.source.get_max_pages() - 1
+        else:
+            self.view.current_page += self.direction
+        kwargs = await self.view.get_page(self.view.current_page)
+        await interaction.response.edit_message(**kwargs)
+
+
+class SimpleMenu(discord.ui.View):
```
phenom4n4n, line 73: I think Views should be suffixed with View (`SimpleMenuView`? `PaginatedView`?) so it isn't confused with menus from the old ext library.

TrustyJAID: I don't see any confusion here? SimpleMenu could be put into the utils.menus namespace but since it's a view I think it makes more sense inside the new utils.views namespace.

Jackenmen (commented):
- Buttons aren't used if the bot doesn't have react perms.
- `[p]helpset usebuttons` and `[p]helpset usemenus` are currently separate settings but it might make sense for the setting to work as a tribool. Either way, the user should be informed (in help, maybe also in command responses) that setting both doesn't work/is required for buttons to function (depending on what you choose to do here)

TrustyJAID:
> * Buttons aren't used if the bot doesn't have react perms.

Should we use buttons if menus are toggled and the bot does not have reaction perms despite buttons not being set then?

> * `[p]helpset usebuttons` and `[p]helpset usemenus` are currently separate settings but it might make sense for the setting to work as a tribool. Either way, the user should be informed (in help, maybe also in command responses) that setting both doesn't work/is required for buttons to function (depending on what you choose to do here)

I thought it was clear in the response "if menus are also enabled."

npc203:
I'd support a tri-bool (a single command for toggling the three), one could get fancier by using a select menu for selecting the option perhaps.

TrustyJAID:
My concern with a tri-bool is that there will be 2 groups of people, some who want buttons and some who do not want buttons. With a tri-bool I now have to decide whether True uses buttons or False uses buttons which will then cause some people in either group to be forced to change their setting appropriately. The current way I have it setup the bot will startup the way they're expecting and can then optionally adjust their settings as they wish.

TwinDragon:
I'm with Trusty on this and having it as two separate settings. Though I do like buttons, for menus, I'd prefer a selection menu, but that's not what's being talked about. In this case, I think leaving it up to the bot owner to set whether or not the bot were to use buttons should remain as-is and would be easier from a UX point to help differentiate what they do, rather than having it all in a single setting.

Jackenmen:
What I meant by tribool would be sth like `[p]helpset menus buttons/reactions/disable` (`[p]helpset menus` being the command)

TrustyJAID:
Something I have been thinking about is the possibility of attaching a select menu to the SimpleMenu. The idea being that if toggled the view would include the drop down selector showing up to 25 pages on the menu, if there's more than 25 the select menu would paginate e.g. if there's 50 pages and you're on page 25 it would show pages 13-37 on the drop down selector. Is this something people want included in this? Is this something people would want on `[p]help` as well? If it's included on `[p]help` how should it work? should it be included with all buttons menus or should there be a way to use just buttons with no drop down selector?

TrustyJAID:
I have implemented my idea mentioned above and included options for buttons, buttons with select menu, and select menu only.

redbot/core/utils/views.py
```diff
@@ -0,0 +1,258 @@
...
+            ctx: `commands.Context`
+                The context to start the menu in.
+            ephemeral: `bool`
+                Send the message ephemerally. This only works
+                if the context is from a slash command interaction.
+        """
+        self.author = ctx.author
+        self.ctx = ctx
+        kwargs = await self.get_page(self.current_page)
+        self.message = await ctx.send(**kwargs, ephemeral=ephemeral)
```
Zephyrkul, line 238: I'm not sure if you plan on making this PR depend on a higher version of discord.py, but for the version of discord.py that Red currently installs this raises a `TypeError` for the unexpected `ephemeral=` kwarg.

TrustyJAID: I expect that by the time this PR gets fully reviewed and merged we will have bumped our d.py version to one that includes this kwarg on ctx. I added this as a sort of future-proof in the event cog creators want to use the menu as an ephemeral response from a slash command.

redbot/core/core_commands.py
```diff
@@ -3644,30 +3656,46 @@ async def helpset_resetsettings(self, ctx: commands.Context):
...
+        When "reactions" is passed, `[p]help` will only show one page at a time
+        and will use reactions to navigate between pages.
 
          **Examples:**
-            - `[p]helpset usemenus True` - Enables using menus.
-            - `[p]helpset usemenus` - Toggles the value.
+            - `[p]helpset usemenus reactions` - Enables using reaction menus.
+            - `[p]helpset usemenus buttons` - Enables using button menus.
+            - `[p]helpset usemenus select` - Enables buttons with a select menu.
+            - `[p]helpset usemensus selectonly` - Enables a select menu only on help.
```
Zephyrkul:
```suggestion
            - `[p]helpset usemenus selectonly` - Enables a select menu only on help.
```

redbot/core/utils/views.py
Kowlin: I think SimpleMenu being placed here in views is totally fine and appropriate for its function and goal.

Kowlin (approved):
Working like a charm, thank you.

PR #5635, opened 2022-03-22
Fix embed requested in DMs

TrustyJAID:

This fixes an error in #5576 where DM channels could fail

Yes

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1256,7 +1256,9 @@ class Red(
         if isinstance(channel, dpy_commands.Context):
             command = command or channel.command
             channel = (
-                channel.author if isinstance(ctx.channel, discord.DMChannel) else channel.channel
+                channel.author
+                if isinstance(channel.channel, discord.DMChannel)
+                else channel.channel
             )
 
         if isinstance(channel, discord.TextChannel):
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1269,7 +1269,6 @@ class Red(
 
             if (guild_setting := await self._config.guild(channel.guild).embeds()) is not None:
                 return guild_setting
-
         else:
             user = channel
             if (user_setting := await self._config.user(user).embeds()) is not None:
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1205,7 +1205,7 @@ class Red(
 
     async def embed_requested(
         self,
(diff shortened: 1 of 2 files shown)


Jackenmen (dismissed):
- Type hint for `channel` should be updated to reject `DMChannel` and `GroupChannel` so that their usage can be detected by type checker users.
- This fails for `commands.Context` with `DMChannel` under its `channel` attribute. You should set `channel` to `Context.author` in such a case.
- There's a bunch of formatting churn here that seems unnecessary.

redbot/core/bot.py
```diff
@@ -1240,10 +1247,15 @@ async def get_command_setting(guild_id: int) -> Optional[bool]:
...
+        if isinstance(channel, discord.GroupChannel) or isinstance(channel, discord.DMChannel):
+            # we should not be passing group channels to this function
+            # this only uses global settings
+            raise TypeError("You cannot pass a GroupChannel or DMChannel to this method")
+
         # using dpy_commands.Context to keep the Messageable contract in full
         if isinstance(channel, dpy_commands.Context):
             command = command or channel.command
-            channel = channel.channel
+            channel = channel.author
```
Jackenmen: That's not right for Context with any other `Context.channel` than `DMChannel` such as `TextChannel`

redbot/core/bot.py
```diff
@@ -1240,10 +1247,15 @@ async def get_command_setting(guild_id: int) -> Optional[bool]:
             scope = self._config.custom(COMMAND_SCOPE, command.qualified_name, guild_id)
             return await scope.embeds()
 
+        if isinstance(channel, discord.GroupChannel) or isinstance(channel, discord.DMChannel):
```
```suggestion
        if isinstance(channel, (discord.GroupChannel, discord.DMChannel)):
```

redbot/core/bot.py
```diff
@@ -1240,10 +1247,19 @@ async def get_command_setting(guild_id: int) -> Optional[bool]:
...
+            # we should not be passing group channels to this function
+            # this only uses global settings
```
Jackenmen: I'm not entirely sure this comment makes that much sense in here as it did when it had a different place to be in so it might make sense to remove it, your call.

Jackenmen (commented):
Code-wise seems fine, I'll test this later today or tomorrow.

Jackenmen (approved):
Works fine, thanks for the fix.

PR #5641, opened 2022-03-29
Add automodule reference for the antispam module

Kreusada:

Edits docstrings in the antispam module to be sphinx-like, and consequently adds an entry for the module into the utils framework documentation.

Yes


redbot/core/utils/antispam.py
```diff
@@ -8,15 +8,15 @@
...
-    Custom class which is more flexible than using discord.py's
-    `commands.cooldown()`
-
-    Can be intialized with a custom set of intervals
-    These should be provided as a list of tuples in the form
-    (timedelta, quantity)
-
-    Where quantity represents the maximum amount of times
-    something should be allowed in an interval.
+    An object that counts against intervals to prevent spam.
```
Jackenmen: I think this description is unclear. I think it's because the definition of "intervals" in this context isn't obvious. This probably applies to other docstrings as well.

redbot/core/utils/antispam.py
```diff
@@ -45,14 +45,13 @@ def __interval_check(self, interval: AntiSpamInterval):
     @property
     def spammy(self):
         """
-        use this to check if any interval criteria are met
+        Check whether any intervals are active.
```
Jackenmen: It's a property so it probably makes more sense to refer to it as an attribute:
```suggestion
        Whether any intervals are active.
```
As said in other review comments, I feel it's not that clear what "intervals" means.

redbot/core/utils/antispam.py
```diff
@@ -45,14 +45,13 @@ def __interval_check(self, interval: AntiSpamInterval):
...
         """
         return any(self.__interval_check(x) for x in self.__intervals)
 
     def stamp(self):
         """
-        Use this to mark an event that counts against the intervals
-        as happening now
+        Mark an event against the intervals as happening now.
```
Jackenmen: I don't really know how to describe this function better but I do think it's not easy to figure out what it means if you don't already understand what this class does.

Jackenmen (changes requested):
I've made a bunch of comments about the clarity of the docstrings. I don't want to require it for this PR but I do feel that some code example in class docstring or method docstrings might be helpful in understanding what this class can be used for and how it can be used.

redbot/core/utils/antispam.py
```diff
@@ -8,15 +8,17 @@
...
 
-    Can be intialized with a custom set of intervals
-    These should be provided as a list of tuples in the form
-    (timedelta, quantity)
-
-    Where quantity represents the maximum amount of times
-    something should be allowed in an interval.
+    Attributes
+    ----------
+    intervals : List[Tuple[datetime.timedelta, int]]
```
Jackenmen, line 71: I think that this a good description. It's just missing the information about the default intervals which are used when the list is empty.

redbot/core/utils/antispam.py
```diff
@@ -8,15 +8,17 @@
 
 class AntiSpam:
     """
-    Custom class which is more flexible than using discord.py's
-    `commands.cooldown()`
+    An object that counts against temporary "intervals"
+    to reduce spam. These intervals can be set to incline
+    in duration for each interval that is stamped.
```
Jackenmen: I think that this really needs an example:
```suggestion
    A class that can be used to count the number of events that happened
    within specified intervals. Later, it can be checked whether the specified
    maximum count for any of the specified intervals has been exceeded.

    Examples
    --------
    Tracking whether the number of reports sent by a user within a single guild is spammy:

    .. code-block:: python

        class MyCog(commands.Cog):
            INTERVALS = [
                # More than one report within last 5 seconds is considered spam.
                (datetime.timedelta(seconds=5), 1),
                # More than 3 reports within the last 5 minutes are considered spam.
                (datetime.timedelta(minutes=5), 3),
                # More than 10 reports within the last hour are considered spam.
                (datetime.timedelta(hours=1), 10),
                # More than 24 reports within a single day (last 24 hours) are considered spam.
                (datetime.timedelta(days=1), 24),
            ]

            def __init__(self, bot):
                self.bot = bot
                self.antispam = {}

            @commands.guild_only()
            @commands.command()
            async def report(self, ctx, content):
                # We want to track whether a single user within a single guild
                # sends a spammy number of reports.
                key = (ctx.guild.id, ctx.author.id)

                if key not in self.antispam:
                    # Create an instance of the AntiSpam class with given intervals.
                    self.antispam[key] = AntiSpam(self.INTERVALS)
                    # If you want to use the default intervals, you can use: AntiSpam([])

                # Check if the user sent too many reports recently.
                # The `AntiSpam.spammy` property is `True` if, for any interval,
                # the number of events that happened within that interval
                # exceeds the number specified for that interval.
                if self.antispam[key].spammy:
                    await ctx.send(
                        "You've sent too many reports recently, please try again later."
                    )
                    return

                # Make any other early-return checks.

                # Record the event.
                self.antispam[key].stamp()

                # Save the report.
                # self.config...

                # Send a message to the user.
                await ctx.send("Your report has been submitted.")
```

I've also tried to further improve the clarity of the class description in the suggestion as well.
You're welcome to further improve this if you can think of anything.

redbot/core/utils/antispam.py
```diff
@@ -8,15 +8,17 @@
...
 
-    Can be intialized with a custom set of intervals
-    These should be provided as a list of tuples in the form
-    (timedelta, quantity)
-
-    Where quantity represents the maximum amount of times
-    something should be allowed in an interval.
+    Attributes
```
Jackenmen: These are not attributes, they're parameters:
```suggestion
    Parameters
```

redbot/core/utils/antispam.py
```diff
@@ -45,14 +47,16 @@ def __interval_check(self, interval: AntiSpamInterval):
     @property
     def spammy(self):
         """
-        use this to check if any interval criteria are met
+        Whether any antispam intervals are active.
```
Jackenmen: It might be worth improving this further with the description that you can find in the comment above the `AntiSpam.spammy` usage in the code example I suggested for the class docstring.

redbot/core/utils/antispam.py
```diff
@@ -45,14 +47,16 @@ def __interval_check(self, interval: AntiSpamInterval):
...
         """
         return any(self.__interval_check(x) for x in self.__intervals)
 
     def stamp(self):
         """
-        Use this to mark an event that counts against the intervals
-        as happening now
+        Mark against the list of antispam intervals.
```
Jackenmen: Maybe this should just say that we're recording an event (or event timestamp)? It probably wouldn't hurt to mention that it records this event as something that happened *now*.

redbot/core/utils/antispam.py
```diff
@@ -45,14 +47,16 @@ def __interval_check(self, interval: AntiSpamInterval):
...
+
+        The stamp will last until the corresponding interval duration
```
Jackenmen, line 121: This looks alright.

docs/framework_utils.rst
```diff
@@ -74,8 +74,16 @@ Common Filters
...
     :members:
 
+==========
 Utility UI
 ==========
 
 .. automodule:: redbot.core.utils.views
     :members:
+
+========
```
Jackenmen: It doesn't seem to make sense for these two to have a higher heading than the other utility modules, I think.
```suggestion
Utility UI
==========

.. automodule:: redbot.core.utils.views
    :members:

```

Jackenmen (dismissed):

redbot/core/utils/antispam.py
Kreusada: Happy with the changes and example you've added, thanks

redbot/core/utils/antispam.py
Flame442:
```suggestion
        Whether, for any interval, the number of events that happened
        within that interval exceeds the number specified for that interval.
```

redbot/core/utils/antispam.py
```diff
@@ -45,14 +102,16 @@ def __interval_check(self, interval: AntiSpamInterval):
...
-        use this to check if any interval criteria are met
+        Whether any antispam intervals are active.
         """
         return any(self.__interval_check(x) for x in self.__intervals)
 
     def stamp(self):
         """
-        Use this to mark an event that counts against the intervals
-        as happening now
+        Mark against the list of antispam intervals.
```
Flame442:
```suggestion
        Mark an event timestamp against the list of antispam intervals happening right now.
        Use this after all checks have passed, and your action is taking place.
```

redbot/core/utils/antispam.py
```diff
@@ -8,15 +8,72 @@
...
+                # Send a message to the user.
+                await ctx.send("Your report has been submitted.")
+
+    Parameters
+    ----------
+    intervals : List[Tuple[datetime.timedelta, int]]
+        A list of tuples in the format (timedelta, int),
+        where the timedelta represents the length of the interval,
+        and the int represents the maximum number of times something
+        can happen within that interval.
```
Flame442, line 75:
```suggestion
        A list of tuples in the format (timedelta, int),
        where the timedelta represents the length of the interval,
        and the int represents the maximum number of times something
        can happen within that interval.
        
        If an empty list is provided, the following defaults will be used:
        - 3 per 5 seconds
        - 5 per 1 minute
        - 10 per 1 hour
        - 24 per 1 day
```
Not certain about the exact formatting of this one, but something like this.

Flame442 (changes requested):
Added some direct suggestions for handling Jack's review comments. Other than what Jack already said, this looks fine to me.

Kreusada:
Appreciate you making those changes for me, Flame

There is a preview of the docs [here](https://red-discordbot--5641.org.readthedocs.build/en/5641/framework_utils.html#module-redbot.core.utils.antispam) if you want to check. My other thought is, do you think the AntiSpamInterval namedtuple should be excluded from the automodule? I don't see a reason for it to be documented, correct me if I'm wrong.

Flame442:
I believe `AntiSpamInterval` is just an internal used for this class to do its work, so it *should* be considered private. Our conventions for privatization aren't followed perfectly throughout the codebase, but I'd say it should at least be excluded from the docs for now, if not also made more explicitly private in the code with leading `_`s.

Flame442 (approved):
Thanks for handling those changes!

PR #5643, opened 2022-03-30
Fix an incorrect comparison in audio

Drapersniper:
Signed-off-by: Draper <27962761+Drapersniper@users.noreply.github.com>

Fix an issue with `Managed Lavalink node RAM allocation ignored due to system limitations, please fix this by setting the correct value with 'ff-llset heapsize'.` being logged unintentionally.

No


Flame442 (approved):
The warning is gone and audio is not any more broken than before!

PR #5647, opened 2022-03-31
[Modlog] Check if guild is unavailable

kpopdev:

Fix for  `AttributeError: 'NoneType' object has no attribute 'guild_permissions'` error.


Jackenmen (approved):
Seems reasonable, thanks.

PR #5654, opened 2022-04-01
Best Red PR to ever exist

Drapersniper:

This PR is revolutionary and is guaranteed to reduce the number of issues in #support by at least 500%.
Not to mention the changes included here will also improve the support ETA for users as there will be fewer users and fewer issues and users to deal with.

With 100% fewer random-ass issues which shouldn't happen.

Also includes a 2-fold performance improvement due to fewer config calls and listeners.

Yes


tests/cogs/test_permissions.py
```diff
@@ -7,7 +7,6 @@ def test_schema_update():
             "owner_models": {
                 "cogs": {
                     "Admin": {"allow": [78631113035100160], "deny": [96733288462286848]},
-                    "Audio": {"allow": [133049272517001216], "default": "deny"},
```
Jackenmen, line 10: Instead of removing this test, please replace the cog name instead.

tests/cogs/test_permissions.py
```diff
@@ -46,7 +45,6 @@ def test_schema_update():
                 str(GLOBAL): {"78631113035100160": True, "96733288462286848": False},
                 "43733288462286848": {"24231113035100160": True, "35533288462286848": False},
             },
-            "Audio": {str(GLOBAL): {"133049272517001216": True, "default": False}},
```
Jackenmen, line 49: Instead of removing this test, please replace the cog name instead.

Jackenmen (changes requested):

PR #5656, opened 2022-04-01
Use Guild.fetch_ban() over Guild.bans()

Jackenmen:

In all of these cases, there isn't really much point in fetching the whole ban list over fetching single bans. This should make it run faster for guilds with a big ban list *and* it addresses the breaking change that Discord made yesterday that added paginations to fetch bans endpoint.

No


TrustyJAID (approved):
LGTM

PR #5664, opened 2022-04-03
Include tag distance and commit hash in dev versions when possible

Jackenmen:

This turns `3.5.0.dev1` into `3.5.0.dev172+gc9a09719` when Red detects that it's installed/running from git repository.
The exact form is: `3.5.0.dev{distance}+g{short_commit_hash}[.dirty]`
- `{distance}` is the distance from the closest tag, e.g. if two commits have been made between `3.4.12` and the current commit (not including `3.4.12` itself), this would be `2`
- `{short_commit_hash}` is the short hash of the current commit 
- `.dirty` is optionally added suffix when Red is run from a repository that has some unstaged changes

Same as with other projects that do this sort of thing, the version string won't be turned when installing a dev version from a location that isn't a git repository, e.g. from `https://github.com/Cog-Creators/Red-DiscordBot/tarball/V3/develop`.

Yes


TrustyJAID (dismissed):
I like this change!

TrustyJAID (approved):

PR #5674, opened 2022-04-06
Use different exit codes for critical errors vs configuration errors

### LINKED ISSUE #5069: Use a different exit code for configuration errors

Bot Core

It might be worth using different exit code(s) (different from 1 that is) when Red has stopped due to an error that will certainly not be fixed just by a restart:
- missing intents
- no owner set (related to #4926)
- no token/prefix set when running with `--no-prompt`
- invalid token
- incorrect cli arguments passed (this is already mostly done by argparse)

This should also come with an update to auto-restart guides so that these different exit codes prevent continuous auto-restarting.

### Anything else?

The part with auto-restart guides might not be achievable on Mac - it doesn't seem that launchd can differentiate based on exit codes whether it should restart or not (it seems to only be able to differentiate between zero and non-zero), perhaps a special compatibility mode flag could be added?

If we want to get suggested by some source on the numbers to use for the exit codes, we can look [here](https://km.kkrach.de/p_exit_and_error_codes/) and [here](https://www.freedesktop.org/software/systemd/man/systemd.exec.html#Process%20Exit%20Codes) though it doesn't really seem like there's that much conventions for this and we can just come up with our own exit code numbers.

Jackenmen:

Fixes #5069 
Adds new possible exit codes: `2` (invalid cli usage) and `78` (configuration error)
These are now used appropriately in the bot which allows performing a different action depending on this value.
I updated the systemd service to work better now that we use different exit codes for critical errors (where restart is *usually* desired) and invalid CLI usage/configuration errors (where restart should always be not desired).
According to launchd's (Mac's service manager) manual, there is no option to discern between different non-zero exit codes, it treats them all the same. This means that Mac's service configuration will continue to work the same as it did before, sadly with no improvements.

Yes

### DIFF
--- a/redbot/core/cli.py
+++ b/redbot/core/cli.py
@@ -2,13 +2,28 @@ import argparse
 import asyncio
 import logging
 import sys
+from enum import IntEnum
 from typing import Optional
 
 import discord
 from discord import __version__ as discord_version
 
-from redbot.core.bot import ExitCodes
 from redbot.core.utils._internal_utils import cli_level_to_log_level
+
+
+# This needs to be an int enum to be used
+# with sys.exit
+class ExitCodes(IntEnum):
+    #: Clean shutdown (through signals, keyboard interrupt, [p]shutdown, etc.).
+    SHUTDOWN = 0
+    #: An unrecoverable error occurred during application's runtime.
+    CRITICAL = 1
+    #: The CLI command was used incorrectly, such as when the wrong number of arguments are given.
+    INVALID_CLI_USAGE = 2
+    #: Restart was requested by the bot owner (probably through [p]restart command).
+    RESTART = 26
+    #: Some kind of configuration error occurred.
+    CONFIGURATION_ERROR = 78  # Exit code borrowed from os.EX_CONFIG.
 
 
 def confirm(text: str, default: Optional[bool] = None) -> bool:
--- a/redbot/core/cli.py
+++ b/redbot/core/cli.py
@@ -7,6 +7,7 @@ from typing import Optional
 import discord
 from discord import __version__ as discord_version
 
+from redbot.core.bot import ExitCodes
 from redbot.core.utils._internal_utils import cli_level_to_log_level
 
 
@@ -25,7 +26,7 @@ def confirm(text: str, default: Optional[bool] = None) -> bool:
             value = input(f"{text}: [{options}] ").lower().strip()
         except (KeyboardInterrupt, EOFError):
             print("\nAborted!")
-            sys.exit(1)
+            sys.exit(ExitCodes.INVALID_CLI_USAGE)
         if value in ("y", "yes"):
             return True
         if value in ("n", "no"):
--- a/redbot/core/cli.py
+++ b/redbot/core/cli.py
@@ -39,7 +39,10 @@ def confirm(text: str, default: Optional[bool] = None) -> bool:
     while True:
         try:
             value = input(f"{text}: [{options}] ").lower().strip()
-        except (KeyboardInterrupt, EOFError):
+        except KeyboardInterrupt:
+            print("\nAborted!")
+            sys.exit(ExitCodes.SHUTDOWN)
+        except EOFError:
             print("\nAborted!")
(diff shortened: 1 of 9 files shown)


Jackenmen:
I went with restart *only* on 1 and 26 *but* I could alternatively do restart on *all* but 0, 2, and 78. The question is which one makes more sense here. It's not a given that we don't add more exit codes in the future and those codes would probably be non-restartable ones which would mean that the current state of this PR is preferred? I don't think adding *another* exit code on which we would want to restart is that likely in comparison.

redbot/core/cli.py
```diff
@@ -25,7 +41,7 @@ def confirm(text: str, default: Optional[bool] = None) -> bool:
             value = input(f"{text}: [{options}] ").lower().strip()
         except (KeyboardInterrupt, EOFError):
             print("\nAborted!")
-            sys.exit(1)
+            sys.exit(ExitCodes.INVALID_CLI_USAGE)
```
Flame442, line 47: This doesn't match with other places where this error code is used. This is generally used for invalid usage of CLI flags, not the user ctrl c-ing. Other places which handle ctrl c, like when the user is prompted to enter their token, it uses `SHUTDOWN`. I think this makes more sense, since the user is explicitly trying to stop whatever is happening, even if it's done forcefully.

redbot/setup.py
```diff
@@ -30,7 +31,7 @@
     config_dir.mkdir(parents=True, exist_ok=True)
 except PermissionError:
     print("You don't have permission to write to '{}'\nExiting...".format(config_dir))
-    sys.exit(1)
+    sys.exit(ExitCodes.CRITICAL)
```
Flame442: This should probably be `CONFIGURATION_ERROR`, since restarting won't give permission to write to this dir.

redbot/setup.py
```diff
@@ -77,7 +78,7 @@ def get_data_dir(*, instance_name: str, data_path: Optional[Path], interactive:
             "We were unable to check your chosen directory."
             " Provided path may contain an invalid character."
         )
-        sys.exit(1)
+        sys.exit(ExitCodes.CRITICAL)
```
Flame442: This should probably be `CONFIGURATION_ERROR`, since restarting won't make the path valid.

Flame442 (changes requested):
A few of the exit codes used in certain places don't make sense to me, and I think they need a second look to see what they should be set to.

redbot/setup.py
Jackenmen: I think this should be `INVALID_CLI_USAGE` as it concerns current user input rather than the existing configuration which is what all `CONFIGURATION_ERROR`s are for.

`CRITICAL` doesn't seem to make much sense here though, yeah.

redbot/setup.py
Jackenmen: Yeah, probably makes more sense. I'm guessing I *probably* used `CRITICAL` because it concerns `redbot-setup` which isn't a daemon process (unlike `redbot`) anyway and you shouldn't restart it after failure.

It's also possible that I didn't want to use `CONFIGURATION_ERROR` in `redbot-setup` at all, maybe because places where it could make sense don't concern Red's configuration. I don't know...

I think `CONFIGURATION_ERROR` probably makes more sense here, I'll make the change.

redbot/core/cli.py
Jackenmen: This is also a handler for `EOFError` which can occur in a non-interactive context where STDIN is closed. Therefore, it's possible that when used in a script, `redbot-setup` wasn't supposed to be aborted but due to incorrect usage it was. We could perhaps split this into two exception handlers instead though.

Jackenmen:
Addressed the review, I've done one unrelated change to a message in redbot-setup failure mentioning running as admin because it bothered me:
[`85a4c47` (#5674)](https://github.com/Cog-Creators/Red-DiscordBot/pull/5674/commits/85a4c477fc6ba9720b0e1375a7bdc19459e752d6)

Flame442 (approved):

PR #5678, opened 2022-04-09
Make `controls` in `menu()` optional

Jackenmen:

When `controls` are not passed to `menu()` it will now use `DEFAULT_CONTROLS` or only show the close menu button if there's only one page.
I updated all usage in core as well, with the exception of help where even though we use default controls, we use the menu in a bit more special way which doesn't allow us to swap it.

No


TrustyJAID (approved):
Looks good to me!

PR #5683, opened 2022-04-15
Add global buttons to base menus

TrustyJAID:
This adds a bot setting for using buttons which is also used by cores menus util.

~~This is blocked by #5634~~

Yes


redbot/core/utils/menus.py
```diff
@@ -62,6 +87,15 @@ async def menu(
     RuntimeError
         If either of the notes above are violated
     """
+    if ctx.message.id in _active_menus:
```
phenom4n4n: What if someone creates multiple menus in the same command invocation?

TrustyJAID: This is still possible, assuming you're using the menu in the intended way e.g. awaiting it. You really shouldn't be sending the menu out in its own task and sending another one immediately after from the same context though since you can modify the existing menu however you want, minus the controls.

Kowlin (approved):

PR #5685, opened 2022-04-19
Allow sending the file in follow-up message in ACL upload commands

### LINKED ISSUE #5473: Allow sending the file in follow-up message in ACL upload commands

ACL upload commands in Permissions cog

The current error message in ACL upload commands ("You must upload a file.") is not very clear about having to run the command file again with the file being attached.

I suggest that rather than improve the error message, we improve the user experience by allowing to send the ACL file in a follow-up message, similarly to how `[p]playlist upload` and `[p]triviaset custom upload` work.

### Anything else?

`[p]set avatar` (and perhaps some other commands?) is another example that doesn't allow sending the file in a follow-up message so it might be worth changing that as well.

matcha19:
Changes permissions so that it allows users to send the ACL file in a follow-up message, similarly to how [p]playlist upload and [p]triviaset custom upload work. 

This fixes #5473.

Yes

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -632,11 +632,23 @@ class Permissions(commands.Cog):
     ) -> None:
         """Set rules from a YAML file and handle response to users too."""
         if not ctx.message.attachments:
-            await ctx.send(_("You must upload a file."))
-            return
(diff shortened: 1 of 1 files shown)


Jackenmen (approved):
Works great, thanks!

Beep boop 🤖

PR #5692, opened 2022-05-06
Remove `bordered()` utility function

Jackenmen:

This has probably never been used outside of Core which makes it a perfect candidate for private namespace *but*, since we also no longer use it, we can just drop this code entirely.

Yes


Kowlin (approved):
Hah, that's an relic from the past before we transitioned to Rich for the CLI.

Jackenmen:
Seems like "this has probably never been used outside of Core" is technically wrong however, I don't think that should really stop us as it doesn't really fit into chat formatting utils...

https://github.com/Kreusada/Kreusada-Cogs/blob/b1e66746ed39d208a6f157419cc1a05a48769119/captcha/commands/settings.py#L195

(don't let the `forms` fool you, it's `chat_formatting`)

PR #5693, opened 2022-05-08
Prevent `/` being used in bot or server prefixes

Kreusada:

Prevents the use of "/" for the bot and guild's prefixes.

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -138,7 +138,7 @@ async def _edit_prefix(red, prefix, no_prompt):
             if not prefixes:
                 print("You need to pass at least one prefix!")
                 continue
-            if prefix.startswith("/"):
+            if any(prefix.startswith("/") for prefix in prefixes):
                 print(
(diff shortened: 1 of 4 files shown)


phenom4n4n:
I believe this should raise an error in the `Red.set_prefixes` method to prevent external sources from setting `/` as a prefix.

redbot/__main__.py
```diff
@@ -138,6 +138,11 @@ async def _edit_prefix(red, prefix, no_prompt):
             if not prefixes:
                 print("You need to pass at least one prefix!")
                 continue
+            if prefix.startswith("/"):
```
Flame442: This should be `prefixes`.

`prefix` is false-like here, if it is a string it will always be empty. This will not have the desired effect, and may error (the function isn't typehinted so I don't know for certain if it would be an empty string or `None`).

Flame442 (changes requested):

Flame442 (approved):
Thanks for your contribution!

PR #5698, opened 2022-05-10
Improve performance of pagify()

Jackenmen:

Drops down the (more or less) quadratic time closer to linear.

I want to use this as an opportunity to add unit tests for `pagify()` (to make sure I didn't alter the behavior of it) so this is a draft for now.

Yes

### DIFF
--- a/docs/framework_utils.rst
+++ b/docs/framework_utils.rst
@@ -27,6 +27,10 @@ Chat Formatting
 
 .. automodule:: redbot.core.utils.chat_formatting
     :members:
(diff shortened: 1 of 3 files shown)


Jackenmen:
Alright, I added some tests, as well as support for `operator.length_hint(...)`. The latter required a refactor of `pagify()` into a class but this should be a transparent change.

Flame442 (approved):
Much faster!

Jackenmen:
SPEEEEEEEEED

PR #5701, opened 2022-05-11
Add Fedora 36 and RHEL 8.6 to list in version guarantees

Jackenmen:

Fedora 34 is not yet EOL but it pretty much is going to be by the time 3.4.17 is released so I removed it. I added 36 to the list and updated EOL dates.
I added RHEL 8.6 as well since that's also been released around the same time.

Yes


Jackenmen:
This is just a regular update to our supported OS list + the appropriate bump of the version in the Fedora install guide, gonna bypass the review here.

PR #5705, opened 2022-05-18
Change note to attention

PythonCoderAS:

Changes a note box to an attention box. This is done because the contents of the box should be emphasized. 

Yes

### DIFF
--- a/docs/guide_cog_creation.rst
+++ b/docs/guide_cog_creation.rst
@@ -28,7 +28,7 @@ Depending on what your goal is should help determine which version you need.
 Open a terminal or command prompt and type one of the following
     Stable Version: :code:`python3.8 -m pip install -U Red-DiscordBot`
 
-.. note::
+.. attention::
 
   To install the development version, replace ``Red-DiscordBot`` in the above commands with the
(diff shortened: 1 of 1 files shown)


flaree:
imo development version should not be emphasised to users. It could contain anything from  breaking changes to in development featues.

Not to mention that using development is technically grounds for no support, so more people potentially just blindly copy pasting something because its more emphasised is not great.

PythonCoderAS:
> imo development version should not be emphasised to users. It could contain anything from breaking changes to in development featues.
> 
> Not to mention that using development is technically grounds for no support, so more people potentially just blindly copy pasting something because its more emphasised is not great.

Sorry, I meant that the **danger** should be emphasized more.

Flame442:
image
This is how it currently appears. `Attention` is used to outline the risks of the development version and `Note` is used for the steps to actually install it (with its own little warning as well to be safe). I feel like the danger is stated in multiple places and it is emphasized as appropriate

PythonCoderAS:
I think the "No support" line should at least be moved to the earlier box.

Jackenmen:
Thanks for your contribution but currently we are not interested in this change.

PR #5711, opened 2022-05-27
Prepend emojis to better differentiate between `[p]canrun` responses

Kreusada:

Prepend ✅ or 🚫 to `[p]canrun` responses to better differentiate between responses.

Yes

### DIFF
--- a/redbot/cogs/permissions/permissions.py
+++ b/redbot/cogs/permissions/permissions.py
@@ -11,7 +11,7 @@ from redbot.core import checks, commands, config
 from redbot.core.bot import Red
 from redbot.core.i18n import Translator, cog_i18n
 from redbot.core.utils import can_user_react_in
-from redbot.core.utils.chat_formatting import box
+from redbot.core.utils.chat_formatting import box, error, success
 from redbot.core.utils.menus import start_adding_reactions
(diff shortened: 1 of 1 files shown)


redbot/cogs/permissions/permissions.py
```diff
@@ -262,9 +262,9 @@ async def permissions_canrun(
                 can = False
 
             out = (
-                _("That user can run the specified command.")
+                "\N{WHITE HEAVY CHECK MARK} " + _("That user can run the specified command.")
```
Jackenmen:
```suggestion
                success(_("That user can run the specified command."))
```

redbot/cogs/permissions/permissions.py
```diff
@@ -262,9 +262,9 @@ async def permissions_canrun(
...
                 if can
-                else _("That user can not run the specified command.")
+                else "\N{CROSS MARK} " + _("That user can not run the specified command.")
```
Jackenmen:
```suggestion
                else error(_("That user can not run the specified command."))
```

Jackenmen (changes requested):
I think `success()` and `error()` from chat formatting utilities should be used here instead.

Jackenmen (approved):
Works fine, thanks for the quick update.

PR #5720, opened 2022-06-01
Add Ubuntu 22.04 install guide

Jackenmen:

Adds Ubuntu 22.04 to the install guides and declares it as the new recommendation.
These docs will eventually be updated to not use deadsnakes repository but that can't happen until we add support for Python 3.10. This new install guide follows the new guideline I want to follow: not to use external repositories when not necessary, so git PPA is not being used here.

No


Jackenmen:
These are already used instructions in our support channels so I'm going to bypass QA here and just merge it.

PR #5774, opened 2022-06-11
Use newer RTD config syntax

Jackenmen:

I like shiny new things. And they call it "Legacy `build` specification" now, I don't like being legacy :(

Yes


Jackenmen:
While testing this, I experienced a 26-second slowdown, let's not go down this road for now...

PR #5778, opened 2022-06-15
Exclude `files_from_attatch` member from Tunnel automodule

Kreusada:

Excludes the `files_from_attatch` member from Tunnel automodule, as this "alias" doesn't need to be documented aswell.

Yes

### DIFF
--- a/docs/framework_utils.rst
+++ b/docs/framework_utils.rst
@@ -67,6 +67,7 @@ Tunnel
 
 .. automodule:: redbot.core.utils.tunnel
     :members: Tunnel
+    :exclude-members: files_from_attatch
 
 Common Filters
 ==============


Fabian-Evolved:
why is it 'files_from_attatch' and not 'files_from_attach'?

Flame442:
When this function was first made, it was misspelled in this manor. To avoid a breaking change, it was renamed to the new spelling but maintained an alias to the old spelling. That alias does not matter, and does not need to be documented, so this PR serves to *hide* the alias.

Jackenmen:
> why is it 'files_from_attatch' and not 'files_from_attach'?

As somewhat explained in the PR description, it's an alias for `files_from_attach` that's been left in Red for backward compatibility after the typo-ed `files_from_attatch` was renamed to `files_from_attach`. The point of this PR is to exclude it from the documentation as it doesn't really make sense to show the wrongly spelled version there.

Kowlin (approved):
Thank you! *Insert cat GIF, but too lazy to look one up.*

PR #5781, opened 2022-06-22
[Core] Use the `discord.User` converter for the dm command to avoid accepting only the id of a user.

AAA3A-AAA3A:

Hello,
Here is a PR to allow the dm command to use the `discord.User` converter to support usernames and mentions as valid command arguments.
We had talked about this quickly on the main Red server and you were not against the idea.
I know that the changes for dpy2 are more important. I didn't open a discussion because the change doesn't really require much thought and is just a minor change to a basic command.
Thanks in advance for your help,
AAA3A

Yes

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -4059,8 +4059,14 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
             - `[message]` - The message to dm to the user.
         """
 
-        if user.bot:
-            await ctx.send(_("Sorry, the user {} is a bot.").format(user))
+        if user.bot or not user.mutual_guilds:
+            await ctx.send(
+                _(
+                    "User is a bot, or user not found.. "
+                    "You can only send messages to people I share "
+                    "a server with."
+                )
+            )
             return
         prefixes = await ctx.bot.get_valid_prefixes()
         prefix = re.sub(rf"<@!?{ctx.me.id}>", f"@{ctx.me.name}".replace("\\", r"\\"), prefixes[0])
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -4046,7 +4046,7 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         async def convert(self, ctx: commands.Context, argument: str):
             if not self._get_id_match(argument) or re.match(r'<@!?([0-9]{15,20})>$', argument):
                 arg = argument
-                if arg[0] == '@':
+                if arg.startswith('@'):
                     # Remove first character
                     arg = arg[1:]
                 if not (len(arg) > 5 and arg[-5] == '#'):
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -4058,16 +4058,6 @@ async def dm(self, ctx: commands.Context, user_id: int, *, message: str):
...
-        destination = self.bot.get_user(user_id)
-        if destination is None or destination.bot:
-            await ctx.send(
-                _(
-                    "Invalid ID, user not found, or user is a bot. "
-                    "You can only send messages to people I share "
-                    "a server with."
-                )
-            )
-            return
```
yamikaitou, line 4070: You should retain the check to see if the user is a Bot. It'd also be nice to retain the check to see if the user shares a server with the bot.

AAA3A-AAA3A: I agree with the verification if it is a bot or not. On the other hand, `discord.User` is a converter for users who have access to the bot, which means, under normal conditions, that there is no need to check.

AAA3A-AAA3A: Both `discord.User` and `bot.get_user` do not use the api. The converter will never give `None` as the user.

yamikaitou: As of d.py 1.6, the converter does indeed perform an API call if the user does not exist in the cache. In my testing as well, I was able to successfully get the user data of someone not shared with the bot

Flame442: `bot.get_user` does not, but iirc the `discord.User` converter *does* resort to a `bot.fetch_user` call if its `get` does not find the user. The `fetch_user` endpoint is also pretty restrictive, but this *is* a bot owner command, so I don't know how huge of a deal that is.

AAA3A-AAA3A: So the user may not share a server with the bot? That's a bit silly. Dpy should have a converter that only works on shared users.

AAA3A-AAA3A: This is corrected. And I tested.

aikaterna:
What happens when your bot can see five users named "John" that are not in your server, and you try to target "John" and it blindly resolves to a different one than you wanted? While this sort of PR may be convenient for you in what you are thinking for your use cases, IDs provide a way of contacting a user that is 100% not going to be misconstrued and I feel like it's less nebulous than trying to provide a name where the bot chooses with no feedback.

AAA3A-AAA3A:
Change the converter of `discord.User` to not support untagged names if multiple users are found.

aikaterna:
> Change the converter of `discord.User` to not support untagged names if multiple users are found.

To me, this sentence reads like you're suggesting that we should go and make a PR to discord.py over this, which is not exactly ideal and most likely won't be happening.

AAA3A-AAA3A:
Make a subclass of `discord.ext.commands.converters.UserConverter`. It could return a raise under certain conditions and then call `super().convert()`.

AAA3A-AAA3A:
Hello,
I have subclassed the user converter so that it blocks an argument without a specified tag. This code necessarily works because it first checks whether it is an id or not, normally, before spotting the #, exactly as in dpy.
(Tested under normal conditions under the latest development version of Red 3.5 but also of dpy2, using a patch file.)
I don't know if this is really good for Core or if the method used is good (maybe add an equivalent to Red or dpy?).
Thanks in advance,
AAA3A

image

Jackenmen:
I don't think we really want this, sorry.

PR #5794, opened 2022-07-03
[Dev] Fix `__repr__()` errors in REPL when referencing an instance of a class & catch Exception

Kreusada:

I have noticed that when `__repr__()` is to raise an exception, the `[p]repl` command fails, and the exception is explicitly raised instead of being sent as a message in the session. However, this only happens when the instance is referenced (`object()`) and not when the `repr()` function is used (`repr(object())`).

Take the following code, for example:

```py
class A(object):
    def __repr__(self) -> str:
        raise Exception
```

If we were to type `A()`, Exception would be explicitly raised as a command error. However, typing `repr(A())` would behave as you would expect.

I discovered this with my own code in a REPL session:

```py
def __repr__(self) -> str:
    attrs = {n[1:]: getattr(self, n) for n in self.__slots__}
    join = ", ".join(f"{k}={v!r}" for k, v in attrs.items())
    return f"{self.__class__.__name__}({join})"
```

I had not included the `__slots__` definition within the class, hence why an AttributeError was raised when I typed "MyClass()". I was expecting this error to be printed in the session's channel, not to error the command.

Yes, working

### DIFF
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -204,7 +204,7 @@ class Dev(commands.Cog):
         try:
             with redirect_stdout(stdout):
                 result = await func()
-        except:
+        except Exception:
             printed = "{}{}".format(stdout.getvalue(), traceback.format_exc())
         else:
             printed = stdout.getvalue()
@@ -293,7 +293,7 @@ class Dev(commands.Cog):
                     else:
                         result = executor(code, env)
                     result = await self.maybe_await(result)
-            except:
+            except Exception:
                 value = stdout.getvalue()
                 msg = "{}{}".format(value, traceback.format_exc())
             else:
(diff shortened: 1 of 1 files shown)


redbot/core/dev_commands.py
```diff
@@ -299,7 +299,10 @@ async def repl(self, ctx):
             else:
                 value = stdout.getvalue()
                 if result is not None:
-                    msg = "{}{}".format(value, result)
+                    try:
+                        msg = "{}{}".format(value, result)
+                    except:
```
Flame442: If this is the way this is handled (and there isn't some other part of the code that can be changed to address this issue), this should *at least* be catching the actual error, rather than a bare `except`.

Flame442 (changes requested):

redbot/core/dev_commands.py
Kreusada: I had left this except bare because the error being raised in `__repr__()` could be any exception. This is also the place where the value from `__repr__()` is formatted so that's why the issue is being addressed here.

Flame442: It should at least be `except Exception` in that case, as to not catch `KeyboardInterrupt` and other `BaseException` exceptions and prevent the bot from exiting as expected.

Flame442 (approved):
That works, thanks!

PR #5795, opened 2022-07-05
Renamed the country of Turkey to Türkiye

TheThomanski:

Renamed all strings referring to the country of Turkey to Türkiye as they changed their name in the United Nations.

Yes


Flame442 (approved):
Thanks!

PR #5797, opened 2022-07-07
Add `core/_debuginfo.py` to labeler configuration

Kreusada:

I noticed that `redbot/core/_debuginfo.py` was not present in the labeler config, and probably should be.

Yes


Kowlin (approved):
Thank you for the labeler change.

PR #5798, opened 2022-07-09
Update documentation about messages intent

RheingoldRiver:

Update to the documentation and adds a new screenshot that's valid through August 31 to better inform users about the Messages privileged intent.

Yes, but I didn't build the docs


docs/bot_application_guide.rst
```diff
@@ -67,9 +67,9 @@ Enabling Privileged Intents
     .. image:: /.resources/bot-guide/discord_bot_tab.png
         :alt: The bot tab in the application page.
 
-5. Scroll down to the "Privileged Gateway Intents" section, enable both privileged intents and save your changes.
+5. Scroll down to the "Privileged Gateway Intents" section, enable all three privileged intents and save your changes. (Note: The "message content" intent must be enabled on or before August 31, 2022, or your bot will cease functioning once Discord begins to enforce it.)
```
Jackenmen: IMO this note is unnecessary - users should just always enable all three privileged intents and having this note could potentially make someone overthink it and not enable it.
```suggestion
5. Scroll down to the "Privileged Gateway Intents" section, enable all three privileged intents and save your changes.
```

RheingoldRiver: kk, I wasn't sure if someone might get confused about the screenshot but that's a good point, I removed it

Jackenmen (changes requested):
Would be nice if the screenshot included the section title since the documentation refers to it.
Maybe also there could be a little bit of margin on that screenshot? I feel like it makes it a bit harder to read with text being so close to the end of the image.
Though it's going to be hard to read regardless since there's a lot text there. Maybe you could use the browser's zoom feature before making the screenshot? 125% or 150% might look better at the cost of the image height.

RheingoldRiver:
kk I took this at 130% zoom and added some margin, as well as the section title

Jackenmen (approved):
Looks good, thanks!

PR #5802, opened 2022-07-11
[Core] Add `--unload-cogs` cli flag.

### LINKED ISSUE #5796: Add a `--unload-cog [UNLOAD_COGS...]` cli flag
### Type of feature request

Other

Adds a way to unload a cog when starting red using the cli - `--unload-cog [UNLOAD_COGS...]`
Kinda similar to `--load-cogs [LOAD_COGS...]` but does the opposite, instead of loading the cogs, it unloads the specified cog.

### Anything else?

So I was having some issues lately with a third party cog which was not letting my bot start and was crashing it on boot, and it was impossible to unload that single cog before starting up the bot, and it's hard to remember what cogs I have loaded at times, and using the `--no-cogs` flag just starts the bot without any cog being loaded on start. It'll be really convenient to have a cli option to unload one or more cogs at boot if some cog is causing the bot to not boot up or similar issue to what I've faced in the past.

AAA3A-AAA3A:

Hello,
Following issue #5796 on this repo, this PR allows to add a cli command line argument that prevents all the cogs mentioned from being loaded at Red startup.
As explained in the related issue, users sometimes want to unload a single cog that is causing problems, without having to quote all the other cogs to be loaded.
I don't know if it's better to allow `--load-cogs` to add the packages anyway.
The code was tested under the development version of Red 3.5 and under the latest commit of dpy2 with a patch file. The agument was used or not, with a valid or invalid package name and with `--no-cogs` and `load-cogs` arguments.
I guess there is some documentation somewhere that I need to change. Where is it?
Thank you in advance,
AAA3A

Fixes #5796 

Yes

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1140,11 +1140,8 @@ class Red(
             if self._cli_flags.load_cogs:
                 packages.update(dict.fromkeys(self._cli_flags.load_cogs))
             if self._cli_flags.unload_cogs:
-                for x in self._cli_flags.unload_cogs:
-                    if x not in self._cli_flags.unload_cogs:
-                        self._cli_flags.unload_cogs.remove(x)
-                        continue
-                    del packages[x]
+                for package in self._cli_flags.unload_cogs:
+                    packages.pop(package, None)
 
         system_changed = False
         machine = platform.machine()
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1140,9 +1140,8 @@ class Red(
             if self._cli_flags.load_cogs:
                 packages.update(dict.fromkeys(self._cli_flags.load_cogs))
             if self._cli_flags.unload_cogs:
-                for x in self._cli_flags.unload_cogs:
-                    if x in self._cli_flags.unload_cogs:
-                        del packages[x]
+                for package in self._cli_flags.unload_cogs:
+                    packages.pop(package, None)
 
         system_changed = False
         machine = platform.machine()
@@ -1174,7 +1173,7 @@ class Red(
             # Load permissions first, for security reasons
             try:
                 packages.move_to_end("permissions", last=False)
-            except ValueError:
+            except KeyError:
                 pass
 
             to_remove = []
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1203,7 +1203,7 @@ class Red(
             for package in to_remove:
                 del packages[package]
         if packages:
-            log.info("Loaded packages: " + ", ".join(packages.keys()))
+            log.info("Loaded packages: " + ", ".join(packages))
         else:
             log.info("No packages were loaded.")
 
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1140,6 +1140,10 @@ class Red(
             if self._cli_flags.load_cogs:
                 packages.extend(self._cli_flags.load_cogs)
 
+            if self._cli_flags.unload_cogs:
+                self._cli_flags.unload_cogs = [x for x in self._cli_flags.unload_cogs if x in packages]
+                packages.remove(*self._cli_flags.unload_cogs)
+
         system_changed = False
         machine = platform.machine()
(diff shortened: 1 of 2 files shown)


redbot/core/bot.py
```diff
@@ -1135,10 +1135,16 @@ async def _pre_connect(self) -> None:
...
 
             if self._cli_flags.load_cogs:
-                packages.extend(self._cli_flags.load_cogs)
+                packages.update(dict.fromkeys(self._cli_flags.load_cogs))
+            if self._cli_flags.unload_cogs:
+                for x in self._cli_flags.unload_cogs:
+                    if x not in self._cli_flags.unload_cogs:
+                        self._cli_flags.unload_cogs.remove(x)
+                        continue
+                    del packages[x]
```
Jackenmen: We don't use the list in `self._cli_flags.unload_cogs` so there is no reason to remove packages from it. Additionally, this wouldn't actually work properly because you're trying to remove an item from a list while iterating over it.
```suggestion
                for package in self._cli_flags.unload_cogs:
                    packages.pop(package, None)
```

redbot/core/bot.py
```diff
@@ -1169,11 +1175,12 @@ async def _pre_connect(self) -> None:
...
             # Load permissions first, for security reasons
             try:
-                packages.remove("permissions")
+                del packages["permissions"]
             except ValueError:
                 pass
             else:
-                packages.insert(0, "permissions")
+                packages["permissions"] = None
+                packages.move_to_end("permissions", last=False)
```
Jackenmen: This can be simplified:
```py
            try:
                packages.move_to_end("permissions", last=False)
            except KeyError:
                pass

File: redbot/core/bot.py [hunk 3]
```diff
@@ -1197,9 +1204,11 @@ async def _pre_connect(self) -> None:
                     await self.remove_loaded_package(package)
                     to_remove.append(package)
             for package in to_remove:
-                packages.remove(package)
-            if packages:
-                log.info("Loaded packages: " + ", ".join(packages))
+                del packages[package]
+        if packages:
+            log.info("Loaded packages: " + ", ".join(packages.keys()))
```
```suggestion
            log.info("Loaded packages: " + ", ".join(packages))
```

Jackenmen (changes requested):

redbot/core/bot.py
```diff
@@ -1135,10 +1135,14 @@ async def _pre_connect(self) -> None:
...
-                packages.extend(await self._config.packages())
+                packages.update(dict.fromkeys(await self._config.packages()))
 
             if self._cli_flags.load_cogs:
-                packages.extend(self._cli_flags.load_cogs)
+                packages.update(dict.fromkeys(self._cli_flags.load_cogs))
+            if self._cli_flags.unload_cogs:
+                for x in self._cli_flags.unload_cogs:
+                    if x in self._cli_flags.unload_cogs:
+                        del packages[x]
```
Jackenmen: - you've done an `if x in self._cli_flags.unload_cogs` check while you probably wanted to check if a package is in `packages` instead
    - I suggest using `packages.pop(package, None)` instead of the whole if
- `x` isn't a good variable name, use `package` instead
```suggestion
                for package in self._cli_flags.unload_cogs:
                    packages.pop(package, None)
```

redbot/core/bot.py
```diff
@@ -1169,11 +1173,9 @@ async def _pre_connect(self) -> None:
         if packages:
             # Load permissions first, for security reasons
             try:
-                packages.remove("permissions")
+                packages.move_to_end("permissions", last=False)
             except ValueError:
```
```suggestion
            except KeyError:
```

Jackenmen (changes requested):

AAA3A-AAA3A:
An error from me. Sorry.
I will make these changes in 2 hours because I no longer have my computer at hand.
Thanks.

AAA3A-AAA3A:
I am back and the two problems you pointed out are fixed.

Jackenmen (approved):
Looks good to me, thank you for the PR!

PR #5806, opened 2022-07-26
feat: add a simple http server

virtual-designer:
Added an HTTP server so that it can be pinged to keep the bot alive


virtual-designer:
Sorry I've created this pull request accidentally.

PR #5821, opened 2022-08-12
[3.4] [Audio] Update application.yml for new Lavalink.jar

aikaterna:
### Description of the changes
We will be using a new Lavalink.jar on build 1350 and later that includes a Spring Boot version update from 2.1.8 to 2.6.6. In this version change, some of the logging functions have changed so they have been moved to their new designation.

### Have the changes in this PR been tested?
Yes, I installed a copy of Red with this file from a GH repo and Red started up, it replaced my old application.yml with this file and then the newer Lavalink.jar started to create log files again.

PR #5822, opened 2022-08-12
[3.4] [Audio] Update Lavalink.jar build

aikaterna:
This changes audio's manager to download and use build 1350 of our Lavalink.jars.

No, I PRed this without explicitly testing this change.


aikaterna:
Changelog: This change/update addresses a cipher change that was needed to be able to find tracks.

Flame442 (approved):
The bot properly installs jar version `1350`, and the jar works as expected (including reencoding) with the exception of the issues fixed by #5821. That PR should be merged at the same time this one is.

PR #5830, opened 2022-08-15
Get rid of localized guild feature list in serverinfo

Jackenmen:

An attempt at lowering the requirement for constant maintenance of the guild feature list. This PR makes it so that instead of maintaining a guild features list, we show the guild features as given to us by the API. Mostly. I've made 3 exceptions where the automatic casing didn't really make sense to me.

Before:
image

After:
image

As can be seen on the screenshots, this PR makes `[p]serverinfo 1` show more features than before but to me, it kind of seems like an upside... Feel free to comment if you disagree.

Yes

### DIFF
--- a/redbot/cogs/general/general.py
+++ b/redbot/cogs/general/general.py
@@ -414,14 +414,26 @@ class General(commands.Cog):
                 inline=False,
             )
 
+            excluded_features = {
+                # available to everyone since forum channels private beta
+                "THREE_DAY_THREAD_ARCHIVE",
+                "SEVEN_DAY_THREAD_ARCHIVE",
+                # rolled out to everyone already
+                "NEW_THREAD_PERMISSIONS",
+                "TEXT_IN_VOICE_ENABLED",
+                "THREADS_ENABLED",
+            }
             custom_feature_names = {
-                "TEXT_IN_VOICE_ENABLED": "Text in Voice enabled",
                 "VANITY_URL": "Vanity URL",
                 "VIP_REGIONS": "VIP regions",
             }
+            features = sorted(guild.features)
+            if "COMMUNITY" in features:
+                features.remove("NEWS")
             feature_names = [
                 custom_feature_names.get(feature, " ".join(feature.split("_")).capitalize())
-                for feature in sorted(guild.features)
+                for feature in features
+                if feature not in excluded_features
             ]
             if guild.features:
                 data.add_field(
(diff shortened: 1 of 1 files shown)


Kowlin:
Thread archive duration, Threads active, New thread permissions and VIP regions can also be removed since they no longer apply.
Also News falls under Community now as it's a mainline community feature and the same applies to Enabled Discovery before with Discovery, as enabled discovery before only handles if it should show the discovery splash page.

Jackenmen:
I'm unsure whether it's worth making the change if we're just going to switch it from include list to exclude list. I guess it at least means that new features will show immediately so there is an upside there.

Kowlin:
I think the change to an exclude list would be better due the fact that Discord *will* keep adding more feature flags over the years, without attempts at cleaning up their old flags.

``NEWS`` is given with ``COMMUNITY``,
``THREE_DAY_THREAD_ARCHIVE``, and ``SEVEN_DAY_THREAD_ARCHIVE`` have become baseline within Discord ever since Discord started rolling out Forum channels in limited private beta,
``VIP_REGIONS`` is an partner flag that Legacy Partners carried over along with a permanent ``VANITY_URL`` so is largely attached to ``PARTNERED``
``NEW_THREAD_PERMISSIONS``, ``THREADS_ENABLED``, and ``TEXT_IN_VOICE_ENABLED`` are considered not applicable because they have been rolled out entirely

The same can be seen with user flags, such as ``SPAMMER``.

Jackenmen:
I added an exclude list.
Excluded:
- THREE_DAY_THREAD_ARCHIVE (available to everyone since forum channels private beta)
- SEVEN_DAY_THREAD_ARCHIVE (available to everyone since forum channels private beta)
- NEW_THREAD_PERMISSIONS (rolled out to everyone already)
- TEXT_IN_VOICE_ENABLED (rolled out to everyone already)
- THREADS_ENABLED (rolled out to everyone already)

> `NEWS` is given with `COMMUNITY`

I made it so `COMMUNITY` flag is present, `NEWS` is removed, does that make sense? News channels were a thing before community servers so some servers may have it without that flag.

> VIP regions can also be removed since they no longer apply.

Red doesn't have `PARTNERED` flag but it does have `VIP_REGIONS`. Legacy partners can have a higher bitrate in voice channels *without* the need for a boost level so this flag still has a meaning.

I think `ENABLED_DISCOVERABLE_BEFORE` actually means that a guild has been in Discovery at least once so it's not really a legacy flag. We could maybe consider showing it only if the guild isn't *currently* in Discovery (so when it doesn't have `DISCOVERABLE` flag)

Kowlin:
I honestly love those options and they all make sense to me.

Kowlin (approved):

PR #5843, opened 2022-09-04
Show source code in tracebacks sent by Dev cog commands

Jackenmen:

Fixes an incredibly annoying thing with Dev cog's commands - the tracebacks returned by it do not include the source lines!
image

With this PR, you'll finally be able to tell where in your (sometimes maybe a little too long) eval lies the problem:
image

This PR features:
- source code in the traceback
- fixed line numbers (no more trying to figure out what that line number is trying to refer to exactly...)
- stripping of frames that are actually part of Red's code, not the code passed to the command

Yes


Jackenmen:
I reorganized the dev cog to improve testability, added a bunch of tests, and fixed a couple of issues based on them:
- source code of syntax errors sometimes didn't show up (may have been dependent on the complexity of the parsed expression)
- fine-grained error locations in tracebacks were off-by-one if the error was in the last line and that line didn't end with `\n`
- the result of `[p]debug` command was not shown if it was `None` (regression in comparison to current behavior)
- start code block regex didn't account for how Discord renders a code block with leading empty lines (it doesn't render them so line numbers need to account for that)

redbot/core/dev_commands.py
```diff
@@ -33,7 +37,272 @@
...
+        exc_type = type(exc)
+        tb = exc.__traceback__
+        for x in range(skip_frames):
+            if tb is None:
+                break
+            tb = tb.tb_next
+
+        # To mimic linecache module's behavior,
+        # all lines (including the last one) should end with \n.
+        source_lines = [f"{line}\n" for line in self._source.splitlines()]
```
Zephyrkul, line 265:
```suggestion
        source_lines = self._source.splitlines(keepends=True)
```

Jackenmen: Not equivalent, see this commit and the comment above this line:
https://github.com/Cog-Creators/Red-DiscordBot/pull/5843/commits/3877520958c44963e4ed055125d5980fa4907c46

Jackenmen: The difference can most easily be seen by checking against this test case:
https://github.com/Jackenmen/Red-DiscordBot/blob/f6f7bd5e9e4032aec2c8449b0cebd08f1712bc51/tests/core/test_dev_commands.py#L234-L243
```
    async def _run_dev_output(
        monkeypatch: pytest.MonkeyPatch,
        source: str,
        result: str,
        *,
        debug: bool = False,
        eval: bool = False,
        repl: bool = False,
    ) -> None:
        source = textwrap.dedent(source)
        result = textwrap.dedent(result)
        monkeypatch.setattr("redbot.core.dev_commands.sanitize_output", lambda ctx, s: s)
    
        if debug:
            output = _get_dev_output(source)
            await output.run_debug()
>           assert str(output) == result
E           AssertionError: assert 'Traceback (m...ion by zero\n' == 'Traceback (m...ion by zero\n'
E             Skipping 89 identical leading characters in diff, use -v to show
E               )
E             -         ~~^~~
E             +          ~~^~~
E             ? +
E               ZeroDivisionError: division by zero

tests/core/test_dev_commands.py:162: AssertionError
```

Kreusada (commented):
I've tested these changes and they seem good to me, I don't have any additional comments (although maybe to be picky, `ctx` parameters could be type hinted).

On a separate note, do you think that `cleanup_code` should be added as a chat_formatting util? I've used this a lot by simply pasting this function in my code, maybe it would be helpful as a util.

image

japandotorg:
> On a separate note, do you think that cleanup_code should be added as a chat_formatting util? I've used this a lot by simply pasting this function in my code, maybe it would be helpful as a util.

+1 from me on this

AAA3A-AAA3A:
I have tested this change a bit and it works.

> On a separate note, do you think that cleanup_code should be added as a chat_formatting util? I've used this a lot by simply pasting this function in my code, maybe it would be helpful as a util.

I also think that `cleanup_code` should become a utility as it is used a lot.

Jackenmen:
> On a separate note, do you think that `cleanup_code` should be added as a chat_formatting util? I've used this a lot by simply pasting this function in my code, maybe it would be helpful as a util.

I'm not sure, I've seen it used in a few different scenarios in 3rd-party cogs so I see some merit to this request. If this were to be pursued, I think it should be a part of a separate PR.

Kreusada:
> I'm not sure, I've seen it used in a few different scenarios in 3rd-party cogs so I see some merit to this request. If this were to be pursued, I think it should be a part of a separate PR.

Agreed, I'm happy to make a separate PR implementing this if/when it is wanted.

Kreusada (approved):
LGTM 👍

PR #5848, opened 2022-09-15
Correction of annotations in Config related methods

madebylydia:

Improve annotations in Config.

- When awaiting a `Value`/`Group`, this will no longer raises `_ValueCtxManager is not awaitable` (Due to invalid annotation of `__await__`)
  Error in capspam.py, compaining that ValueCtxManager is not awaitable
- Correction of some invalid annotations
- Use method overload for `Config.get_conf` to either include `cog_instance` or `cog_name`, or raise an error if both are `None` (See comment below)

Yes

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -5,16 +5,13 @@ import logging
 import pickle
 import weakref
 
-from collections import (
-    Awaitable,
-    MutableMapping
-)
-from contextlib import AbstractAsyncContextManager
-
 from typing import (
     Any,
+    Awaitable,
+    AsyncContextManager,
     Dict,
     Generator,
+    MutableMapping,
     Type,
     Optional,
     Tuple,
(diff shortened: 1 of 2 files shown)


redbot/core/config.py
```diff
@@ -657,15 +661,39 @@ def __init__(
...
+        ...
+
+    @overload
+    @classmethod
+    def get_conf(
+        cls,
+        cog_instance: Optional[Cog],
+        identifier: int,
+        force_registration: bool = False,
+        cog_name: str = "",  # The bool of an empty string is False
```
madebylydia, line 683: This is sadly the only way I've found to not introduce any breaking changes, since those are args and not kwargs, else `Non-default argument follows default argument` is raised.
I guess it is a fine for solution. For now.
Probably it can be considered to make them positional for Red 3.5? As this imply a breaking change along with discord.py 2.0.

PR #5850, opened 2022-09-17
Add missing period which caused ignore list to fail

alexratman:

When testing 3.5 I tried to list all ignored channels and got an error which referenced `AttributeError: 'Core' object has no attribute 'bot_ignored_cache'` bot_ignored_cache is only mentioned here and appears to be a typo. I've tested this locally and it appears to have fixed the error as the ignore list now appears correctly.

Yes


Jackenmen (approved):
Thanks!

PR #5851, opened 2022-09-18
Add Traceback prompt

### LINKED ISSUE #5621: Make `traceback` prompt if you want it to print all pages, if there's more than say 5

^traceback

Occasionally an error will have *egregiously* long text, and using `^traceback` currently is not gated in the number of messages it sends. Adding a confirmation if you want it to print, in the case that an error message is *truly* gigantic would be nice. (Recently I restarted my bot because it was printing the entire source code of a wiki page that had failed to save; I don't even know how many pages it would have been, but it just would not stop.)

### Anything else?

As a bonus, the page count could be configurable, but really I'm just looking for something covering truly outlandish edge cases here, so a quick fix hard-coding even like 15 or 20 would be fine by me.

yuansheng1549:
Allowing `traceback` command to have prompts if there's a lot of error messages 

Yes

Closes #5621 
Blocked by #5942


redbot/core/core_commands.py
```diff
@@ -1435,21 +1436,27 @@ async def traceback(self, ctx: commands.Context, public: bool = False):
...
-            destination = ctx.channel
+
+        copied_ctx = copy(ctx)
+        copied_ctx.channel = (
+            ctx.channel
+            if public
+            else (
+                ctx.author.dm_channel if ctx.author.dm_channel else await ctx.author._get_channel()
+            )
+        )
```
Jackenmen: I get the purpose of this but this is extremely hacky for what it's used for... If we *have to* override a context's channel I think we could do away without a copy but I feel like we shouldn't have to. Maybe it's time to have a `send_interactive` utility that isn't tied to a Context object :thinking:

Jackenmen (changes requested):
A couple of things:
- I'm currently unsure if `send_interactive()` should be part of `redbot.core.utils` (and whether `prompt` is the right name for the submodule it should sit in) or if perhaps it should be a part of the bot object (`Red` class) as are some other send methods such as `send_to_owners()` and `send_filtered()`
    - Since I'm unsure, I suggest not making any changes related to that yet. However, if you do have any opinion on this, I'd be happy to hear it.
- It seems that you've copied an older version of `send_interactive()`, e.g. it doesn't have a `join_character` parameter
- You've changed the behavior of the utility to accept a message with *any* author - you should instead have an argument for specifying the user that the message needs to be authored by
- Ideally, we should support passing pretty much any messageable, that is: `discord.TextChannel`, `discord.VoiceChannel`, `discord.DMChannel`, `discord.User`, `discord.Member`, `commands.Context`, and `discord.Thread`
    - I realize that this is currently impossible due to what `MessagePredicate` supports. I've made a PR (#5942) to rectify this issue, you can assume that it will be merged before your change and change the PR accordingly (which should mean modifying type hints and simplifying the change to `[p]traceback` command).

yuansheng1549:
I think it should be part of the bot object

Jackenmen (dismissed):
When used in combination with #5942, this works as expected. I'll merge this once #5942 gets approved and merged.

Jackenmen (approved):

PR #5854, opened 2022-09-30
Fix unintended moving *between* VCs when channel muting

### LINKED ISSUE #4993: [Mute] Mute/unmute moves users through all channels

3.4.9

### Cog name

Mutes

### Command name

mute

Mute moves user from one channel to another, and then back.

*Or through all channels, and then back.*

### What actually happened?

Mute moved user from one channel to another, moving through all channels in progress.

Final channel was not same as initial channel.

1. Create a server with 3 or more voice channels
2. Make user join one of them
3. Mute that user

### Anything else?

Before mute:
before mute - voice channels
After mute:
after mute - voice channels
Audit log:
audit log - 5 moves for 5 channels

Jackenmen:

Fixes #4993

Yes

### DIFF
--- a/redbot/cogs/mutes/mutes.py
+++ b/redbot/cogs/mutes/mutes.py
@@ -1704,7 +1704,7 @@ class Mutes(VoiceMutes, commands.Cog, metaclass=CompositeMetaClass):
             speak=False,
         )
         send_reason = None
-        if user.voice and user.voice.channel:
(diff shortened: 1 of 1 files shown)


Jackenmen:
Now tested, seems to work fine.

TrustyJAID (approved):
Looks good to me!

PR #5856, opened 2022-10-01
[Streams] Add toggleable button to stream alerts

### LINKED ISSUE #5646: [Streams] Togglable button for streamalerts
### Type of feature request

Cog, Command

With dpy2 coming with tons of nice features, i think a button for `open stream` could be nice to have a toggle under `[p]streamset` for those who wants a button to show on streamalerts.

Example:
s2Mo8Kp

### Anything else?

_No response_

karlsbjorn:

This PR adds a toggleable `[p]streamset livebutton` setting which creates a button under each stream alert (created either through `[p]streamalert` or `[p]twitchstream`/`[p]youtubestream` and the like) that will link to the relevant stream.
It is off by default.

image

Closes #5646 

I'm also not sure if `livebutton` is the best name for this setting, I couldn't think of anything better. Maybe something like `buttons` or `usebuttons` looks better? Would love to hear feedback.

Yes


karlsbjorn:
~~Since #5683 has now been merged, should the toggle for the button be taken from `[p]set usebuttons` instead of having it be a separate setting?~~

~~In case not,~~ I've renamed the setting to `[p]streamset usebuttons` for consistency.

EDIT: After a short discussion on Discord, the setting should stay like it currently is since `[p]set usebuttons` is global, while `[p]streamset usebuttons` should be per-guild.

redbot/cogs/streams/streams.py
```diff
@@ -281,15 +282,32 @@ async def check_online(
...
-            await ctx.send(embed=embed)
+                if use_buttons:
+                    stream_url = embed.url
+                    view = discord.ui.View()
+                    view.add_item(
+                        discord.ui.Button(
+                            label=_("Open Stream"), style=discord.ButtonStyle.link, url=stream_url
+                        )
+                    )
+            await ctx.send(embed=embed, view=view if use_buttons else None)
```
Jackenmen: Code above is identical, this can be deduplicated by making this fragment not part of the `if`. To simplify this further, `view` can be set to `None` by default as then we don't have to do a ternary if inside `ctx.send()` call.

Also, the view needs to have a timeout set to `None` as the default is 3 minutes.
```suggestion
            view = None
            if use_buttons:
                stream_url = embed.url
                view = discord.ui.View(timeout=None)
                view.add_item(
                    discord.ui.Button(
                        label=_("Open Stream"), style=discord.ButtonStyle.link, url=stream_url
                    )
                )
            await ctx.send(embed=embed, view=view)
```

redbot/cogs/streams/streams.py
```diff
@@ -281,15 +282,32 @@ async def check_online(
...
             if isinstance(info, tuple):
                 embed, is_rerun = info
+                if use_buttons:
+                    stream_url = embed.url
+                    view = discord.ui.View()
+                    view.add_item(
+                        discord.ui.Button(
+                            label=_("Open Stream"), style=discord.ButtonStyle.link, url=stream_url
+                        )
+                    )
```
Jackenmen: Code below is identical, this can be deduplicated.
```suggestion
```

redbot/cogs/streams/streams.py
```diff
@@ -705,6 +723,19 @@ async def ignore_schedule(self, ctx: commands.Context):
...
 
+    @streamset.command(name="usebuttons")
+    @commands.guild_only()
+    async def use_buttons(self, ctx: commands.Context):
+        """Toggle whether to use buttons for stream alerts."""
+        guild = ctx.guild
+        current_setting: bool = await self.config.guild(guild).use_buttons()
+        if current_setting:
+            await self.config.guild(guild).use_buttons.set(False)
+            await ctx.send(_("I will not use buttons in stream alerts."))
```
Jackenmen:
```suggestion
            await ctx.send(_("I will no longer use buttons in stream alerts."))
```

redbot/cogs/streams/streams.py
```diff
@@ -773,10 +804,20 @@ async def _send_stream_alert(
...
+            view.add_item(
+                discord.ui.Button(
+                    label=_("Open Stream"), style=discord.ButtonStyle.link, url=stream_url
+                )
+            )
         m = await channel.send(
             content,
             embed=embed,
             allowed_mentions=discord.AllowedMentions(roles=True, everyone=True),
+            view=view if use_buttons else None,
```
Jackenmen: Simplification by removal of the ternary if operator and replacing it with always setting initial value of `view` to `None`.

Also, the view needs to have a timeout set to `None` as the default is 3 minutes.
```suggestion
        use_buttons: bool = await self.config.guild(channel.guild).use_buttons()
        view = None
        if use_buttons:
            stream_url = embed.url
            view = discord.ui.View(timeout=None)
            view.add_item(
                discord.ui.Button(
                    label=_("Open Stream"), style=discord.ButtonStyle.link, url=stream_url
                )
            )
        m = await channel.send(
            content,
            embed=embed,
            allowed_mentions=discord.AllowedMentions(roles=True, everyone=True),
            view=view,
```

Jackenmen (changes requested):
Functionally works fine (aside from a small timeout issue which is mentioned in the review comment), I just have a couple of stylistic suggestions.

redbot/cogs/streams/streams.py
karlsbjorn: From my testing I actually found that timeouts don't matter for `discord.ButtonStyle.link` style buttons, they're usable past the timeout which is why I didn't set one. You're right though that it's still better to set it explicitly.

Jackenmen: Hmm, you're right, we actually shouldn't set a timeout to `None` because then it forever stays in the memory. We should just keep the default so that it's at least removed from the memory after 3 minutes.

The reason why I suggested this change was that the buttons disappeared from the message after a bit which I thought was related to the timeout (but it isn't because that would require you to explicitly edit the message in `on_timeout()`)... I'm actually not sure why it happened *and if* it actually happened (maybe I confused messages?).

Would you mind removing the `timeout=` parameter?

karlsbjorn: Done! You might've had `[p]streamset autodelete` toggled on and didn't notice the message getting deleted? Not sure

Jackenmen:
I think that "Watch the stream" might be a better button caption, what do you think?

image
image

karlsbjorn:
Sure, sounds more natural to me

Jackenmen (approved):
Works as expected.

PR #5862, opened 2022-10-03
Add `[p]warnings server` command to list all warnings in a guild

### LINKED ISSUE #5164: [Warnings] Add a [p]warnsguild command

Warnings

A new command to warns, ie **[p]warnsguild** this would show the list of users who have been warned in a guild, this could be useful if u forgot which all user's currently have a warn and how many so we could take action as if a server is big u may forget which all user's u warned , what this would do is return a paginated embed with a list of users (only who have been warned) and with the no of warns they have and maybe also the total no points , this way we could run this cmd in a while and see the list ie members with more nos of warns will be placed higher on the list , this way if we have a rule ie 3 warns = temp mute or 5 warns = kick we could check the user's warns and initiate the action ourselves , this way we could easily even remove their history with casesfor command because we already know some mod action had taken place.

### Anything else?

That's all

Flame442:
Fixes #5164

I vaguely based the formatting of the output on the output of `[p]leaderboard`. The specific formatting and command name can be changed. I don't know if the command should be sorted by count, points, or if it shouldn't sort at all, right now it just doesn't sort.

Yes


redbot/cogs/warnings/warnings.py
```diff
@@ -561,6 +561,36 @@ async def warnings(self, ctx: commands.Context, member: Union[discord.Member, in
                     ),
                 )
 
+    @warnings.command(name="guild")
+    @commands.guild_only()
+    @checks.admin()
+    async def warnings_guild(self, ctx: commands.Context):
+        """List all members with warnings in this guild."""
```
Kowlin: For end user UI experience this should probably be renamed to "server", and aliased to "guild" as a command name.

Jackenmen (approved):
LGTM

PR #5863, opened 2022-10-03
[Cleanup] Pass reason for bulk message deletion to audit log

### LINKED ISSUE #5861: Specify audit log reason in cleanup cog

Cleanup

Delete messages endpoint supports specifying audit log reason and discord.py 2.0 added a `reason` kwarg to `Messageable.delete_messages(...)` in order to support it. I think that the cleanup cog should be specifying the audit log reason in all places where it deletes messages.

This change will require an addition of a `reason` kwarg to `redbot.core.utils.mod.mass_purge(...)` and usage of that new kwarg in all calls to this function in the Cleanup cog.

### Anything else?

What should the contents of the audit log reason be? Just whatever `get_audit_reason(...)` utility function returns?

leetfin:

Fixes #5861 

No


redbot/core/utils/mod.py
```diff
@@ -12,6 +12,7 @@
 async def mass_purge(
     messages: List[discord.Message],
     channel: Union[discord.TextChannel, discord.VoiceChannel, discord.Thread],
+    reason: str = None,
```
Flame442: The linked issue specified this should be a kwarg, and I do think it would be better as one.

leetfin: Would `**reason: str,` be the correct way to specify this as a kwarg here?

Jackenmen: To make something a keyword-only argument, it needs to be specified after `*` symbol in the argument list, e.g:
```suggestion
    *,
    reason: Optional[str] = None,
```

redbot/core/utils/mod.py
```diff
@@ -29,6 +31,8 @@ async def mass_purge(
         The messages to bulk delete.
     channel : `discord.TextChannel`, `discord.VoiceChannel`, or `discord.Thread`
         The channel to delete messages from.
+    reason : `str` - Optional
```
Jackenmen:
```suggestion
    reason : `str`, optional
```

Kowlin (approved):
Despite Discord not accepting reasons for mass deleted messages (while saying its supported by [their docs](https://discord.com/developers/docs/resources/channel#bulk-delete-messages)) this still gets send through properly. And is handled fine.

PR #5865, opened 2022-10-05
Always allow licenseinfo to be run with a mention prefix

### LINKED ISSUE #5460: [p]licenseinfo and slash commands
The deadline is coming, Discord will soon remove message content for most bots, forcing us to migrate to slash commands.

It has been told countless times that Red does not support public bots and it is fully up to us if we want to host a large public bot, which most of us already understood by using a custom fork.

However, there is a problem with that: the `[p]licenseinfo` command. It has been made for traditional commands and cannot be modified. So, when slash commands kick in, is Discord responsible for blocking access to this command?

Not supporting slash commands is perfectly fine, but in the current state of the bot, it is outright illegal to have a slash-only bot. The command is inaccessible and requires editing to work.

----

This is not a specific feature request, I'm asking for clarifications as to how we should implement this command on slash-commands only bots. I see multiple ways of doing this:

- Allow and require the bot/fork owner to implement his slash version of licenseinfo, not touching the original text
- Provide a slash-command compatible version of licenseinfo. Red doesn't manage interactions by default and people use plenty of different libs, so that would be kinda difficult to implement
- Respond to the bot mention, aka `@Redbot licenseinfo` will always respond, even if the bot mention isn't a prefix

Flame442:
Fixes #5460

Yes


Flame442:
This PR has been modified to override the prefix in `process_commands`, rather than in the prefix function ([ref](https://discord.com/channels/133049272517001216/718148684629540905/1029415548108427394)).

redbot/core/bot.py
```diff
@@ -1589,6 +1589,20 @@ async def process_commands(self, message: discord.Message, /):
...
+            # will always work for it.
+            if not ctx.valid:
+                for m in (f"<@{self.user.id}> ", f"<@!{self.user.id}> "):
+                    if message.content.startswith(m):
+                        ctx.view.undo()
+                        ctx.view.skip_string(m)
+                        invoker = ctx.view.get_word()
+                        if invoker in ("licenseinfo", "licenceinfo"):
+                            ctx.command = self.all_commands.get(invoker)
+                            ctx.prefix = m
```
Zephyrkul:
```suggestion
                        if invoker in ("licenseinfo", "licenceinfo"):
                            ctx.command = self.all_commands.get(invoker)
                            ctx.prefix = m
                        break
```

Kowlin (approved):
Works perfectly and don't have any complaints 👍

PR #5866, opened 2022-10-06
[Modlog API] Stop modlog.get_case() from erroring if no modlog channel is set up

### LINKED ISSUE #5644: [Modlog API] modlog.get_case() fails if no modlog channel is set up.

3.5.0dev1

Get a specific modlog case from the modlog API using ``await modlog.get_case(case_no, guild, bot)``.

The method to return a case.

### What actually happened?

A RuntimeError got raised because no modlog channel was set up.
img

1. Do NOT set up a modlog channel
2. Create a modlog case (example, mute a user)
3. Try getting the case you just created. (e.g. with `[p]case <case_no>`)
4. Realize that something is broken

### Anything else?

image
I tried getting the case using the built-in `[p]case` command as well as through my own cog with the above result.

Dav-Git:
Finally found some time to address this annoying lil bug.

Decided to have the modlog API methods try to get a modlog channel and explicitly pass a guild to from_json instead of relying on the existence of a modlog channel.

Yes

Obviously this closes #5644


Flame442 (changes requested):
> For the 2nd usecase, the modlog channel is not actually required. It is already possible for the message attribute to be None if
> 1. no message has been provided to Case.from_json()
> 2. no message ID has been saved with the case (for example when no modlog channel exists)
> 
> Therefore we could easily just make the mod_channel argument optional and check before trying to get a message.

Quote from the linked issue. Looks like you forgot to add this check. This is currently erroring when trying to get cases with a message (for all cases made when a channel existed afaik).
```py
File "\redbot\core\modlog.py", line 865, in get_case
    return await Case.from_json(mod_channel, bot, case_number, case, guild=guild)
File "\redbot\core\modlog.py", line 692, in from_json
    message = mod_channel.get_partial_message(message_id)
AttributeError: 'NoneType' object has no attribute 'get_partial_message'
```

However, please hold off until I hear back from @jack1142 on how to handle this. I think #4977 is fragile when the modlog channel is changed, and I want to make sure the implementation of `Case.from_json` is set before you make any changes to this PR.

Flame442:
You can add the `if mod_channel is not None` check to `Case.from_json` as planned.

Flame442 (approved):
Thanks for handling this!

PR #5870, opened 2022-10-07
[Mod] Hide config migration commands

Flame442:
`[p]movedeletedelay` and `[p]moveignoredchannels` are owner only commands which migrate Mod's config to a newer schema that should only be run when the owner is told to do so. Because other similar commands (ie `[p]cog reinstallreqs`, `[p]modlogset fixcasetypes`) are hidden, and the fact that the Red versions which would use the old schema are likely no longer usable, this PR hides these two commands.

Yes


Kowlin (approved):
Nothing will break here and moves the commands to be in line with the other ones. Thanks.

PR #5874, opened 2022-10-11
Test implicit subclass type conversion in config defaults/sets

### LINKED ISSUE #5585: Config no longer converts non standard objects

3.5.0.dev1

Pull a registered default (Counter) from config and add it to a current counter, then set it back.

Correctly cast the the object back to Counter, it only cast it back to dict.

### What actually happened?

The object was cast by a dict, and thus errored when trying to add to a counter.

image

1. Load commandstats on current dev version.
2. Run any commandstats command or just watch the logs as it updates on an interval.

### Anything else?

@jack1142 

https://canary.discord.com/channels/133049272517001216/718148684629540905/945755778353102929 (discussion in #AC)

Flame442:
#5557 modified config to enforce JSON serializability in defaults, as before this change non-serializable objects could be used as defaults but could not be passed to config set methods ([ref](https://discord.com/channels/133049272517001216/160386989819035648/937360807434518569)). As #5585 points out, this change caused a previous "feature" of config defaults passing serializable data to subclasses of equivalent python builtins to stop working. Due to numerous gotchas associated with that "feature" ([ref](https://discord.com/channels/133049272517001216/718148684629540905/945755778353102929)), restoring it is not something we are looking to support.

Our current method of serializing data accepts subclasses as *inputs* and implicitly converts them to their respective JSON serializable type, but always returns a basic builtin as its *output*. This PR adds a test to explicitly test this functionality, so similar unintended consequences of a simple change do not break expected behavior without it being acknowledged, as #5557 did.

Supersedes / Closes #5585

Yes, confirmed that reverting https://github.com/Cog-Creators/Red-DiscordBot/commit/b0ab6186ef20ef16d97d9a2f8771417b08ea27f3 causes this test to fail:
```
>       assert type(stored_value) is dict
E       AssertionError: assert <class 'collections.Counter'> is dict
E        +  where <class 'collections.Counter'> = type(Counter())

tests\core\test_config.py:601: AssertionError
```
The test passes on the latest commit.


tests/core/test_config.py
```diff
@@ -593,6 +593,20 @@ async def test_raw_with_partial_primary_keys(config):
     assert await config.custom("CUSTOM", "primary_key").identifier() is False
 
 
+@pytest.mark.asyncio
+async def test_cast_subclass_default(config):
+    from collections import Counter
```
Jackenmen: Is there a reason for this import not to be at the top?

tests/core/test_config.py
```diff
@@ -593,6 +593,20 @@ async def test_raw_with_partial_primary_keys(config):
...
+
+    config.register_global(foo=Counter({}))
```
Jackenmen, line 600: I think we should assert that the value stored in `defaults` is not a Counter instance as well.
```suggestion
    config.register_global(foo=Counter({}))
    assert type(config.defaults["GLOBAL"]["foo"]) is dict
    assert config.defaults["GLOBAL"]["foo"] == {}
```

tests/core/test_config.py
```diff
@@ -593,6 +593,20 @@ async def test_raw_with_partial_primary_keys(config):
     assert await config.custom("CUSTOM", "primary_key").identifier() is False
 
 
+@pytest.mark.asyncio
+async def test_cast_subclass_default(config):
```
Jackenmen, line 598: I think an issue reference for more context would make sense here:
```suggestion
async def test_cast_subclass_default(config):
    # regression test for GH-5557/GH-5585
```

Jackenmen (changes requested):

tests/core/test_config.py
Flame442: https://github.com/Cog-Creators/Red-DiscordBot/blob/233278053b906bcbf3fd748b0df89c85fa4ed867/tests/cogs/test_mod.py#L14-L15
I saw imports that are used for a test imported in that test in other places. I assumed since it was used like that elsewhere, that it might need to be like that, but if it should be at the top of the file I can put it there instead.

Jackenmen: It looks like a code-style thing but we're not consistent about it at all (for comparison, see Downloader, Trivia, or any core tests which put imports at the top). Let's put the import at the top!

Jackenmen (approved):
Looks good.

PR #5876, opened 2022-10-11
Stop using deprecated set-output GH Actions command

Jackenmen:

https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/
https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

No


Kowlin (approved):
Checks are passing and saw the same PR over at dpy. Merge when ready

PR #5880, opened 2022-10-13
Use custom groups in config instead of a big list

### LINKED ISSUE #5242: Migrate Alias's config from a list of entries to a `name->alias_data` mapping

Alias

The current data structure seems inefficient. This isn't much of an issue when calling an alias since we maintain a cache in the cog but it is not ideal for edits or deletions which require finding the alias in the entries list. Using a custom group instead would be preferred.

Low-priority issue but would IMO be nice to change this at some point.

### Anything else?

_No response_

Dav-Git:

Move aliases config from a list per guild to custom groups.
Additionally potentially change the cache structure to better fit the new config model.
Add a migration to migrate existing Alias Data to the new config forma.

Yes

This closes #5242

### DIFF
--- a/redbot/cogs/alias/alias.py
+++ b/redbot/cogs/alias/alias.py
@@ -133,14 +133,13 @@ class Alias(commands.Cog):
             return
 
         if await self.config.entries():
-            async with self.config.entries() as global_alias_list:
-                for a in global_alias_list:
-                    await self.config.custom("Alias", None, a["name"]).command.set(a["command"])
-                    await self.config.custom("Alias", None, a["name"]).creator.set(a["creator"])
-                global_alias_list.clear()
+            async for a in self.config.entries():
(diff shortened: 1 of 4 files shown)


Drapersniper:
You should group Config calls to avoid repeated calls to the same scope - alias is one of the few areas of red that these optimisations have a real world benefit

Dav-Git:
Agreed. Optimizations like this will follow when I have it generally working.
But thanks for the heads up :)

Dav-Git:
No idea what broke the docs test... Works on my Machine(TM) and the error in the actions output isn't really telling me much.

Flame442:
I noticed something about a timeout trying to reach a URL in the docs test output, so I just reran the test and it came back positive. Seems like a temporary outage caused it to fail. All set now.

PR #5889, opened 2022-10-22
[Trivia] Add non-punctuated alias

japandotorg:

Trivia is one of the most used cog in my bot, and I've noticed that people don't really use punctuations when playing the game, because of which they loose points, I'm not against the concept of having punctuations, but having a non-punctuated alias feels like should be there.

Yes


Kreusada (approved):
Thanks for making these changes 😁

PR #5892, opened 2022-10-25
Add guild specific command for helpset maxpages

### LINKED ISSUE #3147: Allow Setting [p]helpset maxpages On a Per-Server Basis
# Feature request

- [ ] Cog
- [X] Command
- [ ] API functionality

The `[p]helpset maxpages` command is a command in the Core cog. Currently, you can only set this globally. I would instead prefer to be able to set this on a per-server basis, as multiple servers use the same bot that we host on their behalf and each of them have different needs for this setting.

Dav-Git:

See above
Closes #3147

Yes


Jackenmen (commented):
I briefly looked through this PR, and it appears to be fine as a standalone feature. However, if we were to have other server-specific settings, we might need to instead consider refactoring `HelpSettings` a bit. Marking for discussion for 3.6.1.

PR #5893, opened 2022-10-28
Use Discord's relative timestamps as command cooldown countdown

karlsbjorn:
Changes the message sent when invoking a command that's on cooldown. Instead of a static "in x seconds" to next available invocation, it uses a relative Discord timestamp.

DiscordCanary_aWZsJX7nuF

EDIT: I've also added relative timestamps to `[p]payday`'s cooldown message in this PR since making a separate one just for that while this one is still open seems a bit pointless.

Yes


redbot/core/events.py
```diff
@@ -295,10 +295,10 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
                 await bot.invoke(new_ctx)
                 return
-            if delay := humanize_timedelta(seconds=error.retry_after):
-                msg = _("This command is on cooldown. Try again in {delay}.").format(delay=delay)
-            else:
-                msg = _("This command is on cooldown. Try again in 1 second.")
+            delay = discord.utils.format_dt(
+                datetime.now(timezone.utc) + timedelta(seconds=error.retry_after), "R"
+            )
+            msg = _("This command is on cooldown. Try again {delay}.").format(delay=delay)
```
Jackenmen: I like `relative_time` more as a name for the replacement field, it's probably going to be clearer to translators about what it will contain, especially if used consistently.
```suggestion
            relative_time = discord.utils.format_dt(
                datetime.now(timezone.utc) + timedelta(seconds=error.retry_after), "R"
            )
            msg = _("This command is on cooldown. Try again {relative_time}.").format(
                relative_time=relative_time
            )
```

Jackenmen (changes requested):
I like this change! It does take away some of the precision as it only shows the highest unit and as such, there's no difference between 3m30s and 2m31s as both will show as 3 minutes *but* I guess we'll see if people complain about it.

I made one small review comment that needs to be addressed before this can be merged.

Jackenmen (approved):
LGTM, thanks for the timely follow-up.

PR #5894, opened 2022-10-29
Core - Add invoke error message customisation

### LINKED ISSUE #5622: Make CommandInvokeError error message customisable

Core

Currently when a command errors out, it sends the following message to all users:
image

However "Check your console or logs for details." for end users doesn't make much sense, just for the bot owner(s) though it makes sense.
My suggestion would be to make this message customisable, like we do on disabled commands. And by default let that message.
Maybe some templating to send the command name, and "Check your console or logs for details." if the command invoker was the owner too.

### Anything else?

_No response_

PredaaA:

Closes #5622.

Yes


PredaaA:
Trivia cog also had a "Check your console or logs for details." message, it's only displaying if the context author was the bot owner now, to follow what's being done on Audio cog.

redbot/core/events.py
```diff
@@ -244,15 +244,14 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
-            ).format(command=ctx.command.qualified_name)
             exception_log = "Exception in command '{}'\n" "".format(ctx.command.qualified_name)
             exception_log += "".join(
                 traceback.format_exception(type(error), error, error.__traceback__)
             )
             bot._last_exception = exception_log
+
+            message = await bot._config.invoke_error_msg()
+            message.replace("{command}", ctx.command.qualified_name)
             await ctx.send(inline(message))
```
Kreusada:
```suggestion
            message = await bot._config.invoke_error_msg()
            message = message.replace("{command}", ctx.command.qualified_name)
            await ctx.send(inline(message))
```

`str.replace` returns a copy of the string like many other `str` methods, it doesn't update the actual string. You'd need to reassign this copy to the same variable or pass the copy straight to the function below.

redbot/core/bot.py
```diff
@@ -134,6 +134,7 @@ def __init__(self, *args, cli_flags=None, bot_dir: Path = Path.cwd(), **kwargs):
             invite_commands_scope=False,
             disabled_commands=[],
             disabled_command_msg="That command is disabled.",
+            invoke_error_msg="`Error in command {command}. Check your console or logs for details.`",
```
Jackenmen: I think we should have `None` as the default and use the default message if config value is `None`. This way, we can continue to support translations in the default message *and* we can even only show the "Check your console or logs for details" if the command invoker was a bot owner, with the same check as in trivia cog.
```suggestion
            invoke_error_msg=None,
```

redbot/core/core_commands.py
```diff
@@ -3670,6 +3670,26 @@ async def use_buttons(self, ctx: commands.Context, use_buttons: bool = None):
...
+        """
+        if msg is not None and len(msg) >= 1000:
+            return await ctx.send(_("The message must be less than 1000 characters."))
+
+        if msg is not None:
+            await self.bot._config.invoke_error_msg.set(msg)
+            content = _("Succesfully updated the error message.")
+        else:
+            await self.bot._config.invoke_error_msg.clear()
+            content = _("Successfully updated the error message back to the default one.")
```
```suggestion
            content = _("Successfully reset the error message back to the default one.")
```

Jackenmen (changes requested):

PredaaA:
Done both requests.

redbot/core/events.py
```diff
@@ -266,16 +266,17 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
             exception_log += "".join(
                 traceback.format_exception(type(error), error, error.__traceback__)
             )
             bot._last_exception = exception_log
-            await ctx.send(inline(message))
+
+            message = (
+                await bot._config.invoke_error_msg()
+                or "`Error in command {command}. Check your console or logs for details.`"
+            )
```
Jackenmen: You didn't actually mark the string as translatable. I would also suggest to make the default message consistent with the one in Trivia cog:
```suggestion
            message = await bot._config.invoke_error_msg()
            if not message:
                if ctx.author.id in bot.owner_ids:
                    message = inline(_("Error in command {command}. Check your console or logs for details."))
                else:
                    message = inline(_("Error in command {command}."))
```

Jackenmen (changes requested):

redbot/core/core_commands.py
```diff
@@ -3670,6 +3670,26 @@ async def use_buttons(self, ctx: commands.Context, use_buttons: bool = None):
         else:
             await ctx.send(_("I will not use buttons on basic menus."))
 
+    @_set.command(name="errormsg")
+    @commands.is_owner()
+    async def _set_errormsg(self, ctx: commands.Context, msg: Optional[str] = None):
```
```suggestion
    async def _set_errormsg(self, ctx: commands.Context, *, msg: str = None):
```

redbot/core/core_commands.py
```diff
@@ -3670,6 +3670,26 @@ async def use_buttons(self, ctx: commands.Context, use_buttons: bool = None):
         else:
             await ctx.send(_("I will not use buttons on basic menus."))
 
+    @_set.command(name="errormsg")
+    @commands.is_owner()
+    async def _set_errormsg(self, ctx: commands.Context, *, msg: str = None):
+        """Set the message that will be sent on uncaught bot errors.
+
+        The message must be less than 1000 characters.
+        Use `{command}` to pass the command name in the message.
```
```suggestion
        Use `{command}` to pass the command name in the message.
        If you omit the `msg` argument, the message will be reset to the default one.
```

redbot/core/core_commands.py
```diff
@@ -3670,6 +3670,26 @@ async def use_buttons(self, ctx: commands.Context, use_buttons: bool = None):
         else:
             await ctx.send(_("I will not use buttons on basic menus."))
 
+    @_set.command(name="errormsg")
+    @commands.is_owner()
+    async def _set_errormsg(self, ctx: commands.Context, *, msg: str = None):
```
Jackenmen, line 3675: This command needs to be added to the core commands guide.

Jackenmen (changes requested):

Jackenmen (approved):
Let's goooo.

PR #5902, opened 2022-11-11
Allow send_interactive to upload a file containing all content

### LINKED ISSUE #5901: Add file output for code evaluation commands & Better long output handling

Dev

Currently when evaluating code with a large response, it would be shown in multiple messages. (when typing more)

Dropping a file would be way more convenient. 

Furthermore instead of typing more, a button will be nice. When clicking the button, the button disappears and a new message with a new button appears. (if there is further output)

### Anything else?

_No response_

TrustyJAID:

This allows `send_interactive` to optionally upload a file containing all content in the next message when `file` is said instead of `more`. This resolves #5901. 

Yes

### DIFF
--- a/redbot/core/commands/context.py
+++ b/redbot/core/commands/context.py
@@ -4,6 +4,7 @@ import asyncio
 import contextlib
 import os
 import re
+from io import StringIO
 from typing import Iterable, List, Union, Optional, TYPE_CHECKING
 import discord
 from discord.ext.commands import Context as DPYContext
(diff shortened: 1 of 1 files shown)


redbot/core/commands/context.py
```diff
@@ -211,6 +218,12 @@ async def send_interactive(
                         # or channel is a DM
                         with contextlib.suppress(discord.HTTPException):
                             await query.delete()
+                    if resp.content == "file":
```
Flame442: Currently, input such as `FILE` is treated identically to `more`.

Due to the predicate `lower_contained_in` being used, `resp.content` may contain capitals. This should either inspect the `result` attribute of the predicate to see if it is `1`, or also casefold the content before comparing here.

Flame442 (changes requested):

Flame442 (approved):

PR #5918, opened 2022-12-01
Core - #5891 Recovering Server Prefixes

### LINKED ISSUE #5891: Add some way to recover when someone forgets what they set their server prefix to

Core

Currently, it's possible to "lock yourself out" in a server if you set a server prefix and forget it. Global prefixes won't work, in best case the bot owner has enabled `--mentionable` and you can use mention prefix in the server but if the bot owner did not do that then you don't really have any good way of finding out what the prefix is, setting it to something else or resetting it. To alleviate this problem, at least one of those things needs to be possible.

Potential options are:
- adding `[guild]` parameter to `[p]set showsettings` command which could be used in DMs to get server's settings (including server prefix) without knowing the server prefix
- adding a new command for reseting server prefix from DMs (e.g. `[p]set serverprefix reset <guild>`)
- adding (or changing existing) command for setting server prefix from DMs

### Anything else?

_No response_

AntonioNarra:
Resolves #5891 
Modifies [p]set showsettings to accept an optional [guild] parameter in case users forget server prefixes
Modifies [p]set serverprefix to have a typing.Optional [guild] parameter before the prefix(es). Now works in DMs to the bot.

Yes

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3351,10 +3351,16 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
     # -- End Set Ownernotifications Commands -- ###
 
     @_set.command(name="showsettings")
-    async def _set_showsettings(self, ctx: commands.Context):
+    async def _set_showsettings(
+        self, ctx: commands.Context, specified_guild: discord.Guild = None
+    ):
         """
-        Show the current settings for [botname].
+        Show the current settings for [botname]. Accepts optional guild parameter in case must recover prefix.
         """
+        if specified_guild is not None:
+            # override ctx.guild if a guild is provided as an argument
+            ctx.guild = specified_guild
+
         if ctx.guild:
             guild_data = await ctx.bot._config.guild(ctx.guild).all()
             guild = ctx.guild
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3351,19 +3351,16 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
     # -- End Set Ownernotifications Commands -- ###
 
     @_set.command(name="showsettings")
-    async def _set_showsettings(
-        self, ctx: commands.Context, specified_guild: discord.Guild = None
-    ):
+    async def _set_showsettings(self, ctx: commands.Context, server: discord.Guild = None):
         """
-        Show the current settings for [botname]. Accepts optional guild parameter in case must recover prefix.
+        Show the current settings for [botname]. Accepts optional guild parameter if its prefix must be recovered.
         """
-        if specified_guild is not None:
-            # override ctx.guild if a guild is provided as an argument
-            ctx.guild = specified_guild
+        if server is None:
+            server = ctx.guild
 
-        if ctx.guild:
-            guild_data = await ctx.bot._config.guild(ctx.guild).all()
-            guild = ctx.guild
+        if server:
+            guild_data = await ctx.bot._config.guild(server).all()
+            guild = server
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -3351,10 +3351,16 @@ async def _set_ownernotifications_listdestinations(self, ctx: commands.Context):
...
+        self, ctx: commands.Context, specified_guild: discord.Guild = None
+    ):
         """
-        Show the current settings for [botname].
+        Show the current settings for [botname]. Accepts optional guild parameter in case must recover prefix.
         """
+        if specified_guild is not None:
+            # override ctx.guild if a guild is provided as an argument
+            ctx.guild = specified_guild
+
```
Flame442: ~~These changes should not be a part of this PR. This command is only for viewing the **current** settings.~~
Nevermind I forgot what **I** wrote in the issue.... However, the override of `ctx.guild` should be replaced with something similar to what I said in a comment on the other command.

redbot/core/core_commands.py
```diff
@@ -3599,8 +3605,9 @@ async def _set_prefix(self, ctx: commands.Context, *prefixes: str):
 
     @_set.command(name="serverprefix", aliases=["serverprefixes"])
     @checks.admin_or_permissions(manage_guild=True)
-    @commands.guild_only()
-    async def _set_serverprefix(self, ctx: commands.Context, *prefixes: str):
+    async def _set_serverprefix(
+        self, ctx: commands.Context, specified_guild: Optional[discord.Guild], *prefixes: str
```
```suggestion
        self, ctx: commands.Context, server: Optional[discord.Guild], *prefixes: str
```
Just to be consistent with the user-facing name "server" used in the command name.

redbot/core/core_commands.py
```diff
@@ -3613,10 +3620,15 @@ async def _set_serverprefix(self, ctx: commands.Context, *prefixes: str):
...
             - `[p]set serverprefix "@[botname] "` - This uses a mention as the prefix.
             - `[p]set serverprefix ! ? .` - Sets multiple prefixes.
+            - `[p]set serverprefix guild_name ? - Sets prefix for specified guild.
 
         **Arguments:**
             - `[prefixes...]` - The prefixes the bot will respond to on this server. Leave blank to clear server prefixes.
         """
+        if specified_guild is not None:
+            # prefer specified guild over ctx.guild when provided as argument
+            ctx.guild = specified_guild
```
Flame442: Instead of overriding the `ctx` attribute (which affects *anything* using that context object), do something more like
```py
if server is None:
    server = ctx.guild
```
then replace anywhere the command currently uses `ctx.guild` with `server`.

redbot/core/core_commands.py
```diff
@@ -3613,10 +3620,15 @@ async def _set_serverprefix(self, ctx: commands.Context, *prefixes: str):
             - `[p]set serverprefix "! "` - Quotes are needed to use spaces in prefixes.
             - `[p]set serverprefix "@[botname] "` - This uses a mention as the prefix.
             - `[p]set serverprefix ! ? .` - Sets multiple prefixes.
+            - `[p]set serverprefix guild_name ? - Sets prefix for specified guild.
```
```suggestion
            - `[p]set serverprefix "Red - Discord Bot" ? - Sets the prefix for a specific server. Quotes are needed to use spaces in the server name.
```

Flame442 (changes requested):

redbot/core/core_commands.py
AntonioNarra: Right, that makes sense. Added changes to use server variable rather than modifying ctx.guild as mentioned below.

Flame442 (approved):
Thanks for making those changes, this is all set now!

PR #5919, opened 2022-12-02
Updated excluded features

Kowlin:

Add ``PRIVATE_THREADS`` to the excluded list of features, as this feature is now available to the public after being removed from T2 boosting

Yes

### DIFF
--- a/redbot/cogs/general/general.py
+++ b/redbot/cogs/general/general.py
@@ -422,6 +422,8 @@ class General(commands.Cog):
                 "NEW_THREAD_PERMISSIONS",
                 "TEXT_IN_VOICE_ENABLED",
                 "THREADS_ENABLED",
+                # available to evertyone sometime after forum channel release
+                "PRIVATE_THREADS",
             }
             custom_feature_names = {
                 "VANITY_URL": "Vanity URL",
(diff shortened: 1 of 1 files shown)


redbot/cogs/general/general.py
```diff
@@ -422,6 +422,8 @@ def _bitsize(num: int):
                 "NEW_THREAD_PERMISSIONS",
                 "TEXT_IN_VOICE_ENABLED",
                 "THREADS_ENABLED",
+                # available to evertyone sometime after forum channel release
```
Zephyrkul:
```suggestion
                # available to everyone sometime after forum channel release
```

Jackenmen:
For completeness, here are other features that should NOT be excluded:
- `APPLICATION_COMMAND_PERMISSIONS_V2` - based on the documentation when this flag is present, the new behavior is NOT enabled so it will actually gradually disappear as servers opt-in/mandatory rollout starts
- `AUTO_MODERATION` - this shows up and disappears depending on whether a server has at least one rule set up (also present if a rule is disabled but not removed)

Jackenmen (approved):
Looks fine.

PR #5920, opened 2022-12-02
Bump discord.py to version 2.1.0

Kowlin:

For as far as I'm aware no additional changes are needed for this bump. However, would be a good idea to double check.

Yes

### DIFF
--- a/redbot/cogs/mod/kickban.py
+++ b/redbot/cogs/mod/kickban.py
@@ -538,7 +538,9 @@ class KickBanMixin(MixinMeta):
                     banned.append(user_id)
                 else:
                     try:
-                        await guild.ban(user, reason=audit_reason, delete_message_seconds=days * 86400)
+                        await guild.ban(
+                            user, reason=audit_reason, delete_message_seconds=days * 86400
+                        )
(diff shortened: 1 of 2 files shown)


Jackenmen (changes requested):
Should do a swap from `delete_message_days` to `delete_message_seconds` when `Guild.ban()`/`Member.ban()` is used (passed argument needs to be multiplied by `86400`) as the former is deprecated. Based on code search, this argument is only used in `redbot/cogs/mod/kickban.py` file.

---

We could also consider allowing people to specify other intervals in commands such as `[p]ban`, `[p]massban`, and `[p]tempban` though in the `[p]tempban` case it may make the parameter syntax more ambiguous due to temp ban duration and delete seconds allowing the same syntax (timedelta converter). Since it's not a very clear cut, I recommend that you make an issue for that where we can discuss it further.

Kowlin:
Okay that was the thing that I absolutely glanced over and blanked out while reading the change log, whoops 😅

Jackenmen:
To be fair, the changelog doesn't mention the deprecation anywhere :)

Jackenmen (approved):
Seems fine now, thanks for making the issue for tracking the new possibility for ban commands.

PR #5921, opened 2022-12-02
Add optional delay for [p]audioset dc

### LINKED ISSUE #5350: [Audio] Timeout value for "dc" setting

Audio

Currently it seems like the alternatives are that the bot either stays in the channel until its empty, and then disconnects after a `emptydisconnect` time, or it disconnects immediately after the queue goes empty, with the `dc` setting.

I think it would make a lot of sense if you could configure a timeout for the `dc` setting, in the form of "Wait X amount of seconds before disconnecting on empty queue". This would allow you to have the bot disconnect when its done playing, but at the same time also reduce the amount of unnecessary disconnects. I often see new songs being queued directly after the current one is done, and the bot has just 1 second ago disconnected.

### Anything else?

_No response_

AntonioNarra:
Closes #5350 
Gives users the option to specify a delay before the bot disconnects on an empty queue with `[p]audioset dc <delay>`

Yes

### DIFF
--- a/redbot/cogs/audio/core/__init__.py
+++ b/redbot/cogs/audio/core/__init__.py
@@ -133,6 +133,7 @@ class Audio(
             emptydc_timer=0,
             emptypause_enabled=False,
             emptypause_timer=0,
+            emptyqueuedc_timer=0,
             jukebox=False,
             jukebox_price=0,
             maxlength=0,
--- a/redbot/cogs/audio/core/commands/audioset.py
+++ b/redbot/cogs/audio/core/commands/audioset.py
@@ -614,23 +614,33 @@ class AudioSetCommands(MixinMeta, metaclass=CompositeMetaClass):
     @command_audioset.command(name="dc")
     @commands.guild_only()
     @commands.mod_or_permissions(manage_guild=True)
-    async def command_audioset_dc(self, ctx: commands.Context):
+    async def command_audioset_dc(self, ctx: commands.Context, seconds: int = 0):
         """Toggle the bot auto-disconnecting when done playing.
 
         This setting takes precedence over `[p]audioset emptydisconnect`.
+        Users can optionally specify a delay.
         """
+        if seconds < 0:
+            return await self.send_embed_msg(
+                ctx, title=_("Invalid Time"), description=_("Seconds can't be less than zero.")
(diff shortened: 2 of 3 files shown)


Zephyrkul:
As it stands in this PR, if the queue empties (thus starting the timer for disconnecting), there's no way to stop that timer even if more songs are enqueued.

AntonioNarra:
> As it stands in this PR, if the queue empties (thus starting the timer for disconnecting), there's no way to stop that timer even if more songs are enqueued.

That's true. It's possible to interrupt the sleep timer whenever a new song is queued with `[p]autoplay`, `[p]bumpplay`, and `[p]play`, although I'm not sure if that's the best fix for this problem.

ltzmax:
> > As it stands in this PR, if the queue empties (thus starting the timer for disconnecting), there's no way to stop that timer even if more songs are enqueued.
> 
> 
> 
> That's true. It's possible to interrupt the sleep timer whenever a new song is queued with `[p]autoplay`, `[p]bumpplay`, and `[p]play`, although I'm not sure if that's the best fix for this problem. 

Tbh not sure if this is good idea either but you could also do `0 to disable the timer` or `-1`.

Drapersniper:
> > As it stands in this PR, if the queue empties (thus starting the timer for disconnecting), there's no way to stop that timer even if more songs are enqueued.
> 
> That's true. It's possible to interrupt the sleep timer whenever a new song is queued with `[p]autoplay`, `[p]bumpplay`, and `[p]play`, although I'm not sure if that's the best fix for this problem.

And previous, or any command really, if the queue is empty and someone is interacting with the cog in any way it should dc during if you are adding a timer

Drapersniper:
> > > As it stands in this PR, if the queue empties (thus starting the timer for disconnecting), there's no way to stop that timer even if more songs are enqueued.
> > 
> > 
> > That's true. It's possible to interrupt the sleep timer whenever a new song is queued with `[p]autoplay`, `[p]bumpplay`, and `[p]play`, although I'm not sure if that's the best fix for this problem.
> 
> Tbh not sure if this is good idea either but you could also do `0 to disable the timer` or `-1`.

0 should disabled the timer yeah to keep the current behaviour

PR #5932, opened 2022-12-20
Fix ordering of 3.0.0.dev1 and 3.0.0a1.dev1 versions

Jackenmen:

According to PEP 440, `3.0.0.dev1` actually comes *before* `3.0.0a1.dev1`. This PR makes sure that our VersionInfo class is consistent with that.

To make sure that our test cases are consistent with PEP 440, I also check that comparison between `packaging.version.Version` objects gives the same result.

Yes


tests/core/test_version.py
```diff
@@ -16,10 +17,16 @@ def test_version_working():
 
 # When adding more of these, ensure they are added in ascending order of precedence
 version_tests = (
+    "3.0.0.dev1",
+    "3.0.0.dev2",
+    "3.0.0a32.dev12",
+    "3.0.0a32",
     "3.0.0a32.post10.dev12",
+    "3.0.0a32.post10",
     "3.0.0rc1.dev1",
```
Flame442: Add one `b`eta string to the tests, since that is technically supported but also not currently tested, and this is good to go.

Flame442 (changes requested):

Flame442 (approved):

PR #5942, opened 2022-12-28
Allow any Messageable in MessagePredicate's channel parameter

Jackenmen:

Allows passing any `Messageable` to `channel` parameter of `MessagePredicate.same_context` and by extension any methods that use it as well. In particular, this allows a developer to pass `User` and `Member` objects to `channel` parameter without having to fetch `DMChannel` first.

This concerns any predicate that support both DM and guild channels. Those that only support guild channels have not been modified as such change wouldn't really benefit users in any way.

Yes


Flame442 (approved):
Thanks!

PR #5952, opened 2023-01-04
[Audio] Use more of the newer Managed/Unmanaged terminology

aikaterna:
This PR holds a correction for #5944, where the change should have been Managed -> Unmanaged instead of Managed -> External.

Changed the wording in some llsetup commands' docstrings to try to be more consistent with this idea.

Kreusada and I had talked about fully changing the External/Internal wording to Unmanaged/Managed when working on Audio's docs previously. I had said I wanted to keep some of the old External wording to try to make the change a bit more transparent for users since we have been referring to them as external/internal for years, but I changed my mind a little on that and wanted to also provide the corresponding docs changes. Let me know if this should be a separate PR for docs.

No, please double check the docs formatting.

### DIFF
--- a/docs/cog_guides/audio.rst
+++ b/docs/cog_guides/audio.rst
@@ -376,7 +376,7 @@ necessary modifications.
 Firstly, stop all Red bots. For each bot using Audio:
 
 1. Start the bot.
-2. Run the command ``[p]llset external``.
+2. Run the command ``[p]llset unmanaged``.
 3. Stop the bot.
 
 Next, open a command prompt/terminal window. Navigate to ``<datapath>/cogs/Audio`` for any of your bot
@@ -806,7 +806,7 @@ Do not use quotes in these commands. For example, ``[p]llset host 192.168.10.101
 
 .. code-block:: none
 
-    [p]llset external
+    [p]llset unmanaged
     [p]llset host "yourlavalinkip"
     [p]llset port "port"
     [p]llset password "password"
@@ -3252,7 +3252,7 @@ llsetup
 **Description**
 
 Manage Lavalink node configuration settings. This command holds all commands to
-manage an unmanaged (external) or managed Lavalink node.
+manage an unmanaged (user-managed) or managed (bot-managed) Lavalink node.
 
 .. warning::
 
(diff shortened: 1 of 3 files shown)


docs/cog_guides/audio.rst
```diff
@@ -3261,21 +3261,28 @@ manage an unmanaged (external) or managed Lavalink node.
...
+    [p]llsetup unmanaged
 
 **Description**
 
-Toggle using external Lavalink nodes - requires an existing external Lavalink node for
-Audio to work, if enabled. This command disables the managed Lavalink server: if you do
-not have an external Lavalink node you will be unable to use Audio while this is enabled.
+Toggle using unmanaged (user-managed) Lavalink nodes - requires an existing Lavalink 
+node for Audio to work, if enabled. This command disables the managed (bot-managed) 
+Lavalink server: if you do not have an unmanaged Lavalink node available, you will be
```
Flame442: Instead of "...if you do not have an unmanaged Lavalink node **available**...", perhaps something like "set up", that better implies that it requires explicit user action (unlike the managed server)?

redbot/cogs/audio/core/commands/llset.py
```diff
@@ -152,19 +152,19 @@ async def validate_input(cog, arg):
...
 
-    @command_llset.command(name="external", aliases=["unmanaged"])
-    async def command_llset_external(self, ctx: commands.Context):
-        """Toggle using external Lavalink nodes - requires an existing external Lavalink node for Audio to work, if enabled.
+    @command_llset.command(name="unmanaged", aliases=["external"])
+    async def command_llset_unmanaged(self, ctx: commands.Context):
+        """Toggle using external (unmanaged) Lavalink nodes - requires an existing Lavalink node for Audio to work, if enabled.
 
-        This command disables the managed Lavalink server, if you do not have an external Lavalink node you will be unable to use Audio while this is enabled.
+        This command disables the managed Lavalink server. If you do not have another Lavalink node to access, you will be unable to use Audio while this is enabled.
```
Flame442: Same comment above, about being more clear that the user probably has to set something up if they use this setting.

Flame442 (changes requested):

Flame442 (approved):

PR #5953, opened 2023-01-04
Remove llsetup alias for llset

aikaterna:

Rename `llsetup` to `llset` and don't keep `llsetup` as an alias. Hopefully, this will avoid users from trending towards thinking that they must set up something on Lavalink to make it work initially.

No


Jackenmen:
Do we really need the llsetup alias at all?

aikaterna:
IMO it could be removed.

Jackenmen:
I think we should remove it then.

Kreusada (commented):
The audio docs will also need to be updated as the llset commands are documented or referenced with the previously proper llsetup name. I can make a separate PR if that's less hassle for you

Jackenmen (approved):
I updated all occurrences of `llsetup` with `llset` and removed the alias swap for `[p]llset external` as I think it's better suited for #5952. Thanks for the PR.

PR #5955, opened 2023-01-05
[Trivia] Fix string interpolation in test errors

Flame442:
`test_trivia_lists` prints the value of `SchemaError` objects without running the object through `format_schema_error` first. This causes values that should be interpolated to be left as is, such as `{key}`. This PR fixes this issue, making the test output once again useful.

To test this PR, insert a new trivia file with valid YAML that does not follow the trivia file schema, and compare the outputs on the current build vs with this PR.

Yes


Jackenmen (approved):
Great :+1:

PR #5969, opened 2023-01-12
Add public `positive_int` and `finite_float` converters

### LINKED ISSUE #5939: Make positive_int converter a public API
### Type of feature request

API functionality

Move `positive_int` converter (along with `PositiveInt` NewType) that's present in `cleanup` and `economy` cogs to `redbot.core.commands.converter`. Other than adding documentation, this should really only require moving things around.

### Anything else?

This was suggested before here:
https://github.com/Cog-Creators/Red-DiscordBot/pull/4583#issuecomment-725577831

AAA3A-AAA3A:

Hello,

This PR adds the `positive_int` and `finite_float` converters to `redbot.core.commands.converter`.
Since `positive_int` is an instance of `redbot.core.commands.Range`, the appropriate error messages have been added to `bot.on_command_error` with type `commands.RangeError`.

Testing was done today in #testing the core server.
Thanks in advance and have a nice day,
AAA3A

Closes #5939

Yes


redbot/core/commands/converter.py
```diff
@@ -70,6 +71,22 @@
...
+    """
+    This converts a user provided string into a finite float.
+    """
+    try:
+        ret = float(arg)
+    except ValueError:
+        raise BadArgument(_("`{arg}` is not a number.").format(arg=arg))
+    if not math.isfinite(ret):
+        raise BadArgument(_("`{arg}` is not a finite number.").format(arg=ret))
+    return ret
```
Zephyrkul:
```suggestion
if TYPE_CHECKING:
    finite_float = float
else:

    def finite_float(arg: str) -> float:
        """
        This converts a user provided string into a finite float.
        """
        try:
            ret = float(arg)
        except ValueError:
            raise BadArgument(_("`{arg}` is not a number.").format(arg=arg))
        if not math.isfinite(ret):
            raise BadArgument(_("`{arg}` is not a finite number.").format(arg=ret))
        return ret
```

redbot/cogs/cleanup/cleanup.py
```diff
@@ -684,9 +685,7 @@ def check(m):
     @commands.guild_only()
     @checks.mod_or_permissions(manage_messages=True)
     @commands.bot_has_permissions(manage_messages=True)
-    async def cleanup_duplicates(
-        self, ctx: commands.Context, number: positive_int = PositiveInt(50)
-    ):
+    async def cleanup_duplicates(self, ctx: commands.Context, number: positive_int = int(50)):
```
Kreusada:
```suggestion
    async def cleanup_duplicates(self, ctx: commands.Context, number: positive_int = 50):
```

redbot/cogs/cleanup/converters.py
```diff
@@ -15,18 +15,3 @@ async def convert(self, ctx: Context, argument: str) -> int:
             return int(argument)
 
         raise BadArgument(_("{} doesn't look like a valid message ID.").format(argument))
-
-
-PositiveInt = NewType("PositiveInt", int)
```
Kreusada, line 20: You can remove the NewType import now as it is not used anywhere else in this file

redbot/core/cog_manager.py
```diff
@@ -7,7 +7,7 @@
 from typing import TYPE_CHECKING, Union, List, Optional
```
Kreusada:
```suggestion
from typing import Union, List, Optional
```
`TYPE_CHECKING` is no longer used inside the cog_manager file, we can remove its import.

redbot/core/events.py
```diff
@@ -241,10 +243,116 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
+                    message = _(
+                        "Provided value `{value}` for argument `{parameter_name}` must be a string no more than {maximum}."
+                    )
+                elif error.minimum is not None and error.maximum is None:
+                    message = _(
+                        "Provided value `{value}` for argument `{parameter_name}` must be a string no less than {minimum}."
+                    )
+                elif error.maximum is not None and error.minimum is not None:
+                    message = _(
+                        "Provided value `{value}` for argument `{parameter_name}` must be a string between {minimum} and {maximum}."
```
Kreusada:
```suggestion
                        "Provided value `{value}` for argument `{parameter_name}` must be a string with a length of no more than {maximum}."
                    )
                elif error.minimum is not None and error.maximum is None:
                    message = _(
                        "Provided value `{value}` for argument `{parameter_name}` must be a string with a length of no less than {minimum}."
                    )
                elif error.maximum is not None and error.minimum is not None:
                    message = _(
                        "Provided value `{value}` for argument `{parameter_name}` must be a string with a length of between {minimum} and {maximum}."
```

These strings should be changed for when the Range annotation is `str`, because a "string no more than [...]" isn't grammatically fitting. I've made a suggestion but feel free to improvise if you have a better idea.

redbot/core/events.py
```diff
@@ -241,10 +243,116 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
+                            "Provided value `{value}` for argument `{parameter_name}` must be a number no more than {maximum}."
+                        )
+                    elif converter.min is not None and converter.max is None:
+                        message = _(
+                            "Provided value `{value}` for argument `{parameter_name}` must be a number no less than {maximum}."
+                        )
+                    elif converter.max is not None and converter.min is not None:
+                        message = _(
+                            "Provided value `{value}` for argument `{parameter_name}` must be a number between {minimum} and {maximum}."
+                        )
```
Kreusada, line 338: You haven't handled when the converter annotation is `str` here. That means the code will fall past this section and UnboundLocalError will be raised because `message` isn't defined. You should add a condition here which accounts for when the converter annotation is a string. You should emulate the behaviour you defined above for when RangeError is raised. You should also take into account my grammatical comments I made above, and ensure that stays consistent in his handling of the converter error.

I suggest using an else block here instead of an additional elif, as we only need to account for three types (str, float, int).

You can test these changes by creating a command which uses a `str` annotation in the Range converter. For example:

```py
@commands.command()
async def test(ctx, arg: dpy_commands.converter.Range[str, None, 5]):
    await ctx.tick()

# [p]test helloworld
```

Currently, this raises an implicit error which looks like this:

image

redbot/core/events.py
```diff
@@ -241,10 +243,116 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
+                elif error.maximum is not None and error.minimum is not None:
+                    message = _(
+                        "Provided value `{value}` for argument `{parameter_name}` must be a number between {minimum} and {maximum}."
+                    )
+            elif isinstance(error.value, str):
+                count = len(error.value)
+                if count == 1:
+                    value = _("one character")
+                else:
+                    value = _("{count} characters").format(count=count)
```
Kreusada: Why are we doing this? It makes the response looks very absurd. Can the error value itself not be formatted here?
image

redbot/core/events.py
```diff
@@ -241,10 +243,116 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
             await ctx.send(msg)
             if error.send_cmd_help:
                 await ctx.send_help()
+        elif isinstance(converter, commands.RangeError):
```
Kreusada:
```suggestion
        elif isinstance(error, commands.RangeError):
```

Kreusada (changes requested):
There are a few changes needed for this PR which stood out to me, including part of the implementation which is missing. Could you address these changes, and then re-test them? I will also test them thoroughly when I review your changes again.

AAA3A-AAA3A:
Hello,
Thank you for your review of this PR.
All requested changes have been pushed.
Thanks in advance,
AAA3A

AAA3A-AAA3A:
After #6002 was merged, to fix an error on the local clone of my Red fork, I force push the commits. I had forgotten to save the `redbot.core.commands.converter` file, so I lost it. I just made the changes again.

Kreusada (commented):
You should update `redbot/core/commands/__init__.py`, importing both `finite_float` and `positive_int` respectively. You should also add these two converter names to `redbot.core.commands.converter.__all__`.

AAA3A-AAA3A:
> You should update `redbot/core/commands/__init__.py`, importing both `finite_float` and `positive_int` respectively. You should also add these two converter names to `redbot.core.commands.converter.__all__`.

Ah, I had thought of adding them to `redbot.core.commands`, but I thought it was a bad idea. I'll do it tomorrow.

redbot/core/events.py
```diff
@@ -241,10 +239,112 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
+                    maximum=error.maximum,
+                    minimum=error.minimum,
+                    parameter_name=ctx.current_parameter.name,
+                )
+            )
+            return
         elif isinstance(error, commands.BadArgument):
+            if isinstance(converter, commands.Range):
+                if converter.annotation is int:
+                    if (converter.min == 0 or converter.min == 1) and converter.max is None:
```
Kreusada: Whilst this check makes sense, if the converter minimum were to be 1, and the user passes "0" to the converter, the message would still ask for a positive integer despite zero being a positive integer. Is the `== 1` check really needed?

image

redbot/core/events.py
```diff
@@ -241,10 +239,112 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
+                        message = _("Argument `{parameter_name}` must be a positive integer.")
+                    elif converter.min is None and converter.max is not None:
+                        message = _(
+                            "Argument `{parameter_name}` must be an integer no more than {maximum}."
+                        )
+                    elif converter.min is not None and converter.max is None:
+                        message = _(
+                            "Argument `{parameter_name}` must be an integer no less than {maximum}."
```
Kreusada:
```suggestion
                            "Argument `{parameter_name}` must be an integer no less than {minimum}."
```
Here, we check if the converter's max is None, so it will always read "no less than None", instead of our desired result. It just requires changing this part of the string in the suggestion.

redbot/core/events.py
```diff
@@ -241,10 +239,112 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
+                        )
+                    elif converter.min is not None and converter.max is None:
+                        message = _(
+                            "Argument `{parameter_name}` must be a number no less than {maximum}."
+                        )
+                    elif converter.max is not None and converter.min is not None:
+                        message = _(
+                            "Argument `{parameter_name}` must be a number between {minimum} and {maximum}."
+                        )
+                else:
```
Kreusada:
```suggestion
                elif converter.annotation is str:
```

redbot/core/events.py
```diff
@@ -241,10 +239,112 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
+                    elif converter.min is not None and converter.max is None:
+                        message = _(
+                            "Argument `{parameter_name}` must be an integer no less than {maximum}."
+                        )
+                    elif converter.max is not None and converter.min is not None:
+                        message = _(
+                            "Argument `{parameter_name}` must be an integer between {minimum} and {maximum}."
+                        )
+                elif converter.annotation is float:
+                    if (converter.min == 0 or converter.min == 1) and converter.max is None:
```
Kreusada: Please apply the above comment to here aswell.

redbot/core/events.py
```diff
@@ -241,10 +239,112 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
+                        message = _("Argument `{parameter_name}` must be a positive number.")
+                    elif converter.min is None and converter.max is not None:
+                        message = _(
+                            "Argument `{parameter_name}` must be a number no more than {maximum}."
+                        )
+                    elif converter.min is not None and converter.max is None:
+                        message = _(
+                            "Argument `{parameter_name}` must be a number no less than {maximum}."
```
Kreusada:
```suggestion
                            "Argument `{parameter_name}` must be a number no less than {minimum}."
```

redbot/core/events.py
```diff
@@ -241,10 +239,112 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
+                    )
+                elif error.minimum is not None and error.maximum is None:
+                    message = _(
+                        "Argument `{parameter_name}` must be a number no less than {maximum}."
+                    )
+                elif error.maximum is not None and error.minimum is not None:
+                    message = _(
+                        "Argument `{parameter_name}` must be a number between {minimum} and {maximum}."
+                    )
+            else:
```
Kreusada:
```suggestion
            elif isinstance(error.value, str):
```

redbot/core/events.py
```diff
@@ -241,10 +239,112 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
             await ctx.send(msg)
             if error.send_cmd_help:
                 await ctx.send_help()
+        elif isinstance(error, commands.RangeError):
+            if isinstance(error.value, int):
+                if (error.minimum == 0 or error.minimum == 1) and error.maximum is None:
```
Kreusada: Whilst this check makes sense, if the minimum were to be 1, and "0" is passed to the converter, the message would still ask for a positive integer despite zero being a positive integer. Is the `== 1` check really needed?

redbot/core/events.py
```diff
@@ -241,10 +239,112 @@ async def on_command_error(ctx, error, unhandled_by_cog=False):
...
+                    message = _("Argument `{parameter_name}` must be a positive integer.")
+                elif error.minimum is None and error.maximum is not None:
+                    message = _(
+                        "Argument `{parameter_name}` must be an integer no more than {maximum}."
+                    )
+                elif error.minimum is not None and error.maximum is None:
+                    message = _(
+                        "Argument `{parameter_name}` must be an integer no less than {maximum}."
```
Kreusada:
```suggestion
                        "Argument `{parameter_name}` must be an integer no less than {minimum}."
```

Kreusada (changes requested):
There are still changes that are needed before this can be approved and merged. Have a look at the review comments, and test your changes afterwards if you can. Should be good to go if those changes are made correctly.

Kreusada (approved):
Thanks for working on this, looks all set now 🎉

PR #5970, opened 2023-01-13
[modlog] Use new audit log event

Zephyrkul:

Uses the new audit log event instead of polling the audit logs on member ban / unban. Depends on an update to discord.py.

- [X] Yes


Flame442 (approved):
This works fine, and greatly simplifies the code. ~~However, this will cause a potential ""breaking"" change, in the sense that the bot requires the `view_audit_log` permission bit in order to get this event for a server. We should probably note this in the changelogs as a minor thing to be aware of if we are going to go forward with this change.~~ It already needed it to read from the audit log after the event was triggered :P

Kowlin:
Hhhmmmmmm. While I'm a bit wary about losing the on ban stuff, I think overall this is a very welcome change and I'm fine with merging it like this.

PR #5978, opened 2023-02-02
Allow tempban by ID and ban downgrade

### LINKED ISSUE #5915: [tempban] Allow tempbans by `UserID` instead of `Member`
### Type of feature request

Command

- Command: **`tempban`**
- Cog: **`Mod`**

Hi!
By default, the `tempban` command checks for the user to be present in the server before banning them temporarily. That means you **cannot** ban any user who's not present in the server. Which you should be able to.

> I don't think it's discord limitation since the you can do the same with other bots and even you can `ban` users with the `ban` from the same **cog**, __regardless__ of their presence in the server.

So it'd be great to have that functionality on red, allowing `tempban` to check for `UserID` instead of member so that we can ban people after they leave the server temporarily.
Thank you!

### Anything else?

_No response_

raidensakura:

- tempban will allow banning user by ID
- If using the command would cause a downgrade from permanent ban to temporary ban, a prompt will appear asking if command invoker wants to proceed
- If the user is already tempbanned, a prompt will appear asking if command invoker wants to update the duration of the current tempban.
- Prompt will be skipped if `ctx.assume_yes` evaluates to `True`

Fixes #5915 

Yes.

### Additional Notes
This is my first PR to this repo and I'm quite new to Python. Suggestions for improvement to my PR are very welcome.


Zephyrkul:
> - Prompt will be skipped if `ctx.assume_yes` evaluates to `True`

I don't think allowing tempban to be run on a schedule would be best here.

raidensakura:
> I don't think allowing tempban to be run on a schedule would be best here.

I'm only implementing this based on Jack's last statement in the issue discussion, it can be raised for discussion again if needed.

I personally think it's okay to include it since prompt only appear if user is already banned anyway, my concern is that whether it should be the default choice to downgrade instead of skipping it completely.

Jackenmen:
> > * Prompt will be skipped if `ctx.assume_yes` evaluates to `True`
> 
> I don't think allowing tempban to be run on a schedule would be best here.

It's not for scheduling, it's to allow for implementing a command that skips the prompt if someone prefers it that way.

redbot/cogs/mod/kickban.py
```diff
@@ -665,9 +672,64 @@ async def tempban(
...
+
+            if response.content.lower().startswith("y"):
+                with contextlib.suppress(discord.NotFound):
+                    await prompt.delete()
+                with contextlib.suppress(discord.HTTPException):
+                    await response.delete()
+                return True
+            else:
+                await ctx.send("Cancelled.")
+                return False
```
Jackenmen: This can be vastly simplified by using the `ConfirmView`:
```suggestion
            content = (
                _(
                    "User is already tempbanned (ban expires {relative_time}), do you want to proceed?"
                ).format(relative_time=f"<t:{banned_until}:R>")
                if banned_until
                else _("This will downgrade the ban to a tempban, do you want to proceed?")
            )
            view = ConfirmView(ctx.author)
            view.message = await ctx.send(content)
            await view.wait()
            return view.result
```
Also, you weren't using i18n for the strings (i.e. they weren't wrapped in `_()`), the above suggestion fixes that as well.
Bu the way, considering this inner function is only used once, it may make more sense to just put this code directly where the function is used.

redbot/cogs/mod/kickban.py
```diff
@@ -638,18 +644,19 @@ async def tempban(
...
-        elif not await is_allowed_by_hierarchy(self.bot, self.config, guild, author, member):
-            await ctx.send(
-                _(
-                    "I cannot let you do that. You are "
-                    "not higher than the user in the role "
-                    "hierarchy."
+        elif in_server:
+            if not await is_allowed_by_hierarchy(self.bot, self.config, guild, author, member):
+                await ctx.send(
+                    (
```
Jackenmen: Use i18n
```suggestion
                    _(
```

redbot/cogs/mod/kickban.py
```diff
@@ -665,9 +672,64 @@ async def tempban(
...
+            await guild.fetch_ban(member)
+        except discord.NotFound:
+            pass
+        else:
+            async with self.config.guild(guild).current_tempbans() as current_tempbans:
+                banned_until = False
+                if member.id in current_tempbans:
+                    banned_until = await self.config.member_from_ids(
+                        guild.id, member.id
+                    ).banned_until()
```
Jackenmen: No need to use the context manager when we're only reading the value:
```suggestion
            current_tempbans = await self.config.guild(guild).current_tempbans()
            banned_until = False
            if member.id in current_tempbans:
                banned_until = await self.config.member_from_ids(
                    guild.id, member.id
                ).banned_until()
```

redbot/cogs/mod/kickban.py
```diff
@@ -665,9 +672,64 @@ async def tempban(
...
+            if not await prompt_downgrade(banned_until):
+                return
+
+        await self.config.member_from_ids(guild.id, member.id).banned_until.set(
+            unban_time.timestamp()
+        )
         async with self.config.guild(guild).current_tempbans() as current_tempbans:
-            current_tempbans.append(member.id)
+            if not member.id in current_tempbans:
```
Jackenmen: The `not in` operator is more readable when testing whether something is not in a collection:
```suggestion
            if member.id not in current_tempbans:
```

Jackenmen (changes requested):

Jackenmen (dismissed):
Looks good, thanks!

redbot/cogs/mod/kickban.py
```diff
@@ -665,9 +672,38 @@ async def tempban(
...
+                        " do you want to proceed?"
+                    ).format(relative_time=f"<t:{int(banned_until)}:R>")
+                    if banned_until
+                    else _("This will downgrade the ban to a tempban, do you want to proceed?")
+                )
+                view = ConfirmView(ctx.author)
+                view.message = await ctx.send(content, view=view)
+                await view.wait()
+                if not view.result:
+                    return
```
Jackenmen, line 700:
```suggestion
                    await ctx.send(_("Okay, I will not be doing that then."))
                    return
```

Jackenmen (approved):

PR #5989, opened 2023-02-14
docs: fix typos

Lioness100:

This PR fixes a few random typos (this was a trial run for [a CLI I made](https://github.com/Lioness100/rspell))
I think "bumpped_shuffle" is a typo, but it was used in a lot of places in the code, so I didn't want to change anything without clearing it with a maintainer.

Also, I know this edits both code and comments, and modifies a lot of different sections of the code, so let me know if you need me to split it up into more modular PRs. :)

No

### DIFF
--- a/redbot/cogs/mutes/mutes.py
+++ b/redbot/cogs/mutes/mutes.py
@@ -428,7 +428,7 @@ class Mutes(VoiceMutes, commands.Cog, metaclass=CompositeMetaClass):
         for result in results:
             if not result:
                 continue
-            _mmeber, channel, reason = result
+            _mmember, channel, reason = result
             unmuted_channels.remove(channel)
         modlog_reason = _("Automatic unmute")
 
@@ -1571,7 +1571,7 @@ class Mutes(VoiceMutes, commands.Cog, metaclass=CompositeMetaClass):
             "user": user,
         }
         # TODO: This typing is ugly and should probably be an object on its own
-        # along with this entire method and some othe refactorization
+        # along with this entire method and some other refactorization
         # v1.0.0 is meant to look ugly right :')
         if permissions.administrator:
             ret["reason"] = _(MUTE_UNMUTE_ISSUES["is_admin"])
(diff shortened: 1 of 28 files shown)


redbot/cogs/trivia/trivia.py
```diff
@@ -154,7 +154,7 @@ async def triviaset_allowoverride(self, ctx: commands.Context, enabled: bool):
             )
 
     @triviaset.command(name="usespoilers", usage="<true_or_false>")
-    async def trivaset_use_spoilers(self, ctx: commands.Context, enabled: bool):
```
Flame442, line 157: All changes in this file are inaccurate, as the command name is the literal `triviaset`, standing for "trivia setting".

redbot/cogs/mutes/mutes.py
```diff
@@ -428,7 +428,7 @@ async def _auto_channel_unmute_user_multi(
         for result in results:
             if not result:
                 continue
-            _mmeber, channel, reason = result
+            _mmember, channel, reason = result
```
Flame442:
```suggestion
            _member, channel, reason = result
```

Flame442 (changes requested):
Much more accurate than most other automated spellchecking tools that have made PRs to Red, I'll give you that!

redbot/cogs/trivia/trivia.py
Lioness100: I'm sorry, I think my intent was correct but I didn't fix it correctly. The typo was that it's "trivaset" (missing the "i" after the "v"). The problem was that I only fixed the trivia typo and forgot to add "set" with it.

Flame442: Ah, yep I see that now. If you want to make that change but keep `set` in the command name, that would be fine!

Flame442 (approved):
Thanks!

Lioness100:
Np! Just as a note, do you know if "bumpped_shuffle" is a typo?

Issue #5766, opened 2022-06-07
Fix PR titles made by prepare/publish release workflows when releasing from 3.x branch


Jackenmen:
PRs made against the 3.x branch need to have their PR title be prefixed with `[3.x] ` which the prepare and publish release workflows currently don't know about. This has been an issue for a while but I always leave it for later so now I'll just make an issue instead.

PR #5990, opened 2023-02-14
Fix PR titles made by prepare/publish release workflows

BigPeep0doo:
We have added an action in the .github/workflow/prepare_release.yml and .github/workflow/publish_release.yml files to display the PR title with the major and minor version prefix. This PR is from the issue #5766.

Yes

### DIFF
--- a/.github/workflows/prepare_release.yml
+++ b/.github/workflows/prepare_release.yml
@@ -45,7 +45,7 @@ jobs:
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           commit-message: Automated Crowdin downstream
-          title: "[i18n] Automated Crowdin downstream"
+          title: "[${{ steps.semver_parser.outputs.major }}.${{ steps.semver_parser.outputs.minor }}][i18n] Automated Crowdin downstream"
           body: |
             This is an automated PR that is part of Prepare Release automated workflow (2 out of 2).
             Please ensure that there are no errors or invalid files are in the PR.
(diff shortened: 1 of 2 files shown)


Jackenmen (changes requested):
Hi, thanks for taking the interest in improving our automation!

There are some things that need to be addressed before this can be merged:
- `[3.x] ` prefix should only be added when the PRs are made against `3.x` branches
- I'd rather have a simple script using `actions/github-script` than use another action such as `booxmedialtd/ws-action-parse-semver` that you used, especially considering that it's not a widely used action. This should actually make it a lot easier to implement the logic for only adding a prefix when PR is against `3.x` branch since it allows you to either return an empty string or appropriate `"[3.x] "` prefix string and use that in the PR title and the commit message

Jackenmen:
In case it helps, here's an example of a PR that should have a prefix:
- https://github.com/Cog-Creators/Red-DiscordBot/pull/5304

and here's an example of a PR that should not have a prefix:
- https://github.com/Cog-Creators/Red-DiscordBot/pull/5145

Jackenmen:
I realize that you're currently still working on this but I noticed that you're trying to use the value of `new_version` which is not what the existence of the prefix depends on. The existence of the prefix depends merely on whether the workflow dispatch was done against a branch starting with `3.`, the actual version number does not matter here. Since the prefix depends on the branch name, the semver parsing is also unnecessary here.

Issue #5976, opened 2023-01-31
Add functionality to mention app commands


Luc1412:
### Type of feature request

API functionality

### Description of the feature you're suggesting

Currently mentioning an app command is pain with plain discord.py. You would need to fetch all commands from api and search for the right command. This can be improved by implementing a cache. 

A tag from discord.py: 
> To have your bot mention its own commands without hardcoding them, you'll need to use either tree.sync or tree.fetch_commands to get the commands from discord then .mention the command(s) you want.
It's recommended to subclass CommandTree to override these methods and cache the results to avoid unnecessary api calls.

The best way would be adding such cache in the tree. This cache is either supplied on sync or when triggering manually an update with a method.

Furthermore a method to get a mention would be nice. E.g. `Bot.get_app_command_mention(name: str, guild_id: Optional[int]) -> Optional[str]` 
Usage: 
- `bot.get_app_command_mention('ban')`
- `bot.get_app_command_mention('cog update')`
- `bot.get_app_command_mention('load', guild_id=133049272517001216)`

PR #5992, opened 2023-02-18
Application Command Manager

### LINKED ISSUE #5672: Add command for syncing app commands
### Type of feature request

Command

#### Cog: 
Downloder or Admin

#### Description
A command to sync the bots application commands as it’s the preferred way to sync commands. 
This command should include a guild id option for syncing a specific guilds app commands 

#### Restrictions
This command should have the same permissions as loading/unloading cogs 

### Anything else?

_No response_

Flame442:
Closes #5672

#### Checklist for this PR
- [x] `[p]slash enable <command> [type]` - Enables a specific app command.
- [x] `[p]slash disable <command> [type]` - Disables a specific app command.
- [x] `[p]slash enablecog <cog>` - Enables all app commands from a cog.
- [x] `[p]slash disablecog <cog>` - Disables all app commands from a cog.
- [x] `[p]slash list` - Lists all app commands the bot can see, and whether they are enabled.
- [x] `[p]slash sync` - Syncs app commands to discord.
- [x] Slash command error handling.
- [x] Notifications on cog update warning users about modified app commands.
- [x] Documentation
  - [x] Document `RedTree`.
  - [x] Document new commands.

#### Future work
- Guild commands
  - Currently this just auto-accepts any command that is in the guild scope. Might want to make guild scoped commands be managed and/or add the ability for guild owners to enable/disable a command at the guild level.
- Translation support
  - I don't really know how to do this, or if it's even really feasible.
- Override `app_commands` with Red objects, similar to existing overrides for `ext.commands`.
  - Subclass `app_commands.errors.BotMissingPermissions`
    - This is necessary if we want `permissions` to work with slash commands/hybrid commands. Since slash commands can be restricted with built in discord permissions, I don't think it's a major concern to get `permissions` working with them. Still, it might be something we want to do in the future.
  - Local caching of `AppCommand` objects (#5976)
    - Add an `id` attribute filled from the config values stored in this PR to the `Command` subclass.
  - Support commands that are required to be enabled at all times, and will prevent a cog load if they cannot be. (#6018)
    - Add a kwarg to the `@app_commands.command()` deco which causes the command to bypass normal load logic.
    - This has the benefits of allowing some commands that a CC denotes as "required", while still supporting a system for non-required commands to be added without entirely blocking a cog from loading.
- Handle slash versions of hybrid commands not working in DMs (#5995)
  - Either the bot needs to send a message explaining why it isn't responding, or we need to figure out how to get around the `PartialMessageable` limitation in general.
- Check Red specific global checks in `app_command` commands. (#6015)
  - `block`/`allowlist` settings, possibly the `permissions` cog, and possibly discord permissions.

Lightly


Luc1412:
~~You can link #5672 as it will be solved with this PR.~~ Additionally including #5976 could be an option while already implementing a custom tree.

Flame442:
This is now ready for review.

Luc1412:
I'd say for #5976, 2 things would be missing:
- Fetch/Cache commands on bot start
- Util function to mention command with a possible signature of `def mention_command(name: str, guild: Optional[Guild] = None`. Name could be like: "ban" or "commands list" (supporting subcommands)

Flame442:
Fetching on bot start is an API call that will only be used if someone actually wants to use an `AppCommand`, so there's no reason to do it every bot start if it isn't going to get used by a cog. Lazy loading, like it is right now with `get_or_fetch`, makes it so API calls are only made when strictly necessary.

I don't see any reason to add a specific mention util on top of a util that gives the `AppCommand` object from the same information, as it is trivial to just access the `.mention` attribute yourself.

I didn't follow the exact "specs" of that issue, I adapted them to be more reasonable while still giving the same functionality.

Luc1412:
> Fetching on bot start is an API call that will only be used if someone actually wants to use an `AppCommand`, so there's no reason to do it every bot start if it isn't going to get used by a cog. Lazy loading, like it is right now with `get_or_fetch`, makes it so API calls are only made when strictly necessary.

Fetching all commands on startup can be an unnecessary api request, if you don't use cogs with app commands, however a cheap api request could simplify the process for developers that create app commands cogs and even might save requests.

App Commands Cog devs got 2 options:
1.) Fetch all app commands on cog load, if the cache is empty
2.) Fetch commands with `get_or_fetch`  on demand.
The 2nd option could be more expensive, than 1 simple request.

> I don't see any reason to add a specific mention util on top of a util that gives the AppCommand object from the same information, as it is trivial to just access the .mention attribute yourself.

Also the workflow for mentioning app commands isn't that seamless:
```py
cmd = self.bot.tree.get_or_fetch('testcommand')
cmd.mention
# or with fetching all commands on startup
self.bot.tree.cached_app_commands.get(('testcommand', None, 'slash')).mention
```
1st option always require an async function, 2nd option is just really inconvenient.

or even worse with sub commands
 ```py
cmd = self.bot.tree.get_or_fetch('testcommand')
subcmd = [o for o in cmd if isinstance(o, discord.app_commands.AppCommandGroup) and o.name == 'testsub'][0]
subsubcmd = [o for o in subcmd if isinstance(o, discord.app_commands.AppCommandGroup) and o.name == 'testsubsub'][0]
subsubcmd.mention

cmd.mention
# or with fetching all commands on startup
cmd = self.bot.tree.cached_app_commands.get(('testcommand', None, 'slash'))
subcmd = [o for o in cmd if isinstance(o, discord.app_commands.AppCommandGroup) and o.name == 'testsub'][0]
subsubcmd = [o for o in subcmd if isinstance(o, discord.app_commands.AppCommandGroup) and o.name == 'testsubsub'][0]
subsubcmd.mention
```
This can be really painful and require too much code. Espacally if you need to check if commands/subcommand exist.

Alternative having:
 ```py
 self.bot.tree.get_mention('testcommand')
 ```
or 
 ```py
 self.bot.tree.get_mention('testcommand testsub testsubsub')
 ```
is way more convenient.

**Comment by Flame442** [comment]
We had some discussion on how best to handle #5976, and the solution is out of scope for this PR. The eventual solution will likely be caching the IDs of commands in config on `sync` and imbedding that ID in the `Command` object.

**Comment by Zephyrkul** [comment]
I have some major concerns with this PR as it currently stands, which have to do with this PR's design breaking a major promise that discord.py follows: extensions are never loaded in a partial state - either the entire extension succeeds and is accessible, or the entire extension fails and is fully ejected from the bot. It's ultimately up to the org to decide if this is a promise they want to break, as well as how far they want to break it. I must strongly state, however, that breaking this promise to the extent that this PR does will lead to major pain points for both 3rd-party cog creators and end users alike. As an example, a 3rd party cog could reference another command as a failure state (e.g. one command needing settings that are applied via another command), yet if that other command isn't loaded, it can result in confusion when that other command doesn't show up in the bot's app commands list. Cog creators (as well as core devs if they ever decide to add any app or hybrid commands to core) would have to code around this limitation, something that not all of them would do, or even be aware they would need to do.

In my eyes, an extension should (and, in general, *does*) work to accomplish one singular goal, e.g. automoderation, playing music, etc. Some 3rd party extensions invariably break this principle, but those are few and far between - forcing all 3rd party creators to resolve this issue due to the mere *possibility* that this could happen is not a good approach in my eyes. Allowing an extension to only be partially added to the bot's tree would result in said extension being only partially able to serve that goal via the slash commands the extension attempts to make available - and if an extension happens to be slash-only, this can be disastrous.

There are currently 3 routes that I can think of by which this PR allows end users to break this promise. 

The first is obvious - extensions can be loaded and users can decide against adding an extension's commands to the tree. As this is the entire point of this PR and a necessity for a modular bot such as Red, I have no issue with this; especially since this PR is well-documented as doing exactly this.
The second route is by enabling only some commands from an extension via `[p]slash enable`. This can be avoided by checking if the command is part of an extension and pointing the user to the relevant command.
The third route is by loading an extension's commands via `[p]slash enablecog`, yet not all commands could be added. Rather than rolling back like discord.py does when failing to add an extension, this PR instead leaves the cog in its partial state, having added only some commands completely arbitrarily. Even if the org decides to stick with the broken promise that this PR puts forward, this will have to change - nowhere else does Red do something like this, only adding some commands at random and leaving it up to the user to figure out the rest.

I also have a minor concern, that being that `tree.sync` is a heavily rate-limited endpoint. While the helpstring for `[p]slash sync` does mention this, all that a novice user would see is the bot not responding (not even typing) for up to a minute or longer. As well, from experience I can tell you that users may be tempted to unnecessarily run `[p]slash sync` on every launch. It's up to you if you want to address this misconception or not in this PR. Finally, another thing that might be worth documenting is the fact that global app commands don't necessarily show up immediately. Modern Discord clients will be notified immediately, and most others will be notified within a few minutes. By now, though, most if not all clients should see the new commands immediately.
You can see my own attempts to educate users in my own sync cog here, if you want a reference to an approach to this: https://github.com/Zephyrkul/FluffyCogs/blob/dpy2/sync/__init__.py#L106

File: redbot/core/core_commands.py [hunk 1]
```diff
@@ -1940,6 +1941,297 @@ async def reload(self, ctx: commands.Context, *cogs: str):
...
+            guild = guild.id
+        try:
+            await self.bot.tree.sync(guild=guild)
+        except discord.Forbidden as e:
+            # Should only be possible when syncing a guild, but just in case
+            if not guild:
+                raise e
+            await ctx.send(
+                _(
+                    "I need the `applications.commands` scope in this server to be able to do that. Reinvite me with that scope first."
```
Zephyrkul: It may be worth noting here that kicking the bot is unnecessary - users will think "reinvite" implies that they need to kick the bot first.
Pointing to `[p]inviteset commandscope` might also be worth doing.

Flame442:
Thank you for your feedback. I have updated `[p]slash enablecog` to only enable any commands if it can enable all commands in the cog, added a cooldown to `[p]slash sync` with a custom error message similar to your example, and updated the `applications.commands` message to give a better idea of what needs to be done to resolve the issue.

Regarding the second route of this PR that can break the promise that extensions are never loaded in a partial state, I believe that this is a necessary trade off to enable customization for bot owners and to allow Cog Creators to add app commands to their cogs without concern for the command limits.

If core mandates that either all or no app commands from a cog are loaded, then cogs would be discouraged from adding app commands (especially context menu app commands) unless strictly necessary. If a cog does not limit themselves in this case, it would be unable to have any slash commands due to even slightly exceeding a limit. For example, any cog that decides to add a context menu command must hope that it is not installed on a bot that already has 5 commands of the same scope, or all application commands from that cog will be unable to be enabled.

This PR is implemented in such a way as to give Cog Creators the freedom to add as many app commands to their cogs as they want, with the understanding that not every bot owner will want every command from every cog. It allows Cog Creators to offer lots of app commands for bot owners to choose from, without worrying that they will hit the command limit and be locked out of using slash entirely. For convenience, all commands from a cog can be enabled with the `enable/disablecog` commands, for cases where all commands from a cog are wanted.

To your point about a command being referenced but not actually being available, this is no different than if the command is disabled with `[p]command disable` or the `permissions` cog, both features that allow bot (/server) owners to configure exactly what commands their bot will use. While this PR is default-disable for commands, rather than default-enable, like text commands, the same principle that commands are not guaranteed to be available applies. I do not see this as a problem unique to this PR.

Zephyrkul:
`[p]slash enablecog` and related commands will likely need to utilize a lock, max concurrency check, or some form of atomically / synchronously updating the enabled count beforehand, so that running it multiple times doesn't result in race conditions raising a CommandLimitReached error (and leaving the slash commands in an arbitrary state).

Kowlin (dismissed):
This is working great, just gonna wait for a sanity check from @TrustyJAID! 👍

Kowlin (dismissed):
This is working great, just gonna wait for a sanity check from TrustyJAID! 👍

Rapptz:
>Regarding the second route of this PR that can break the promise that extensions are never loaded in a partial state, I believe that this is a necessary trade off to enable customization for bot owners and to allow Cog Creators to add app commands to their cogs without concern for the command limits.

I think a good middle ground for this is to provide a tag that would label a command as "essential" and if the essential command is disabled/not there then it fails to load completely. Or it can be flipped so that there's a tag that marks it as "optional".

redbot/core/bot.py
```diff
@@ -34,6 +34,7 @@
 from types import MappingProxyType
 
 import discord
+from discord.enums import AppCommandType
```
Rapptz: Please don't import from modules that aren't `discord`, `discord.abc`, `discord.utils`, `discord.ui`, `discord.app_commands`, or `discord.ext`.

redbot/core/tree.py
```diff
@@ -0,0 +1,324 @@
+import discord
+from discord.abc import Snowflake
+from discord.utils import MISSING, _is_submodule
```
Rapptz: Please don't use this. It's private for a reason. This has and will break again in the future.

redbot/core/tree.py
```diff
@@ -0,0 +1,324 @@
+import discord
+from discord.abc import Snowflake
+from discord.utils import MISSING, _is_submodule
+from discord.enums import AppCommandType
```
Rapptz: Please don't import from modules that aren't `discord`, `discord.abc`, `discord.utils`, `discord.ui`, `discord.app_commands`, or `discord.ext`.

redbot/core/tree.py
```diff
@@ -0,0 +1,324 @@
...
+            self._disabled_context_menus[(command, None, type.value)] = com
+
+    @staticmethod
+    async def _send_from_interaction(interaction, *args, **kwargs):
+        """Util for safely sending a message from an interaction."""
+        ctx = await interaction.client.get_context(interaction)
+        try:
+            await ctx.send(*args, ephemeral=True, **kwargs)
+        except discord.HTTPException:
+            await ctx.channel.send(*args, **kwargs)
```
Rapptz: Note this can be done without context:

```suggestion
        if interaction.response.is_done():
            if interaction.is_expired():
                return await interaction.channel.send(*args, **kwargs)
            return await interaction.followup.send(*args, ephemeral=True, **kwargs)
        return await interaction.response.send_message(*args, ephemeral=True, **kwargs)
```

redbot/core/tree.py
```diff
@@ -0,0 +1,324 @@
...
+            )
+            await self._send_from_interaction(interaction, msg, delete_after=error.retry_after)
+        elif isinstance(error, CheckFailure):
+            await self._send_from_interaction(
+                interaction, _("You are not permitted to use this command.")
+            )
+        else:
+            log.exception(type(error).__name__, exc_info=error)
+
+    def _remove_with_module(self, name: str, *args, **kwargs) -> None:
```
Rapptz, line 312: Note this method is private so it'll probably break in the future.

Rapptz (commented):
Just a quick glance on things from an API hygiene POV.

From a different POV, I'm not really sure how I feel about the partial loads like I've mentioned before long ago. I get the limits are a thing and there needs to be some sort of solution to the problem. One of the ones I've thought about involve tagging the commands as optional or not and when installing a cog the user can pick which optional components they want to add or have it fail if the slash commands themselves can't fit within the limit. It's a middle ground that allows extension developers to pick what functionality they expect to be there without it being taken from them should they do something with it at runtime.

Anyway these are just random musings. I didn't spend much time on this review, only like 5 minutes or so.

TrustyJAID:
After reading through the comments here I think there are some minor things we can address before this gets merged. @Zephyrkul makes a good point about the promise of partial loading. I still believe this is the best path forward for a modular Red the way this PR currently is but there could be cases where a cog creator may make strictly slash command only cogs and it would be inconvenient to have the cog fail to function because we intervened when attempting to add it. Perhaps it would be best to add a new `info.json` key that can let a cog author designate that a certain application command(s) *must* be loaded with this cog when the cog is installed as a middle ground. If that cog happens to add a command that puts the bot over the command limit then fail to load the cog entirely rather than leave the cog in a partial state.

redbot/core/core_commands.py
```diff
@@ -1940,6 +1941,357 @@ async def reload(self, ctx: commands.Context, *cogs: str):
...
+        for name in to_add_slash:
+            await self.bot.enable_app_command(name, discord.AppCommandType.chat_input)
+        for name in to_add_message:
+            await self.bot.enable_app_command(name, discord.AppCommandType.message)
+        for name in to_add_user:
+            await self.bot.enable_app_command(name, discord.AppCommandType.user)
+
+        # Update the tree with the new list of enabled cogs
+        await self.bot.tree.red_check_enabled()
+        await ctx.send(_("Enabled all commands from `{cog_name}`.").format(cog_name=cog_name))
```
TrustyJAID: I would like to see this list how many top level commands were added and list the names of the top level commands added.

redbot/core/core_commands.py
```diff
@@ -1940,6 +1941,357 @@ async def reload(self, ctx: commands.Context, *cogs: str):
...
+        if not count:
+            await ctx.send(
+                _(
+                    "Couldn't find any enabled commands from the `{cog_name}` cog. Use `{prefix}slash list` to see all cogs with slash commands."
+                ).format(cog_name=cog_name, prefix=ctx.prefix)
+            )
+            return
+        await self.bot.tree.red_check_enabled()
+        await ctx.send(
+            _("Disabled {count} commands from `{cog_name}`.").format(
```
TrustyJAID: I would like to see this list which commands were disabled to make it more clear which commands were part of the cog.

redbot/core/core_commands.py
```diff
@@ -1940,6 +1941,357 @@ async def reload(self, ctx: commands.Context, *cogs: str):
...
+    async def slash_sync(self, ctx: commands.Context, guild: discord.Guild = None):
+        """Syncs the slash settings to discord.
+
+        Settings from `[p]slash list` will be synced with discord, changing what commands appear for users.
+        This should be run sparingly, make all necessary changes before running this command.
+
+        **Arguments:**
+            - `[guild]` - If provided, syncs commands for that guild. Otherwise, syncs global commands.
+        """
+        try:
```
TrustyJAID: It may be appropriate to return early if `ctx.assume_yes` is enabled to prevent automating this command through other cogs.
```suggestion
        """
        if ctx.assume_yes:
            return
        try:
```

Drapersniper:
> After reading through the comments here I think there are some minor things we can address before this gets merged. @Zephyrkul makes a good point about the promise of partial loading. I still believe this is the best path forward for a modular Red the way this PR currently is but there could be cases where a cog creator may make strictly slash command only cogs and it would be inconvenient to have the cog fail to function because we intervened when attempting to add it. Perhaps it would be best to add a new `info.json` key that can let a cog author designate that a certain application command(s) _must_ be loaded with this cog when the cog is installed as a middle ground. If that cog happens to add a command that puts the bot over the command limit then fail to load the cog entirely rather than leave the cog in a partial state.

Info.json would work fine for published cogs, perhaps instead we can provide a decorator that tags a command as necessary at the code level that way this would work for both published and unpublished cogs

redbot/core/core_commands.py
```diff
@@ -1940,6 +1941,357 @@ async def reload(self, ctx: commands.Context, *cogs: str):
...
+            await ctx.send(_("Command type must be one of `slash`, `message`, or `user`."))
+            return
+
+        current_settings = await self.bot.list_enabled_app_commands()
+        current_settings = current_settings[command_type]
+
+        if command_name in current_settings:
+            await ctx.send(_("That application command is already enabled."))
+            return
+
```
TrustyJAID, line 1995: We should validate the command being enabled against the disabled list as well here to prevent confusion when enabling individual commands. This will prevent users from being able to enable commands pre-emptively before adding a cog but it's better than them being confused why enabling a command didn't work and the list command showing extraneous unknown commands. If you want to get real fancy you can add a fuzzy search if the user makes a typo but that can be another time.

redbot/core/core_commands.py
```diff
@@ -1940,6 +1941,357 @@ async def reload(self, ctx: commands.Context, *cogs: str):
...
+                msg += diff + command_type.ljust(7) + " | " + name + "\n"
+            msg += "\n"
+
+        pages = pagify(msg, delims=["\n\n", "\n"])
+        pages = [box(page, lang="diff") for page in pages]
+        await menu(ctx, pages)
+
+    @slash.command(name="sync")
+    @commands.cooldown(1, 60)
+    async def slash_sync(self, ctx: commands.Context, guild: discord.Guild = None):
```
TrustyJAID, line 2284: It might be worth wrapping the content of this command in `ctx.typing()` so the user knows something is happening immediately and then return a message instead of just ticking the original message with a reaction. We could go so far as to provide the number of commands synced so the user has a direct reference to where they're at on the command limit instead of guessing.

TrustyJAID (changes requested):
Most things seem fine. I would like to see Danny's review addressed as well as a few things here I mentioned.

Drapersniper:
Another question here, should the slash control commands also be special cased like the licenseinfo commands to allow them to be run via mention? 

While I know red doesn't officially supports slash only/public bots this could be a convenient way to allow slash-only/public bots based on red to sync their commands without the message content intent.

Flame442:
`_remove_with_module` needs to be overridden in the subclass to get the same functionality it performs for the superclass with the added internal command containers. Technically it could be omitted, but then the cases where a module raises an exception in the loading process, but added commands to the tree would leave dangling objects, which isn't the best. It's understandable that it's internal, however that entire object has to meddle with internals in order to function, so I think it needs to stay.

I know Zeph was pointing out that the current `max_concurrency` solution can technically be bypassed by someone using `[p]slash enable` at the same time as `[p]slash enablecog`. With how fast `[p]slash enablecog` runs, I would say that doing that is far too much of an edge case to warrant additional logic to prevent that from happening. Yes, it technically could happen, but all that it would cause is a cog to be partially enabled, which is not even a huge deal since it doesn't actually _break_, just has some number of commands that would not be _enabled_, which could be viewed with `[p]slash list` to fix. I personally chalk that up to PEBCAK.

In terms of adding a way to make a command required-enabled in order to load a cog, I would not be comfortable adding that to `info.json`, partially because non-downloader code would need to be checking that value, and partially because of the fact Draper pointed out that it would not be applicable to local cogs. We are most likely going to need to subclass `Command` and override the respective `app_command` decorators (similar to our overrides to `ext.commands` objects) in order to implement #5976, and it would probably be best to add that as a kwarg to the decorators. The tree can easily read that attribute to skip the normal logic of enabling and disabling. For the same reasons that logic is not included in this PR, that functionality will be included in a followup PR. This PR is already large and hard enough to review as it is.

I have no opinion on special casing `[p]slash` to work with the mention prefix, so I will wait to add that until further discussion.
EDIT: I forgot that the command can still be used in DMs, as Kowlin pointed out. Since that's still an option, I think that there is no reason to add the override for this command, since it does not need to be _always_ available, it just needs to be available in general.

All other comments up to this point should be addressed at this point.

Kowlin:
> I have no opinion on special casing [p]slash to work with the mention prefix, so I will wait to add that until further discussion.

Theoretically it doesn't hurt making it an always available command, but at the same time. We're talking about a owner only command which is just fine being run in DMs, I expect a bot owner to know their own prefix and not rely solely on slash commands. So for me its a pass on making it always available.

TrustyJAID (approved):
This looks good to me!

Kowlin (approved):
Gonna go ahead and merge this. Since all the outstanding issues have been resolved.

PR #5998, opened 2023-03-18
Bump just Discord.py to 2.2.2

### LINKED ISSUE #5965: [Cleanup] after command takes a very long time

3.5.0.dev1 (at commit https://github.com/Cog-Creators/Red-DiscordBot/commit/f13d910f66dbb38cbf546666807f4eec660ae279)

### Cog name

Cleanup

### Command name

cleanup after

If the `after` message ID was close to the most recent message, then the messages should be deleted almost instantly (<5 seconds).

### What actually happened?

The messages were deleted after 5 minutes.

1. Choose a channel with an extremely long message history.
2. Choose a message close to the most recent message, and grab its ID.
3. Run `[p]cleanup after <ID>`.
4. Observe that it takes a long time before the messages are deleted. You may also notice HTTP 429 in the console, but this may be due to other cogs I have loaded.

### Anything else?

I am filing this for posterity. This should be addressed when there is a minor version bump for discord.py, namely including the fix for https://github.com/Rapptz/discord.py/pull/9093 .

Kowlin:
Discord.py 2.1.0 -> 2.2.2
Closes #5965

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -30,7 +30,7 @@ commonmark==0.9.1
     # via rich
 contextlib2==21.6.0
     # via schema
-discord-py==2.1.0
+discord-py==2.2.2
     # via
     #   -r base.in
     #   red-lavalink


Injabie3:
~~Disregard, I cannot read~~

TrustyJAID (approved):
Looks good to me.

PR #6001, opened 2023-03-21
Allow passing multiple cogs to `slash enablecog` and `slash disablecog`

karlsbjorn:
This PR allows for more than one cog to be passed as an argument to `slash enablecog` and `slash disablecog`, similarly to `load` and `reload`.

Yes


redbot/core/core_commands.py
```diff
@@ -2060,43 +2060,59 @@ async def slash_disable(
...
         **Arguments:**
-            - `<cog_name>` - The cog to enable commands from. This argument is case sensitive.
+            - `<cog_names>` - The cogs to enable commands from. This argument is case sensitive.
         """
         enabled_commands = await self.bot.list_enabled_app_commands()
         to_add_slash = []
         to_add_message = []
         to_add_user = []
 
+        successful_cogs = set()
```
Flame442, line 2101: This variable doesn't actually do anything because there is no way for an individual cog to be unsuccessful how you have it here. The command currently enables all cogs requested or none of the cogs requested, if any would cause it to fail.

It makes no special casing for if an *individual* cog has no commands or puts the tree over the command limit, only if *all* cogs requested have no commands or put the tree over the limit.

In order to allow the command to enable only a subset of cogs, while still either enabling or not enabling each individual cog, you would probably want to loop over a helper method of the current code for this method to do that, rather than trying to integrate logic for multiple cogs into each step at the same time. The existing code is supposed to be a single atomic step, which limits what can be done when integrating support for multiple cogs.

redbot/core/core_commands.py
```diff
@@ -2151,47 +2167,68 @@ async def slash_enablecog(self, ctx: commands.Context, cog_name: str):
...
-                name, guild_id, com_type = key
-                await self.bot.disable_app_command(name, discord.AppCommandType(com_type))
-                removed.append(name)
+            for cog_name in cog_names:
+                if self._is_submodule(cog_name, com.module):
+                    name, guild_id, com_type = key
+                    await self.bot.disable_app_command(name, discord.AppCommandType(com_type))
+                    removed.append(name)
+                    removed_cogs.add(cog_name)
         if not removed:
```
Flame442, line 2229: Since this one has nothing that would cause it to not be able to disable any particular command, there's no worries about this one being atomic. Still, this command currently does not properly display when an *individual* cog did not have any commands, only when *all* cogs did not have commands, due to this line. If you changed this to something like this, that would be sufficient for this particular command:
```py
        failed_cogs = set(cog_names) - removed_cogs
        if failed_cogs:
            # use failed_cogs instead of cog_names for processing & user output
            ...
        if not removed_cogs:
            return
```

Flame442 (approved):

PR #6005, opened 2023-03-27
Fix the handling of PartialMessageable channels

### LINKED ISSUE #5995: Ensure global slash commands don't break in DMs

3.5.0.dev

I was trying to confirm readiness of my bot instance to upgrade to Red 3.5 once it is released by using any of the global slash commands in DM to my V3/develop test bot instance, whether provided by my own cog `inatcog` from https://github.com/dronefly-garden/dronefly/tree/dpy2, currently in development to support Red 3.5, or any other third-party cog.

The Discord docs say that this should work, that global slash commands can be used in DM, so long as both you and the bot are both members of at least one server together, which is the case here.

### What actually happened?

Instead, I received an error, `The application did not respond`.

In fact, using Trusty's `weather` cog is not my goal, but provides a convenient example that I know fails. My own cog produces the same result, but development of that is in flux, so I'd rather not use that as an example. These instructions could be adapted to use any other cog, so long as it provides global slash commands:

1. Install `weather` from https://github.com/TrustyJAID/Trusty-cogs/tree/dpy-2.0
2. Use both message commands (e.g. `[p]weather halifax`) and slash commands (e.g. `/weather location halifax`) from the cog in a server channel to confirm proper operation. If the API key is not set yet, each command should respond with a message saying the api key needs to be set.
3. Try the same thing in a DM to the bot. Observe that the message command produces a response from the bot, but the global slash command does not.

### Anything else?

In discussion with Zephyrkul about the issue, we determined that this is what prevents the command from working: https://github.com/Cog-Creators/Red-DiscordBot/blob/2ab204438e2d3d45c9e29f6a78a32aa579a042bf/redbot/core/bot.py#L821-L824

While that might be expedient to not consider these messages as eligible, if that's the case, at the very least, I think they should not be published as available commands. That is, when a user presses `/` in a DM to the bot, they should not see them if they are known to not work. Otherwise, the bot owner and/or cog developer will be bothered by support requests when this happens. While cog developers can opt to publish their commands as guild only instead to avoid the issue, there's little recourse left to the bot owner unless they can fork the cog themselves to change them to guild-only.

I'm not attached to a particular outcome: I'd be happy with anything that doesn't unduly delay the release of Red 3.5. Here are four possibilities:

1. If global slash commands can be supported in DMs, that's outcome I'd prefer. 
    - Many users of my cog have found it difficult to remember the whole of my cog's expansive command set. Slash commands will help with discovery of the right command through command autocompletion.
2. If global slash commands can simply be dropped from the command tree because they're not easily supportable in time for Red 3.5, that would be an acceptable outcome.
3. Red could make commands guild-only always, and simply not support global slash commands.
4. Advise the cog developers preparing their cogs for Red 3.5 that global slash commands won't work yet, and leave it to them to make their cogs provide guild-only commands to avoid the issue.

Kowlin:
Closes #5995.

~~This PR adds in a new public function that works in a similar vain such as ``get_or_fetch_user/member``. Which looks for a channel in the cache and if it isn't found tries to retrieve it from Discord.~~
This is initially added to resolve the issues found in #5995, where a ``PartialChannel`` doesn't resolve permissions which we (I think) require for some command handling.

Yes*~Ish*

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -825,9 +825,8 @@ class Red(
         # We do not consider messages with PartialMessageable channel as eligible.
         # See `process_commands()` for our handling of it.
 
-        # 27-03-2023: Addendum, Fetching the channel here is gonna cost us some API calls.
-        # But so far seems like the better solution.
-        # On hybrid commands this may in a very rare scenario cause a timeout. But that's assuming the worst
+        # 27-03-2023: Addendum, DMs won't run into the same issues that guild partials will,
+        # so we can safely continue to execute the command.
         if (
             isinstance(channel, discord.PartialMessageable)
             and channel.type is not discord.ChannelType.private
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -824,8 +824,15 @@ class Red(
 
         # We do not consider messages with PartialMessageable channel as eligible.
         # See `process_commands()` for our handling of it.
+
+        # 27-03-2023: Addendum, Fetching the channel here is gonna cost us some API calls.
+        # But so far seems like the better solution.
+        # On hybrid commands this may in a very rare scenario cause a timeout. But that's assuming the worst
         if isinstance(channel, discord.PartialMessageable):
-            return False
+            try:
+                channel = await self.get_or_fetch_channel(channel.id)
+            except discord.DiscordException:
+                return False
 
         if guild:
             assert isinstance(channel, (discord.TextChannel, discord.VoiceChannel, discord.Thread))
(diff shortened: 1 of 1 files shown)


redbot/core/bot.py
```diff
@@ -824,7 +824,14 @@ async def message_eligible_as_command(self, message: discord.Message) -> bool:
 
         # We do not consider messages with PartialMessageable channel as eligible.
         # See `process_commands()` for our handling of it.
-        if isinstance(channel, discord.PartialMessageable):
+
+        # 27-03-2023: Addendum, Fetching the channel here is gonna cost us some API calls.
+        # But so far seems like the better solution.
+        # On hybrid commands this may in a very rare scenario cause a timeout. But that's assuming the worst
```
Flame442:
```suggestion
        # 27-03-2023: Addendum, DMs won't run into the same issues that guild partials will,
        # so we can safely continue to execute the command.
```

Flame442 (approved):
All set!

PR #6006, opened 2023-03-27
Add redbot.core.app_commands namespace

Jackenmen:

Nothing special in there right now, just future-proofing by allowing ourselves to override stuff in there and not require users to change their imports (or add another one) in the future. Cog creators should be using this instead of `discord.app_commands`. It is expected that the existing usage of `discord.app_commands` will be updated before the final release is available.

Additionally, I updated the existing test for `redbot.core.commands` to no longer ignore hybrid commands now that they're implemented.

Yes


Flame442 (approved):
Thanks!

PR #6010, opened 2023-04-02
Subclass app_command commands and allow force enabling app commands

Flame442:
Subclasses `Command`, `Group`, `HybridCommand`, `HybridGroup`, and `GroupCog`, and replaces associated decorators, in order to make it easier to add extra attributes to them. Also adds a kwarg `red_force_enable` to these objects and their decorators, which disables the default behavior of `RedTree` of not adding commands immediately to the tree and requiring `[p]slash enable` for them to be enabled. As such, commands created with this flag enabled will attempt to add themselves to the tree at cog load, which will prevent the cog from being loaded if the command limit is reached, rather than allowing the cog to load but preventing the commands from being added.

All changes relating to `__new__` and metaclasses are necessary complexity due to how the duel purpose `GroupCog` is implements passing kwargs to the `Group` object it represents. The current code adds the ability to use `class MyCog(GroupCog, group_red_force_enable=True)` to set this flag. If it would be preferable, most of this complexity can be removed, however it will then be impossible to create a `GroupCog` which uses Red's `Group` subclass. The alternative method of creating a `Group` by creating a `Group` object and using `bot.tree.add_command(group)` would still work.

When testing this PR, the main thing to test is that all methods of creating app commands still function as expected with no (internal) errors, and that the command properly bypasses the enabling logic when the flag is set.

Yes


Flame442:
The code changes in this PR have been completed, so I marked it as ready for review. There's probably some docs changes that need to get done as well before this can be merged, but I have not yet started on those.

Flame442:
Superseded by #6018.

PR #6013, opened 2023-04-11
Prohibit zero-length prefixes in `[p]set [server]prefix`

japandotorg:

Disallow empty prefixes being a thing.
For example:
image

yes
image


Jackenmen:
Seems like this should also be done to `[p]set prefix`?

japandotorg:
> Seems like this should also be done to `[p]set prefix`?

yeah, totally forgot about that command

Kreusada (approved):
LGTM, it works well. Thanks for implementing this.

PR #6015, opened 2023-04-13
Add global checks to app commands

Flame442:
Adds checks for `[p]ignore`, `[p]allowlist`, and `[p]blocklist` to app commands.

Does *not* add checks for whether or not the bot has `send_messages` (or similar) permissions in the channel, since interacting via the `interaction` object does not require those permissions. These checks should still apply for hybrid commands, since those use the `commands` system of checks, so CCs coding specifically `app_command.command` commands should be able to account for this where necessary.

I opted to make the fail cases send an ephemeral message, as far as I know responding to an interaction does not count against the bot's rate limits, and it's just generally good practice to always respond to interactions if the bot is able to.

`ignored_channel_or_guild` was modified to accept an `Interaction` in addition to the `Message` and `Context` it already supported. It also has a (technically breaking) change to accept `PartialMessageable` objects if their `.type` is `discord.ChannelType.private`. It uses the same logic as for `PrivateChannel`s, allowing app commands used in DMs to not be caught by this filter.

Yes

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -847,7 +847,7 @@ class Red(
         return True
 
     async def ignored_channel_or_guild(
-        self, ctx: Union[commands.Context, discord.Message]
+        self, ctx: Union[commands.Context, discord.Message, discord.Interaction]
     ) -> bool:
         """
         This checks if the bot is meant to be ignoring commands in a channel or guild,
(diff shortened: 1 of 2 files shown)


TrustyJAID (approved):
This looks good to me.

PR #6018, opened 2023-04-14
Allow force enabling app commands using flag in extras

Flame442:
Alternative to #6010

Unlike that PR, this PR does not subclass any of the `app_commands` objects to add kwargs to their constructors. Instead, it relies on the same data being passed via `Command.extras`, serving the same purpose once it is there. A list of differences in the two options are listed below.

| | #6010 | This PR |
| --- | --- | --- |
| Enabling a command | `@app_commands.command(`<br />`  red_force_enable=True`<br />`)` | `@app_commands.command(`<br />`  extras={"red_force_enable": True}`<br />`)` |
| Accessing the setting | `command.red_force_enable` | `command.extras.get(`<br />`  "red_force_enable",`<br />`  False`<br />`)` |
| Avoids subclassing? | No | Yes |

Still not certain how to approach the UX on `[p]slash disable`ing a force enabled cog, or displaying force enabled cogs in `[p]slash list`. The UX is, however, implemented the same on both this PR and #6010 from the perspective of a user.

Yes

### DIFF
--- a/docs/cog_guides/core.rst
+++ b/docs/cog_guides/core.rst
@@ -4168,7 +4168,14 @@ slash list
 **Description**
 
 List the slash commands the bot can see, and whether or not they are enabled.
+
 This command shows the state that will be changed to when ``[p]slash sync`` is run.
+Commands from the same cog are grouped, with the cog name as the header.
+
+The prefix denotes the state of the command:
+- Commands starting with ``- `` have not yet been enabled.
+- Commands starting with ``+ `` have been manually enabled.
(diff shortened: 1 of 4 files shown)


TrustyJAID:
I think we should at least explain to the user attempting to disable a force enabled commands that it is force enabled and the command will not be removed. Cog creators have some requirements on this and we shouldn't restrict them too much on this regard. We can handle the administration side by requiring approved cog creators that choose to utilize force enabled commands to explain it in the cog install string.

TrustyJAID (approved):
This is looking good to me.

PR #6021, opened 2023-04-15
Privatize APIs by renaming or removing them from `__all__`

Jackenmen:

- write down new version guarantees
- add `__all__` to public modules that are missing it
- add missing items in modules that already have `__all__`
- remove names that aren't supposed to be public from `__all__`:
    - removed `create_temp_config()`, `load_basic_configuration()`, `core_data_path()` from `redbot.core.data_manager.__all__`
    - removed `set_locale()`, `reload_locales()` from `redbot.core.i18n.__all__`
    - removed `_update_event_loop_policy()` and `MIN_PYTHON_VERSION` from `redbot.__all__`
- keep `redbot.core.data_manager.storage_details()` as public API but update it to deep copy underlying storage details dict
- (can be reverted) add `data_path()` and `metadata_file()` functions to `redbot.core.data_manager` module
- (can be reverted) remove long deprecated `redbot.core.data_manager.load_bundled_data()`
- privatize modules in `redbot.core` by adding `_` prefix to them:
    - `cog_manager`
    - `settings_caches`
    - `global_checks`
    - `events`
    - `cli`
    - `rpc`
    - `drivers`
- privatize `Config.driver`
- add notes about unsupported instantiation to `redbot.core.config` and `redbot.core.modlog` classes

Yes


redbot/core/errors.py
```diff
@@ -7,6 +7,18 @@
...
 
+__all__ = (
+    "RedError",
+    "PackageAlreadyLoaded",
+    "CogLoadError",
+    "BankError",
+    "BalanceTooHigh",
+    "BankPruneError",
+    "ConfigError",
+    "StoredTypeError",
```
Flame442, line 18: `MissingExtraRequirements`?

redbot/core/generic_casetypes.py
```diff
@@ -3,6 +3,25 @@
...
+    "voiceunban",
+    "voicemute",
+    "channelmute",
+    "servermute",
+    "voiceunmute",
+    "channelunmute",
+    "serverunmute",
+    "voicekick",
+    "warning",
+)
```
Flame442: `all_generics`?

redbot/core/data_manager.py
```diff
@@ -9,21 +9,18 @@
...
 from . import commands
-from .cli import ExitCodes
+from ._cli import ExitCodes
 
-__all__ = [
-    "create_temp_config",
-    "load_basic_configuration",
+__all__ = (
     "cog_data_path",
-    "core_data_path",
```
Flame442, line 21: Not certain if this should be private. At least one existing 3rd party cog uses it (Toby's `errorlogs`), and I'm not sure of the benefits and drawbacks of keeping it public.

Flame442 (dismissed):
This is all set as is, but I've listed a few things that might make sense to add/remove.

redbot/core/errors.py
Jackenmen: Purposefully omitted, I want to move it to `redbot.core._drivers` at a later time.

redbot/core/generic_casetypes.py
Jackenmen: Good idea.

redbot/core/data_manager.py
Jackenmen: The idea behind it was that 3rd-party cogs should be using `cog_data_path()` so there should be no reason to use `core_data_path()` but the use case of `core_data_path() / "logs"` is not a bad one. Of course there's the `data_path() / "core/logs"` but I don't see a reason to have API users jump through these hoops. Gonna revert this removal.

Flame442 (approved):

PR #6024, opened 2023-04-15
Split public and private config/driver APIs

Jackenmen:

Split out `get_latest_confs()` and `migrate()` from `redbot.core.config` to `redbot.core._config` module as they don't really have a sensible use outside our own code

Yes


redbot/setup.py
```diff
@@ -20,10 +20,16 @@
     create_backup as red_create_backup,
     cli_level_to_log_level,
 )
-from redbot.core import config, data_manager, _drivers
-from redbot.core._cli import ExitCodes
+from redbot.core import config, data_manager
+from redbot.core._config import migrate
+from redbot.core.cli import ExitCodes
```
Flame442:
```suggestion
from redbot.core._cli import ExitCodes
```
ref #6021

Flame442 (changes requested):

Flame442 (approved):
cats

PR #6033, opened 2023-04-19
[Audio] Update Lavalink.jar build

aikaterna:

This changes audio's manager to download and use version 3.7.5 of our Lavalink.jar.

No

Lavalink changelog for us:
```
Fixed Twitch playback.
Plain word search queries should work again.
```

### DIFF
--- a/redbot/cogs/audio/manager.py
+++ b/redbot/cogs/audio/manager.py
@@ -235,7 +235,7 @@ class LavalinkVersion:
 
 
 class ServerManager:
-    JAR_VERSION: Final[str] = LavalinkVersion(3, 7, 5):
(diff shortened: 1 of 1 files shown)


Drapersniper:
@aikaterna there will be a 3.7.6 with the m1 fixes topi is current working on that so may be an idea to hold as late as possible to merge this if 3.7.6 is released before 3.5 is release then we can update to it 

https://github.com/freyacodes/Lavalink/pull/865

aikaterna:
I'm helping Topi with that. I don't feel it will be ready in time for what we need.

Jackenmen (approved):
Thanks for working on this! Looks good.

PR #6079, opened 2023-04-19
Use the contributor list URL instead of the org in [p]info

aikaterna:

I changed the info command to use the full contributor list for "improved by many" instead of linking to the org. I feel like we should be acknowledging everyone that has contributed any kind of work on the bot and not just the org.

Yes, but would appreciate someone taking a look before merging.


Kowlin (dismissed):
Looks good

Kowlin (approved):
Still looking good. But with style now!

PR #6098, opened 2023-04-22
Bump platformdirs to 3.x and remove the migration code for 2.x

Jackenmen:

This code never existed in stable Red (3.4.x) and was only necessary because platformdirs 2.0 used a different path for `user_config_dir` than appdirs 1.x did. Apparently they decided that the previous way was actually the correct one so this being here will actively hurt us if we release it since we'll then have to deal with the migration when we inevitably will have to upgrade to platformdirs 3.x or newer.

See: https://github.com/platformdirs/platformdirs/compare/2.6.0...3.2.0#diff-c62b68d21c20415287fcac2ff0dd4a8e67ec181ee45fea497215143a885e4513

No

### DIFF
--- a/redbot/core/data_manager.py
+++ b/redbot/core/data_manager.py
@@ -51,14 +51,6 @@ if _system_user:
         config_dir = appdir.site_data_path
 
 config_file = config_dir / "config.json"
-if not config_file.exists() and sys.platform == "darwin":
-    # backwards compatibility with the location given by appdirs 1.4.4 (replaced by platformdirs 2)
-    # which was the same as user_data_path
-    # https://platformdirs.readthedocs.io/en/stable/changelog.html#platformdirs-2-0-0
-    _old_config_location = appdir.user_data_path / "config.json"
-    if _old_config_location.exists():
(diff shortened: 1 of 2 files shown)


Jackenmen:
The expected behavior is that this will work as nothing ever happened between 3.4 and 3.5.

If a macOS user had `3.5.0.dev301+g519acedf` or newer installed, the instances they created/removed since they installed that version would not be visible anymore until they manually readd them (or manually copy the file to the proper location). It shouldn't really matter to us since `V3/develop` is a development version and such things, even if rare, can be expected.

aikaterna:
Tested on an M1 Mac (Big Sur 11.7.2 / 3.5.0.dev353+ge1f5d65d test instance) and an Intel Mac (Catalina 10.15.7 / 3.5.0.dev360+gd8d4b4f1 test instance). Things behaved as expected where instance names made in 3.4.x were available with this branch installed, but instances made on the mentioned 3.5 dev versions were not visible.

aikaterna (approved):

PR #6101, opened 2023-04-25
Fix visual issues with Discord's new Markdown 

Vexed01:

Discord has started rolling out enhanced Markdown to some smaller guilds, including my testing server. I have noticed that many of the help strings no longer format properly when Discord tries to turn them into a list. The issue is where lists are indented beyond the text preceding the list. This PR primarily fixes help strings but a few numbered lists are also adjusted to ensure they format properly.

Current code with an extra indent with the list than the line preceding it:
https://github.com/Cog-Creators/Red-DiscordBot/blob/ec55622418810731e1ee2ede1569f81f9bddeeec/redbot/core/core_commands.py#L3105-L3111

Current formatting in Discord:
DiscordCanary_bDzUA3GKSI

This PR's format in Discord without extra indents:
DiscordCanary_S6R631cndO

~~I understand not everyone currently has access to the new formatting so if you want a screenshot just ping me, or someone else in testing might be able to who has a server with it or the experiment enabled.~~

~~Not sure on an ETA for rollout but for now best to be cautious and not include anything which would look bad on clients without the new formatting (eg # headings will be supported). Maybe if this PR waits for clients to fully support it (would it be able to fit in 3.5.0 now it's been frozen? Is it only feature frozen?) then I will look into user # headings in the help menu instead of bold. But in its current state nothing will look out of place to users without the new Markdown styling.~~

~~I can confirm in the guild that has it I can see the new formatting on mobile (Android) so mobile isn't being left behind like sometimes in the past.~~

Discord they have rolled this out to every guild, so it might be worth looking at adding some # headings but maybe in another PR to _improve_ visuals while this one _fixes_ visual issues.

No, for help strings there are so many that I have checked formatting works by using the help command on a subset, and I have manually looked at every edited line (not a blind find and replace).
Yes, all numbered lists have been verified with the bot in Discord by using the associated command.


redbot/core/core_commands.py
```diff
@@ -1117,17 +1117,18 @@ async def embedset(self, ctx: commands.Context):
...
+          1. Channel override - `[p]embedset channel`
+          2. Server command override - `[p]embedset command server`
+          3. Server override - `[p]embedset server`
+          4. Global command override - `[p]embedset command global`
+          5. Global setting  -`[p]embedset global`
+
+        - In DM context:
+          1. User override - `[p]embedset user`
+          2. Global command override - `[p]embedset command global`
+          3. Global setting - `[p]embedset global`
```
Vexed01, line 1131: This currently formats as a lettered list. There is the chance that in the future it will change to a numbered list so I'm leaving it as numbers.

image

redbot/cogs/mod/kickban.py
```diff
@@ -889,9 +889,10 @@ async def unban(
         """Unban a user from this server.
 
         Requires specifying the target user's ID. To find this, you may either:
-         1. Copy it from the mod log case (if one was created), or
-         2. enable developer mode, go to Bans in this server's settings, right-
-        click the user and select 'Copy ID'."""
+        1. Copy it from the mod log case (if one was created), or
+        2. enable developer mode, go to Bans in this server's settings, right-
+          click the user and select 'Copy ID'.
```
Vexed01: Pre-emptive comment - there need to be two spaces here (L894) to ensure the text is indented to the same level as the list. Here Discord treats four spaces as a double indent which does not look good. Four would fit in better with the indentation for the actual code but then Discord for some strange reason assumes the first three spaces mean a single indent and the fourth space is shown as an actual space:
image

Kowlin (dismissed):

Jackenmen (approved):
I can't test whether this renders properly with the Markdown so I'll have to trust you on that but I did verify that it doesn't alter how it looks with the old Markdown. Thanks!

PR #6102, opened 2023-04-27
Add `header`, `hyperlink` and `subtext` utilities

Kreusada:

Adds new `header` and `hyperlink` utilities, showcased [here](https://support.discord.com/hc/en-us/articles/210298617-Markdown-Text-101-Chat-Formatting-Bold-Italic-Underline-).

Their implementation could be up for discussion if anyone has alternative ideas. I could also add a bullet point util to this PR, but I held back due to not knowing how to best implement embedded bullets.

Yes


Kowlin:
The chance is big that Discord.py will add these to their library, however I'm not opposed to merging this.

Jackenmen:
> The chance is big that Discord.py will add these to their library

That's unlikely, discord.py doesn't have utilities for *constructing* Markdown since most of the time it's trivial to just do it yourself, often shorter. Our whole chat_formatting module is proof of d.py not having such utilities.

Kowlin:
Probably should be updated to also include Subtext 😅

Kreusada:
> Probably should be updated to also include Subtext 😅

Good idea. I've just updated the PR to include this 👍

Kowlin (approved):

PR #6103, opened 2023-04-27
Fix AttributeError from slash command tree

Vexed01:
Fix an AttributeError when adding a slash command that already exists.

Edits a few lines to use the imported CommandAlreadyRegistered exception instead of trying to use the non-existant one in discord.errors.

image

Yes


Jackenmen (approved):
Thanks!

PR #6104, opened 2023-04-27
Fix help breaking when removing a command without removing its cog

Flame442:
#5354 introduced a bug in cogs which dynamically override commands from existing cogs. If a command is removed from the bot, but the cog is not removed, the updated logic would still attempt to display it in help. This would cause help to silently fail or to error, depending on OS. This PR reverts that change, as there is no good way to filter from the cog object itself while accounting for differences from the bot's command list.

Yes

### DIFF
--- a/redbot/core/commands/help.py
+++ b/redbot/core/commands/help.py
@@ -274,10 +274,7 @@ class RedHelpFormatter(HelpFormatterABC):
     async def get_cog_help_mapping(
         self, ctx: Context, obj: commands.Cog, help_settings: HelpSettings
     ):
-        if obj is None:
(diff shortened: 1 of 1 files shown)


aikaterna (approved):
Works for me, tested on a current dev bot vs this PR.

Repro steps for both that I used on a bot with no custom cogs:
Test that help sends
Eval out a bot.remove_command for the info command
Use help again

japandotorg:
Can confirm that this fixes the “help silently failing” issue

Jackenmen (approved):
It's just a revert so looks good to me :)

PR #6020, opened 2023-04-15
Move private things in modlog and `redbot.core.errors`

Jackenmen:
### Description of the changes

Part of the remainder from #6021 - moving things around in `redbot.core.modlog` and `redbot.core.errors`.

Changes are described in the commit titles.

### Have the changes in this PR been tested?

Yes

PR #5600, opened 2022-03-06
Migration to discord.py 2.0

Jackenmen:
### Description of the changes

This is the work on migrating Red to discord.py 2.0.

If you want to test our work on migration to discord.py 2.0, you can test these changes by installing this PR's branch in a virtual environment with the command:
```
python -m pip install -U --force-reinstall git+https://github.com/Cog-Creators/Red-DiscordBot@V3/dpy2_unstable#egg=Red-DiscordBot
```
Be sure to report any issues you run into, this is a huge changeset so it's possible that I missed something.

Things that still need doing (but might be done in separate PR later):
- Quite a bit of user-facing docstrings refer to the current channel
  when they also mean thread (should we fix this?).
  - Developer-facing docs do too but I don't think that's an issue.
- Perhaps something we missed?

### Have the changes in this PR been tested?

No (Partially yes, but I have not tested all the changes that have been made here)

### Information for reviewers

The code review is best done by going through individual commits (there's a nice next button when you are in a commit view on a PR so it's quite convenient) as they are well-scoped.

As for functional review... You have to test on the whole changeset but you can still leverage the commit separation so that you test a single thing at the time though if you think it's going to be easier to just go through the whole files changed list, that seems fine to me too.

PR #6113, opened 2023-05-03
Remove unused ConversionFailure

Jackenmen:

I've apparently accidentally readded it to `__all__` in #6020? Weirdly enough, I didn't remove it in #5600 either... Us no longer wrapping `BadArgument` in `ConversionFailure` is already mentioned in the "Backwards incompatible changes in Red 3.5" document so there's no need to alter it?

Let me know if you think that it's fine to get this into 3.5 or if we should go with a standard deprecation warning on usage and remove it in 3.6, even though it's not used by anything at this point.

No

### DIFF
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -43,7 +43,6 @@ from discord.ext.commands import (
     Greedy,
 )
 
-from .errors import ConversionFailure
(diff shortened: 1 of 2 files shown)


Flame442 (approved):
It isn't *that* much more breaking, since while it was importable it would not have actually functioned for type checks, so I think it's fine.

PR #6116, opened 2023-05-03
Show serverlocked setting inside `[p]set showsettings`

Kreusada:

Currently, there is no proper way to see whether the bot has been serverlocked or not. This PR adds a line which is displayed inside `[p]set showsettings`, to show whether the bot is serverlocked or not. It first checks whether the Admin cog is loaded (if not, then serverlock cannot function, so it will show as False), and then it accesses the cogs config to see whether serverlock is enabled.

image

Yes

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3842,7 +3842,7 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         locale = global_data["locale"]
         regional_format = global_data["regional_format"] or locale
         colour = discord.Colour(global_data["color"])
-        
+
         admin_cog = self.bot.get_cog("Admin")
         serverlocked = await admin_cog.config.serverlocked() if admin_cog else False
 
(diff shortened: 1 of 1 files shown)


Jackenmen:
Why is this not done in the Admin cog?

Kreusada:
> Why is this not done in the Admin cog?

I suppose because there wasn't an existing showsettings command for Admin, and that the serverlock setting affects the `[p]invite` command which is part of Core. Did you have an idea in how to implement this better?

Flame442:
The invite command is only really a helper for generating invite URLs. Disabling it doesn't strictly _prevent_ people from inviting the bot, since the URL can be easily generated manually using only the bot ID. `[p]invite` does interface with the serverlock setting, but mostly just as a warning to the person using it that it isn't going to actually work without disabling the serverlock.

Admin doesn't otherwise have a huge need for a showsettings command, and the only ways to know the setting currently are to toggle/set it or know that `[p]invite` will warn you about it, so I don't think it would hurt to add some way to check it. The core settings page isn't the worst place to put it, though it is a little confusing to make it the only (afaik) command in `set showsettings` that isn't modified in `set` and that relies on another cog being loaded to be configured.

Kreusada:
After giving this some thought, I think this change may cause confusion and isn't really necessary.

PR #6130, opened 2023-05-05
d.py 2.3 / pomelo changes

Jackenmen:

Let's open this early to have some time to discuss this.

List of changes:
- Updated global unique user finder in Audio to match by global name as well
- Replaced manual construction of user handle in Audio and Modlog with `str(author)`
- Updated audit log reasons in Admin to use `str(author)` rather than `author.name`
- Updated logs in Cleanup, Mod, and Trivia to use user handles
- Updated usage of `User/Member.name` with `display_name` attribute in places where we're supposed to show the display name (`[p]serverinfo`, `[p]leaderboard`, `[p]bankset reset`, `[p]set showsettings`, Red's CLI splash screen, )
- Updated name tracking in Mod to track both the usernames and display names
- Updated last known username tracking in Modlog
- Altered `Red.is_automod_immune()` to consider *all* bots immune

TODO:
- Figure out if there's going to be an endpoint for changing the app/global name and add `[p]set bot globalname` if so

Discussion questions:
- I was forced to update the handling of `User` objects in `Red.is_automod_immune()` - it is no longer possible for me to differ between webhook users and bots (it would be if we did not accept `discord.abc.User` directly - I need a message object to know if something was sent a webhook) and as such I decided to consider all bots to be immune to auto-moderation. Is this fine?

No

### DIFF
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -273,7 +273,7 @@ class Admin(commands.Cog):
         """
         author = ctx.author
         reason = _("{author} ({author.id}) changed the colour of role '{role.name}'").format(
-            author, author.id, role.name
+            author=author, role=role.name
         )
 
         if not self.pass_user_hierarchy_check(ctx, role):
@@ -307,7 +307,7 @@ class Admin(commands.Cog):
         old_name = role.name
         reason = _(
             "{author} ({author.id}) changed the name of role '{old_name}' to '{name}'"
-        ).format(author, author.id, old_name, name)
+        ).format(author=author, old_name=old_name, name=name)
 
         if not self.pass_user_hierarchy_check(ctx, role):
             await ctx.send(_(ROLE_USER_HIERARCHY_ISSUE).format(role=role))
--- a/redbot/cogs/admin/admin.py
+++ b/redbot/cogs/admin/admin.py
@@ -272,7 +272,9 @@ class Admin(commands.Cog):
             `[p]editrole colour Test #ff9900`
         """
         author = ctx.author
-        reason = "{}({}) changed the colour of role '{}'".format(author.name, author.id, role.name)
+        reason = _("{author} ({author.id}) changed the colour of role '{role.name}'").format(
+            author, author.id, role.name
+        )
 
         if not self.pass_user_hierarchy_check(ctx, role):
             await ctx.send(_(ROLE_USER_HIERARCHY_ISSUE).format(role=role))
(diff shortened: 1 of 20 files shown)


Jackenmen:
I updated the description with a list of changes and todo/discussion items.

redbot/cogs/admin/admin.py
```diff
@@ -272,7 +272,9 @@ async def editrole_colour(
             `[p]editrole colour Test #ff9900`
         """
         author = ctx.author
-        reason = "{}({}) changed the colour of role '{}'".format(author.name, author.id, role.name)
+        reason = _("{author} ({author.id}) changed the colour of role '{role.name}'").format(
```
Flame442, line 275: I'm pretty sure this (and similar) will error, because having named fillins requires kwargs in the format call.

Jackenmen: Of course, this is fixed now, thanks!

Jackenmen:
I've updated this PR with most of the things that were previously listed as TODOs. We're now left with potentially having to add `[p]set bot globalname` *if* Discord adds an endpoint for it and figuring out whether the new behavior of `Red.is_automod_immune()` is acceptable. Of course, any feedback about the other changes is welcome too.

redbot/cogs/mod/events.py
```diff
@@ -158,21 +159,31 @@ async def on_message(self, message):
...
-                    name_list.remove(before.name)
-                name_list.append(before.name)
-                while len(name_list) > 20:
-                    name_list.pop(0)
+                self._update_past_names(before.name, name_list)
+        if before.display_name != after.display_name:
+            track_all_names = await self.config.track_all_names()
+            if not track_all_names:
+                return
+            async with self.config.user(before).past_display_names() as name_list:
```
Flame442, line 185:
```py
  File "venv\lib\site-packages\discord\client.py", line 441, in _run_event
    await coro(*args, **kwargs)
  File "venv\lib\site-packages\redbot\cogs\mod\events.py", line 185, in on_user_update
    async with self.config.user(before).past_display_names() as name_list:
  File "venv\lib\site-packages\redbot\core\config.py", line 363, in __getattr__
    raise AttributeError("'{}' is not a valid registered Group or value.".format(item))
AttributeError: 'past_display_names' is not a valid registered Group or value.
```
First thing in console when loading this on a bot. This key needs to be added to the defaults.

Flame442 (changes requested):

redbot/cogs/mod/names.py
```diff
@@ -20,23 +21,23 @@ class ModInfo(MixinMeta):
...
-        names = await self.config.user(user).past_names()
-        nicks = await self.config.member(user).past_nicks()
-        if names:
-            names = [escape_spoilers_and_mass_mentions(name) for name in names if name]
-        if nicks:
-            nicks = [escape_spoilers_and_mass_mentions(nick) for nick in nicks if nick]
-        return names, nicks
+    async def get_names(self, member: discord.Member) -> Tuple[List[str], List[str], List[str]]:
+        user_data = await self.config.user(member).all()
+        usernames, display_names = user_data["past_usernames"], user_data["past_display_names"]
```
Flame442:
```suggestion
        usernames, display_names = user_data["past_names"], user_data["past_display_names"]
```

Flame442 (changes requested):
Seems like everything else works besides this typo.

Flame442 (approved):

PR #6142, opened 2023-05-11
Fix audio menus not respecting choices

### LINKED ISSUE #6136: [Audio] search only picks the 1st option when you have buttons enabled

3.5.1

### Cog name

Audio

### Command name

search

It's supposed to pick the `subsequent` song from the search results **according to** your interaction with buttons. 
I.e.
If you choose to pick 5 (let's say) and press on the following no. 5 button, it should pick the 5th number song, from the search results and not the 1st one.

### What actually happened?

So it only picks the 1st option i.e. 1st song from the search results **even if** you selected the 5th one, no matter which page you're on. Same result.

1. `<p>set usebuttons` (make the buttons on)
2. Join VC & `<p>search <queries>` (any queries)
3. Try to select the results __other than__ the 1st option.
4. You will see it only picks the 1st option even tho you've selected something else other than the 1st option.

Now, if you use the reactions instead, 
> `<p>set usebuttons` (make the buttons off)
> Then run the cmds in the same sequence, it work as how you expect it to be.

### Anything else?

Nope

TrustyJAID:

Resolves #6136

Yes

### DIFF
--- a/redbot/cogs/audio/core/utilities/formatting.py
+++ b/redbot/cogs/audio/core/utilities/formatting.py
@@ -117,6 +117,7 @@ class FormattingUtilities(MixinMeta, metaclass=CompositeMetaClass):
             )
         if not await self.maybe_charge_requester(ctx, guild_data["jukebox_price"]):
             return
+        emoji = str(emoji)
         try:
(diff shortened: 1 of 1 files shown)


TrustyJAID:
Closing in favor of #6143

PR #6143, opened 2023-05-11
Fix menus passing PartialEmoji instead of str

### LINKED ISSUE #6136: [Audio] search only picks the 1st option when you have buttons enabled

TrustyJAID:

Fixes #6136

Yes

### DIFF
--- a/redbot/core/utils/menus.py
+++ b/redbot/core/utils/menus.py
@@ -45,7 +45,7 @@ class _GenericButton(discord.ui.Button):
         message = self.view.message
         page = self.view.current_page
         timeout = self.view.timeout
-        emoji = str(self.emoji)
+        emoji = str(self.emoji) if self.emoji.is_unicode_emoji() else self.emoji
         try:
             await self.func(ctx, pages, controls, message, page, timeout, emoji)
         except Exception:
--- a/redbot/core/utils/menus.py
+++ b/redbot/core/utils/menus.py
@@ -45,7 +45,11 @@ class _GenericButton(discord.ui.Button):
         message = self.view.message
         page = self.view.current_page
         timeout = self.view.timeout
-        emoji = str(self.emoji) if self.emoji.is_unicode_emoji() else self.emoji
+        emoji = (
+            str(self.emoji)
+            if self.emoji.is_unicode_emoji()
+            else (ctx.bot.get_emoji(self.emoji.id) or self.emoji)
+        )
         try:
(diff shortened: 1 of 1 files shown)


Jackenmen (changes requested):
Button-based `menu()` doesn't actually seem to work with `discord.Emoji`, seems like `menu()`'s code for buttons needs to `str(emoji)` before passing it to `_GenericButton`/`discord.PartialEmoji.from_str()`.

In case it helps, this is what I use to test this in full:
```py
from redbot.core.utils.menus import menu, close_menu


async def func(
    ctx: commands.Context,
    pages: list,
    controls,
    message: discord.Message,
    page: int,
    timeout: float,
    emoji: str,
):
    await ctx.send(repr(emoji))
    await menu(ctx, pages, controls, message=message, page=page, timeout=timeout)

it = iter(bot.guilds)
controls = {next(it).emojis[0]: func, next(it).emojis[0]: func, "\N{LEFTWARDS BLACK ARROW}\N{VARIATION SELECTOR-16}": func, "\N{CROSS MARK}": close_menu}

await menu(ctx, ["1", "2", "3"], controls, timeout=None)
```

Jackenmen (changes requested):
The change you've made is not enough - `discord.PartialEmoji.from_str()` is still called on potential `discord.Emoji`s in `menu()`, see lines 158 and 162

Jackenmen:
(if we're to allow non-str inputs to `_GenericButton.__init__`, you should also update the type hint there)

Jackenmen (approved):
Works as expected, thank you!

PR #6150, opened 2023-05-12
Fix max heapsize calculation on 32-bit platforms

### LINKED ISSUE #6137: [Audio] Adjust max heapsize for 32 bit devices

3.5.1

Help Raspberry Pi 3b+ users in support with their bot managing and launching the Lavalink.jar.

I expected the bot to launch, download the jar if needed, start the jar, and then have Audio connect to it.

### What actually happened?

The Lavalink.jar seemed to lock up before fully completing the startup or in some cases, lock up when a connection from Red to Lavalink was initiated. Logs were incomplete where I expected to see Red connection lines or standard Lavalink log lines.

1. Have a 32 bit OS/device with a smaller amount of RAM, like Raspbian 11 on a Pi 3b+ (with 1GB RAM, which was the case for both Red users that reported this in support)
2. Install Red
3. Start the bot and load audio
4. The jar will seemingly not be able to get to a point where Red can connect to it. Running the jar as an externally managed process without an Xmx flag, or where Xmx is equal to 512mb or less, will let it launch and have Red connect (have not tested further yet, this comment is based on what I saw in Red support)

The proposed fix is:

<https://github.com/Cog-Creators/Red-DiscordBot/blob/4daf81aa5b8d9140ae2295548d22566b07f64a71/redbot/cogs/audio/utils.py#L30>

`min(4 * 1024**3, psutil.virtual_memory().total)` may work for this

https://github.com/Cog-Creators/Red-DiscordBot/blob/4daf81aa5b8d9140ae2295548d22566b07f64a71/redbot/cogs/audio/utils.py#L39

```
    _min_max = 512 * 1024**2
    max_ram_allowed = max(max_allocation * 0.5, _min_max)
```

I am not PRing this change myself at this point as maybe someone would like to work on this before I have time to do so. I have a Pi 3b+ in the configuration needed to reproduce this, so if there is further testing needed on a PR please let me know and I can help.

### Anything else?

See https://discord.com/channels/133049272517001216/387398816317440000/1104968352625078313 in Red's support for proof that it reserves 4GB on the current version of Red

See https://discord.com/channels/133049272517001216/387398816317440000/1105198290452086814 for a second user where setting the Xmx smaller helped

Jackenmen:

This fixes:
- `get_max_allocation_size()` potentially returning a number that is bigger than total available RAM on 32-bit platforms
- the max ram value returned by `get_jar_ram_defaults()` being more than half of the available memory on 32-bit platforms

Fixes #6137

No

### DIFF
--- a/redbot/cogs/audio/utils.py
+++ b/redbot/cogs/audio/utils.py
@@ -27,7 +27,7 @@ def get_max_allocation_size(exec) -> Tuple[int, bool]:
         max_heap_allowed = psutil.virtual_memory().total
         thinks_is_64_bit = True
     else:
-        max_heap_allowed = 4 * 1024**3
+        max_heap_allowed = min(4 * 1024**3, psutil.virtual_memory().total)
         thinks_is_64_bit = False
     return max_heap_allowed, thinks_is_64_bit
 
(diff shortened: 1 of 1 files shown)


aikaterna (approved):
Jack showed me before and after images of him testing it on a bot and the heapsize in the displayed launcher line is exactly as I would expect.

PR #6152, opened 2023-05-13
Update bad looking 'diff' code blocks with 'markdown'

### LINKED ISSUE #6003: Remove the colors in Installed Cogs list

cog list

To me, it is very confusing that "Installed" is red and "Uninstalled" is green. I would suggest removing the syntax highlighting from this menu altogether.

Example screenshot:

image

This implies to me that the things in red are things that I do not have available to me. But, maybe I'm calling up this list to see "what's installed and ready to use" as opposed to "what's ready to be installed." In that case, the color-coding is completely backwards.

In fact, even when I am pulling this cog up because I want to install something, I get confused, because to me green should be "already installed and good to go."

### Anything else?

1. This suggestion applies to every similar interface where the diff color coding is used but the POV could be considered in either direction, "what's available to use" or "what's available to install" when the command is called
2. There is a typo in this command (Available cog: -> Available cogs)

rramboer:
Resolves #6003. Changed the cog and repo lists syntax highlighting within Discord to markdown from diff.

image

Yes

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -1333,9 +1333,9 @@ class Downloader(commands.Cog):
         )
 
         if len(installed_cogs_in_repo) > 1:
-            installed_str = _("> Installed Cogs\n{text}").format(text=installed_str)
+            installed_str = _("# Installed Cogs\n{text}").format(text=installed_str)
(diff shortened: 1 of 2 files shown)


Jackenmen (approved):
LGTM.

PR #6158, opened 2023-05-14
[redbot.core.tree] The kwarg `delete_after` doesn't exist in `discord.Interaction.followup.send`.

AAA3A-AAA3A:

Hello,

This PR fixes an error in the error handler for slash commands. It was reported by @Drapersniper in the main Red server (https://canary.discord.com/channels/133049272517001216/718148684629540905/1107334514302275615). The `delete_after` kwarg is used when the error is `redbot.core.app_commands.CommandOnCooldown`, but does not exist if `discord.Interaction.followup.send` is used to send the message, if the interaction has been answered but has not yet expired.
The simple removal of this kwarg for a slash command error in this one case was chosen.

Have a nice day,
AAA3A

Closes #6159

No

### DIFF
--- a/redbot/core/tree.py
+++ b/redbot/core/tree.py
@@ -321,7 +321,7 @@ class RedTree(CommandTree):
             msg = _("This command is on cooldown. Try again {relative_time}.").format(
                 relative_time=relative_time
             )
-            await self._send_from_interaction(interaction, msg, delete_after=error.retry_after)
+            await self._send_from_interaction(interaction, msg)
         elif isinstance(error, CheckFailure):
             await self._send_from_interaction(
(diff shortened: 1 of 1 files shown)


Flame442 (changes requested):
`pop` on a dict with no default raises a `KeyError` if the key was not in the dict. This code will break if this method is ever called with kwargs that do not include delete_after.

AAA3A-AAA3A:
> `pop` on a dict with no default raises a `KeyError` if the key was not in the dict. This code will break if this method is ever called with kwargs that do not include delete_after.

Fixed, sorry. 😅

Flame442 (approved):

PR #6164, opened 2023-05-16
Include command-line arguments in debuginfo

Jackenmen:

I thought it could be helpful in some cases, not sure how others feel about it. It is a bit better on Python 3.10 and higher as we can use `sys.orig_argv` there which is a more precise representation of arguments as received by the Python executable. Note that neither `sys.argv` nor `sys.orig_argv` is necessarily the same as what the user put in the command line, it's just that `sys.orig_argv` allows you to get more information and allows you to differ between some cases that you otherwise couldn't.

image

Yes

### DIFF
--- a/redbot/core/_debuginfo.py
+++ b/redbot/core/_debuginfo.py
@@ -3,6 +3,7 @@ from __future__ import annotations
 import getpass
 import os
 import platform
+import re
 import sys
 from typing import Optional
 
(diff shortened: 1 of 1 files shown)


redbot/core/_debuginfo.py
```diff
@@ -3,6 +3,7 @@
 import getpass
 import os
 import platform
+import re
```
TrustyJAID:
```suggestion
```

TrustyJAID (approved):
This looks good to me!

PR #6166, opened 2023-05-18
Clear the view in `[p]set api` after timeout

AAA3A-AAA3A:

Hello,

This PR disables the "Set API token" button of the `redbot.core.utils.views.SetApiView` view after the 180 seconds timeout. The logic is in the `[p]set api` command and not in the View itself, as Trusty would rather have.

Thanks in advance,
AAA3A

Yes


AAA3A-AAA3A:
As Predä proposed, this PR now removes/clears the View from the message and edits the message ("This API keys setup message has expired.").

Jackenmen (approved):
LGTM, thanks!

PR #6170, opened 2023-05-19
Fix NUMBER_EMOJIS breakages inside Audio

### LINKED ISSUE #6169: Audio NUMBER_EMOJIS tuple breakages from 3.5
This location in Audio currently attempts to mutate `NUMBER_EMOJIS`, which was mutable prior to 3.5, but [is now immutable](https://docs.discord.red/en/stable/incompatible_changes/3.5.html#menus-default-controls-and-reactionpredicate-emojis-use-immutable-types-now). It needs to be updated to properly account for this breaking change.
https://github.com/Cog-Creators/Red-DiscordBot/blob/f47d1dffb34266578b32016439b0b07d751778d8/redbot/cogs/audio/core/utilities/playlists.py#L265-L266

Kreusada:

Closes #6169, fixes where the code attempts to mutate the now immutable `NUMBER_EMOJIS`.

Yes

### DIFF
--- a/redbot/cogs/audio/core/utilities/playlists.py
+++ b/redbot/cogs/audio/core/utilities/playlists.py
@@ -262,8 +262,7 @@ class PlaylistUtilities(MixinMeta, metaclass=CompositeMetaClass):
             colour=await context.embed_colour(),
         )
         msg = await context.send(embed=embed)
-        available_emojis = ReactionPredicate.NUMBER_EMOJIS[1:]
-        available_emojis.append("🔟")
+        available_emojis = ReactionPredicate.NUMBER_EMOJIS[1:] + ("🔟",)
         emojis = available_emojis[: len(correct_scope_matches)]
         emojis.append("\N{CROSS MARK}")
         start_adding_reactions(msg, emojis)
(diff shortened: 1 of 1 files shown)


aikaterna:
After testing this change I ran into another issue on the next line:

```
[08:44:28] ERROR [red] AttributeError
Traceback (most recent call last)
/Users/aikaterna/red38/lib/python3.8/site-packages/discord/ext/commands/core.py:229 in wrapped
❱ 229 ret = await coro(*args, **kwargs)
/Users/aikaterna/red38/lib/python3.8/site-packages/redbot/cogs/audio/core/commands/playlists.py:830 in command_playlist_info
❱ 830 playlist, playlist_arg, scope = await self.get_playlist_match(
/Users/aikaterna/red38/lib/python3.8/site-packages/redbot/cogs/audio/core/utilities/playlists.py:267 in get_playlist_match
❱ 267 emojis.append("\N{CROSS MARK}")

AttributeError: 'tuple' object has no attribute 'append'
```

You could do something like this to resolve it:

```
emojis = available_emojis[: len(correct_scope_matches)]
emojis += ("\N{CROSS MARK}",)
```

I'm not going to add this as a suggestion in case there is a cleaner way to do this or another way you'd like to go about it.

Jackenmen (approved):
I switched this to a simple list() call which should make it work as it would have before, thanks for the PR.

PR #6173, opened 2023-05-23
Pass `timeout` to `SimpleMenu` 

OmLanke:
Pass the `timeout` parameter to `SimpleMenu`. 

The `timeout` parameter is currently used by the reactions based menu.
Since it wasn't passed to `SimpleMenu`, buttons based menu used the default timeout of 180 seconds.

No


Jackenmen (changes requested):
This should update the `view.timeout` with the value of the `timeout` argument after line 108 inside the `if` to allow changes to the timeout after the view was created.

Jackenmen (approved):
Thanks for making a contribution to our project!

PR #6176, opened 2023-05-25
Add ConfirmView utility

### LINKED ISSUE #6174: A ConfirmView similar to SimpleView
### Type of feature request

API functionality, Other

So now that we already have a view for buttons for menus with `SimpleView`, would've been great to add something with confirmations as well, like `ConfirmView` to confirm the actions with buttons as well. right now they're only the normal reactions so an option to have them buttons as well would've been good!

### Anything else?

_No response_

TrustyJAID:
closes #6174

Yes

### DIFF
--- a/redbot/core/utils/views.py
+++ b/redbot/core/utils/views.py
@@ -11,7 +11,7 @@ from redbot.core.commands.converter import get_dict_converter
 if TYPE_CHECKING:
     from redbot.core.commands import Context
 
-__all__ = ("SimpleMenu", "SetApiModal", "SetApiView")
+__all__ = ("ConfirmView", "SimpleMenu", "SetApiModal", "SetApiView")
 
 _ = Translator("UtilsViews", __file__)
 
--- a/redbot/core/utils/views.py
+++ b/redbot/core/utils/views.py
@@ -529,6 +529,7 @@ class ConfirmView(discord.ui.View):
 
     @discord.ui.button(label=_("Yes"), style=discord.ButtonStyle.green)
     async def confirm_button(self, interaction: discord.Interaction, button: discord.ui.Button):
+        # Warning: The Sphinx documentation for this method/attribute does not use this docstring.
         """
         A `discord.ui.Button` to confirm the message.
 
@@ -561,6 +562,7 @@ class ConfirmView(discord.ui.View):
 
     @discord.ui.button(label=_("No"), style=discord.ButtonStyle.secondary)
     async def dismiss_button(self, interaction: discord.Interaction, button: discord.ui.Button):
+        # Warning: The Sphinx documentation for this method/attribute does not use this docstring.
         """
         A `discord.ui.Button` to dismiss the message.
 
--- a/redbot/core/utils/views.py
+++ b/redbot/core/utils/views.py
@@ -540,7 +540,9 @@ class ConfirmView(discord.ui.View):
             view.confirm_button.style = discord.ButtonStyle.red
             view.confirm_button.label = "Delete"
             view.dismiss_button.label = "Cancel"
-            view.message = await ctx.send("Are you sure you want to remove #very-important-channel?", view=view)
+            view.message = await ctx.send(
+                "Are you sure you want to remove #very-important-channel?", view=view
+            )
             await view.wait()
             if view.result:
                 await ctx.send("Channel #very-important-channel deleted.")
@@ -567,7 +569,9 @@ class ConfirmView(discord.ui.View):
             view.confirm_button.style = discord.ButtonStyle.red
             view.confirm_button.label = "Delete"
             view.dismiss_button.label = "Cancel"
-            view.message = await ctx.send("Are you sure you want to remove #very-important-channel?", view=view)
+            view.message = await ctx.send(
+                "Are you sure you want to remove #very-important-channel?", view=view
+            )
             await view.wait()
             if view.result:
                 await ctx.send("Channel #very-important-channel deleted.")
--- a/redbot/core/utils/views.py
+++ b/redbot/core/utils/views.py
@@ -444,40 +444,46 @@ class ConfirmView(discord.ui.View):
         If this is omitted anyone can click yes or no.
     timeout: float
         The timeout of the view in seconds. Defaults to ``180`` seconds.
+    disable_buttons: bool
+        Whether to disable the buttons after the timeout. Defaults to ``False``.
 
     Examples
     --------
     Using the view::
 
         view = ConfirmView(ctx.author)
-        msg = await ctx.send("Are you sure you about that?", view=view)
+        view.message = await ctx.send("Are you sure you about that?", view=view)
+        # attach the message to the view this way to ensure clean removal
+        # when the view is finished.
         await view.wait()
         if view.result:
             await ctx.send("Okay I will do that.")
         else:
             await ctx.send("I will not be doing that then.")
-            if view.result is None:
-                await msg.edit(view=None)
-                # remove the buttons if they're not pressed.
 
     Disabling if the buttons are not pressed::
 
-        view = ConfirmView(ctx.author)
-        msg = await ctx.send("Are you sure you about that?", view=view)
+        view = ConfirmView(ctx.author, disable_buttons=True)
+        view.message = await ctx.send("Are you sure you about that?", view=view)
         await view.wait()
         if view.result:
             await ctx.send("Okay I will do that.")
         else:
             await ctx.send("I will not be doing that then.")
-            if view.result is None:
-                view.yes_button.disabled = True
-                view.no_button.disabled = True
-                await msg.edit(view=view)
 
     Attributes
     ----------
     result: Optional[bool]
         The result of the confirm view.
+    author: Optional[discord.abc.User]
+        The author of the message who is allowed to press the buttons.
+    message: Optional[discord.Message]
+        The message the confirm view is sent on. This can be set while
+        sending the message. This can also be left as ``None`` in which case
+        if the view is never interacted with nothing will happen in ``on_timeout``
+    disable_buttons: bool
+        Whether to disable the buttons instead of remove the buttons if the message
+        has been seen by the view.
     """
 
     def __init__(
@@ -485,26 +491,45 @@ class ConfirmView(discord.ui.View):
         author: Optional[discord.abc.User] = None,
         *,
         timeout: float = 180.0,
+        disable_buttons: bool = False,
     ):
         if timeout is None:
             raise TypeError("This view should not be used as a persistent view.")
         super().__init__(timeout=timeout)
         self.result: Optional[bool] = None
         self.author: Optional[discord.abc.User] = author
+        self.message: Optional[discord.Message] = None
+        self.disable_buttons = disable_buttons
+
+    async def on_timeout(self):
+        if self.message is None:
+            return
+            # we can't do anything here if message is none
+        if self.disable_buttons:
+            # prioritize disabling buttons since it's considered custom
+            # behavior
+            self.disable()
+            await self.message.edit(view=self)
+            return
+        await self.message.edit(view=None)
+
+    def disable(self):
+        self.yes_button.disabled = True
+        self.no_button.disabled = True
 
     @discord.ui.button(label=_("Yes"), style=discord.ButtonStyle.green)
     async def yes_button(self, interaction: discord.Interaction, button: discord.ui.Button):
         self.result = True
         self.stop()
-        await interaction.response.edit_message(view=None)
 
     @discord.ui.button(label=_("No"), style=discord.ButtonStyle.red)
     async def no_button(self, interaction: discord.Interaction, button: discord.ui.Button):
         self.result = False
         self.stop()
-        await interaction.response.edit_message(view=None)
 
     async def interaction_check(self, interaction: discord.Interaction):
+        if self.message is None:
+            self.message = interaction.message
         if self.author and interaction.user.id != self.author.id:
             return False
         return True
--- a/redbot/core/utils/views.py
+++ b/redbot/core/utils/views.py
@@ -444,6 +444,8 @@ class ConfirmView(discord.ui.View):
         anyone can click yes or no.
     default: bool
         The default result value. Defaults to False.
+    timeout: float
+        The timeout of the view in seconds. Defaults to 180 seconds.
 
     Attributes
     ----------
@@ -461,8 +463,16 @@ class ConfirmView(discord.ui.View):
             await ctx.send("I will not be doing that then.")
     """
 
-    def __init__(self, author: Optional[discord.abc.User] = None, default: bool = False):
-        super().__init__()
+    def __init__(
+        self,
+        author: Optional[discord.abc.User] = None,
+        default: bool = False,
+        *,
+        timeout: float = 180.0,
+    ):
+        if timeout is None:
+            raise TypeError("This view should not be used as a persistent view.")
+        super().__init__(timeout=timeout)
         self.result = default
         self.author: Optional[discord.abc.User] = author
 
--- a/redbot/core/utils/views.py
+++ b/redbot/core/utils/views.py
@@ -442,8 +442,6 @@ class ConfirmView(discord.ui.View):
     author: Optional[discord.abc.User]
         The user who you want to be interacting with the confirmation.
         If this is omitted anyone can click yes or no.
-    default: bool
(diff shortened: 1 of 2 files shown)


flaree (commented):
Would it be worth adding an on_timeout that sets the buttons to disabled?

TrustyJAID:
I am open to customizing the timeout but I will not be adding support for disabling the buttons automatically. That would require sending the message through the view which I feel takes away from the simplicity of this. The buttons when interacted with will remove the buttons from the message but timeout will do nothing and it should be up to the user to handle that I think.

flaree:
Huh, I have a confirm that doesn't send through a view and works fine.

```py
async def on_timeout(self):
        for item in self.children:
            item.disabled = True
```

TrustyJAID:
That would disable the button and prevent any response but would still require editing the original message which this does not have any reference to.

TrustyJAID:
I have reworked this slightly so we can know if the buttons were not pressed so users can clean them up however they wish, whether that be disabling the buttons or removing them after the timeout.

Jackenmen:
It shouldn't be required (i.e. we should only try doing it if not `None`) but I think that it would be simple enough to allow setting `view.message` attribute after the message is sent and remove the view from the message `on_timeout` in such a case. I would also suggest updating the relevant example to use that, i.e.
```py
view = ConfirmView(ctx.author)
view.message = await ctx.send("Are you sure you about that?", view=view)
await view.wait()
if view.result:
    await ctx.send("Okay I will do that.")
else:
    await ctx.send("I will not be doing that then.")
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,105 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+        super().__init__(timeout=timeout)
+        self.result: Optional[bool] = None
+        self.author: Optional[discord.abc.User] = author
+        self.message: Optional[discord.Message] = None
+        self.disable_buttons = disable_buttons
+
+    async def on_timeout(self):
+        if self.message is None:
+            return
+            # we can't do anything here if message is none
```
Jackenmen: That's a weird order, comments generally are placed before the action they talk about:
```suggestion
            # we can't do anything here if message is none
            return
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,105 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+        if self.disable_buttons:
+            # prioritize disabling buttons since it's considered custom
+            # behavior
```
Jackenmen: I don't know what this is supposed to convey, making it `if not self.disable_buttons` with an early return would have been no different so I don't see any prioritization here.

redbot/core/utils/views.py
```diff
@@ -431,3 +431,105 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+
+    @discord.ui.button(label=_("Yes"), style=discord.ButtonStyle.green)
+    async def yes_button(self, interaction: discord.Interaction, button: discord.ui.Button):
+        self.result = True
+        self.stop()
+
+    @discord.ui.button(label=_("No"), style=discord.ButtonStyle.red)
+    async def no_button(self, interaction: discord.Interaction, button: discord.ui.Button):
+        self.result = False
+        self.stop()
```
Jackenmen: These have been changed to no longer remove the view from the message, why is that?

redbot/core/utils/views.py
```diff
@@ -431,3 +431,105 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+        Whether to disable the buttons after the timeout. Defaults to ``False``.
+
+    Examples
+    --------
+    Using the view::
+
+        view = ConfirmView(ctx.author)
+        view.message = await ctx.send("Are you sure you about that?", view=view)
+        # attach the message to the view this way to ensure clean removal
+        # when the view is finished.
```
Jackenmen: That's a weird order, comments generally are placed before the action they talk about. 
The comment also seemed a bit unclear so here's a suggested rephrase:
```suggestion
        # Attach the message to the view after sending it.
        # This way, the view will be automatically removed
        # from the message after the timeout.
        view.message = await ctx.send("Are you sure you about that?", view=view)
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,105 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+
+    Parameters
+    ----------
+    author: Optional[discord.abc.User]
+        The user who you want to be interacting with the confirmation.
+        If this is omitted anyone can click yes or no.
+    timeout: float
+        The timeout of the view in seconds. Defaults to ``180`` seconds.
+    disable_buttons: bool
+        Whether to disable the buttons after the timeout. Defaults to ``False``.
```
Jackenmen: It would probably make sense to explain the default behavior:
```suggestion
        Whether to disable the buttons instead of removing them from the message after the timeout.
        Defaults to ``False``.
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,105 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+        The result of the confirm view.
+    author: Optional[discord.abc.User]
+        The author of the message who is allowed to press the buttons.
+    message: Optional[discord.Message]
+        The message the confirm view is sent on. This can be set while
+        sending the message. This can also be left as ``None`` in which case
+        if the view is never interacted with nothing will happen in ``on_timeout``
+    disable_buttons: bool
+        Whether to disable the buttons instead of remove the buttons if the message
+        has been seen by the view.
```
Jackenmen:
```suggestion
        Whether to disable the buttons instead of removing them on timeout
        (if the ``message`` attribute has been set on the view).
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,105 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+    Attributes
+    ----------
+    result: Optional[bool]
+        The result of the confirm view.
+    author: Optional[discord.abc.User]
+        The author of the message who is allowed to press the buttons.
+    message: Optional[discord.Message]
+        The message the confirm view is sent on. This can be set while
+        sending the message. This can also be left as ``None`` in which case
+        if the view is never interacted with nothing will happen in ``on_timeout``
```
Jackenmen: This may be unnecessary but the sentence read weirdly to me:
```suggestion
        nothing will happen in ``on_timeout``, if the view is never interacted with.
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,105 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+            # prioritize disabling buttons since it's considered custom
+            # behavior
+            self.disable()
+            await self.message.edit(view=self)
+            return
+        await self.message.edit(view=None)
+
+    def disable(self):
+        self.yes_button.disabled = True
+        self.no_button.disabled = True
```
Jackenmen: Is this meant to be a public method? If so, please document it, otherwise prefix it with `_`.

redbot/core/utils/views.py
```diff
@@ -431,3 +431,105 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+        view.message = await ctx.send("Are you sure you about that?", view=view)
+        # attach the message to the view this way to ensure clean removal
+        # when the view is finished.
+        await view.wait()
+        if view.result:
+            await ctx.send("Okay I will do that.")
+        else:
+            await ctx.send("I will not be doing that then.")
+
+    Disabling if the buttons are not pressed::
```
Jackenmen:
```suggestion
    Auto-disable the buttons after timeout if nothing is pressed::
```

Jackenmen (changes requested):

redbot/core/utils/views.py
```diff
@@ -11,7 +11,7 @@
 if TYPE_CHECKING:
     from redbot.core.commands import Context
 
-__all__ = ("SimpleMenu", "SetApiModal", "SetApiView")
+__all__ = ("ConfirmView", "SimpleMenu", "SetApiModal", "SetApiView")
```
Jackenmen: nitpick: we try to keep items in `__all__` in definition order to make it somewhat easier to check if anything is missing in `__all__`:
```suggestion
__all__ = ("SimpleMenu", "SetApiModal", "SetApiView", "ConfirmView")
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,147 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+        self.result = False
+        self.stop()
+        # respond to the interaction so the user does not see "interaction failed".
+        await interaction.response.defer()
+        # call `on_timeout` explicitly here since it's not called when `stop()` is called.
+        await self.on_timeout()
+
+    async def interaction_check(self, interaction: discord.Interaction):
+        if self.message is None:
+            self.message = interaction.message
```
Jackenmen, line 614: Looking at the documentation, it seems that it's consistent with it but I do wonder if we actually want to make the behavior of the view's timeout action dependent on whether some random person other than the author decides to click one of the buttons under the message? Would it not make more sense to this API's consumers to expect that not setting `view.message` will fully prevent it from removing/disabling the buttons on the message?

redbot/core/utils/views.py
```diff
@@ -431,3 +431,147 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+    Parameters
+    ----------
+    author: Optional[discord.abc.User]
+        The user who you want to be interacting with the confirmation.
+        If this is omitted anyone can click yes or no.
+    timeout: float
+        The timeout of the view in seconds. Defaults to ``180`` seconds.
+    disable_buttons: bool
+        Whether to disable the buttons instead of removing them from the message after the timeout.
+         Defaults to ``False``.
```
Jackenmen: This may seem like a nitpick but this is actually a very important fix to how Sphinx ends up displaying this:
```suggestion
        Defaults to ``False``.
```
image

redbot/core/utils/views.py
```diff
@@ -431,3 +431,147 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
         return await interaction.response.send_modal(
             SetApiModal(self.default_service, self.default_keys)
         )
+
+
+class ConfirmView(discord.ui.View):
+    """
+    A simple ``discord.ui.View`` used for confirming something.
```
Jackenmen:
```suggestion
    A simple `discord.ui.View` used for confirming something.
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,147 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+    Attributes
+    ----------
+    result: Optional[bool]
+        The result of the confirm view.
+    author: Optional[discord.abc.User]
+        The author of the message who is allowed to press the buttons.
+    message: Optional[discord.Message]
+        The message the confirm view is sent on. This can be set while
+        sending the message. This can also be left as ``None`` in which case
+        nothing will happen in ``on_timeout``, if the view is never interacted with.
```
Jackenmen:
```suggestion
        nothing will happen in `on_timeout()`, if the view is never interacted with.
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,147 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+            self.confirm_button.disabled = True
+            self.dismiss_button.disabled = True
+            await self.message.edit(view=self)
+        else:
+            await self.message.edit(view=None)
+
+    @discord.ui.button(label=_("Yes"), style=discord.ButtonStyle.green)
+    async def confirm_button(self, interaction: discord.Interaction, button: discord.ui.Button):
+        """
+        A ``discord.ui.Button`` to confirm the message.
```
Jackenmen:
```suggestion
        A `discord.ui.Button` to confirm the message.
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,147 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+
+        Examples
+        --------
+        Changing the style and label of this ``discord.ui.Button``::
```
Jackenmen:
```suggestion
        Changing the style and label of this `discord.ui.Button`::
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,147 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+        self.stop()
+        # respond to the interaction so the user does not see "interaction failed".
+        await interaction.response.defer()
+        # call `on_timeout` explicitly here since it's not called when `stop()` is called.
+        await self.on_timeout()
+
+    @discord.ui.button(label=_("No"), style=discord.ButtonStyle.secondary)
+    async def dismiss_button(self, interaction: discord.Interaction, button: discord.ui.Button):
+        """
+        A ``discord.ui.Button`` to dismiss the message.
```
Jackenmen:
```suggestion
        A `discord.ui.Button` to dismiss the message.
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,147 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+
+        Examples
+        --------
+        Changing the style and label of this ``discord.ui.Button``::
```
Jackenmen:
```suggestion
        Changing the style and label of this `discord.ui.Button`::
```

redbot/core/utils/views.py
```diff
@@ -431,3 +431,147 @@ async def auth_button(self, interaction: discord.Interaction, button: discord.Bu
...
+        The result of the confirm view.
+    author: Optional[discord.abc.User]
+        The author of the message who is allowed to press the buttons.
+    message: Optional[discord.Message]
+        The message the confirm view is sent on. This can be set while
+        sending the message. This can also be left as ``None`` in which case
+        nothing will happen in ``on_timeout``, if the view is never interacted with.
+    disable_buttons: bool
+        Whether to disable the buttons isntead of removing them on timeout
+        (if the ``message`` attribute has been set on the view).
```
Jackenmen:
```suggestion
        (if the `message` attribute has been set on the view).
```

Jackenmen (changes requested):
There's mostly a bunch of docstring fixes left but I also commented with one implementation question.

redbot/core/utils/views.py
TrustyJAID: I suppose my original thinking was to prefer cleanup of the view in as many situations as possible. For example, a cog sends the view without explicitly setting the message attribute, if the issuer of the command doesn't click the button but someone else does it would be better to then cleanup the view even if it doesn't actually respond to them. I view it as better in that the buttons no longer sit around unusable for anyone. I could move it under the author check to be more consistent and keep everything else the same I suppose. Though now I also wonder if I should be responding to someone who isn't the author with a message as well.

Jackenmen: > Though now I also wonder if I should be responding to someone who isn't the author with a message as well.

I haven't thought about it but it would definitely make it seem not broken to bystanders. And it would be more consistent with all the other views/modals that we already have (just the simple menu and set api so far).

> I could move it under the author check to be more consistent and keep everything else the same I suppose.

It seems like an alright way of handling it. But with your explanation, I honestly am not sure if we should do that or just leave it as is (since I guess that the option of not trying to do the cleanup at all is not something we're interested in).

TrustyJAID: I added a response if someone other than the author attempts to interact with the buttons. I left in setting the message object from any interaction though. I am much happier with the way this would work now and I would rather prioritize removing or disabling the view in as many situations as possible. Leaving broken buttons lying around outside of development and testing just feels bad.

Jackenmen (approved):
This new utility works as expected so from my side, this is good to go.

I would, however, like you to look at the rendered documentation and assess whether the current version of it is acceptable:
https://red-discordbot--6176.org.readthedocs.build/en/6176/framework_utils.html#redbot.core.utils.views.ConfirmView

Basically, I looked at the rendered documentation and thought that showing the documentation from the base class on our overridden version of the methods was confusing and have instead opted to document it with the actual behavior of them. Additionally, I documented `confirm/dismiss_button` as a regular attribute of the `Button` type since that's what it actually ends up being on the instance of this view.

If everything seems fine to you, feel free to just merge this.

PR #6185, opened 2023-06-12
Bump dependencies

Jackenmen:

Meaningful changes:
- orjson wheels are built for base x86-64 instruction set again
    - Should we back out on dropping support for it considering that this was the only dependency that was causing problems for us?
- the search in sphinx-rtd-theme now works again

Yes

macOS x86_64: https://github.com/Jackenmen/Red-Install-Tests/actions/runs/5336827530
Everything else: https://cirrus-ci.com/build/5771016132100096

### DIFF
--- a/requirements/base.txt
+++ b/requirements/base.txt
@@ -10,7 +10,7 @@ aiohttp-json-rpc==0.13.3
     # via -r base.in
 aiosignal==1.3.1
     # via aiohttp
-apsw==3.41.2.0
+apsw==3.42.0.0
     # via -r base.in
 async-timeout==4.0.2
     # via aiohttp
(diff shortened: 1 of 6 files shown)


Drapersniper:
> Should we back out on dropping support for it considering that this was the only dependency that was causing problems for us?

Is it just a temporary thing or have they committed to support it long term? if it's just a temporary thing it probably makes sense to drop it in 3.6.

If it's a long-term support decision then keeping it is beneficial to users.

Flame442 (dismissed):
No strong feelings on (officially) supporting/not the old instruction set, it affected ~10 people total who spoke up so it's not a *wide* problem...

Kowlin:
No hard feelings either way from me regarding Orjson's changes.

Kowlin:
Do we want to include a bump to Discord.py v2.3.1?

Jackenmen:
Yes, I'll update this PR with more up-to-date version bumps later since I still am not sure when the release will actually be happening.

Jackenmen:
I bumped the deps again since I'm considering doing a release within the following days.

Flame442 (approved):

PR #6186, opened 2023-06-12
Update readme with information about Discord

### LINKED ISSUE #6183: Update Readme to include a section explaining what is Discord.

Category: Docs - Readme

Since Discord is required in order to use Red Discord Bot. I believe it would be beneficial to new users/contributors who wants to learn more about this project, despite not knowing what Discord is. This section will include a short description of what Discord is and provide links to download the software (depending on the user's operating system).

### Anything else?

_No response_

blam1992:

Adds a short description of Discord and how to download it within the overview section.

Closes #6183

Yes


blam1992:
Any ideas as to why the 2 checks got cancelled?

NoPlagiarism:
Just reminder that this PR closes issue #6183

Drapersniper:
Is the download URL necessary if we are already linking it to the Discord main website? it feels like something akin to the following is enough.

"Red is a interactable bot for [Discord](https://discord.com/) using its [API](https://discord.com/developers/docs/intro) through the [Discord.py](https://discordpy.readthedocs.io/en/stable/index.html) library."

If the goal is to assist newcomers understand what Red is for, this should be sufficient, as this is directing you to the service main page where they themselves educate new users as to what they are this would also nail the Discord API docs reference and Discord.Py reference.

Feels like speaking at a high level and providing the relevant link for each of the 3 primary dependencies, is sufficient rather than going into too much detail that users can get themselves through previously referenced links.


Edit:
Because going into too much default here would be akin to discussing the Python dependency or async or Postgres/JSON.

blam1992:
I agree, the download link might not be necessary if we are linking users to the main website. I think it's a good idea to keep the short description of Discord, as it can be helpful for people who are new to the software or who want a quick refresher.

README.md
```diff
@@ -64,6 +64,10 @@ liking, making it completely customizable. This is a *self-hosted bot* – meani
 to host and maintain your own instance. You can turn Red into an admin bot, music bot, trivia bot,
 new best friend or all of these together!  
 
+Red is built to be used in [Discord](https://discord.com/), a voice, video, and text chat application that allows users to 
+communicate with their friends and communities. Discord utilizes servers that contain chat rooms and 
+voice channels that can be accessed through invite links.
```
Jackenmen, line 69: I think that the second sentence here may be non-obvious, "servers" in the context that Discord uses it is IMO not an everyday use of the term. "spaces" is probably better. The meaning I would try to convey here is:
> Discord allows any of its users to create spaces (servers) with text and video chat rooms (channels) for their friend groups or communities.

Part of this is already said in the first sentence so it would be good to not say the same thing twice too.

Jackenmen (changes requested):

Flame442:
Superseded by #6650

PR #6193, opened 2023-06-19
Update Audio terminology in LICENSE

aikaterna:
The LICENSE file refers to "audio.py", as it was only the one file when this wording was added in 2018. This terminology change describes the Audio module generically instead.

No


Jackenmen (approved):
Seems fine.

PR #6199, opened 2023-07-08
Allow humanize_timedelta to parse negative timedeltas

TrustyJAID:
This allows humanize_timedelta to still return something if the delta is in the negatives. A bit of a niche use case but I did encounter it since the default str representation of timedeltas isn't great. 

This adds a new kwarg on humanize_timedelta if you want to explicitly pass `-` or some other string to represent a negative timedelta. By default this will return `negative 5 hours` as an example.

Yes


TrustyJAID:
Closing in favor of #6350.

PR #6203, opened 2023-07-13
Fix CancelledError handling in done callbacks of cog init tasks

Jackenmen:

Stops the done callback from raising on `CancelledError`, returning silently instead.

Yes

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -98,14 +98,19 @@ class Downloader(commands.Cog):
 
     def create_init_task(self):
         def _done_callback(task: asyncio.Task) -> None:
-            exc = task.exception()
-            if exc is not None:
+            try:
+                exc = task.exception()
+            except asyncio.CancelledError:
+                pass
+            else:
+                if exc is None:
+                    return
                 log.error(
                     "An unexpected error occurred during Downloader's initialization.",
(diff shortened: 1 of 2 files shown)


redbot/cogs/downloader/downloader.py
```diff
@@ -98,14 +98,19 @@ async def red_delete_data_for_user(self, **kwargs):
...
+                if exc is None:
+                    return
                 log.error(
                     "An unexpected error occurred during Downloader's initialization.",
                     exc_info=exc,
                 )
-                self._ready_raised = True
-                self._ready.set()
+            self._ready_raised = True
+            self._ready.set()
```
TrustyJAID, line 113: Wouldn't this make it so that some methods don't properly wait until this task is complete?

Jackenmen: `initialize()` calls `self._ready.set()`

TrustyJAID: Actually no, looking closer this would set the event before initialize is actually done. So anything waiting on `_ready` won't properly wait since this is now outside the done callback. The same exists on Mutes now too.

TrustyJAID: Oh no I am blind on the gh diff.

TrustyJAID (approved):

PR #6208, opened 2023-07-23
Indicate when a core cog is overridden in [p]cogs and debuginfo

Jackenmen:

Since core cogs can be overridden by 3rd-party cogs, this PR adds information about which cogs are overridden to debug info *and* provides the base implementation for `[p]cogs` to build on top on for @aikaterna who said that she could do the UX part. The UX isn't part of my interest here but I'll just mention that the two options that were brought up are:
- the third section that shows only when there are any overridden cogs
- different formatting (bold/italics/etc.) for the overridden cogs in the loaded and unloaded sections
	- this may prove a bit more difficult for the non-embed variant of the command

In both of these cases, it would be important that the user can clearly tell what this is so either the section title or additional footer with a properly worded sentence explaining it would be needed.

Yes

### DIFF
--- a/redbot/core/_cog_manager.py
+++ b/redbot/core/_cog_manager.py
@@ -2,9 +2,9 @@ import contextlib
 import keyword
 import pkgutil
 from importlib import import_module, invalidate_caches
-from importlib.machinery import ModuleSpec
+from importlib.machinery import FileFinder, ModuleSpec
 from pathlib import Path
-from typing import Union, List, Optional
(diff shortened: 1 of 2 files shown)


Jackenmen:
Whether we want to show this to the user in `[p]cogs` as well is technically up for debate. I think that it would be a useful thing to have but only provided that we can get the UX right, otherwise, it's best to just leave it in debuginfo.

PR #6209, opened 2023-07-23
Improve unreleased Command.is_enabled() and document related methods

### LINKED ISSUE #4130: [API] add ability to check for disabled command

Jackenmen:

Changes the unreleased `Command.is_enabled()` method (see #5552 where that was introduced) to have a more reasonable public API. I also added documentation to this method and some additional preservations to `Command.disable/enable_in()` methods. The public aspect of this API was not touched on during the previous review which means that we would be intentionally exposing an API we did not intend to. This means that this PR should be considered a release blocker.

This resolves #4130 and closes #5001. Note that these methods are named opposite than in both of mentioned issues - `is_enabled()` instead of `is_disabled()`. I'm not necessarily sure that I prefer it but it is consistent with the existing `enabled` attribute so it probably makes more sense this way.

Yes

### DIFF
--- a/redbot/core/commands/commands.py
+++ b/redbot/core/commands/commands.py
@@ -461,11 +461,32 @@ class Command(CogCommandMixin, DPYCommand):
             if not change_permission_state:
                 ctx.permission_state = original_state
 
-    def is_enabled(self, ctx) -> bool:
+    def is_enabled(self, guild: Optional[discord.abc.Snowflake] = None) -> bool:
+        """
+        Check if the command is enabled globally or in a guild.
+
+        When guild is provided, this method checks whether
+        the command is enabled both globally and in the guild.
+
+        This is generally set by the settings managed with
+        the ``[p]command enable/disable global/server`` commands.
+
+        Parameters
+        ----------
+        guild : discord.abc.Snowflake, optional
+            The guild to check that the command is enabled in.
+            If this is ``None``, this will check whether
+            the command is enabled globally.
+
+        Returns
+        -------
+        bool
+            ``True`` if the command is enabled.
+        """
         if not self.enabled:
             return False
-        if ctx.guild:
-            if self._disabled_in.has(ctx.guild.id):
+        if guild is not None:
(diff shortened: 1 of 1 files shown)


redbot/core/commands/commands.py
```diff
@@ -530,7 +551,15 @@ async def can_see(self, ctx: "Context"):
         return True
 
     def disable_in(self, guild: discord.Guild) -> bool:
-        """Disable this command in the given guild.
+        """
+        Disable this command in the given guild.
+
+        This is generally called by the settings managed with
+        the ``[p]command disable global/server`` commands and
+        any changes made outside of that will not persist after cog
```
Flame442:
```suggestion
        the ``[p]command disable global/server`` commands.
        Any changes made outside of that will not persist after cog
```
Sort of a run on, but should this even be public if this doesn't behave as expected without support from the cog?

redbot/core/commands/commands.py
```diff
@@ -552,6 +580,13 @@ def disable_in(self, guild: discord.Guild) -> bool:
     def enable_in(self, guild: discord.Guild) -> bool:
         """Enable this command in the given guild.
 
+        This is generally called by the settings managed with
+        the ``[p]command disable global/server`` commands and
+        any changes made outside of that will not persist after cog
```
Flame442:
```suggestion
        the ``[p]command disable global/server`` commands.
        Any changes made outside of that will not persist after cog
```
Same as above.

Flame442 (changes requested):

redbot/core/commands/commands.py
Jackenmen: No, but we can't remove it until 3.6 and we should probably also provide alternative, safe APIs for command disabling in such cases. We should track it as an issue.

Flame442 (approved):

PR #6216, opened 2023-07-25
[CustomCom] Add `last_editor` in `[p]cc show` and add the alias `[p]cc info` to this command.

AAA3A-AAA3A:

Hello,

This PR adds the last editor of a CC to the output of the `[p]cc show` command. If the CC was created before this PR was merged, the cog will use `editors[-1]`. This allows the author of a CC to know who modified it and to check before removing a change.
This PR also adds the alias `[p]cc info` to this command, as I think this is the command name that comes to mind when you think of "displaying information about a CC".

Thanks in advance,
Have a nice day,
AAA3A

Yes


Jackenmen (changes requested):
I feel that it would be better for us to handle this entirely through the `editors` list rather than special-case the last editor - i.e. have a migration that changes `editors` from a list of IDs to a list of dictionaries in the form of: `{"id": 123, "name": None}` and update the implementation to append new editor as `{"id: editor.id, "name": str(editor)}` or, if they're already on the list, move them to the end (essentially making the list be in edit order).

PR #6100, opened 2023-04-22
Bump dependencies

Jackenmen:
### Description of the changes

A preliminary check didn't reveal any issues with these dep bumps. platformdirs bump (#6098) and Black style bump (#6099). Leaving this PR as a draft until I have time to read through the changes in the dependencies more closely.

Not much happened with breaking changes but here's a list of base deps that had *some* breaking changes that are at least notable:
- charset_normalizer 3.0.0 - https://github.com/Ousret/charset_normalizer/blob/master/CHANGELOG.md
- rapidfuzz 3.0.0 - https://maxbachmann.github.io/RapidFuzz/changelog_link.html 

### Have the changes in this PR been tested?

No

PR #6224, opened 2023-08-08
Fix poor fuzzy results due to changes in rapidfuzz 3.0

### LINKED ISSUE #6129: The fuzzy search for local search doesn't seem to be matching fully

3.5.1

### Cog name

Audio

### Command name

local search

Having a file structure as follows: "localtracks/numbered folder/artist - album.m4a"
I expected the command to return file results when searching a term like the artist or album name.
For example searching "tool" would return "295/Tool - Lateralus.m4a" etc.

### What actually happened?

Searching album or artist names like "tool" and "lateralus" brings back "No matches.".
It does not matter what artist or album names I search it always says no matches.

1. Copy/create tracks in "localtracks" folder in the audio cog directory
2. Boot redbot/load audio cog
3. Use command [prefix]local search [term]

### Anything else?

All of my files are named "Artist - Album.m4a" (I'm not doubling the file extension) inside numbered folders. i.e. 001, 002, etc.
Using the "local folder" command does work. ie. Typing "local folder 295" returns and plays the 295th folder as expected.

# Debug Info for Red:

## System Metadata:
CPU Cores: 12 (AMD64)
RAM: 13.7GB/31.9GB (42.8%)

## OS variables:
OS version: Windows 10 (version 10.0.19045)
User: Name
Python executable: C:\Users\Name\Music\redenv\Scripts\python.exe
Python version: 3.11.3 (64bit)
Pip version: 23.1.2
Red version: 3.5.1
Discord.py version: 2.2.3

## Red variables:
Instance name: Namebot
Owner(s): 137861053592764416 (Sohl#6054)
Disabled intents: None
Storage type: JSON
Data path: C:\Users\Name\Music\Namebot
Metadata file: C:\Users\Name\AppData\Local\Red-DiscordBot\Red-DiscordBot\config.json

Jackenmen:

I missed this notable regression caused by the lack of ANY pre-processing being done on strings in rapidfuzz 3.0 that I upgraded us to in #6100. This PR simply updates our usage of rapidfuzz to use the `rapidfuzz.utils.default_process()` function that preprocesses strings in the same way that rapidfuzz < 3 did *and* fuzzywuzzy/thefuzz do *by default*. This should close the gap that I accidentally created between 3.4.19 and 3.5.0.

Yes

### DIFF
--- a/redbot/cogs/audio/core/utilities/local_tracks.py
+++ b/redbot/cogs/audio/core/utilities/local_tracks.py
@@ -7,7 +7,7 @@ import discord
 import lavalink
 from red_commons.logging import getLogger
 
-from rapidfuzz import process
+import rapidfuzz
 from redbot.core import commands
 from redbot.core.i18n import Translator
 from redbot.core.utils import AsyncIter
@@ -116,7 +116,9 @@ class LocalTrackUtilities(MixinMeta, metaclass=CompositeMetaClass):
         to_search_string = {
             i.local_track_path.name for i in to_search if i.local_track_path is not None
         }
-        search_results = process.extract(search_words, to_search_string, limit=50)
+        search_results = rapidfuzz.process.extract(
+            search_words, to_search_string, limit=50, processor=rapidfuzz.utils.default_process
+        )
         search_list = []
         async for track_match, percent_match, __ in AsyncIter(search_results):
             if percent_match > 85:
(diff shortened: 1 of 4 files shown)


aikaterna:
Looks like this resolves #6129 for the example given, and the comment on that issue about casing is related as this change makes it easier to find things with a lowercased search vs. a title cased object. I was able to test and confirm this fix works regarding audio queue and local track search queries but was unable to find cases to verify success for customcom or fuzzy help searching.

aikaterna (approved):

PR #6226, opened 2023-08-09
Add custom status support

PredaaA:

This PR adds support for custom status in `[p]set status` command group.
Custom status for bots are now supported since August 8th 2023 https://github.com/discord/discord-api-docs/pull/6345.

Yes.

### DIFF
--- a/docs/cog_guides/core.rst
+++ b/docs/cog_guides/core.rst
@@ -3769,6 +3769,35 @@ Maximum length for a competing status is 128 characters.
 **Arguments:**
     - ``[competing]`` - The text to follow ``Competing in``. Leave blank to clear the current activity status.
 
+.. _core-command-set-status-custom:
+
+""""""""""""""""""""
+set status custom
+""""""""""""""""""""
+
+.. note:: |owner-lock|
+
+**Syntax**
+
+.. code-block:: none
+
+    [p]set status custom [text]
+
+**Description**
+
+Sets Red's custom status.
+
+This will appear as ``<text>``.
+
+Maximum length for a custom status is 128 characters.
+
+**Examples:**
+    - ``[p]set status custom`` - Clears the activity status.
+    - ``[p]set status custom Running cogs...``
+
+**Arguments:**
+    - ``[text]`` - The custom status text. Leave blank to clear the current activity status.
+
 .. _core-command-set-status-dnd:
 
 """"""""""""""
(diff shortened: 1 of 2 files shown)


redbot/core/core_commands.py
```diff
@@ -3212,6 +3212,43 @@ async def _set_my_status(self, ctx: commands.Context, status: discord.Status):
...
+        if activity:
+            await ctx.send(_("Custom status set to `{name}`.").format(name=name))
+        else:
+            await ctx.send(_("Custom status cleared."))
+
+    async def _set_my_status(self, ctx: commands.Context, status: discord.Status):
+        game = ctx.bot.guilds[0].me.activity if len(ctx.bot.guilds) > 0 else None
+        await ctx.bot.change_presence(status=status, activity=game)
+        return await ctx.send(_("Status changed to {}.").format(status))
+
```
Jackenmen: This is unnecessary, it's already in our codebase looking exactly the same.
```suggestion
```

Jackenmen (changes requested):
Can we make the argument name something like `text` instead of `name`? Also, the argument name in the Arguments section of the docstring is wrong.

Jackenmen (changes requested):
You also need to update `docs/cog_guides/core.rst`.

PredaaA:
> You also need to update `docs/cog_guides/core.rst`.

Done

docs/cog_guides/core.rst
```diff
@@ -3769,6 +3769,35 @@ Maximum length for a competing status is 128 characters.
 **Arguments:**
     - ``[competing]`` - The text to follow ``Competing in``. Leave blank to clear the current activity status.
 
+.. _core-command-set-status-custom:
+
+""""""""""""""""""""
+set status custom
+""""""""""""""""""""
```
Jackenmen: nit, we keep the header's over/underline of the same length as the header text
```suggestion
"""""""""""""""""
set status custom
"""""""""""""""""
```

Jackenmen (approved):
LGTM.

Rapptz:
I think this got merged too quick cause the API is still subject to change due to the awkwardness of it.

Jackenmen:
I saw approval on the dapi docs PR so I assumed it's good to go 😬

DJTOMATO:
Not sure if should be reported as a "bug"
but the latest red update allows us to set a custom status without a "activity as playing/listening, etc"

``!set status custom Invite me with !invitar | !help`` 

I have noticed that when music is being played by the bot, this command get resetted or cleared, having to set the status manually again.
https://bae.lena.moe/7wDTNa7VBjjo.png

PR #6230, opened 2023-08-10
Fix Picarto channel avatar in Streams cog

Leo40Git:

Fixes the thumbnail in embeds for Picarto channels starting a stream.
This PR gets the avatar URL from the `avatar` field in the returned data, instead of trying to make our own URL.

No


Jackenmen (approved):
Looks good to me, thanks!

PR #6231, opened 2023-08-11
Remove type-breaking annotation on cog_i18n decorator

laggron42:

This `type` annotation voids all typing when using this decorator, thus all instances and explicit typing of the class would be interpreted as `type` by linters.

It was replaced by a TypeVar, preserving the original type.

Yes

### DIFF
--- a/redbot/core/i18n.py
+++ b/redbot/core/i18n.py
@@ -8,7 +8,7 @@ import logging
 import discord
 
 from pathlib import Path
-from typing import Callable, TYPE_CHECKING, Union, Dict, Optional
+from typing import Callable, TYPE_CHECKING, Union, Dict, Optional, TypeVar
 from contextvars import ContextVar
 
 import babel.localedata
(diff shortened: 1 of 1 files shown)


redbot/core/i18n.py
```diff
@@ -358,11 +358,13 @@ def get_babel_regional_format(regional_format: Optional[str] = None) -> babel.co
 # noinspection PyPep8
 from . import commands
 
+_TypeT = TypeVar("_TypeT", bound=type)
+
 
 def cog_i18n(translator: Translator):
```
Jackenmen:
```suggestion
def cog_i18n(translator: Translator) -> Callable[[_TypeT], _TypeT]:
```

redbot/core/i18n.py
```diff
@@ -358,11 +358,13 @@ def get_babel_regional_format(regional_format: Optional[str] = None) -> babel.co
...
     """Get a class decorator to link the translator to this cog."""
 
-    def decorator(cog_class: type):
+    def decorator(cog_class: _TypeT):
```
Jackenmen:
```suggestion
    def decorator(cog_class: _TypeT) -> _TypeT:
```

Jackenmen (changes requested):

Jackenmen (approved):

PR #6244, opened 2023-08-31
[Help command] Utilize copy to clipboard

TrustyJAID:

This reorganizes the display of help to allow users to more easily click the copy to clipboard on code blocks to copy the command input. Basically this just moves the "Syntax:" outside of the code block leaving most things the same. I also moved aliases outside of the base command for convenience but it's now inside its own code block.

Yes


TrustyJAID:
Here's some examples of how this looks now. 
image
image
image
On desktop you can click the code block to copy the contents of the code block. Users will still need to remove parts of command arguments to work properly. But this still makes it easier to copy/paste command input.

Jackenmen:
I like the idea though I feel like it falls apart when most commands have some arguments you need to pass. `Ctrl+Backspace` doesn't work too badly but I'm not entirely convinced to be honest.

I would be interested in learning whether anyone tried using this feature for a while and if they actually ended up using the ability to copy the command or if it was just a gimmick.

Also, I would welcome more feedback on whether this way of showing command syntax and aliases is perhaps considered by the users to look better on its own. If so, then the copy-paste benefit is just a bonus and the question of the usefulness of this feature isn't too important since we can just make the UI look better and be more functional at the same time.

---

Btw, if we wanted to optimize for mobile more, we could put the command name with the prefix without anything else in one of the embed areas (author name, description, field name/value, footer) since you can tap and hold to copy any part of the embed on mobile :smile: I don't think there's necessarily a suitable place for this anywhere in the embed for it to not look off but I am pretty sure this would be a bigger usability win than the ability to copy the command on desktop.

BenCos17:
> I would be interested in learning whether anyone tried using this feature for a while and if they actually ended up using the ability to copy the command or

I use it quite a bit on my bot
It's pretty handy on longer commands tbh

Kreusada (commented):
I have tested these changes and they work great. I'd be interested to hear about your stance on [Jack's suggestions](https://github.com/Cog-Creators/Red-DiscordBot/pull/6244#issuecomment-2002648617) before approving and merging this PR.

Jackenmen (approved):

PR #6247, opened 2023-09-06
[Streams] Switch to Twitch's newer follower endpoint

### LINKED ISSUE #6246: [streams] Use of deprecated helix/users/follows endpoint

3.5.4

### Cog name

Streams

### Command name

twitchstream

Use of current API endpoint

### What actually happened?

Just received an email from Twitch:

> Hello,
>
> We are reaching out as we have identified that your application(s) below have made requests to the deprecated "helix/users/follows" Twitch API endpoint in the past month and may be affected by the scheduled shutdown if further action is not taken.
> 
> (my Red bot)
> 
> This “Get Users Follows” Twitch API endpoint was deprecated in February 2023 and replaced by “[Get Channel Followers](https://dev.twitch.tv/docs/api/reference/#get-channel-followers)” and “[Get Followed Channels](https://dev.twitch.tv/docs/api/reference/#get-followed-channels).”  To continue receiving follow count, receiving follower information, or checking for a follow, please make sure your application is updated to use the new API URLs and that your authentication flow includes the new scopes if required.

1. Follow a Twitch stream
2. wait a month(?) until Twitch emails about it
3.
...

### Anything else?

Looks like just changing out `TWITCH_FOLLOWS_ENDPOINT` in `streamtypes.py` to `helix/channels/followers`, and looking for the `total` key in the response would get the total # of followers of that broadcast.

karlsbjorn:

Switches to the newer non-deprecated follower endpoint for Twitch since the one we're using now is being decommissioned.
Closes #6246

Yes, with `[p]twitchstream`

### DIFF
--- a/redbot/cogs/streams/streamtypes.py
+++ b/redbot/cogs/streams/streamtypes.py
@@ -27,7 +27,7 @@ from redbot.core.utils.chat_formatting import humanize_number, humanize_timedelt
 TWITCH_BASE_URL = "https://api.twitch.tv"
 TWITCH_ID_ENDPOINT = TWITCH_BASE_URL + "/helix/users"
 TWITCH_STREAMS_ENDPOINT = TWITCH_BASE_URL + "/helix/streams/"
-TWITCH_FOLLOWS_ENDPOINT = TWITCH_BASE_URL + "/helix/channels/follows"
+TWITCH_FOLLOWS_ENDPOINT = TWITCH_BASE_URL + "/helix/channels/followers"
 
 YOUTUBE_BASE_URL = "https://www.googleapis.com/youtube/v3"
 YOUTUBE_CHANNELS_ENDPOINT = YOUTUBE_BASE_URL + "/channels"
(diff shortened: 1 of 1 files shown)


palmtree5 (dismissed):
This solution is actually not enough as the way we authenticate with Twitch's API at the moment results in an app access token, which is not enough to use `/helix/channels/followers` as that endpoint requires a user access token that has the `moderator:read:followers` scope and given we only use this information to get a follower count, it might be better to just remove it for the moment, or replace it with some other useful piece of information

karlsbjorn:
Not having the scope should be fine for our use case here, no? This is from the Twitch docs here: https://dev.twitch.tv/docs/api/reference/#get-channel-followers

image

Works fine in my testing. I definitely don't have that scope:

DiscordCanary_hXi4wv5GT2

palmtree5:
ya know, I think I read that line once, months ago when the change was first announced, but then when I read it when writing that review I missed that line lmao. Yeah, this should be perfect then

palmtree5 (approved):

PR #6254, opened 2023-10-07
feat: use f-string instead of format

madebylydia:

This PR propose to change the LMGTFY link from `https://lmgtfy.app` to `https://letmegooglethat.com` as it seems that the old domain seems to experience problems that will never be resolved.

It's been weeks since the HSTS error has been happening.

image

Also remove the `s` query parameter, it seems unused.

No, but should be fine.


madebylydia:
Closing for reopening... GitHub freaked out...

PR #6007, opened 2023-03-27
Update general.py / Fix LMGTFY

DJTOMATO:
the current LMGTFY function uses <https://lmgtfy.app>, which used to work in the past (a year ago) but right now just sends you to a "custom search engine" with no results and not google like it used to be, It's sketchy at best, and could even be tracking everyone using it.

<https://bae.lena.moe/i45u9c05kjub.png>
<https://bae.lena.moe/867osbhjqrcz.png>

The current function redirects you to: <https://lmgtfy.app/?q=Red-DiscordBot&qtype=search>
which has zero results, and this been like that for months

The second alternative I've found could be https://letmegooglethat.com/?q=Red-DiscordBot

<https://bae.lena.moe/kcjdlenr47kg.png>
<https://bae.lena.moe/4z55xwg8874e.png>
But this still redirects you to<https://google.gprivate.com/search.php?search?q=Red-DiscordBot> a private search engine.

So, as of today, the command doesn't work and even lost the comedic value. It has no purpose/functionality as of now.

I've come across this project and has a demo we could use, which, doesn't track you, and actually redirects you to Google. <https://github.com/NatoBoram/lmgtfy>

<https://natoboram.github.io/lmgtfy/search?q=Red-DiscordBot&btnK=Google+Search> 
<https://bae.lena.moe/5eu9hpx3fboj.png>
<https://bae.lena.moe/jl6tcg24iyie.png>

This PR changes the URL generated to the third alternative I propose and actually revives the command.

### Have the changes in this PR been tested?

Yes
No

PR #6255, opened 2023-10-07
[General] Update LMGTFY link

madebylydia:

This PR propose to change the LMGTFY link from `https://lmgtfy.app` to `https://letmegooglethat.com` as it seems that the old domain seems to experience problems that will never be resolved.

It's been weeks since the HSTS error has been happening.

image

Also remove the `s` query parameter, it seems unused.

No, but should be fine.

### DIFF
--- a/redbot/cogs/general/general.py
+++ b/redbot/cogs/general/general.py
@@ -220,7 +220,7 @@ class General(commands.Cog):
     async def lmgtfy(self, ctx, *, search_terms: str):
         """Create a lmgtfy link."""
         search_terms = escape(urllib.parse.quote_plus(search_terms), mass_mentions=True)
-        await ctx.send("https://letmegooglethat.com/?q={}".format(search_terms))
+        await ctx.send(f"https://letmegooglethat.com/?q={search_terms}")
 
     @commands.command(hidden=True)
     @commands.guild_only()
--- a/redbot/cogs/general/general.py
+++ b/redbot/cogs/general/general.py
@@ -220,7 +220,7 @@ class General(commands.Cog):
     async def lmgtfy(self, ctx, *, search_terms: str):
         """Create a lmgtfy link."""
         search_terms = escape(urllib.parse.quote_plus(search_terms), mass_mentions=True)
-        await ctx.send("https://lmgtfy.app/?q={}&s=g".format(search_terms))
+        await ctx.send("https://letmegooglethat.com/?q={}".format(search_terms))
 
     @commands.command(hidden=True)
     @commands.guild_only()
--- a/redbot/cogs/general/general.py
+++ b/redbot/cogs/general/general.py
@@ -220,7 +220,7 @@ class General(commands.Cog):
     async def lmgtfy(self, ctx, *, search_terms: str):
         """Create a lmgtfy link."""
         search_terms = escape(urllib.parse.quote_plus(search_terms), mass_mentions=True)
-        await ctx.send(f"https://letmegooglethat.com/?q={search_terms}")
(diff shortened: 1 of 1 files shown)


Flame442:
The old URL is still technically working, but it's certificates have expired which makes it very hard to actually use the site now. The `s` query param is used by that site to specify it should redirect to a google search, rather than some other search engine. The site you have changed it to may not support that option.

The new site redirects to `gprivate.com` instead of `google.com`. Considering it has an about page that is unfilled out, a merch page underneath, and redirects to a non-google site, I'm hesitant to say this site is entirely safe. Considering the history of this command, and the fact that there doesn't seem like a great replacement site, this might be time to just cut the command.

madebylydia:
Tell me what's the taken decision and I'll make the changes. 👍

TrustyJAID:
I think it might be a good idea to just cut the middle man entirely and link straight to google instead. `https://www.google.com/search?q={search_term}` that way there's no issues and we maintain the functionality.

Flame442:
Yeah, I'd agree that's a good compromise.

Jackenmen:
I don't agree with cutting the middle-man since it doesn't quite have the same power as "let me google that for you" animation but the `letmegooglethat.com` site (or `googlethatforyou.com` which is really the same thing) is not an acceptable choice either...

https://lmgtfy2.com seems to work nicely? Not sure if we want the variant with "Internet Explainer", I kinda like it but it may be too much lol

TrustyJAID:
Sure it doesn't have the same power but is it really worth keeping around for a silly joke? I think cutting the middle-man also helps prevent potential abuse. For example what if whatever site we choose to use decides to let their domain lapse and someone else buys it to exploit anyone using that service? That's outside our control and leads to us having to find yet another alternative like we are now. I think the best approach would be just to leave the joke in the command name and skip this hassle in the future.

Jackenmen:
> but is it really worth keeping around for a silly joke?

Absolutely.

However, we could also expand the command to do the explainer on its own :)

madebylydia:
I understand both point of view. However, that would totally beat the command's joke to directly use Google's website. That's -1 from me. It's `[p]lmgtfy`, not `[p]google`. :wink:

Also agree that `lmgtfy2.com` seems like the best option for now, but I also have little doubt to the website's future considering the copyright says "2022-2023", to take back on what @TrustyJAID said about website's availability.
I don't think it's really a bad thing nor a big time-consuming task to find another alternative to the LMGTFY provider, if needed.

I'll update the branch to use Jacknmen's host, but I'm still listening to talk regarding this change. Also, the command explanation required.

Flame442 (approved):
Good enough for now, we can readdress this if (when) this site goes down.

BenCos17:
happened to come across https://lmgt.org/ also just now browsing reddit which seems to also work btw

DJTOMATO:
I've come across this project and has a demo we could use, which, doesn't track you, and actually redirects you to Google. https://github.com/NatoBoram/lmgtfy

https://natoboram.github.io/lmgtfy/search?q=Red-DiscordBot&btnK=Google+Search


https://github.com/Cog-Creators/Red-DiscordBot/pull/6007/commits

DJTOMATO:
Installed a test version
https://lmgtfy.lena.moe


```py
import urllib.parse

search_string = "What is Red-DiscordBot"
encoded_search_string = urllib.parse.quote(search_string)

link = f"[Here's your link](https://lmgtfy.lena.moe/lmgtfy/search?q={encoded_search_string}&btnK=Google+Search)"
await ctx.send(link)

madebylydia:
You should consider making a new issue/create a new PR to consider changing the modification made from this PR.

PR #6259, opened 2023-10-14
Prefer force_registration=true in Config documentation

Jackenmen:

Adds `force_registration=True` to all `Config.get_conf()` usage examples with the exception of "V2 Data Usage" section where no defaults are registered and this may be unwanted.

Additionally, I did a slight rephrasing and clarification to the documentation that mentions `force_registration=True`.

Yes

### DIFF
--- a/docs/framework_config.rst
+++ b/docs/framework_config.rst
@@ -60,11 +60,14 @@ Then, in the class's :code:`__init__` function, you need to get a config instanc
 The ``identifier`` in :py:meth:`Config.get_conf` is used to keep your cog's data separate
 from that of another cog, and thus should be unique to your cog. For example: if we
 have two cogs named :code:`MyCog` and their identifier is different, each will have
-its own data without overwriting the other's data. Note that, as shown by most of the examples
+its own data without overwriting the other's data.
+
+Note that, as shown by most of the examples
 in this document, it is also possible to force registration of a data key before allowing you
 to get and set data for that key by adding :code:`force_registration=True` after identifier.
 When this is set to :code:`False` (the default), the default value for any key that isn't registered
-will be :code:`None`.
+will be :code:`None`. When this is set to :code:`True` (as shown in this document), attempting
+to read from or write to any key that isn't registered will raise an :code:`AttributeError`.
 
 After we've gotten that, we need to register default values:
 
--- a/docs/framework_config.rst
+++ b/docs/framework_config.rst
@@ -25,7 +25,7 @@ Basic Usage
 
     class MyCog(commands.Cog):
         def __init__(self):
-            self.config = Config.get_conf(self, identifier=1234567890)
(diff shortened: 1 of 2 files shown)


docs/framework_config.rst
```diff
@@ -55,23 +55,24 @@ Then, in the class's :code:`__init__` function, you need to get a config instanc
...
+            self.config = Config.get_conf(self, identifier=1234567890, force_registration=True)
 
 The ``identifier`` in :py:meth:`Config.get_conf` is used to keep your cog's data separate
 from that of another cog, and thus should be unique to your cog. For example: if we
 have two cogs named :code:`MyCog` and their identifier is different, each will have
-its own data without overwriting the other's data. Note that it is also possible
-to force registration of a data key before allowing you to get and set data for
-that key by adding :code:`force_registration=True` after identifier (that defaults
-to :code:`False` though)
+its own data without overwriting the other's data. Note that, as shown by most of the examples
```
Flame442:
```suggestion
its own data without overwriting the other's data.

Note that, as shown by most of the examples
```
I would line break here, since it's a different topic and my other suggestion makes the `force_registration` part longer.

docs/framework_config.rst
```diff
@@ -55,23 +55,24 @@ Then, in the class's :code:`__init__` function, you need to get a config instanc
...
+in this document, it is also possible to force registration of a data key before allowing you
+to get and set data for that key by adding :code:`force_registration=True` after identifier.
+When this is set to :code:`False` (the default), the default value for any key that isn't registered
+will be :code:`None`.
```
Flame442:
```suggestion
will be :code:`None`. When this is set to :code:`True` (as shown in this document), attempting
to read from or write to any key that isn't registered will raise an :code:`AttributeError`.
```
I also think that in the `Config.force_registration` description, it should specify that the error raised is an `AttributeError`. Generally CCs _shouldn't_ be trying to catch this, however it is better to be clear about what will happen so they understand that error comes from this setting.

docs/framework_config.rst
```diff
@@ -304,7 +305,7 @@ the built-in Economy credits::
 
     class Pets(commands.Cog):
         def __init__(self):
-            self.config = Config.get_conf(self, 1234567890)
+            self.config = Config.get_conf(self, 1234567890, force_registration=True)
```
Flame442, line 311: Should this use `identifier` as a kwarg, to match other examples?

Flame442 (changes requested):
The `V2 Data Usage` block example does not have `force_registration` added to it. Do we even still need that block, and does it have a reason to not use `force_registration`?

docs/framework_config.rst
Jackenmen: Variety in examples shows that there are multiple valid ways of using the method - seems fine to me.

Flame442 (approved):

PR #6261, opened 2023-10-16
Fix bot shutdown hang on connection error (or other unhandled exceptions)

### LINKED ISSUE #5780: Investigate shutdown handling from unhandled exceptions in main bot task
It would seem that Red may infinitely hang during the handling of such exceptions.

An example:
```
[2022-06-21 06:44:47] [CRITICAL] red.main: The main bot task didn't handle an exception and has crashed
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/__main__.py", line 466, in red_exception_handler
    red_task.result()
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/__main__.py", line 360, in run_bot
    await red.start(token, bot=True)
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/redbot/core/bot.py", line 1187, in start
    await self.login(*args)
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/discord/client.py", line 511, in login
    await self.http.static_login(token.strip(), bot=bot)
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/discord/http.py", line 300, in static_login
    data = await self.request(Route('GET', '/users/@me'))
  File "/home/ubuntu/red-venv/lib/python3.8/site-packages/discord/http.py", line 265, in request
    raise DiscordServerError(r, data)
discord.errors.DiscordServerError: 500 Internal Server Error (error code: 0): <html>
<head><title>500 Internal Server Error</title><script async src='/cdn-cgi/bm/cv/669835187/api.js'></script></head>
<body bgcolor="white">
<center><h1>500 Internal Server Error</h1></center>
<hr><center>nginx</center>
<script type="text/javascript">(function(){window['__CF$cv$params']={r:'71eac6d9c8a49b46',m:'_iRxzcZJICzOmK6RBDUCqIPQZX>
</html>
[2022-06-21 06:44:47] [WARNING] red.main: Attempting to die as gracefully as possible...
[2022-06-21 06:44:47] [INFO] red.main: Shutting down from unhandled exception
```

CyberKiller:
Change red_exception_handler() from calling shutdown_handler() to sys.exit()
fixes #5780 

Yes

### DIFF
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -472,7 +472,7 @@ def red_exception_handler(red, red_task: asyncio.Future):
     except Exception as exc:
         log.critical("The main bot task didn't handle an exception and has crashed", exc_info=exc)
         log.warning("Attempting to die as gracefully as possible...")
-        asyncio.create_task(shutdown_handler(red))
+        sys.exit(ExitCodes.CRITICAL)
 
 
 def main():
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -403,10 +403,6 @@ async def run_bot(red: Red, cli_flags: Namespace) -> None:
             " (and --co-owner if you need more than one) flag\n"
         )
         sys.exit(ExitCodes.CONFIGURATION_ERROR)
-    except Exception as e:
-        # Catch all other exceptions here to prevent shutdown hangs
-        log.critical(e)
-        sys.exit(ExitCodes.CRITICAL)
 
     return None
 
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -403,7 +403,10 @@ async def run_bot(red: Red, cli_flags: Namespace) -> None:
             " (and --co-owner if you need more than one) flag\n"
         )
         sys.exit(ExitCodes.CONFIGURATION_ERROR)
-
+    except Exception as e:
+        log.critical(e)
+        sys.exit(ExitCodes.CRITICAL)
+        
     return None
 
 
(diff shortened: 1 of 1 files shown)


CyberKiller:
On further thought, I realise that a catch all exception will probably not be acceptable, as it will catch exceptions during the running of the bot and not just at start up so I am going to convert this pull request to a draft and look for a better solution to this.

Jackenmen:
Errors while the bot is running don't generally propagate this far and are handled earlier by things such as loop's exception handler. If they do propagate here, something more serious happened and we do probably want to shutdown after. If connection errors propagate here then the placement is likely fine.

CyberKiller:
> Errors while the bot is running don't generally propagate this far and are handled earlier by things such as loop's exception handler. If they do propagate here, something more serious happened and we do probably want to shutdown after. If connection errors propagate here then the placement is likely fine.

So do you think a catch all exception is appropriate or would catching more specific exceptions be better?

Jackenmen:
It may be fine, can't really comment on that until I have a chance to test/play with it.

CyberKiller:
I suspect if any exception doesn't get caught by run_bot() while the bot is starting up then the hang will occur. ~~The question is does having a catch all exception that calls sys.exit() prevent shutdown_handler() from doing what it needs to do?~~

CyberKiller:
I had a little more of a poke around and I think I found out the real cause of the hang: if red_exception_handler() catches an exception and calls shutdown_handler(), then the code never falls through back into main().
Replacing the call to shutdown_handler() in red_exception_handler() with sys.exit() allows the code to fall back though to main() and allows it to call the shutdown_handler() and run the finally block at the end of main() therefore stopping the loop correctly without hanging.
As for why it was behaving in this way I have no idea.

PR #6265, opened 2023-11-01
Add warning to [p]set api if <>s are included in secret

### LINKED ISSUE #6253: Add warning to [p]set api if <>s are included in secret
### Type of feature request

Command

When users are instructed to set an API key with `[p]set api`, they are often told to fill in parts of the command by marking a part with something like `<keyname>` (for example, `[p]set api youtube api_key <your_api_key_here>`). When this is done, the user is expected to replace the entire word, including the `<>`s. Sometimes users do not realize this, and enter their key within `<>`s. This often breaks the API service in an unexpected way, which can be confusing for users.

We can't reject keys that are wrapped in `<>`s, because they may be valid for a strange service out there, however we should warn the user to double check if their key is actually supposed to be formatted like that. All secrets should be checked to see if they start with `<` and end with `>`, and if so the user should be warned that they may have failed to replace the entire block, with a warning added to the confirmation message such as:
> You may have failed to properly format the key. If you were told to enter a key with an example such as `[p]set api youtube api_key <your_api_key_here>`, and your API key was `HREDFGWE`, make sure to run `[p]set api youtube api_key HREDFGWE`, and **not**  `[p]set api youtube api_key <HREDFGWE>`

There's a small concern that a key might be something entirely stupid like `<>` itself, and this would technically reveal information about the key, but I believe that is a relatively tiny concern.

### Anything else?

_No response_

giplgwm:

Closes #6253
For every given secret, check if its wrapped in <>, if it is, add a message to an embed with a warning that that secret may not be set properly. I tested this on a local instance of the bot and it seems to be working as expected. The warning is also logged

Yes

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3633,7 +3633,20 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         else:
             if ctx.bot_permissions.manage_messages:
                 await ctx.message.delete()
+            print(tokens)
+            embed = discord.Embed()
+            for dict_key, token in tokens.items():
+                if token.startswith("<") and token.endswith(">"):
+                    angle_bracket_warning = _(
+                        "Your `{dict_key}` was set including the <> symbols. If this was an accident you may experience"
+                        " issues with `{service}`. In that case, please try setting your `{dict_key}` again with"
+                        " the < and > removed"
+                    ).format(dict_key=dict_key, service=service)
+                    log.warning(angle_bracket_warning)
+                    embed.add_field(name="Warning", value=angle_bracket_warning)
             await ctx.bot.set_shared_api_tokens(service, **tokens)
+            if embed:
+                await ctx.send(embed=embed)
             await ctx.send(_("`{service}` API tokens have been set.").format(service=service))
 
     @_set_api.command(name="list")
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3633,7 +3633,6 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         else:
             if ctx.bot_permissions.manage_messages:
                 await ctx.message.delete()
-            print(tokens)
             embed = discord.Embed()
             for dict_key, token in tokens.items():
                 if token.startswith("<") and token.endswith(">"):
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3634,19 +3634,21 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
             if ctx.bot_permissions.manage_messages:
                 await ctx.message.delete()
             embed = discord.Embed()
-            for dict_key, token in tokens.items():
+            for api_service_name, token in tokens.items():
                 if token.startswith("<") and token.endswith(">"):
-                    angle_bracket_warning = _(
-                        "Your `{dict_key}` was set including the <> symbols. If this was an accident you may experience"
-                        " issues with `{service}`. In that case, please try setting your `{dict_key}` again with"
-                        " the < and > removed"
-                    ).format(dict_key=dict_key, service=service)
+                    angle_bracket_warning = (
+                        "You may have failed to properly format your {api_service_name}. If you were told to enter a key"
+                        " with an example such as [p]set api {service} api_key <your_api_key_here>, and your API key"
+                        " was HREDFGWE, make sure to run [p]set api {service} api_key HREDFGWE, and not "
+                        " [p]set api {service} api_key <HREDFGWE>"
+                    ).format(api_service_name=api_service_name, service=service)
                     log.warning(angle_bracket_warning)
-                    embed.add_field(name="Warning", value=angle_bracket_warning)
+                    embed.add_field(name=_("Warning"), value=_(angle_bracket_warning))
             await ctx.bot.set_shared_api_tokens(service, **tokens)
-            if embed:
-                await ctx.send(embed=embed)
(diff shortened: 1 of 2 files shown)


redbot/core/core_commands.py
```diff
@@ -3633,7 +3633,19 @@ async def _set_api(
...
+                if token.startswith("<") and token.endswith(">"):
+                    angle_bracket_warning = _(
+                        "Your `{dict_key}` was set including the <> symbols. If this was an accident you may experience"
+                        " issues with `{service}`. In that case, please try setting your `{dict_key}` again with"
+                        " the < and > removed"
+                    ).format(dict_key=dict_key, service=service)
+                    log.warning(angle_bracket_warning)
+                    embed.add_field(name="Warning", value=angle_bracket_warning)
             await ctx.bot.set_shared_api_tokens(service, **tokens)
+            if embed:
```
Drapersniper: This will be always the case with this command so this check is obsolete

You merge the card.send to send both the string and the new embed by specifying keyword arguments in that function

redbot/core/core_commands.py
```diff
@@ -3633,7 +3633,19 @@ async def _set_api(
...
                 await ctx.message.delete()
+            embed = discord.Embed()
+            for dict_key, token in tokens.items():
+                if token.startswith("<") and token.endswith(">"):
+                    angle_bracket_warning = _(
+                        "Your `{dict_key}` was set including the <> symbols. If this was an accident you may experience"
+                        " issues with `{service}`. In that case, please try setting your `{dict_key}` again with"
+                        " the < and > removed"
+                    ).format(dict_key=dict_key, service=service)
+                    log.warning(angle_bracket_warning)
```
Drapersniper, line 3645: Log messages shouldn't be translated and should remain on the same language as the software

redbot/core/core_commands.py
```diff
@@ -3633,7 +3633,19 @@ async def _set_api(
         else:
             if ctx.bot_permissions.manage_messages:
                 await ctx.message.delete()
+            embed = discord.Embed()
+            for dict_key, token in tokens.items():
+                if token.startswith("<") and token.endswith(">"):
+                    angle_bracket_warning = _(
+                        "Your `{dict_key}` was set including the <> symbols. If this was an accident you may experience"
```
Drapersniper: You should explicitly clarify that it is wrapped by <> current sentence implies that < or > is found in the token/key

Drapersniper: For translatable strings, I would personally suggest naming the formatting variable something contextual so that translators are better aware of the translation context for example "api_service_name"

redbot/core/core_commands.py
```diff
@@ -3633,7 +3633,19 @@ async def _set_api(
...
+                        " issues with `{service}`. In that case, please try setting your `{dict_key}` again with"
+                        " the < and > removed"
+                    ).format(dict_key=dict_key, service=service)
+                    log.warning(angle_bracket_warning)
+                    embed.add_field(name="Warning", value=angle_bracket_warning)
```
Drapersniper: Field name should be translated

giplgwm:
@Drapersniper I added the changes requested in your comments as another commit on the PR, thanks

redbot/core/core_commands.py
```diff
@@ -3633,8 +3633,22 @@ async def _set_api(
...
             if ctx.bot_permissions.manage_messages:
                 await ctx.message.delete()
+            embed = discord.Embed()
+            for api_service_name, token in tokens.items():
+                if token.startswith("<") and token.endswith(">"):
+                    angle_bracket_warning = (
+                        "You may have failed to properly format your {api_service_name}. If you were told to enter a key"
+                        " with an example such as [p]set api {service} api_key <your_api_key_here>, and your API key"
+                        " was HREDFGWE, make sure to run [p]set api {service} api_key HREDFGWE, and not "
+                        " [p]set api {service} api_key <HREDFGWE>"
```
Kreusada, line 3643: Wrapping the command examples with inline marks may be appropriate to improve readability.

```suggestion
                        "You may have failed to properly format your {api_service_name}. If you were told to enter a key"
                        " with an example such as `[p]set api {service} api_key <your_api_key_here>`, and your API key"
                        " was `HREDFGWE`, make sure to run `[p]set api {service} api_key HREDFGWE` and not "
                        " `[p]set api {service} api_key <HREDFGWE>`."
```

redbot/core/core_commands.py
```diff
@@ -3633,8 +3633,22 @@ async def _set_api(
...
+                    ).format(api_service_name=api_service_name, service=service)
+                    log.warning(angle_bracket_warning)
+                    embed.add_field(name=_("Warning"), value=_(angle_bracket_warning))
             await ctx.bot.set_shared_api_tokens(service, **tokens)
-            await ctx.send(_("`{service}` API tokens have been set.").format(service=service))
+            await ctx.send(
+                _("`{service}` API tokens have been set.").format(service=service),
+                embed=embed if len(embed.fields) > 0 else None,
```
Kreusada, line 3650: It isn't necessary to give this warning inside of an embed, when there is already text content (it would make things more complicated, such as having to check if an embed is requested in the current channel). Instead, you could add line breaks under the original message and put the warning text there. Something like this would suffice:
```py
angle_bracket_warning = None # define this before iterating through tokens
for api_service_name, token in tokens.items():
    ...
```
```py
message = _("`{service}` API tokens have been set.").format(service=service)
if angle_bracket_warning:
    message += "\n\n" + _("**Warning:** ") + _(angle_bracket_warning)
await ctx.send(message)
```

Remember to remove the `embed` definition on line 3636.

Kreusada (changes requested):
Just a couple of suggested changes to improve readability.

Jackenmen:
Superseded by #6651

PR #6278, opened 2023-12-29
Add methods for getting app command IDs/mentions from cache

### LINKED ISSUE #5976: Add functionality to mention app commands

Flame442:
Adds `bot.get_app_command_id` and `bot.get_app_command_mention` which pull from the ID cache built when `RedTree.sync`ing.

Potential future work:
- Hook into `CommandTree.fetch_commands` to update the cache manually using the fetched command data
- Migrate the current data structure to allow storing guild specific app command IDs on guild `sync`, and allow these new methods to be used to pull from that cache as well

Fixes #5976

Yes


Jackenmen:
> Potential future work:
> * Migrate the current data structure to allow caching on guild `sync`, and allow these new methods to be used to pull from that cache as well

Having such a cache now would mean we don't need to make these methods asynchronous and such a change to method signature is not really something we can do once the API is released. Is there any reason we wouldn't want those methods to be synchronous?

Flame442:
I don't mean a memory cache, I mean storing in config _guild specific_ command ids. Currently, only global app command ids are stored, since that is all the `[p]slash` system cares about.
(Edited original to be more clear)

Jackenmen:
Oh yeah, I missed that this remark is about guild stuff. Either way, I am still wondering about the in-memory cache thing to get a synchronous API, it does seem a bit unfortunate that it requires awaiting right now.

Daniel07468 (commented):
Redbot

Jackenmen (approved):
LGTM. Sorry that I'm only coming to this *now*, two releases later than I said I would...

PR #6286, opened 2024-01-14
Add a start_dm option to SimpleMenu

### LINKED ISSUE #6282: Add a start_dm option to SimpleMenu
In order to support `[p]help` with a button menu in DMs, `SimpleMenu` needs to be able to be sent as a DM to a particular user [[Ref](https://github.com/Cog-Creators/Red-DiscordBot/pull/5375#issuecomment-1874666094)]. The current `SimpleMenu.start` method accepts a `Context` object, which is incompatible with this behavior. `SimpleMenu` should be given a new method `.start_dm(user: discord.User)`, which sends the menu as a direct message to that user.

**Developer note:**
`SimpleMenu` currently stores `self.ctx` in `.start`, however it does not appear to ever be used. The other two attributes set by `.start` (`message` and `author`) will be possible to set in this method.

Unblocks #5375

yeetbruises:

- Added start_dm(user) method to SimpleMenu in "Views.py".
- - This method allows for a direct message to be sent by the bot, which bypasses the usage of "ctx" as a parameter. This also unblocks issue #5375. My testing found the buttons do work in DMs following implementation of the new method.
- Fixes #6282 

**Yes**

I tested the method by making changes to "help.py" under the send_pages() method. 

**The image below shows the modified code in that file.**

Screenshot 2024-01-14 at 5 15 57 PM

**DM result**

Screenshot 2024-01-14 at 5 18 12 PM


Flame442 (approved):
Thanks!

As an FYI, the example screenshot of your testing had a line `get_user = ctx.bot.get_user(ctx.author.id)`. `ctx.author` is already a `User` object (even if used in a guild, since the `Member` object you will get is a subclass of `User`), so you don't need the extra cache lookup in that case.

PR #6289, opened 2024-02-01
Ensure the repo list is included in backups

### LINKED ISSUE #6288: Backup does not include list of repos

3.5.5

I was in the process of working on #5703 and created a backup in order to work on the restore process

The `repos.json` file to be present in the backup

### What actually happened?

I opened up the backup to look at the repos file generated to work on restoring the repos and cogs that the `repos.json` file being created during the backup process was not present in the specified folder

1. Create a backup
2. Open the backup in something like 7-zip
3. Navigate to the directory and see the file is missing
...

### Anything else?

I believe this to be due to the exclusions, as I ran the code that pulls the list of files to add in an eval and found that the file was not present when I listed the files to be backed up, but was present when listing excluded files.

```py
from pathlib import Path
import os
from redbot.core import data_manager

data_path = Path(data_manager.core_data_path().parent)

exclusions = [
    "__pycache__",
    "Lavalink.jar",
    os.path.join("Downloader", "lib"),
    os.path.join("CogManager", "cogs"),
    os.path.join("RepoManager", "repos"),
    os.path.join("Audio", "logs"),
]

out_list = []
for f in data_path.glob("**/*"):
    if not any(ex in str(f) for ex in exclusions) and f.is_file():
        out_list.append(str(f))
print("\n".join(out_list))
```

the exclusion in particular of `RepoManager/repos` is having the effect of excluding `RepoManager/repos` (the thing intended to be excluded), AND `RepoManager/repos.json` (which needs to be in the backup)

palmtree5:

Fixes #6288 by just manually adding the file to the list of files to be included


palmtree5:
Superseded by #6709

PR #6296, opened 2024-02-22
Update autostart_systemd.rst

pablo8itall:
When creating the name of the of the systemd file you need to add the instance name after the @ symbol, e.g. red@MyInstance.service, as the service file uses the %I for this information.

added some text explaining that the instance name needs to be used when naming the file.

Yes

### DIFF
--- a/docs/autostart_systemd.rst
+++ b/docs/autostart_systemd.rst
@@ -32,9 +32,9 @@ Next, your python :code:`path` can be fetched with the following commands:
     $ pyenv shell <virtualenv_name>
     (redenv) $ pyenv which python
 
-Then create the new service file:
+Then create the new service file (replace InstanceName with the name of your Instance):
 
-:code:`sudo nano /etc/systemd/system/red@.service`
+:code:`sudo nano /etc/systemd/system/red@InstanceName.service`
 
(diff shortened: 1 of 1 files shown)


yamikaitou:
This is actually incorrect. The file needs to be saved as we had it but when you interact with the service, you include the instancename.

palmtree5:
As stated by @yamikaitou, this is incorrect. What we're doing here is creating a template file that one can then instantiate multiple units off of. To quote from the documentation for [`systemd.unit`](https://www.freedesktop.org/software/systemd/man/latest/systemd.unit.html)

> As mentioned above, a unit may be instantiated from a template file. This allows creation of multiple units from a single configuration file. If systemd looks for a unit configuration file, it will first search for the literal unit name in the file system. If that yields no success and the unit name contains an "@" character, systemd will look for a unit template that shares the same name but with the instance string (i.e. the part between the "@" character and the suffix) removed.

So with that, when you attempt to start `instancename` using `systemctl start red@instancename`, it is going to find our unit template `red@.service` and instantiate a service named `red@instancename` from it.

I am closing this PR as there is no need to change the current instructions on this

PR #6299, opened 2024-02-28
Fix spelling error in "Multiple Errors Detected" error message.

BlizzardTheWolf:
Fixed a spelling mistake that said "temporally" instead of "temporarily".

Yes!


Kreusada:
I believe this may not be a spelling mistake. Temporarily means "for a limited time; not permanently". _Temporally_ means “with regard to time; having to do with time". These are two different words and _temporally_ works fine here IMO.

BlizzardTheWolf:
> I believe this may not be a spelling mistake. Temporarily means "for a limited time; not permanently". _Temporally_ means “with regard to time; having to do with time". These are two different words and _temporally_ works fine here IMO.

Ah makes sense!

PR #6300, opened 2024-03-02
Update ModLog to include info about `[p]modlogset`

### LINKED ISSUE #6182: Update docs/docstring to make [p]modlogset's new home more clear
Now that `[p]modlogset` has been moved to be under `core_commands.py` with 3.5, it is somewhat unclear that this command exists when looking at the help output for the `Modlog` cog and the [`Modlog` docs](https://docs.discord.red/en/stable/cog_guides/modlog.html). A reference to this command should probably be put in the `Modlog` cog. I would guess that best place for this to go is probably the cog's docstring.

This change might also make sense for `[p]bankset` and the `Economy` cog, though I think that one is a bit less confusing.

BlizzardTheWolf:
Changed the docstrings of the ModLog class to include info about the `[p]modlogset` command, so it's more clear on how to change modlog settings.

Yes


Kreusada (approved):
LGTM

PR #6302, opened 2024-03-06
Fix TypeError from parse_timedelta and parse_relativedelta when months or years is allowed.

NoobInDaHause:
This PR fixes the TypeError from timedelta when you allow months or years. When a user tries to input months and/or years the timedelta throws a TypeError saying that months or years is an invalid keyword argument. I have made it so the months or years is converted to days.

Yes

### DIFF
--- a/redbot/core/commands/converter.py
+++ b/redbot/core/commands/converter.py
@@ -133,6 +133,14 @@ def parse_timedelta(
     ]
     params = _parse_and_match(argument, allowed_units)
     if params:
+        if params.get("years"): # convert years to days
+            years = params.pop("years")
+            params.setdefault("days", 0)
+            params["days"] += 30 * years
+        if params.get("months"): # convert months to days
+            months = params.pop("months")
+            params.setdefault("days", 0)
+            params["days"] += 365 * months
         try:
             delta = timedelta(**params)
         except OverflowError:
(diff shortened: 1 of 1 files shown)


Flame442:
`parse_timedelta` is already documented as only supporting `weeks, days, hours, minutes, seconds` due to the fact that it returns a `timedelta` which only accepts those units. `parse_relativedelta` is documented as accepting `months` and `years` as the `relativedelta` object it returns is able to parse those natively, so there is no need for preprocessing of the data.

I'm not quite sure what this PR looks to address, besides possibly adding support for `months` and `years` in `parse_timedelta`. I don't think that is a good idea however, since the length of a month or year depends on the starting point from the delta (some months are 31 days, others are 30 days, leap years are 366 days, etc). `relativedelta` is designed to handle these edge cases, and this PR prevents it from knowing when specifically `months` or `years` are requested, instead forcing a hardcoded number of days. `timedelta` is not designed to handle `months` or `years`, and currently is documented as such.

NoobInDaHause:
I see, I am not a professional programmer but I just thought of doing this since you can allow "years" and "months" from `allowed_units` and I forgot about the part where months and years have different days. Anyway guess we'll be sticking to it for now thank you for your time flame.

PR #6312, opened 2024-03-21
Bump dependencies

Jackenmen:

This PR does a few things:
- drops `aiodns` dependency
    - While aiohttp does have an aiodns implementation, it has been
disabling it unconditionally for the last 8 years:
https://github.com/aio-libs/aiohttp/commit/9fbb7d708375e0ba01d435c1e8cf41912381c0fc
- bumps all dependencies before release as usual
    - this doesn't include Black 24 bump which affects code style
    - this required setting an upper-bound for pytest-asyncio to `0.21.*` (+ pytest <8 due to incompatibility) - there is a known issue with `0.23.*` on the upstream side for now
- updates scripts for generating dependency lists to consider differences between Python versions

Partially yes, see results:

Test run of updated install instructions:
- macOS x86_64 12 and 13 (14 on GitHub is M1 so same as Cirrus CI): https://github.com/Jackenmen/Red-Install-Tests/actions/runs/8369137984
- everything else: https://cirrus-ci.com/build/4744265834692608

### DIFF
--- a/.github/workflows/run_pip_compile.yaml
+++ b/.github/workflows/run_pip_compile.yaml
@@ -20,12 +20,27 @@ jobs:
       - name: Set up Python 3.8.
         uses: actions/setup-python@v4
         with:
-          python-version: '3.8'
+          python-version: |
+            3.11
+            3.10
+            3.9
+            3.8
 
-      - name: Install dependencies
+      - name: Install dependencies on Linux/macOS
+        if: matrix.os != 'windows-latest'
         run: |
-          python -m pip install -U pip
-          python -m pip install -U pip-tools
+          python3.11 -m pip install -U pip pip-tools
+          python3.10 -m pip install -U pip pip-tools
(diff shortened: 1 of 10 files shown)


Jackenmen:
Postgres is failing hard with asyncpg v0.29.0, it seems.

Jackenmen:
Alright, I figured it out, it had nothing to do with asyncpg and everything to do with the testing framework (or more specifically, pytest-asyncio plugin). I added an upper-bound for relevant packages and we can reevaluate it in the future since it's classified as an upstream issue.

Flame442 (approved):
Technically it isn't against our version guarantees, but removing dependencies has the potential to break 3rd party cogs. `pytz` is currently used by at least Trusty and Aika, though both include it in `requirements.txt`. I didn't do a super exhaustive search, so I don't know if there are others that don't explicitly include it. If that isn't something we are worried about, this can be merged.

PR #6330, opened 2024-03-27
Ask for confirmation when path passed to addpath seems incorrect

Jackenmen:

This is one of the most common issues encountered by beginner cog creators trying to use `[p]addpath`, let's try to nudge them in the right direction!

Suggestions on the message's content welcome, the distance between the question and the buttons doesn't sit right with me but I wasn't sure how to address this. Would it make sense to repeat the question?

Yes

### DIFF
--- a/docs/guide_cog_creation.rst
+++ b/docs/guide_cog_creation.rst
@@ -147,7 +147,7 @@ have successfully created a cog!
         ------ __init__.py
         ------ coolcog.py
     
-    You would then use :code:`[p]addpath D:\red-cogs` to add the path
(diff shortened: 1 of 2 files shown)


yamikaitou:
What about adding on to this to make sure the path isn't within the Data or Install paths?

Jackenmen:
> What about adding on to this to make sure the path isn't within the Data or Install paths?

Good idea, I disallowed anything under data and core paths completely (the install path is part of the data path).

Flame442 (approved):

PR #6342, opened 2024-04-01
[Trivia] Fix descriptions on Star Wars and Star Trek

aikaterna:

We trusted you for the 5 years you sat dormant in your dev position, Jack, waiting for the day when you could secretly try to make this change without anyone noticing. Hordes of rabid fans would have descended upon our support, drowning us in the screams of angry nerds. This must be stopped before it goes any further.

No


Jackenmen (approved):
feisty cat

PR #6346, opened 2024-04-04
Increase default timeout for send_interactive to 60

TrustyJAID:
15 seconds to respond to whether you want a file or more pages from `send_interactive` feels just a bit too short. This increases the default to 60 seconds allowing more time for individuals to recognize what the bot is asking you for and respond.

No


Jackenmen (approved):
I don't see why the timeout should be as low as 15 seconds and the proposed 60 seconds alternative seems fine, let's do this!

PR #6348, opened 2024-04-04
Explain how to remove URL restrictions in the play error message

### LINKED ISSUE #6347: [Audio] Audioset restrict user-facing alert message could be improved

Audio

Currently the bot is set to deny non-commerical links by default, which returns a message to the user that "This URL is not allowed." when trying to use something like an http stream or a Discord file link, leading to quite a few support questions relating to how to play these types of links. 

I feel that this message could be improved in letting the user know that the bot owner can toggle off this restriction, but I'm also not sure if there should be more information presented informing people why this restriction is in place before advising to turn it off.

### Anything else?

Maybe something like this, whether the added message is in the body of the embed, or the footer. My vote is for the embed description as the footer placement looks a bit long.

Unable To Play Tracks
That URL is not allowed.
The bot owner can remove this restriction by using [p]audioset restrict.

DJTOMATO:

Implements explanation of how to enable allowed tracks with the current prefix for the bot.

GXyUJmxu4DHk

Fixes #6347

Yes


Jackenmen:
In the future, when fixing an issue, please use "Fixes #number" so that GitHub detects it and auto-closes the issue when the PR fixing it is merged.

redbot/cogs/audio/core/commands/player.py
```diff
@@ -50,7 +50,9 @@ async def command_play(self, ctx: commands.Context, *, query: str):
                 return await self.send_embed_msg(
                     ctx,
                     title=_("Unable To Play Tracks"),
-                    description=_("That URL is not allowed."),
+                    description=_(
+                        "That URL is not allowed. \n\n The bot owner can remove this restriction by using ``{prefix}audioset restrict``"
```
aikaterna:
```suggestion
                        "That URL is not allowed. \n\n The bot owner can remove this restriction by using ``{prefix}audioset restrict``."
```

Jackenmen (approved):

PR #6349, opened 2024-04-09
Improve timedelta conversions

### LINKED ISSUE #6274: [Mutes] Duration should take the first value provided

3.5.5

### Cog name

Mutes

### Command name

mute

`[p]mute <user_id> 1h spamming 3 copypastas within 2 hours`. 

The first time mentioned is `1h`, which means the intended mute time is 1 hour.

### What actually happened?

Our example: `[p]mute <user_id> 1h spamming 3 copypastas within 2 hours`. 

If you mention a time, even after one has been specified, it will use the _last_ time mentioned. In this case, `2 hours` is used as the duration instead of the intended `1h`.

Per `mutes` help screen, the following examples are provided:
```
[p]mute @member1 @member2 spam 5 hours
[p]mute @member1 3 days
```

I would argue that the first mention of a time should be used as the base for the command's duration. In a `reason` you might need to explain that a certain amount of bad behavior took place within a time duration.

For further explanation, the `mute` help provides this:
`[p]mute <users...> [time_and_reason]`

Being picky, `time` is listed before `reason`. Therefore, the line of thinking is that you provide a time before the reason, even if it does not matter when reviewing listed examples.

1. Load `mutes` cog.
2. Execute a `mute` command as such: `[p]mute <user_id> 1h spamming 3 copypastas within 2 hours`
3. Mute will set duration as `2 hours` instead of the `1h` specified first.

### Anything else?

This would not effect the ability to format mutes in the second example, except it would reverse the behavior (second time is now overwritten by first time mentioned).
`[p]mute @member1 @member2 spam 5 hours`

If no other time is passed before the `5 hours` in the command, this would function just the same. Mostly this bug report is looking to change the behavior to accept the first mention of a time, rather than the last as it would make more sense.

Zephyrkul:

Resolves #6274

- Changes core's timedelta / relativedelta converters to permit units in any order, like the Mutes cog
- Negative timedeltas and relativedeltas are now permitted
  - `minimum=` must be explicitly set to `None` for `parse_timedelta` to permit negative timedeltas.
- Updates Mutes' conversion logic to match only at the beginning of the string, the end of the string, or directly after `t=` / `time=`

Yes

### DIFF
--- a/redbot/cogs/mutes/converters.py
+++ b/redbot/cogs/mutes/converters.py
@@ -1,63 +1,74 @@
+from __future__ import annotations
+
 import logging
 import re
-from typing import Union, Dict
+from typing import Optional, TypedDict
 from datetime import timedelta
+from typing_extensions import Annotated
 
 from discord.ext.commands.converter import Converter
 from redbot.core import commands
 from redbot.core import i18n
-
-log = logging.getLogger("red.cogs.mutes")
-
-# the following regex is slightly modified from Red
-# it's changed to be slightly more strict on matching with finditer
-# this is to prevent "empty" matches when parsing the full reason
-# This is also designed more to allow time interval at the beginning or the end of the mute
-# to account for those times when you think of adding time *after* already typing out the reason
-# https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/core/commands/converter.py#L55
-TIME_RE_STRING = r"|".join(
-    [
-        r"((?P<weeks>\d+?)\s?(weeks?|w))",
-        r"((?P<days>\d+?)\s?(days?|d))",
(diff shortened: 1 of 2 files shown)


Zephyrkul:
[Regex101 link](https://regex101.com/r/1wJBOb/) if you want to test the regex easily.

Zephyrkul:
Minor something worth noting, I've intentionally created an `edgematch` variation to `re.search` to only match the timedelta at the beginning or end of the string as per the original intent of the code. That said, the new modified regex should make it possible to simply `re.search` the string to reflect current behavior, while still fixing all the bugs mentioned in #6274. If whoever reviews would prefer the latter approach, let me know.

TrustyJAID (approved):
This appears to be working as expected and improves the versatility of the converter as well.

PR #6353, opened 2024-04-18
Prevent OverflowError from very large timedeltas in Mutes

TrustyJAID:

Prevent Mute's converter from allowing timedeltas that would result in a datetime leading to an OverFlow error.

Yes


TrustyJAID:
To add some context you can achieve this with `[p]mute user 3650000d`. The timedelta would not Overflow but this timedelta is then added to the current date in Mutes and that would end up Overflowing. I could try/catch all of those but I don't think we'll ever need a timedelta larger than a datetime can support.

Jackenmen (approved):

PR #6366, opened 2024-04-25
Added new optional param to [p]muteset senddm for auto unmute dm

### LINKED ISSUE #6262: Separate muteset senddm for mutes and unmutes

mutes

Currently, `[p]muteset senddm` will send notifications for both mutes and unmutes. This proposes that the unmute dm be toggle-able separately from a mute.

Why? Notifying the user they have been unmuted reminds the user of why they were muted (big angry!), but gives them a first class pass to go complain about it in the server and cause issues immediately. 

Without being notified, likely they will forget about it and return days later with a better perspective, also moving past it.

For non-behavioral reasons, this provides more control to the user to decide what notifications are sent to the user versus modlog only.

### Anything else?

While the first point is an issue to be handled by moderation, I ask what the harm is in thwarting this behavior in the first place by not announcing unmutes to the muted.

turbo-jayce:
* Added a new variable dm_auto_unmute as a toggle
* senddm has new optional parameter that can toggle dm's off for auto unmutes
* if parameter isnt specified it will default to value inputted to setdm 
* allows for dm's to be sent for mutes, and no dm's sent for when they are _automatically_ unmuted

No

### Additional Notes
This is my first PR to repo and unsure if this is the proper way to go about this.

### Issue
Closes #6262

### DIFF
--- a/redbot/cogs/mutes/mutes.py
+++ b/redbot/cogs/mutes/mutes.py
@@ -98,6 +98,7 @@ class Mutes(VoiceMutes, commands.Cog, metaclass=CompositeMetaClass):
             "muted_users": {},
             "default_time": 0,
             "dm": False,
+            "dm_auto_unmute": False,
             "show_mod": False,
         }
         self.config.register_global(schema_version=0)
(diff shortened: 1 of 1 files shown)


Flame442:
I know this is still a draft, but here's some tips about what is currently missing to get you on the right track:
- You need to make a config call to get the value of `dm_auto_unmute` before it can be used as a value.
  - It might make sense to move these checks within `_send_dm_notification` by adding a parameter specifying whether it was called for an auto unmute or a manual unmute.
- Rather than adding to the existing command, you should probably make a separate command for configuring this new setting.
- You need to use the `black` autoformatter to style the updated code, in accordance with the `CONTRIBUTING.MD` file.

Kowlin:
Due to the lack of activity on this PR, I'm gonna go ahead and close this.

PR #6367, opened 2024-04-25
commands.Context object to extract the author id 

### LINKED ISSUE #6281: [Audio] Autoplay failure with User scope playlist

3.5.5

### Cog name

Audio

### Command name

autoplay

I expected a Red playlist saved with the User scope would be able to be used as an autoplay playlist.

### What actually happened?

A User scoped playlist, set in [p]audioset autoplay playlist, is never started on bot start nor does [p]autoplay queue a song from that playlist.

In running in verbose mode, this traceback is offered on bot start or while using [p]autoplay.
```
VERBOSE [red.cogs.Audio.api.AudioAPIInterface] Failed to fetch playlist for autoplay
Traceback (most recent call last) 
/Users/aikaterna/red38/lib/python3.8/site-packages/redbot/cogs/audio/apis/interface.py:938 in autoplay
❱ 938 playlist = await get_playlist(
/Users/aikaterna/red38/lib/python3.8/site-packages/redbot/cogs/audio/apis/playlist_interface.py:392 in get_playlist
❱ 392 raise RuntimeError(f"That playlist does not exist for the following scope: {scope}") 
RuntimeError: That playlist does not exist for the following scope: USERPLAYLIST
```

User scoped playlists are saved in the SQL db as the user ID as the value in the scope_id column. However, when loading playlists with the get_playlist function, we are passing in the bot's ID instead of the playlist owner's user ID, which causes the lookup to fail on User scoped playlists.

https://github.com/Cog-Creators/Red-DiscordBot/blob/76abb7cab2c28bda3ad5e939b3d07250cdd99d7f/redbot/cogs/audio/apis/interface.py#L940

The autoplaylist config value we're looking at to start that get_playlist lookup only offers the playlist ID, so not sure if we should be doing another lookup at that point to fetch the true author instead of assuming it's the bot ID.

1. Have the bot running in verbose debug mode to view the traceback
2. Set a User scoped playlist as the playlist in [p]audioset autoplay playlist
3. Use [p]autoplay or set autoplay to start with [p]audioset autoplay toggle and then restart the bot

### Anything else?

Additional context:
The fact that user playlists don't work for auto play is intentional - because auto play is a guild level feature meaning guild and global scoped playlists should work for it, but user not so, as we may exposing user data to users in a server that the playlist author/manager is no longer in.
There should be an error show to the user when trying to set a user scoped playlist as an autoplay playlist.

sxliang3093:
I passed a commands.Context object to extract the author id. I changed the arguments of the autoplay function to encompass this change. This is for issue #6281 

Closes #6281

No

### DIFF
--- a/redbot/cogs/audio/apis/interface.py
+++ b/redbot/cogs/audio/apis/interface.py
@@ -921,7 +921,7 @@ class AudioAPIInterface:
                 )
         return results, called_api
 
-    async def autoplay(self, player: lavalink.Player, playlist_api: PlaylistWrapper):
+    async def autoplay(self, ctx: commands.Context, player: lavalink.Player, playlist_api: PlaylistWrapper):
         """Enqueue a random track."""
         autoplaylist = await self.config.guild(player.guild).autoplaylist()
         current_cache_level = CacheLevel(await self.config.cache_level())
@@ -937,7 +937,7 @@ class AudioAPIInterface:
                     self.bot,
                     playlist_api,
                     player.guild,
-                    player.guild.me,
+                    ctx.author.id
                 )
                 tracks = playlist.tracks_obj
             except Exception as exc:


aikaterna:
Check out the additional context section I wrote on the linked issue. This probably isn't the way this should be resolved.

Flame442:
I know this is still a draft, but here's some tips about what is currently missing to get you on the right track:
- You aren't actually passing `ctx` to the `autoplay` method, you changed its signature without changing any of the places that method is called.
  - This logic wouldn't work for resuming autoplay on bot start, not from a command.
  - Additionally, the notes of the issue you are resolving request a different solution for this issue:
    > The fact that user playlists don't work for auto play is intentional - because auto play is a guild level feature meaning guild and global scoped playlists should work for it, but user not so, as we may exposing user data to users in a server that the playlist author/manager is no longer in.
**There should be an error show to the user when trying to set a user scoped playlist as an autoplay playlist.**
- You need to use the `black` autoformatter to style the updated code, in accordance with the `CONTRIBUTING.MD` file.

Kowlin:
Due to the lack of activity on this PR, I'm gonna go ahead and close this.

PR #6373, opened 2024-05-05
Use YouTube source plugin over the deprecated built-in source

Jackenmen:

This PR switches our auto-generated Lavalink `application.yml` file to use the YouTube source plugin over the deprecated built-in source. The proposed implementation does not define this plugin in the Config defaults under the `yaml` key and instead dynamically generates it in the `ServerManager.process_settings()` method. The upsides here are:
- we can keep the existing (user-configurable) config value for the YouTube source and use that to determine if the plugin should be enabled
- we won't accidentally enable both the built-in source and the plugin
- the YT plugin version constant can be set right under the LL version, making it easier for us to update

The downside is that the defaults are no longer placed in a single spot. It is not ideal but there's a need for some redesign to support plugins anyway so I figure this is likely a better solution, at least for now.

Yes


aikaterna:
Do you think it would be worth having an info level log line on what plugins it's downloading, similar to other log lines exposed on audio/Lavalink.jar startup? Thinking of any potential support info needed there.

redbot/cogs/audio/manager.py
```diff
@@ -295,11 +299,14 @@ def __ge__(self, other: object) -> bool:
 
 class ServerManager:
     JAR_VERSION: Final[str] = LavalinkVersion(3, 7, 11, red=3)
+    YT_PLUGIN_VERSION: Final[str] = "1.1.0"
```
aikaterna:
```suggestion
    YT_PLUGIN_VERSION: Final[str] = "1.4.0"
```

aikaterna (changes requested):
This looks good to go after bumping the plugin version.

aikaterna (approved):

PR #6375, opened 2024-05-06
Fix an issue with autocomplete in ignored channels/servers

### LINKED ISSUE #6374: Check responding to autocomplete interaction: 400 Bad Request - Value must be one of {8}.

3.5.9

Presumably get autocomplete results for something

The results to come through properly

### What actually happened?

The exception seems like it was handled and just logged to sentry, although i'm not certain as I cant seem to repro it on purpose.

I am not sure how to repro this error, it has occurred 19 times within the last 30 days during which i was unaware.

## Raw Stack Trace
```
HTTPException: 400 Bad Request (error code: 50035): Invalid Form Body
In type: Value must be one of {8}.
  File "discord/app_commands/tree.py", line 1089, in wrapper
    await self._call(interaction)
  File "discord/app_commands/tree.py", line 1210, in _call
    if not await self.interaction_check(interaction):
  File "redbot/core/tree.py", line 344, in interaction_check
    await interaction.response.send_message(
  File "discord/interactions.py", line 801, in send_message
    await adapter.create_interaction_response(
  File "discord/webhook/async_.py", line 221, in request
    raise HTTPException(response, data)
```
### Breadcrumbs
image

image

Some insight from Zeph: check is responding to an autocomplete interaction, which only supports `APPLICATION_COMMAND_AUTOCOMPLETE_RESULT` (internally represented by the value 8 to Discord's API) responses

TrustyJAID:
Resolves #6374 by replacing the text response on an autocomplete interaction with choices stating why the command won't function properly

Yes

image

### DIFF
--- a/redbot/core/tree.py
+++ b/redbot/core/tree.py
@@ -334,6 +334,20 @@ class RedTree(CommandTree):
         else:
             log.exception(type(error).__name__, exc_info=error)
 
+    async def _send_interaction_check_failure(
+        self, interaction: discord.Interaction, message: str
+    ):
+        """Handles responding to interaction check failures.
+        Mainly used for when an interaction is an autocomplete and
+        providing the message in the autocomplete response.
+        """
+        if interaction.type is discord.InteractionType.autocomplete:
+            await interaction.response.autocomplete(
+                [discord.app_commands.Choice(name=message[:80], value="None")]
+            )
+            return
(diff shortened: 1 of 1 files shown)


Flame442:
Does this also need to be added for the whitelist/blacklist? Does it make more sense to not use `interaction_check` to block in cases where it is just for autocomplete, not for a full command?

Flame442 (approved):

PR #6382, opened 2024-05-23
Fix info.json keys in approved CC guide

Flame442:
Like many people, I simply referenced other people's cogs to make my `info.json`s, and an extra phantom key from the V2 days made its way into our docs. I removed the `name` key from both the repo and cog requirements, and added the conditional `min_python_version` key since lots of cogs lately have needed that one.

Technically the cookie cutter repo and the policies should also be updated, but I don't want to do that :(

Yes

### DIFF
--- a/docs/guide_cog_creators.rst
+++ b/docs/guide_cog_creators.rst
@@ -41,17 +41,16 @@ Any Cog Creator that does not follow these requirements will have their repo rem
 - Repo-wide ``info.json`` file with the keys
 
   - ``author``
-  - ``name``
   - ``short``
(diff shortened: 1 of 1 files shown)


Jackenmen:
> Technically the cookie cutter repo and the policies should also be updated, but I don't want to do that :(

Arguable - it's still an existing convention to put cog's class name in the `name` key, even if it doesn't currently do anything. I agree that it doesn't really make sense to have it listed in the approved CC guide though.

Jackenmen (approved):

PR #6385, opened 2024-05-31
fix for audio disconnect command not checking perms 

### LINKED ISSUE #6365: [Audio] `disconnect` command should check if the user is actually in the VC

3.5.9

### Cog name

Audio

### Command name

disconnect

- Just like the `summon` command - the `disconnect` cmd should **also** check if the user (who is invoking the command) actually is in the VC before letting them disconnect the bot from the VC - where it's already playing songs **along with** other humans.
image

### What actually happened?

It lets anyone **who is not in the VC**, run the command resulting in the bot getting disconnected from the VC, leading to an unpleasant experience.

1. Connect to a VC 
2. Set the vote percentage to 0 by `audioset vote 0`
3. Summon the bot to join
4. Ask any user or `<p>mock` anyone to run `<p>disconnect`
Result - It should let that user x disconnect the bot without having to be there in the VC.
> The `summon` command ONLY works if you are in the VC - likewise, the disconnect should also work if the user is in the VC, regardless of the status of the `audioset vote`.

### Anything else?

N/A

BenCos17:

fixes #6365

Yes


Flame442:
I think in cases where nobody is in the voice channel, it would make sense to allow someone not in the voice channel to disconnect.
Additionally, you are currently failing the style checks. Please format your PR with `black -l 99` using the instructions in our [CONTRIBUTING.MD](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/CONTRIBUTING.md#43-style) file.

BenCos17:
> I think in cases where nobody is in the voice channel, it would make sense to allow someone not in the voice channel to disconnect. Additionally, you are currently failing the style checks. Please format your PR with `black -l 99` using the instructions in our [CONTRIBUTING.MD](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/CONTRIBUTING.md?rgh-link-date=2024-05-31T21%3A38%3A00Z#43-style) file.

fixed the style stuff btw

BenCos17:
And that should all work 
I've tested it all on my red instance for the past while with no issues

Flame442:
I'm not certain if the current implementation needs to be so complicated. [`self._can_instaskip`](https://github.com/Cog-Creators/Red-DiscordBot/blob/33e0eac741955ce5b7e89d9b8f2f2712727af770/redbot/cogs/audio/core/utilities/player.py#L83) seems to be the utility method used by `Audio` to account for all of the kinds of permissions a user may need to take actions on the player. It would likely be better to use that method instead.

BenCos17:
> I'm not certain if the current implementation needs to be so complicated. [`self._can_instaskip`](https://github.com/Cog-Creators/Red-DiscordBot/blob/33e0eac741955ce5b7e89d9b8f2f2712727af770/redbot/cogs/audio/core/utilities/player.py#L83) seems to be the utility method used by `Audio` to account for all of the kinds of permissions a user may need to take actions on the player. It would likely be better to use that method instead.

true 
I'll have a look at that at some point then 
thanks

Jackenmen:
Superseded by #6768

PR #6387, opened 2024-06-09
Switch to orjson, fallback to stdlib json

japandotorg:
To optimize performance we should utilize orjson, which is already included in [`requirements/base.in`](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/requirements/base.in#L9), but unused in Red.

I have been using a modified version of Red with orjson (replacing stdlib json, no fallback) without any problem and have noticed performance gains so far, I primarily used `orjson.dumps` and `orjson.loads`, while other methods like `json.dump`, `json.load`, `aiohttp.ClientSession(json_serialize=...)` remained unchanged using stdlib json, also never in audio before.

While I don't think a full switch to orjson is necessary, but I think we should give orjson a chance.

No

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -231,6 +231,7 @@
   - redbot/core/_global_checks.py
   - redbot/core/_settings_caches.py
   - redbot/core/_sharedlibdeprecation.py
+  - readbot/core/_json.py
   - redbot/core/utils/_internal_utils.py
   # Tests
   - redbot/pytest/__init__.py
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -231,7 +231,7 @@
   - redbot/core/_global_checks.py
   - redbot/core/_settings_caches.py
   - redbot/core/_sharedlibdeprecation.py
-  - readbot/core/_json.py
+  - redbot/core/_json.py
   - redbot/core/utils/_internal_utils.py
   # Tests
   - redbot/pytest/__init__.py
--- a/redbot/__main__.py
+++ b/redbot/__main__.py
@@ -6,7 +6,6 @@ _early_init()
 import asyncio
 import functools
 import getpass
-import json
 import logging
 import os
 import pip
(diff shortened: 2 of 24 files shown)


Jackenmen:
> It raises `JSONEncodeError` on an integer that exceeds 64 bits by default or, with `OPT_STRICT_INTEGER`, 53 bits.
> **Source:** https://github.com/ijl/orjson?tab=readme-ov-file#serialize

This makes it unsuitable for Config. Additionally, the support for dataclasses and other types would make it harder to migrate back.

japandotorg:
yikes, well that's a bummer.

Is there anything that can be done? or maybe an alternative?

Jackenmen:
Nothing can really be done about this in the current Config API. If there's ever a new one, probably, though I don't know if a new one would even have a JSON driver in the first place.

You could probably sprinkle some orjson usage in other places in Core, just not anything to do with Config. I don't know if it will really give any significant gains though, I don't think we use JSON outside of Config much.

Other than that, I glanced at the diff and I just want to say that there's no point in the `_json` module - `orjson` is not an optional dependency of Red.

japandotorg:
ah I see

> You could probably sprinkle some orjson usage in other places in Core, just not anything to do with Config. I don't know if it will 
really give any significant gains though, I don't think we use JSON outside of Config much.

for now I'll just look into this for this PR, if the org thinks that this is really unnecessary than the PR can be closed

> It raises JSONEncodeError on an integer that exceeds 64 bits by default or, with OPT_STRICT_INTEGER, 53 bits.

perhaps can we use a fallback to `json` incase this happens?

Jackenmen:
> perhaps can we use a fallback to `json` incase this happens?

That is an interesting idea though there are more cases where orjson differs in serialization and that could lead to introducing discrepancies between drivers (and even within the driver in case of the fallback) - for example it serializes UUID types while normal JSON would fail at that point.

Jackenmen:
Per https://github.com/Cog-Creators/Red-DiscordBot/pull/6387#issuecomment-2156377027, we cannot really use this, closing.

PR #6388, opened 2024-06-10
Docs: remove Atom from the list of recommended editors

cswimr:

Atom has been deprecated/archived since December 15 of 2022, and https://atom.io/ now redirects to a GitHub blog post detailing the deprecation.  
I don't think it's a good idea to be recommending software that hasn't been officially supported in over a year, when there's far superior options available (such as the other recommended editors).

No


cswimr:
Alternatively, if we (for whatever reason) would prefer to keep Atom in the docs, we should at least change the url to [the one that actually lets you download the editor](https://atom-editor.cc/) instead of just being a redirect to the GitHub deprecation article.

Jackenmen (approved):
Thanks!

PR #6397, opened 2024-07-08
[RedTree] add `UserFeedbackCheckFailure` for `app_commands`

japandotorg:

Add a `UserFeedbackCheclkFailure` exception class with configurable message which is handled within `RedTree.on_error`

No


redbot/core/tree.py
```diff
@@ -331,6 +332,9 @@ async def on_error(
             await self._send_from_interaction(
                 interaction, _("You are not permitted to use this command.")
             )
+        elif isinstance(error, UserFeedbackCheckFailure):
+            if error.message:
+                await self._send_from_interaction(interaction, error.message)
```
Flame442: This must be before the `CheckFailure` branch, or that branch will trigger first due to `UserFeedbackCheckFailure` being a subclass.

redbot/core/app_commands/errors.py
```diff
@@ -0,0 +1,12 @@
+"""Errors module for the app_commands package."""
+from discord import app_commands
+
+__all__ = ("UserFeedbackCheckFailure",)
```
Flame442: This package will not be public, so `__all__` should not be used. This class name should remain in the `__all__` of `app_commands.__init__`, as it is only guaranteed to exist when imported directly from `redbot.core.app_commands`.

> Lack of `__all__` in the module means that all of its names are considered private APIs.

https://docs.discord.red/en/stable/version_guarantees.html#developer-guarantees

Flame442 (changes requested):

japandotorg:
that should be it, requested changes have been made.

redbot/core/app_commands/errors.py
```diff
@@ -0,0 +1,10 @@
+"""Errors module for the app_commands package."""
+from discord import app_commands
+
+
+class UserFeedbackCheckFailure(app_commands.CheckFailure):
+    """A generic version of CheckFailure."""
```
Jackenmen: The docstring from `commands.CheckFailure` makes more sense than this one:
```suggestion
    """A version of CheckFailure which isn't suppressed."""
```

japandotorg: i was gonna use that as well, but unlike `Red.on_command_error` `app_commands.CheckFailure` isn't really suppressed on `RedTree.on_error` https://github.com/Cog-Creators/Red-DiscordBot/blob/efc99c5dceda82537f75c24f734d4fb4dd4a8e11/redbot/core/tree.py#L336

japandotorg: would there be something more appropriate in this case?

Jackenmen: I see... Maybe something like this?
```suggestion
    """A version of CheckFailure responding with a custom error message."""
```

japandotorg: I suppose this would be a better one for now.

Flame442 (approved):
image

PR #6413, opened 2024-07-21
Get viewer count from stream data instead of user profile data

anopem:
Removed old viewer count from user profile data and added it with stream data instead

Yes


palmtree5 (approved):
Double-checked against Twitch's docs and this looks good to me!

PR #6429, opened 2024-08-08
[Audio] Fix trying to send notify message with no channel object

aikaterna:
A user in Red support submitted this traceback.

image

The player object has a channel reference inside the extra information telling it where to display messages and the channel was deleted or unreachable. This change checks for a channel object before trying to fetch its permissions.

Yesn't

I tested this change and checked for the _has_notify_perms condition returning False when an audioset notify channel is set to a deleted channel. After this change the player continues playing without throwing an error when a track changes and a notify message or autoplay message needs to be sent. There are a lot of points where interacting with audio and sending a command will readjust the notify channel saved in the player, so there isn't much of a need to clear that non-existent channel object as it will do it on its own with most usage.

If someone else would like to do any kind of sanity check on this please feel free. Otherwise I'm mostly confident I didn't break more things. 🤞

### DIFF
--- a/redbot/cogs/audio/core/events/cog.py
+++ b/redbot/cogs/audio/core/events/cog.py
@@ -198,6 +198,8 @@ class AudioEvents(MixinMeta, metaclass=CompositeMetaClass):
         if not guild:
             return
         notify_channel = guild.get_channel_or_thread(player.fetch("notify_channel"))
+        if not notify_channel:
+            return
         has_perms = self._has_notify_perms(notify_channel)
         tries = 0
         while not player._is_playing:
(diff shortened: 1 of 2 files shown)


Jackenmen (changes requested):
1. This changes the return type from `bool` to `Optional[bool]` and it's arguable whether it's more logical to return true or false from this function.
2. The type hint for the `channel` argument already did not allow `None`. Presumably, this was intentional and the issue should be fixed at the call site instead. All but one of the calls to this method already check that the notify channel is not None before calling the method. The outlier can be found at:
https://github.com/Cog-Creators/Red-DiscordBot/blob/903992f48a5e616ebc0efbc21c7b832e4258782c/redbot/cogs/audio/core/events/cog.py#L201

Kowlin (dismissed):
LGTM

Jackenmen (approved):
LGTM.

PR #6430, opened 2024-08-09
Generate default LL server config and attach it to GH release

Jackenmen:

There's few things going on here:
- I split out Audio's manager module slightly - the LL server config (application.yml) and LL managed node version (along with helper classes) functionality are now submodules of `redbot.cogs.audio.manager`
    - I guess it's arguable whether this is in any way better than before (aside from the newly added `generate_server_config()` function) but the purpose was to minimize how much gets imported just to be able to generate the default LL server config
- Package distributions (binary wheel and source dist) are now attached to the GH release once they're built
- Aaand, the reason this PR exists in the first place... We now generate default application.yml during the release publishing process

It's important to note that the first bullet point above changed the location of pinned versions used to start the managed node - now the versions are placed in `redbot/cogs/audio/manager/version_pins.py` (CC: @aikaterna).

Yes

...sort of - I tested changes to Audio *and* the new script but I did not test the GH Actions workflow itself. We'll do that on next release :)


Kowlin (approved):
Yep works like a charm, lets see if GHActions breaks 🙃

PR #6443, opened 2024-09-01
[Downloader] Support `[botname]` substitutions in cog install messages

Kreusada:

Support `[botname]` substitutions in downloader install messages. `[p]` is already supported, it would make sense to support `[botname]` too and to prevent confusion.

Yes


Kowlin (approved):
LGTM, but once we're starting to add more then 2 arguments to replace we really gotta consider a helper function to handle that sort of stuff for us.

PR #6444, opened 2024-09-01
Fix `header` utility docstring

Kreusada:

Fixes the incorrectly named argument in the header function's docstring. #6102


Kowlin (approved):
The one time I don't check doc strings 😅

PR #6457, opened 2024-09-24
Fix permissions fetching for User Installable Bots

Kowlin:

This change fixes permissions regarding discord.Interactions and fetching permissions. Where this would fail where a bot isn't in the server (Yep, that's a thing now!) or doesn't have the channels cached in memory with interactions on User Installable Bots (UIBs) This fix pulls the permissions directly from the interaction, and gives the allowed permissions the interaction can use directly.

This will however sucks for bots trying to see if a person can ignore the ignore list. As its unclear to the bot if the author has manage_guild or similar permissions. But I see no other viable method of solving this.

Partially? (Lmao)


redbot/core/bot.py
```diff
@@ -886,7 +886,13 @@ async def ignored_channel_or_guild(
...
                 raise TypeError("Can't check permissions for non-private PartialMessageable.")
             is_private = True
-        perms = ctx.channel.permissions_for(author)
+        if isinstance(ctx, discord.Interaction):
+            ctx: discord.Interaction
+            perms = ctx.permissions
+            # The permissions here are always linked to the interaction author &
+            # the channel it originated from.
+        else:
+            perms = ctx.channel.permissions_for(author)
```
Jackenmen: `Context` can also be based on an interaction:
```suggestion
        if isinstance(ctx, discord.Message):
            perms = ctx.channel.permissions_for(author)
        else:
            # `permissions` attribute will use permissions from the interaction when possible,
            # or `ctx.channel.permissions_for(author)` for non-interaction contexts.
        	perms = ctx.permissions
```

Jackenmen (changes requested):

Jackenmen (dismissed):

Jackenmen (approved):

SomewhatDamaged:
Huge thanks for this and all the work you do. Upgraded and my slash commands now work in threads (where the bot isn't).

\o/

🎉

PR #6463, opened 2024-10-08
Convert project to poetry

tekulvw:

Converting setup.py + requirements/* to a modern package manager like poetry will bring benefits in dev (simple setup: `poetry install`) and versioning (no custom scripting, just: `poetry set-git-version`).

No


Jackenmen:
We don't currently have any plans to switch to Poetry.
*Some* context is available at https://discord.com/channels/133049272517001216/1293005886389751849, though there are definitely more things that just haven't been written down.
We might start using *some* tool for project management *like* Poetry eventually, but it probably won't be Poetry. Some other choices are hatch, pdm, and uv (the last one has recently been acquired by OpenAI, which makes me wary, though). With uv becoming popular, Poetry is no longer the only Python installer with a universal resolver - uv does too and is A LOT faster than Poetry (though tbf, Poetry can even be slower than pip when it comes to resolution, at least in my limited experience) and there's also some effort being made towards writing a pure Python universal resolver based on PubGrub algorithm (same as uv) that might make its way into pip: https://github.com/notatallshaw/nab

Anyhow, I just don't want to keep this open forever in limbo when I know it's not the direction I want Red to take. If something changes, this can always be revisited.

PR #6468, opened 2024-10-28
guild id most be integer.

kpopdev:
When the input is a string, you will get None. To fix this, it should be an integer.

data = await self.bot.get_cog("something").config.guild(self.bot.get_guild(channel_name)).channel_count.set(how_many_users) File "/root/redenv/lib/python3.11/site-packages/redbot/core/config.py", line 991, in guild return self._get_base_group(self.GUILD, str(guild.id))
                                        ^^^^^^^^

yes
Choose one (remove the line that doesn't apply):
-->
Yes
No

### DIFF
--- a/redbot/core/config.py
+++ b/redbot/core/config.py
@@ -988,7 +988,7 @@ class Config(metaclass=ConfigMeta):
             The guild's Group object.
 
         """
-        return self._get_base_group(self.GUILD, str(guild.id))
+        return self._get_base_group(self.GUILD, int(guild.id))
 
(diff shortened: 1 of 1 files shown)


Jackenmen:
The changes in this PR are unnecessary (and they will break things, as showcased by the tests).

It would seem that you misinterpreted the traceback you got - what you did is you passed `None` to `config.guild()` when what it's expecting is `discord.Guild`. You likely have some function or property that can return `discord.Guild` or `None` and in the case you ran into it, it returned `None` so when you tried using it with `Config`, it failed.

Issue #6445, opened 2024-09-02
Implement warning by user ID.


RdrSeraphim:
### What component of Red (cog, command, API) would you like to see improvements on?

warnings

### Describe the enhancement you're suggesting.

It seems that this was once behavior for the cog a long while ago, but I would like to propose allowing warns by user ID. It would be useful to warn IDs of users who, say, violated a rule but left before any warning could be given.

### Anything else?

I'm not sure how this would intersect with warn actions, I'm not sure how the cog would perform these on absent users. I would also be willing to try and take this enhancement on if that would be preferable. My current line of thought would be to replicate the member param validation from `[p]warnings` and ignore DMing if the member isn't present.

Needless2Say:
Hi, can I take this?

aikaterna:
We don't assign non-members to issues. Feel free to work on it if you wish.

Flame442:
At least as far back as the start of V3, `warnings` has always taken a `Member` as the argument ([ref](https://github.com/Cog-Creators/Red-DiscordBot/commit/c1ac78eea4f036c377f70799112cd544e3a14ee5#diff-1fa0574b7404468882355fe1f64fa85f3efec457da84180904112ddb54f41d17R222)). Perhaps you are thinking of `[p]ban` or similar commands?

As you pointed out, this seems a little hard to address given the nature of a warnings based cog. `warnings` allows actions to run any arbitrary command on the user, and cannot guarantee that the command will succeed if given an ID of a user who has left the server. It would make sense to allow a `[p]warn` that would give enough points to execute the `[p]ban` command to be used on a user who has left the server, since a user who is at that tier should be banned regardless of if they try to rejoin the server, and the ban command is able to accept IDs of members not in the server. However, something like `[p]timeout` which *requires* the user to be in the server would need to wait for them to rejoin (resulting in the timeout length not starting until after they rejoined, for better or for worse). There is no easy way for `warnings` to tell the difference between these cases and handle them individually. The only option that could work is always waiting until the user rejoins, but that requires the bot to be online when the user joins for them to be punished.

I am not certain if that added complexity is worth the effort for this edge case. If a disruptive user is leaving before a moderator is able to take action, it may often make sense to consider them a troll and bypass the server's standard `warnings` track.

Ascensionn:
[...]

PR #6476, opened 2024-11-20
implemented warn by userID

Ascensionn:

As per #6445, added feature to make reddiscordbot be able to warn a user by their userID or their name. It was a simple short change

image
image

### DIFF
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -373,7 +373,7 @@ class Warnings(commands.Cog):
     async def warn(
         self,
         ctx: commands.Context,
-        member: discord.Member,
+        identifier: str, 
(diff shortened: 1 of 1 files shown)


Kowlin:
Hey there,
Thank you for your contribution, however. This functionality is already build in the handler of Discord.py's ``discord.Member`` argument. Where it will check if the author of a command has included a mention, display name, full username or user ID. Therefor this PR is not required.

Flame442:
The linked issue is specific to the case where the user has left the server. In that case, the `Member` converter will not work with an ID, as it only matches users that are currently in the server. The fix for the issue requires additional consideration, as the command expects a fully resolved `Member` object that would not be available if the user is no longer in the server. This change, using `get_member`, also relies on the user being in the server.

PR #6477, opened 2024-11-22
Allow enforcing reason to be filled in Mod cog commands

### LINKED ISSUE #6105: Option to classify reason in moderation actions as required

Mod

An option to classify the reason arg as required. I believe this would be a beneficial option to add to the modularity of Red.

### Anything else?

Reason being, we are trying to move to Red for all moderation uses and being able to require a reason would be nice to ensure its properly logged and not require to use the reason command afterwards.

kevin1015wang:

This PR adds an option to classify reason in moderation actions as required. Users are able to use `!modset requirereason true` to specify if a reason is required for running any moderation action. It adds the requirement to all moderation commands including `ban`, `kick`, `massban`, `tempban`, `softban`, `voicekick`, `voiceunban`, `voiceban`, and `unban`. If a user runs a mod command when the reason setting is enabled but doesn't include a reason, the bot reasons by telling the user that a reason is required.

Closes #6105

Yes, I manually tested my changes by running each moderation command with and without the requirement. The bot successfully executed the command when a reason is provided and rejects the command if a reason isn't provided when reasons are required.

### DIFF
--- a/docs/cog_guides/mod.rst
+++ b/docs/cog_guides/mod.rst
@@ -268,6 +268,28 @@ and reason as to why they were kicked/banned.
 
 .. _mod-command-modset-hierarchy:
 
+"""""""""
+modset reason
+"""""""""
+
+**Syntax**
+
+.. code-block:: none
+
+    [p]modset reason [enabled]
+
+**Description**
+
+Toggle whether a reason is required for mod actions.
+        
+If this is enabled, the bot will require a reason to be provided for all mod actions.
+
+**Arguments**
+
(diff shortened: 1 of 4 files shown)


Zephyrkul:
It would probably be faster to read the `mod_reason` entry from config only after checking if `reason` is `None`, otherwise Red would be reading from disk and immediately discarding the result if `reason` is provided.

Jackenmen (approved):
Looks good! Just made a few nitpicky adjustments.

PR #6480, opened 2024-11-24
Add support for SimpleMenu to use custom select options

### LINKED ISSUE #6455: Allow custom select options in SimpleMenu
### Type of feature request

API functionality

SimpleMenu should allow custom names for the pages instead of the hardcoded "Page X".
The most straight forward solution I can see is to take an optional list of names when creating the menu and if its given, use it instead of generating the "Page X" list.
My cog currently "hacks" the existing behavior using 
```python
menu.select_options = [discord.SelectOption(label=n, value=num) for num, (n,_) in enumerate(menu_items)]
menu.select_menu = menu._get_select_menu()
menu.add_item(menu.select_menu)
```
(menu_items is a list of tuples of (name,content) in this case)
### Anything else?
I can write a pull request myself if needed.

TrustyJAID:

Add the ability to customize the select options on SimpleMenu before sending. This Closes #6455.

Yes

### DIFF
--- a/redbot/core/utils/views.py
+++ b/redbot/core/utils/views.py
@@ -269,6 +269,11 @@ class SimpleMenu(discord.ui.View):
                 Send the message ephemerally. This only works
                 if the context is from a slash command interaction.
         """
+        if self.use_select_menu and self.source.is_paginating():
+            self.remove_item(self.select_menu)
+            # we added a default one in init so we want to remove it and add any changes here
+            self.select_menu = self._get_select_menu()
+            self.add_item(self.select_menu)
         self._fallback_author_to_ctx = True
         if user is not None:
             self.author = user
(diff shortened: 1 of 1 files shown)


Jackenmen (commented):
Won't the `remove_item` + `add_item` change the item order, if `use_select_only` is True?

TrustyJAID:
Do you mean the order of the components, i.e. where the select menu appears in the grid? If so then we don't offer a guarantee on that with this view. They appear in order preferring buttons then the menu last. If someone were to add their own components those would usually come after but could now come before the select menu. This PR is largely just so what was asked can be achieved without accessing private attributes. If people want complete control on the order of the components they should make their own view.

Jackenmen:
I think the order we define in `__init__`:
https://github.com/Cog-Creators/Red-DiscordBot/blob/74b6fdeba01aa49f9714fb4b7993f323479be29c/redbot/core/utils/views.py#L196-L202
will no longer do anything since `start()`/`start_dm()` immediately changes it.

TrustyJAID:
Ah that order, that was a special case so that a close button appeared under the select menu when applied. I can fix the logic so that is accounted for.

TrustyJAID:
Actually upon testing it is working as expected. This is because when we initially define each of the components their row is set automatically and when removed that spot is now free to be refilled. So the order does not change when we define it in init.

Jackenmen:
I see. I don't really have any other complaints so...

Jackenmen (approved):

PR #6481, opened 2024-11-26
`[p]warn` ask to ban when user not in server

### LINKED ISSUE #6445: Implement warning by user ID.

Ascensionn:

image

This is to follow up with what @Flame442 said: "If a disruptive user is leaving before a moderator is able to take action, it may often make sense to consider them a troll and bypass the server's standard warnings track."

Instead of warning them, I check if they are in the discord guild; if they are not, then they will be banned. 

Closes #6445

image

Note: the command here is "[p]wasd" because I tested it by implementing a new temporary cog.

### DIFF
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -402,10 +402,7 @@ class Warnings(commands.Cog):
                     view=confirm,
                 )
                 await confirm.wait()
-                if confirm.result is None:
-                    await ctx.send("No response received within 10 seconds. No action taken.")
-                    return
-                elif confirm.result:
+                if confirm.result:
                     await ctx.guild.ban(user_obj, reason=reason)
                 else:
                     confirm.message = await ctx.send(_("No action taken."))
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -373,7 +373,7 @@ class Warnings(commands.Cog):
     async def warn(
         self,
         ctx: commands.Context,
-        identifier: str, 
+        identifier: str,
         points: UserInputOptional[int] = 1,
         *,
         reason: str,
@@ -389,7 +389,7 @@ class Warnings(commands.Cog):
         member = None
         user = None
         """User can be warned by ID or warned by their name."""
-        
+
         if identifier.isdigit():
             member = ctx.guild.get_member(int(identifier))
             # await ctx.send("Got member by ID")
@@ -397,22 +397,41 @@ class Warnings(commands.Cog):
             member = ctx.guild.get_member_named(identifier)
             # await ctx.send("Got member by name")
         if not member:
-            await ctx.send(f"User with ID `{identifier}` was not found in discord guild.")
+            await ctx.send(f"User `{identifier}` not found in discord guild")
 
             # Because he has not been found, he will be banned.
             try:
                 user = await self.bot.fetch_user(int(identifier))
-                await ctx.send(f"User `{user.name}` is not in the server but has been found globally.")
+                await ctx.send(
+                    f"User `{user.name}` is not in the server but has been found globally. Would you like to ban them instead? [Y/N]"
+                )
+                try:
+
+                    def check(m):
+                        return (
+                            m.author == ctx.author
+                            and m.channel == ctx.channel
+                            and m.content.lower() in {"y", "n"}
+                        )
+
+                    response = await self.bot.wait_for("message", check=check, timeout=30.0)
+                    if response.content.lower() == "y":
+                        await ctx.guild.ban(user, reason=reason)
+                        await ctx.send(f"User `{user.name}` has been banned for: {reason}")
+                    elif response.content.lower() == "n":
+                        await ctx.send("No action taken.")
+                    else:
+                        # This should not happen due to the check, but it's good to cover edge cases
+                        await ctx.send("I do not recognize that answer. No action taken.")
+                except asyncio.TimeoutError:
+                    await ctx.send("No response received. No action taken.")
                 await ctx.guild.ban(user, reason=reason)
-                await ctx.send(f"User `{user.name}` has been banned for: {reason}")
             except discord.Forbidden:
                 await ctx.send("I don't have permission to ban this user.")
             except discord.NotFound:
                 await ctx.send(f"User with ID `{identifier}` not found globally.")
                 return
-
             return
-        
         if member == ctx.author:
             return await ctx.send(_("You cannot warn yourself."))
         if member.bot:
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -391,6 +391,9 @@ class Warnings(commands.Cog):
         if isinstance(user, discord.Member):
             member = user
         elif isinstance(user, int):
+            if not ctx.channel.permissions_for(ctx.guild.me).ban_members:
+                await ctx.send(_("User {user} is not in the server."),format(user=user))
+                return
             user_obj = self.bot.get_user(user) or discord.Object(id=user)
             try:
                 confirm = ConfirmView(ctx.author, timeout=10)
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -373,7 +373,7 @@ class Warnings(commands.Cog):
     async def warn(
         self,
         ctx: commands.Context,
-        member: discord.Member,
+        identifier: str, 
         points: UserInputOptional[int] = 1,
         *,
         reason: str,
@@ -386,6 +386,19 @@ class Warnings(commands.Cog):
         or a custom reason if ``[p]warningset allowcustomreasons`` is set.
         """
         guild = ctx.guild
+
+        """User can be warned by ID or warned by their name."""
+        if identifier.isdigit():
+            member = ctx.guild.get_member(int(identifier))
+            # await ctx.send("Got member by ID")
+        else:
+            member = ctx.guild.get_member_named(identifier)
+            # await ctx.send("Got member by name")
+
+        if not member:
+            await ctx.send(f"User `{identifier}` not found.")
+            return
+        
         if member == ctx.author:
             return await ctx.send(_("You cannot warn yourself."))
         if member.bot:
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -427,9 +427,9 @@ class Warnings(commands.Cog):
                     "The person you're trying to warn is equal or higher than you in the discord hierarchy, you cannot warn them."
                 )
             )
-
         guild_settings = await self.config.guild(ctx.guild).all()
         custom_allowed = guild_settings["allow_custom_reasons"]
+
         reason_type = None
         async with self.config.guild(ctx.guild).reasons() as registered_reasons:
             if (reason_type := registered_reasons.get(reason.lower())) is None:
--- a/redbot/cogs/warnings/warnings.py
+++ b/redbot/cogs/warnings/warnings.py
@@ -15,9 +15,10 @@ from redbot.cogs.warnings.helpers import (
 )
 from redbot.core import Config, commands, modlog
 from redbot.core.bot import Red
-from redbot.core.commands import UserInputOptional
+from redbot.core.commands import UserInputOptional, RawUserIdConverter
 from redbot.core.i18n import Translator, cog_i18n
 from redbot.core.utils import AsyncIter
+from redbot.core.utils.views import ConfirmView
 from redbot.core.utils.chat_formatting import warning, pagify
 from redbot.core.utils.menus import menu
 
@@ -373,7 +374,7 @@ class Warnings(commands.Cog):
     async def warn(
         self,
         ctx: commands.Context,
-        identifier: str,
+        user: Union[discord.Member, RawUserIdConverter],
         points: UserInputOptional[int] = 1,
         *,
         reason: str,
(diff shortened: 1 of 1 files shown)


Flame442:
My original comment was not intended to be complete approval of the idea of having this command default to banning the member if they have left the server. I think it could be useful to have it as an option (whether that be by asking if they should be banned as a prompt when `[p]warn` is used on a member that left, or as a config option that causes `[p]warn` to ban when used on a member that has left), but I would consider it unexpected to happen automatically without any input from the server owner/mod. My comment was mostly intended to address the fact that this issue is one I would consider a fairly rare one to run into, and one that has a fairly easy manual second step to address it.

Regarding the implementation of this fix, the [`RawUserIdConverter`](https://docs.discord.red/en/stable/framework_commands.html#redbot.core.commands.converter.RawUserIdConverter) would be the preferred way to accept both a mention or a user ID, and is the converter used by other parts of the bot with similar behavior (ie, `[p]ban`). 

Additionally, you are currently failing the style checks. Please format your PR with `black -l 99` using the instructions in our [CONTRIBUTING.MD](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/CONTRIBUTING.md#43-style) file.

Ascensionn:
ok, I can add that as a prompt then. The bot should say "[User]" has left the server, so I cannot warn them. Would you like to ban them instead? [Y/N]". If the mod/owner responds with "Y", [User] will be banned. Else, no action will be taken.

I will submit another PR request with the changes.

Ascensionn:
*Reminder: the reason I used "[p]wasd" here is because I had to create a new separate cog in order to test for this kind of command. After I finished testing it, I carefully put in the modified parts.

image

From this image, we can see that the bot eventually timed out after it hasn't received a proper command (either Y or N). 

image
image

The bot only registers the responses it expects. Otherwise it will not do anything and eventually time out.

redbot/cogs/warnings/warnings.py
```diff
@@ -373,7 +373,7 @@ async def actionlist(self, ctx: commands.Context):
     async def warn(
         self,
         ctx: commands.Context,
-        member: discord.Member,
+        identifier: str,
```
Flame442: This should use [`RawUserIdConverter`](https://docs.discord.red/en/stable/framework_commands.html#redbot.core.commands.converter.RawUserIdConverter) to handle the case where the user is no longer in the server. See the [`[p]ban`](https://github.com/Cog-Creators/Red-DiscordBot/blob/33e0eac741955ce5b7e89d9b8f2f2712727af770/redbot/cogs/mod/kickban.py#L373-L396) command for an example. Similar to that command, it should be `Union`ing the `Member` and `RawUserIdConverter`, and having an `isinstance(user, int)` check to perform this banning logic instead of the existing warning behavior.

Additionally, the name of the function parameter defines the name a user will see in `[p]help`, so it should still be called either `member` or `user`, not `identifier`.

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +386,52 @@ async def warn(
...
+        """User can be warned by ID or warned by their name."""
+
+        if identifier.isdigit():
+            member = ctx.guild.get_member(int(identifier))
+            # await ctx.send("Got member by ID")
+        else:
+            member = ctx.guild.get_member_named(identifier)
+            # await ctx.send("Got member by name")
+        if not member:
+            await ctx.send(f"User `{identifier}` not found in discord guild")
```
Flame442: Avoid having multiple `send` calls for the same input, as that spams the API. One way to handle this is to combine strings until one final `send`, but in this case the last `send` you have already says basically the same thing, so this message is redundant.

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +386,52 @@ async def warn(
         or a custom reason if ``[p]warningset allowcustomreasons`` is set.
         """
         guild = ctx.guild
+        member = None
+        user = None
+        """User can be warned by ID or warned by their name."""
```
Flame442: If this is intended to be a part of the help text, it must be within the string directly below the function header. Otherwise, don't use string literals, use comments.

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +386,52 @@ async def warn(
...
+            # await ctx.send("Got member by ID")
+        else:
+            member = ctx.guild.get_member_named(identifier)
+            # await ctx.send("Got member by name")
+        if not member:
+            await ctx.send(f"User `{identifier}` not found in discord guild")
+
+            # Because he has not been found, he will be banned.
+            try:
+                user = await self.bot.fetch_user(int(identifier))
```
Flame442: `fetch_user` is a very heavily ratelimited call, and there is no reason for it to be used in this case since the actual user data isn't useful. As in the `[p]ban` command, just having `self.bot.get_user(ident) or discord.Object(id=user)` is enough to have something to pass to `guild.ban()`. You won't be able to use `user.name`, but just repeating the `.id` or saying "that user" is probably fine.

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +386,52 @@ async def warn(
...
+                await ctx.send(
+                    f"User `{user.name}` is not in the server but has been found globally. Would you like to ban them instead? [Y/N]"
+                )
+                try:
+
+                    def check(m):
+                        return (
+                            m.author == ctx.author
+                            and m.channel == ctx.channel
+                            and m.content.lower() in {"y", "n"}
```
Flame442: Instead of manually defining a check, you should use ~~[`MessagePredicate.yes_or_no`](https://docs.discord.red/en/stable/framework_utils.html#redbot.core.utils.predicates.MessagePredicate.yes_or_no) ([example](https://github.com/Cog-Creators/Red-DiscordBot/blob/33e0eac741955ce5b7e89d9b8f2f2712727af770/redbot/cogs/admin/admin.py#L526-L531)) to handle more ways to say "yes" or "no" as a text response~~.

EDIT: I didn't know we already had a util for this, but now that buttons exist the best option is probably to use those. The [`ConfirmView`](https://docs.discord.red/en/stable/framework_utils.html#redbot.core.utils.views.ConfirmView) utility should make it easy, see the examples in the docs.

Flame442 (changes requested):

Ascensionn:
Implemented the fixes you have mentioned @Flame442 

Here is what happens when you warn the user who is not in the server:

image
image

Here is what happens when it times out after 10 seconds.

image

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +387,34 @@ async def warn(
         or a custom reason if ``[p]warningset allowcustomreasons`` is set.
         """
         guild = ctx.guild
+        member = None
+        if isinstance(user, discord.Member):
+            member = user
+        elif isinstance(user, int):
+            user_obj = self.bot.get_user(user) or discord.Object(id=user)
```
Flame442, line 397: Rather than asking and hoping we have perms, let's add a check to exit slightly earlier if the bot doesn't have them
```suggestion
            if not ctx.channel.permissions_for(ctx.guild.me).ban_members:
                await ctx.send(_("User {user} is not in the server."),format(user=user))
                return
            user_obj = self.bot.get_user(user) or discord.Object(id=user)
```

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +387,34 @@ async def warn(
...
+            try:
```
Flame442: This `try` should only wrap around the `ban`, not the entire block of code.

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +387,34 @@ async def warn(
...
+                confirm = ConfirmView(ctx.author, timeout=10)
```
Flame442:
```suggestion
                confirm = ConfirmView(ctx.author, timeout=30)
```
10s is a bit short.

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +387,34 @@ async def warn(
...
+                confirm.message = await ctx.send(
+                    f"User `{user}` is not in the server but has been found globally. Would you like to ban them instead?",
+                    view=confirm,
```
Flame442:
```suggestion
                    _("User `{user}` is not in the server. Would you like to ban them instead?").format(user=user),
                    view=confirm,
```
Allow the string to be translated (you might need to re-black format after accepting some of these changes).

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +387,34 @@ async def warn(
...
+                )
+                await confirm.wait()
+                if confirm.result is None:
+                    await ctx.send("No response received within 10 seconds. No action taken.")
+                    return
+                elif confirm.result:
```
Flame442:
```suggestion
                if confirm.result:
```
No need to have a whole extra message for a slightly different case.

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +387,34 @@ async def warn(
...
+                    await ctx.guild.ban(user_obj, reason=reason)
+                else:
+                    confirm.message = await ctx.send("No action taken.")
+
+                await confirm.message.add_reaction("\N{WHITE HEAVY CHECK MARK}")
```
Flame442: If the bot does not have `add_reactions` permissions, this will cause it to claim an error occurred while trying to ban a user, which isn't quite true. Either this should use `await ctx.tick()` which handles that case correctly, or it should send a message instead to confirm.

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +387,34 @@ async def warn(
...
+                    view=confirm,
+                )
+                await confirm.wait()
+                if confirm.result is None:
+                    await ctx.send("No response received within 10 seconds. No action taken.")
+                    return
+                elif confirm.result:
+                    await ctx.guild.ban(user_obj, reason=reason)
+                else:
+                    confirm.message = await ctx.send("No action taken.")
```
Flame442:
```suggestion
                    confirm.message = await ctx.send(_("No action taken."))
```
Translations.

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +387,34 @@ async def warn(
...
+
+                await confirm.message.add_reaction("\N{WHITE HEAVY CHECK MARK}")
+
+            except discord.Forbidden:
+                await ctx.send("I don't have permission to ban this user.")
+            except discord.HTTPException:
+                await ctx.send("An error occurred while trying to ban the user.")
```
Flame442:
```suggestion
                await ctx.send(_("I don't have permission to ban this user."))
            except discord.HTTPException:
                await ctx.send(_("An error occurred while trying to ban the user."))
```
Translations.

Flame442 (changes requested):
If the ban is successful, it should create a modlog case ([example](https://github.com/Cog-Creators/Red-DiscordBot/blob/33e0eac741955ce5b7e89d9b8f2f2712727af770/redbot/cogs/mod/kickban.py#L215-L225)).

redbot/cogs/warnings/warnings.py
Ascensionn: yep, now I added a checker

redbot/cogs/warnings/warnings.py
```diff
@@ -392,27 +392,45 @@ async def warn(
             member = user
         elif isinstance(user, int):
             if not ctx.channel.permissions_for(ctx.guild.me).ban_members:
-                await ctx.send(_("User {user} is not in the server."),format(user=user))
+                await ctx.send(_("I cannot ban users. Missing the permission to ban members."))
```
Flame442: I had the string simply stating that the user was not in the server, because when the `[p]warn` command is used, that is the best response the bot can give. Stating as a response that the bot cannot ban users, without explaining _why_ it is trying to ban is confusing. I think it is better to just state that the user isn't in the server without worrying about the fact that it _could_ have attempted to ban.

redbot/cogs/warnings/warnings.py
```diff
@@ -392,27 +392,45 @@ async def warn(
...
 
+            if ctx.channel.permissions_for(ctx.guild.me).add_reactions:
                 await confirm.message.add_reaction("\N{WHITE HEAVY CHECK MARK}")
-
-            except discord.Forbidden:
-                await ctx.send(_("I don't have permission to ban this user."))
-            except discord.HTTPException:
-                await ctx.send(_("An error occurred while trying to ban the user."))
+            else:
+                await ctx.tick()
```
Flame442: `ctx.tick` is a shorthand for `if permissions_for.add_reactions: add_reaction(CHECK)`. That's why I said before you could just replace the `add_reaction` call with a call to `tick`, and let its error handling do the trick.

Flame442 (changes requested):

redbot/cogs/warnings/warnings.py
Ascensionn: oh ok, thank you for the clarification. Yes you're right, it would be strange to see that the bot suddenly prompts up with an error that it cannot ban a person when the [p]warn command is evoked. In that case, would you think it makes more sense for the bot to check "if not ctx.channel.permissions_for(ctx.guild.me).ban_members:" and then put the call to the ban() function in that if statement?

redbot/cogs/warnings/warnings.py
Ascensionn: In that case, do you want me to remove the "try...except"? I think the try..except block will be pointless then. All I need to do is add the check "if not ctx.channel.permissions_for(ctx.guild.me).ban_members:" before calling the ban() function. Because this ensures that there are no permission issues. However, as I didn't write the ban function, there could be another possible error, which is why I wanted to keep the try...except block in there.

redbot/cogs/warnings/warnings.py
Flame442: No. It makes no sense to ask if the user should be banned if the bot is not going to be able to ban them. A simpler failure message and an even earlier exist is sufficient.

redbot/cogs/warnings/warnings.py
Flame442: There is always the potential for error, but if an error you cannot think of happens, you probably _don't_ want to suppress it. You want to be able to identify it so you know if it is something that needs to be dealt with. Any API call can error if the internet goes out, but it doesn't make sense to wrap every API call in a `try` just in case.

I don't think it hurts to keep an `HTTPException` `try` on the `ban` call, since there are other cases that do make sense to catch, such as if the user is already banned. But the bot not having permissions is an obvious one that makes sense to have additional pre-check for. The `try` on the `tick` would be unnecessary.

Ascensionn: Sorry, I misinterpreted what you were correcting, but yes, I see what ctx.tick() does. I just wanted to keep it consistent with what the other commands did, and I saw that they were doing the add_reaction() call. Fixed it.

Flame442: I'm not sure which ones you mean, in this file at least they are all using `tick`.

Ascensionn: nevermind. I swore I saw there was some mention of it. Must've been in a different file.

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +387,49 @@ async def warn(
...
+                except discord.HTTPException as error:
+                    await ctx.send(
+                        _("An error occurred while trying to ban the user. Error: {error}").format(
+                            error=error
+                        )
+                    )
+            else:
+                confirm.message = await ctx.send(_("No action taken."))
+
+            ctx.tick()
```
Flame442:
```suggestion
            await ctx.tick()
```

redbot/cogs/warnings/warnings.py
```diff
@@ -386,6 +387,49 @@ async def warn(
...
+            )
+            await confirm.wait()
+            if confirm.result:
+                try:
+                    await ctx.guild.ban(user_obj, reason=reason)
+                    await modlog.create_case(
+                        self.bot,
+                        guild,
+                        ctx.message.created_at,
+                        "ban",
```
Flame442:
```suggestion
                        "hackban",
```

Flame442 (changes requested):

Flame442 (approved):
Thanks for contributing!

PR #6500, opened 2024-12-28
Single-source supported Java versions in Audio code

Jackenmen:

Updated the code so that the supported Java versions are sourced from constants in one place. This refactoring will be helpful when we implement LL jar updates outside the normal release schedule.

Yes

### Screenshots from testing

image
image
image


aikaterna (approved):
LGTM
cat-high-five-1762052381

PR #6503, opened 2025-01-06
Handle UIB interactions for ignoring channels.

### LINKED ISSUE #6501: Commands Used in Threads as User-Installed App

3.5.14

Receive a command.

Context:
Installed as a user app
Not present on server
Command run in a thread (cannot replicate what is causing this, permission wise, but a user can trigger it every time)

Bot receives command, processes it as normal and passes it to the cog.

### What actually happened?

Function doesn't get called. Following appears in Red's log after a few minutes:

```
[2025-01-03 16:13:18] [ERROR] asyncio: Task exception was never retrieved
future: <Task finished name='CommandTree-invoker' coro=<CommandTree._from_interaction.<locals>.wrapper() done, defined at /home/gel/sourcebot/lib/python3.11/site-packages/discord/app_commands/tree.py:1149> exception=AttributeError("'NoneType' object has no attribute 'id'")>
  File "/home/gel/sourcebot/lib/python3.11/site-packages/discord/app_commands/tree.py", line 1151, in wrapper
    await self._call(interaction)
  File "/home/gel/sourcebot/lib/python3.11/site-packages/discord/app_commands/tree.py", line 1272, in _call
    if not await self.interaction_check(interaction):
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gel/sourcebot/lib/python3.11/site-packages/redbot/core/tree.py", line 361, in interaction_check
    if not (await self.client.ignored_channel_or_guild(interaction)):
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gel/sourcebot/lib/python3.11/site-packages/redbot/core/bot.py", line 920, in ignored_channel_or_guild
    chann_ignored = await self._ignored_cache.get_ignored_channel(channel)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/gel/sourcebot/lib/python3.11/site-packages/redbot/core/_settings_caches.py", line 169, in get_ignored_channel
    cid: int = channel.id
               ^^^^^^^^^^
```

I cannot replicate this myself, though the user who keeps having the problem can reliably trigger it on servers (and they aren't able to get the permissions list for them, though they are trying to contact an admin to do so).

### Anything else?

What it looks like is your code [here](https://github.com/Cog-Creators/Red-DiscordBot/blob/cd0e8750c12a29690837937d989da26375f41ce5/redbot/core/tree.py#L355-L374) is running the check as if the bot was present on the server (which it isn't).

Then [this check](https://github.com/Cog-Creators/Red-DiscordBot/blob/cd0e8750c12a29690837937d989da26375f41ce5/redbot/core/bot.py#L913-L920) attempts to get the parent channel (which isn't available to the bot because it is a user-installed-app), and then tries to call that channel's (`None`) attributes.

There really can't be any checks for ignored (by the bot) channels like this, since it isn't on that server for anyone to run the command to lock channels.

Kowlin:

Fixes #6501, this PR handles User Installed Bot interactions where they will error out on trying to check from Threads, as it cannot get the channels it need. Furthermore, this handling like this will NOT check for categories, as the bot cannot gather those from a context where it can only get information from the interaction. Tough luck...

Yes, but only internally, this must be tested by trying to reproduce the original error from #6501


Flame442:
Fixes the issues for pure app commands, but there is still a remaining error for hybrids in particular. Not sure if we want to just fix that here, or if we want to save that for another PR.
```python
File "venv\lib\site-packages\redbot\core\_global_checks.py", line 8, in check_message_is_eligible_as_command
    return await ctx.bot.message_eligible_as_command(ctx.message)
  File "venv\lib\site-packages\redbot\core\bot.py", line 844, in message_eligible_as_command
    if not can_user_send_messages_in(guild.me, channel):
  File "venv\lib\site-packages\redbot\core\utils\__init__.py", line 724, in can_user_send_messages_in
    perms = channel.permissions_for(obj)
  File "venv\lib\site-packages\discord\threads.py", line 425, in permissions_for
    raise ClientException('Parent channel not found')
discord.errors.ClientException: Parent channel not found
```

Kowlin:
Hybirds are now handled properly! 
image

redbot/core/_global_checks.py
```diff
@@ -5,4 +5,6 @@
 def init_global_checks(bot):
     @bot.check_once
     async def check_message_is_eligible_as_command(ctx: commands.Context) -> bool:
+        if ctx.interaction is not None:
+            return True
```
Jackenmen, line 10: I think we should add a comment here and possibly also in `Red.message_eligible_as_command` so that we don't introduce discrepancies when changing the logic in the future:
```suggestion
        if ctx.interaction is not None:
            # equivalent checks are performed by `RedTree.interaction_check`
            return True
```

Jackenmen (changes requested):
I'm only really concerned about proper documentation here - there used to be just one flow that everything went through. Now that we continue to have to fix things for various interactions and the implementation is split between different places, it would be good to inform the code reader about the traps - see the review comment for a specific suggestion.
It might also be worth updating help text / messages from the `[p]ignore` command such that the user knows that the category ignores won't apply to user-installed commands.

Jackenmen (dismissed):

Jackenmen (approved):

PR #6505, opened 2025-01-19
Basic Docker image

dowerx:

Create a basic Docker image for easy deployment.

Yes

### DIFF
--- a/docker/Dockerfile.release
+++ b/docker/Dockerfile.release
@@ -1,0 +1,36 @@
+FROM debian:bookworm-slim AS build
+
+RUN apt update && apt install -y --no-install-recommends \
+  python3 python3-dev python3-venv build-essential ca-certificates
+
+RUN useradd -u 1000 -U -m red
+USER red
+WORKDIR /home/red
+
+RUN bash -c \
+  'python3 -m venv .redenv && \
+  source .redenv/bin/activate && \
+  pip install Red-DiscordBot && \
+  rm -rf .cache && \
+  redbot-setup \
+    --no-prompt \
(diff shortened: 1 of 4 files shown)


Jackenmen:
Hi, I'm sorry to say that we're currently not interested in maintaining a Docker image and supporting such a hosting method. There is a well-known Docker image from the community maintained by PhasecoreX for people who want to run using Docker regardless of our current stance on it, I'd suggest looking at that, if you're hosting Red with Docker yourself.

peterwilli:
Who doesn't want to maintain docker xD Its literally easier than to have everyone install it themselves

Kowlin:
> Who doesn't want to maintain docker xD Its literally easier than to have everyone install it themselves

Red is in the unique situation where requirements are strongly dependant on 3rd party plugins. Some may require a simple Pypi package. We can handle that within the bot just fine. However others, especially plugins that make images. Require other libraries that cannot be installed with Pip and Pypi. But rather require system level libraries. Stuff we cannot easily install without possibly creating problems down the line, or make potential system level security issues.

If this was a simple solution to solve. We'd have done so already. But due to the complexity of our needs. The lack of general enthusiasm for Docker from our (the maintainers) end. And lack of time and training to support Docker installations. We opted for not having a build in deploy option via Docker.

If you're still dead set on using Docker. We recommend PhasecoreX's Docker installation.

peterwilli:
Wow that was a fast response. Sorry I was mostly surprised to find this initial response. I did find the PhaseCoreX docker image and it works surprisingly well, so I'm continuing with this one and thanks for the amazing package!

PR #6510, opened 2025-02-01
Add `--cog-path` CLI argument

### LINKED ISSUE #6506: Add a CLI argument to `redbot` to add cog paths to `[p]paths`
### Type of feature request

Other

I propose a `--cog-paths` flag for the `redbot` CLI. This flag would add a colon-separated list of cog paths to `[p]paths`, and would not persist through restarts (just like other CLI flags).

Example usage:
`--cog-paths /workspaces/FooCogs:/workspaces/BarCogs`

### Anything else?

So, for some context, I develop my cogs for Red inside of a [Dev Container](https://containers.dev/). [Here's what that looks like.](https://www.coastalcommits.com/cswimr/SeaCogs/src/branch/main/.devcontainer) Then, I use Visual Studio Code's [Run & Debug feature](https://www.coastalcommits.com/cswimr/SeaCogs/src/branch/main/.vscode/launch.json) to actually run Red. The initial setup process for this development environment is as follows:
- Run `redbot-setup`
- Launch Red through the Run & Debug menu
- Once the bot is started, run `[p]addpath /workspaces/SeaCogs` inside of Discord to add the project directory to the bot as a cog path.
  - This step is what would be improved by this feature request.

If this feature request is implemented, I could add `--cog-paths /workspaces/SeaCogs` to my VSCode Run & Debug configuration, removing one initial setup step to get up and running.

Discussed in the Red Discord Server, in #feature-ideas: https://discord.com/channels/133049272517001216/133049878837329920/1332807433650704534

cswimr:

This PR adds an argument, `--cog-path`, to the `redbot` CLI that accepts a path, validates that the provided path exists, and adds the path via `self._cog_mgr.add_path()`, so that cogs can be loaded from that path. The argument can be passed multiple times to add more than one path at a time. These paths are stored in memory and never to disk, so restarting the bot without the argument will "remove" the paths.

Additionally, this PR adds a `persist` argument to `CogManager.add_path()`, which defaults to `True`. The original behavior of appending the added path to a list and then calling `.set_path()` on that list is retained through this argument being set to `True`. Setting this argument to `False` will append the path to a module-level list of paths, stored in memory. These are, as I mentioned earlier, never written to the bot's config, and are effectively discarded on shutdown.

Closes #6506.

Yeah, but not super thoroughly

### DIFF
--- a/redbot/core/_cli.py
+++ b/redbot/core/_cli.py
@@ -220,6 +220,15 @@ def parse_cli_flags(args):
         help="Force unloading specified cogs.",
     )
     parser.add_argument(
+        "--cog-path",
+        type=str,
+        default=[],
+        nargs="+",
+        action="extend",
+        help="Add a specific path to the list of cog paths. "
+        "This can be used multiple times to add multiple paths.",
+    )
+    parser.add_argument(
         "--dry-run",
         action="store_true",
         help="Makes Red quit with code 0 just before the "
--- a/redbot/core/_cog_manager.py
+++ b/redbot/core/_cog_manager.py
@@ -144,8 +144,8 @@ class CogManager:
         ----------
         path : `pathlib.Path` or `str`
             Path to add.
-        persist : `bool`
-            Whether or not the path should be persisted through restarts.
+        persist : `bool`, optional
+            Whether or not the path should be persisted through restarts. Defaults to True.
 
         Raises
         ------
--- a/redbot/core/_cog_manager.py
+++ b/redbot/core/_cog_manager.py
@@ -88,7 +88,7 @@ class CogManager:
             A list of user-defined paths.
 
         """
-        return list(map(Path, deduplicate_iterables(_TEMP_PATHS, await self.config.paths())))
+        return list(map(Path, deduplicate_iterables(await self.config.paths())))
 
     async def set_install_path(self, path: Path) -> Path:
         """Set the install path for 3rd party cogs.
@@ -230,7 +230,9 @@ class CogManager:
                 name=name,
             )
 
-        real_paths = list(map(str, [await self.install_path()] + await self.user_defined_paths()))
+        real_paths = list(
+            map(str, [await self.install_path()] + _TEMP_PATHS + await self.user_defined_paths())
+        )
 
         for finder, module_name, _ in pkgutil.iter_modules(real_paths):
             if name == module_name:
@@ -307,7 +309,7 @@ class CogManager:
 
     async def available_modules(self) -> List[str]:
         """Finds the names of all available modules to load."""
-        paths = list(map(str, await self.paths()))
+        paths = list(map(str, _TEMP_PATHS + await self.paths()))
 
         ret = []
         for finder, module_name, _ in pkgutil.iter_modules(paths):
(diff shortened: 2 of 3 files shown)


redbot/core/_cli.py
```diff
@@ -219,6 +219,15 @@ def parse_cli_flags(args):
         action="extend",
         help="Force unloading specified cogs.",
     )
+    parser.add_argument(
+        "--cog-path",
+        type=str,
+        default=[],
```
cswimr, line 225: I'm not sure what `default=[],` is doing here. I kept it because it was listed in `--co-owners`, but I'm not sure if I actually need it or not.

Kowlin: The default argument over there was a simple way for us to initialize the co-owner list and start from there. As we just check for a raw int in a list for co-owners it simplified development

cswimr: So I guess I probably don't need the default argument then, as the code that actually uses the `cog_path` attribute that this generates checks if the attribute is null before trying to iterate through it (see changes in bot.py). Is that right?

Flame442: https://github.com/Cog-Creators/Red-DiscordBot/blob/a0c1713e785b54dedd83e9167f4882d85c5b7931/redbot/core/bot.py#L221
It saves a line of code later just being able to `.update` without erroring if no co-owners were passed. It doesn't really help or hurt as is.

Flame442:
This PR does not behave as outlined in the issue. `self._cog_mgr.add_path()` is the method used by `[p]addpath` to add a new **permanent** cog path to the bot. The issue described this flag as a **temporary** path which would only exist as long as the path was present. The method does not duplicate paths, so it is not a huge problem, but if we want this flag to specifically be for temporary paths then a new method for doing so will be needed.

cswimr:
Oops, good catch! I'll take another look at this.

cswimr:
So, I propose the following:
Temporary paths would be stored in config, separately from ones added via `[p]addpath`. This is the safest place that I can think of to store them. When the bot is started, that config value would be overwritten with the contents (if any) of the `--cog-path` CLI argument. If nothing is passed to `--cog-path`, then the config value is emptied. A new function would have to be added to `CogManager` to retrieve these temporary paths, and `CogManagerUI` would have to be updated to check this in `[p]paths`. Is there a better way Red has to do something like this? It feels kind of clunky imo but it'd work I think.

Flame442:
Why not just store it in memory during the run the flag is used? Why does the disk need to be used at all?

cswimr:
> Why not just store it in memory during the run the flag is used? Why does the disk need to be used at all?

Latest commit stores it in memory. Assuming I didn't miss anything, should be ready for review now.

Flame442 (changes requested):
`[p]removepath` (and by extension, `bot._cog_mgr.remove_path` abd `.add_path`) has incorrect behavior with the way you have temporarily added the path. These methods pull from `bot._cog_mgr.user_defined_paths` to get the list of user-defined paths to modify. By including the temporary paths in this method's output, they will attempt to `append` to and `remove` from the list containing the temporary paths, then set the list of user paths including any temporary paths that remain after that change. This will cause those remaining paths to become permanent paths.

All of the frontend (`CogManagerUI` cog) and backend (`CogManager` class) logic needs to properly handle not only processing both permanent and temporary user-defined paths, but also listing and modifying them.

cswimr:
> `[p]removepath` (and by extension, `bot._cog_mgr.remove_path` abd `.add_path`) has incorrect behavior with the way you have temporarily added the path. These methods pull from `bot._cog_mgr.user_defined_paths` to get the list of user-defined paths to modify. By including the temporary paths in this method's output, they will attempt to `append` to and `remove` from the list containing the temporary paths, then set the list of user paths including any temporary paths that remain after that change. This will cause those remaining paths to become permanent paths.

Good catch, fixed this in the latest commit.

> All of the frontend (`CogManagerUI` cog) and backend (`CogManager` class) logic needs to properly handle not only processing both permanent and temporary user-defined paths, but also listing and modifying them.

Updated in latest commit to make `[p]paths` list temporary paths - this comes with a change to a localization string, not sure how to regenerate locale files if necessary? Is it just `redgettext`, same as for a third party cog?

Out of curiosity, why should temporary paths be modifiable through Discord? I struggle to see a reason to use this feature in that way, as you would need to re-run the command to add a temporary path after each time the bot restarts. This would also result in any cogs loaded from a previously set temporary path failing to load on restart. With a CLI argument, you can just have your service runner (systemd, docker, etc.) use the argument every time it starts the bot. The temporary paths list is also populated before cogs begin loading at startup when using the CLI argument.

Flame442:
> This comes with a change to a localization string, not sure how to regenerate locale files if necessary?

`.pot` files are automatically generated and pushed to crowdin using a [branch push workflow.](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/.github/workflows/crowdin_upload_strings.yml) Our [release workflow](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/.github/workflows/prepare_release.yml) handles pulling localized strings from crowdin.

> Out of curiosity, why should temporary paths be modifiable through Discord?

In an attempt to speak succinctly, I said something I did not mean. What I should have said is that all frontend and backend logic needs to be updated such that it does not cause unexpected behavior when temporary paths are present. This includes logic that is currently used to modify paths, as it had the incorrect behavior I described. I agree that there is no reason to allow modifying temporary paths within discord.

Flame442 (approved):
Thank you for your contribution!

PR #6517, opened 2025-02-03
Handle invalid locale configuration graciously during startup

Jackenmen:

Fix startup issue on invalid locale configuration:
```
[2025-02-02 22:09:10] [CRITICAL] red.main: The main bot task didn't handle an exception and has crashed
  File "/data/venv/lib/python3.11/site-packages/redbot/core/_i18n.py", line 40, in _get_standardized_locale_name
    locale = Locale.parse(language_code, sep="-")
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/venv/lib/python3.11/site-packages/babel/core.py", line 349, in parse
    raise ValueError(msg)  # `parse_locale` would raise a ValueError, so let's do that here
    ^^^^^^^^^^^^^^^^^^^^^
ValueError: ("Empty locale identifier value: ''\n\nIf you didn't explicitly pass an empty value to a Babel function, this could be caused by there being no suitable locale environment variables for the API you tried to use.",)

During handling of the above exception, another exception occurred:

  File "/data/venv/lib/python3.11/site-packages/redbot/__main__.py", line 470, in red_exception_handler
    red_task.result()
  File "/data/venv/lib/python3.11/site-packages/redbot/__main__.py", line 369, in run_bot
    await red.start(token)
  File "/data/venv/lib/python3.11/site-packages/redbot/core/bot.py", line 1278, in start
    await self._pre_login()
  File "/data/venv/lib/python3.11/site-packages/redbot/core/bot.py", line 1150, in _pre_login
    _i18n.set_global_regional_format(i18n_regional_format)
  File "/data/venv/lib/python3.11/site-packages/redbot/core/_i18n.py", line 60, in set_global_regional_format
    language_code = _get_standardized_locale_name(language_code)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/venv/lib/python3.11/site-packages/redbot/core/_i18n.py", line 42, in _get_standardized_locale_name
    raise ValueError("Invalid language code. Use format: `en-US`")
ValueError: Invalid language code. Use format: `en-US`
```
with config such as:
```
... "locale": "en-US", "regional_format": ""}
```

While an empty string is not something that could have been set by Core, a previously public i18n API for setting locale did not validate its input and as such, some users could have such values in their config files.

No

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -1148,23 +1148,25 @@ class Red(
         try:
             _i18n.set_global_locale(i18n_locale)
         except ValueError:
+            log.warning(
+                "The bot's global locale was set to an invalid value (%r)"
+                " and will be reset to default (%s).",
+                i18n_locale,
+                _i18n.FRESH_INSTALL_LOCALE,
+            )
             i18n_locale = _i18n.FRESH_INSTALL_LOCALE
             await self._config.locale.clear()
-            log.warning(
-                "The bot's global locale was set to an invalid value and has been reset"
-                " to default (%s).",
-                i18n_locale,
-            )
         i18n_regional_format = await self._config.regional_format()
         try:
             _i18n.set_global_regional_format(i18n_regional_format)
         except ValueError:
-            await self._config.regional_format.clear()
             log.warning(
-                "The bot's global regional format was set to an invalid value and has been reset"
-                " to default (which is to inherit global locale, i.e. %s).",
+                "The bot's global regional format was set to an invalid value (%r)"
+                " and will be reset to default (which is to inherit global locale, i.e. %s).",
+                i18n_regional_format,
                 i18n_locale,
             )
+            await self._config.regional_format.clear()
 
     async def _pre_connect(self) -> None:
         """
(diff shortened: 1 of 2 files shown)


redbot/core/bot.py
```diff
@@ -1145,9 +1145,26 @@ async def _pre_login(self) -> None:
...
 
         i18n_locale = await self._config.locale()
-        _i18n.set_global_locale(i18n_locale)
+        try:
+            _i18n.set_global_locale(i18n_locale)
+        except ValueError:
+            i18n_locale = _i18n.FRESH_INSTALL_LOCALE
+            await self._config.locale.clear()
+            log.warning(
+                "The bot's global locale was set to an invalid value and has been reset"
```
Flame442: I probably would like to see it print the old and incorrect value in the warning message for debugging purposes - We don't want a case in the future where this catch bugs out to confuse users if their language gets reset.

Flame442 (changes requested):

redbot/core/bot.py
Jackenmen: Updated.

Flame442 (approved):

PR #6551, opened 2025-04-02
Add support for images inside `[p]info`

### LINKED ISSUE #6132: Enhance [p]info with a custom image.

Core

The following patch:
https://github.com/Cog-Creators/Red-DiscordBot/compare/V3/develop...DJTOMATO:Red-DiscordBot:patch-2

- Adds ``[p]set bot infoimage`` command which adds an image to [p]info command when using embeds
- Won't display if embeds are disabled.
- Defines ``infoimage`` as default variable.

 

### Anything else?

Hi, I'd like to hear org opinion on enhancing info command allowing end-users to further customize the bot with a custom image on the info page.

Kreusada:

Closes #6132. Enables bot owners to configure an image that can be shown inside `[p]info` when embeds are enabled. This will be a useful way to add visual appeal to the bot.

By implementing this, I have created a new group, `[p]set bot custominfo`, with two subcommands of `[p]set bot custominfo text` and `[p]set bot custominfo image`.

Yes

image


docs/cog_guides/core.rst
```diff
@@ -2964,25 +2964,70 @@ set bot custominfo
...
+
+""""""""""""""""""""""""
+set bot custominfo image
+""""""""""""""""""""""""
+
+**Syntax**
+
+.. code-block:: none
+
+    [p]set bot custominfo text [text]
```
Jackenmen, line 3012:
```suggestion
    [p]set bot custominfo image [url]
```

Jackenmen (commented):
I like the idea, but only storing the URL of the image in case of the attachment is IMO problematic, as Discord can drop the image from the CDN if the message is removed.
If we were to allow uploading the image as an attachment, we'd probably have to store the actual image in the data path (and have some additional verification of e.g. the file size being below the non-nitro attachment size limit). I think it might be simpler to just remove the option to upload the image as an attachment.

PR #6567, opened 2025-05-03
[Audio] Update application.yml generation for new sources

aikaterna:
Rearranged source order for optimal playback. 
Added additional sources and their options. 
Added missing options for IOS. 

Yes. (Tested through using the application.yml that I pulled these values from.)

As a reviewer, please double check this against the posted application.yml file in the Red server (I'm unable to attach it here).


Jackenmen (approved):
I made a small change to the key conversion function because the `ANDROID_VR`/`ANDROID_MUSIC` keys got transformed into `ANDROID-VR` and `ANDROID-MUSIC`. Other than that, it looks identical to the application.yml you've posted in the help desk channel.

PR #6576, opened 2025-05-24
Disable Rich tracebacks by default

Jackenmen:

Rich tracebacks trade usability for the "nice look". We don't want that as a default experience. It's been a constant support nightmare due to how much it inflates the text length of tracebacks, doesn't react well to terminal resizing, and, empirically, probably also slows down console logging noticeably...
I feel like this has been a known "ah, yes, this does kinda suck and we'd like to change it at some point" for a while.

Yes

### DIFF
--- a/redbot/core/_cli.py
+++ b/redbot/core/_cli.py
@@ -301,20 +301,31 @@ def parse_cli_flags(args):
         "This flag can be used multiple times to specify multiple intents.",
     )
     parser.add_argument(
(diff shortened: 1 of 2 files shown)


Jackenmen:
One could argue that, since this is marked as useful for development, maybe we should auto-enable it as part of the `--dev` flag, but I think that it would be too opinionated to do that.

Flame442 (approved):
giphy

PR #6577, opened 2025-05-25
Fix checks not properly handling Group commands type

Sutaai:

This PR propose to add the type `Group` to the type variable used in `Requires.get_decorator` (Which is used by [Red's checks](https://docs.discord.red/en/stable/framework_checks.html)).

This was made as creating a command group, in addition of one of Red's provided check, would alert Pylance of incorrect type, expecting type `Command` only.

**Before**

image

**After**
image
image

The correct type is now reflected.
Same issue and fix apply for `@commands.hybrid_group`

Yes

### DIFF
--- a/redbot/core/commands/requires.py
+++ b/redbot/core/commands/requires.py
@@ -33,10 +33,10 @@ from .errors import BotMissingPermissions
 from redbot.core import utils
 
 if TYPE_CHECKING:
-    from .commands import Command
+    from .commands import Command, Group
     from .context import Context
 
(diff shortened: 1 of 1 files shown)


Flame442:
I would be interested to know if swapping the order of your checks on the group would make the error go away - put the `is_owner` below the `@commands.group()`.

Sutaai:
It does, but raise another error that I believe is rather more of an issue from discord.py.
image

This issue (This screenshot) apply to anything as long as `@commands.group()/.command()` is not the last decorator as far as I've seen. Some other combination could work.

PR #6579, opened 2025-05-26
Inject guild locale information in app commands

### LINKED ISSUE #6272: Slash commands do not inject guild locale information
Text commands auto inject the contextual guild into the i18n logic Red uses (`bot._i18n_cache` I think?). App commands (and hybrids) currently do not do this, and fall back to the global setting. It isn't a huge problem since an explicit `set_contextual_locales_from_guild` still fixes it, however it would be nice to automatically have it filled in.

https://github.com/Cog-Creators/Red-DiscordBot/blob/6603cd1a861230d36810b3889fe58c6dd48f9115/redbot/core/_events.py#L478

Flame442:
Previously, app commands had no guild locale information, and would only use the global locale settings. This PR causes the locale information to be injected prior to calling the command, similar to the current logic for text commands.

The chain of app command execution in dpy is not the same as text commands, with the command processing logic a public method that can be explicitly called when `on_message` is overriden. As a result, this is currently being done by overriding a private `CommandTree` method. We are already messing with the internals of this object enough that it's probably fine to mess with it a little more.
1. `state.py` `parse_interaction_create`
2. `CommandTree._from_interaction`
3. `CommandTree._call` <--- injected here

Fixes #6272

No

### DIFF
--- a/redbot/core/tree.py
+++ b/redbot/core/tree.py
@@ -19,7 +19,10 @@ from .app_commands import (
     TransformerError,
     UserFeedbackCheckFailure,
 )
-from .i18n import Translator
+from redbot.core.i18n import (
+    Translator,
+    set_contextual_locales_from_guild,
+)
 from .utils.chat_formatting import humanize_list, inline
 
 import logging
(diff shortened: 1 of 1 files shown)


Kowlin:
I thought that in this case we were invoking the locale from the user's preferred locale that Discord sets? Although I suppose that has the issue of just being very limited

Flame442:
I can't find anywhere we are setting the language using `interaction.locale`, but I might be missing it if it is being done in some other way. I don't _believe_ that is the current behavior, but it might be something to consider as an alternative option. It feels like that method works best for ephemeral commands, but we have no way of knowing whether the command being invoked will be ephemeral or not.

Jackenmen (approved):
LGTM.

PR #6586, opened 2025-07-15
Update the invite Regex pattern with a new accepted format.

AAA3A-AAA3A:

Hello,
Last week we noticed on several servers where I'm staff that a new invite format is displayed by Discord as valid but is not currently supported by most bots, leaving the door open to scam bots. The new format is `discordapp.com/invite\<code>`.
image
I'm using this modified pattern in one of my cogs, while waiting for a modification in Red.
Thanks in advance,
Have a nice day,
AAA3A

Yes

### DIFF
--- a/redbot/core/utils/common_filters.py
+++ b/redbot/core/utils/common_filters.py
@@ -16,7 +16,9 @@ __all__ = [
 # regexes
 URL_RE = re.compile(r"(https?|s?ftp)://(\S+)", re.I)
 
-INVITE_URL_RE = re.compile(r"(discord\.(?:gg|io|me|li)|discord(?:app)?\.com\/invite)[\/\\](\S+)", re.I)
+INVITE_URL_RE = re.compile(
+    r"(discord\.(?:gg|io|me|li)|discord(?:app)?\.com\/invite)[\/\\](\S+)", re.I
+)
 
 MASS_MENTION_RE = re.compile(r"(@)(?=everyone|here)")  # This only matches the @ for sanitizing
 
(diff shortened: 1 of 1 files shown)


Flame442:
This isn't really a "new" format, it is a byproduct of the fact that chrome (and chromium based browsers, like the discord client) assume backslash characters are typo'd forwards slash characters. Maybe this became a "workaround" recently due to invite links embeding even when not a clickable URL... Technically any of the slashes could be backslashes and still be a valid URL for a chromium based browser, though only some cases actually render the invite embed in the discord client.

Any of the following will trigger the embed:
`discord.com/invite\red`
`discordapp.com/invite\red`
`https://discord.gg\red` (visually appears with a forward slash since the link is clickable)

But these will not trigger the embed:
`discord.gg\red` on its own without `https://`
`discord.com\invite/red`
`discordapp.com\invite/red`

Interestingly, putting backslashes in `https:\\` does not cause the link to become clickable or for the invite embed to appear, despite chromium based browsers accepting it as a valid URL.

I wouldn't trust any of this behavior to never change again. The safest thing to do is probably to union every slash in both the invite URL and the link URL regexs to match either a forward or a backslash, since either way someone is clearly trying to advertise, even if an embed doesn't appear.

AAA3A-AAA3A:
Yes, it's not a "new" format, but it's something that has been detected by Discord as an invite since very recently and that some scams are starting to use.
Would you like me to add the union for each slash, and to the other pattern as well?

Twentysix26:
It's probably a good time to mention that this regex also catches `promos.discord.gg` links, which are not server invites (https://github.com/Twentysix26/x26-Cogs/issues/57)
This could be solved with a negative lookbehind, which should result in
`(?<!promos\.)(discord\.(?:gg|io|me|li)|discord(?:app)?\.com\/invite)[\/\\](\S+)`

Jackenmen (approved):
Thanks!

PR #6587, opened 2025-07-16
Remove unused direct dependencies orjson, Brotli and zstandard from configuration files

karyotakisg:

As part of our ongoing research on Python dependency management we noticed a potential improvement in your project’s dependency management.

Specifically, the direct dependencies `orjson`, `zstandard` and  `Brotli` are specified as a requirement in the `base.in` and `base.txt` files, when in reality they are not needed.

Hope this is helpful!

Best regards

Yes, `tox` was used.

### DIFF
--- a/requirements/base.in
+++ b/requirements/base.in
@@ -2,11 +2,9 @@ aiohttp<3.10
 aiohttp-json-rpc
 apsw
 babel
-Brotli
 click
 discord.py
 markdown
-orjson
 packaging
 platformdirs
 psutil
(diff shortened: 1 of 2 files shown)


Flame442:
For brotli, aiohttp considers it an optional dependency and gains increased performance/functionality when it is installed. For orjson, discord.py considers it an optional dependency and gains increased performance parsing data from the discord API. I don't know for certain for z standard, but I imagine we have it for a similar reason as the other two. It could potentially be documented better that they are not unused dependencies.

Jackenmen:
These are used - not by our code directly but by our dependencies. They don't get added as transitive dependencies by pip-compile because they're optional dependencies and we don't list the relevant extras they're included in as those extras don't allow for the same granularity (i.e. I can't include aiohttp[speedups] without also including aiodns, which we don't want).
Closing as invalid.

Issue #6442, opened 2024-09-01
[Audio] Toggle setting the currently playing song as a voice channel status


EternalllZM:
### What component of Red (cog, command, API) would you like to see improvements on?

Audio

### Describe the enhancement you're suggesting.

Current functionality of audioset status according to docs:
[p]audioset status - Show the now playing song in the bot’s status, or show how many servers the bot is playing music on, if more than one.

This command and functionality could be extended (sub command?) to also display the currently playing song as a voice channel status.

![image](https://github.com/user-attachments/assets/9a4a4efe-1bb5-4947-a740-8979f6250d6e)

### Anything else?

While the last comment on this issue mentions channel status may be replaced by hang status - recently the hang status was completely rolled back to 0% and cements channel status as what is likely staying instead https://github.com/discord/discord-api-docs/pull/6398. 

Unfortunately I couldn't find anything showing this endpoint been implemented yet, so even if this idea is approved, it may have to stay blocked for a while.

Jackenmen:
It is currently unclear what Discord will do with this - this feature went through 3 different API implementations, I believe that the feature is presently in A/B testing in clients (i.e. only present for *some* users) and not too long ago there was a similar feature that allowed individual members to set voice statuses but seems to no longer be available (at least for me).
We'll have to see what this settles on first. Presumably, it won't be implemented in our upstream library before that time either so it will definitely be blocked for a long while.

bryceoj:
Any update on this whether this is something that can implemented?  

Interested in having audio update the voice channel status to the current song playing.  Should be easily implemented alongside the existing audio status for the bot's status, but instead change the voice channel status.

Using the channel.edit:
https://discordpy.readthedocs.io/en/latest/api.html?highlight=voice%20channel#discord.VoiceChannel.edit

Flame442:
This seems like a reasonable feature if someone is interested in developing it. Not sure if it would be better to make this an alternative option within `[p]audioset status` (similar to `[p]set usemenus`) or as an entirely separate toggle so both could be enabled. The main downside to using the bot status is that it is a global setting, so this would allow similar functionality for all servers the bot is playing in.

bryceoj:
@Flame442 yeah even just an additional toggle separate from `[p]audioset status`  could easily just be `[p]audioset voicestatus`

I would easily attempt to make a PR for this change, just let me know.

bryceoj:
@Flame442 started working on this. 
[...]

PR #6591, opened 2025-08-01
Audio status to voice channel 

bryceoj:

Add enhancement to audio cog to enable Red to update the Voice Channel status to the current playing song.

Added an additional command to allow for both Bot status updates ([p]audioset status) and Voice Channel status updates ([p]audioset voicestatus)

Capture
Screenshot 2025-07-31 213328
Screenshot 2025-07-31 213212

Related: Cog-Creators/Red-DiscordBot#6442

Yes

### DIFF
--- a/redbot/cogs/audio/core/events/lavalink.py
+++ b/redbot/cogs/audio/core/events/lavalink.py
@@ -124,6 +124,7 @@ class LavalinkEvents(MixinMeta, metaclass=CompositeMetaClass):
             current_track, self.local_folder_current_path
         )
         status = await self.config.status()
+        vc_status = await self.config.vc_status()
         prev_song: lavalink.Track = player.fetch("prev_song")
         await self.maybe_reset_error_counter(player)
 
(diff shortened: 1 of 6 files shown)


redbot/cogs/audio/core/utilities/player.py
```diff
@@ -65,6 +65,23 @@ async def get_active_player_count(self) -> Tuple[Optional[str], int]:
             playing_servers = 0
         return get_single_title, playing_servers
 
+    async def get_current_track_title(self) -> str:
```
Jackenmen, line 68: Since this method reuses code from `get_active_player_count()`, we should update `get_active_player_count()` to just call it.

redbot/cogs/audio/core/events/lavalink.py
```diff
@@ -227,6 +228,20 @@ async def lavalink_event_handler(
...
+            await self.update_voice_channel_presence(voice_channel, track_title)
+
+        if event_type == lavalink.LavalinkEvents.TRACK_END and vc_status:
+            await asyncio.sleep(1)
+            if not player.is_playing:
+                log.debug("Track ended for %s, updating voice channel status", guild_id)
+                voice_channel = current_channel
+                track_title = await self.get_current_track_title()
+                await self.update_voice_channel_presence(voice_channel, track_title)
+
```
Jackenmen, line 244: I think merging the code from these two `if` statements with the code right above might be a good idea. The coed doesn't seem *that* different + we'd avoid an additional `sleep` call.

Jackenmen (commented):
The implementation looks mostly fine but there is some discussion to be had about how configurable this should be (global and/or per-server), what the default should be (I personally want us to at least consider having it enabled by default), and how we should handle the destructiveness of editing the voice status (i.e. if the user of a channel sets their own status and then the bot ends up overriding it).

rishabhsingroha:
wow

NiyahVE:
Looking forward to this still, nice work :)

abogatskiy:
Eagerly awaiting 🥺

Fabian-Evolved:
Also waiting for this, but I think they are no longer developing this bot.

yamikaitou:
This PR is still in Draft status, so it cannot be merged.

EternalllZM:
I don't think this PR had much left to do outside of a few fixes. Unfortunate.

bryceoj:
@EternalllZM Feel free to re-attempt it.  The changes are there.  I didn't have the time to make the requested adjustments.

PR #6593, opened 2025-08-06
Mod cog: Option to show an extra field with custom content on the ban embed

### LINKED ISSUE #5591: [Mod] Allow setting a custom DM message for [p]modset dm
### Type of feature request

Command

This command would be included in the `mod` cog.
The specific subcommand would be under the `modset` command

This command would either add another subcommand to the existing `modset dm` and split the two commands:
ex: `modset dm true/false (or) toggle` & `modset dm setmessage` or be a standalone command under `modset`. 

It would most likely be a setting restricted to bot administrator privilege or above.

The function would be that of a pre-set message to DM to the punished user. The most useful case would be to provide instructions for what punished users can do to either appeal their punishment, or take other action such as "calm down" (as if that ever happens). 

Extremely useful for communities providing appeal servers, as Discord does not have a native way to allow banned users to appeal. However, customizable enough to fit any necessary community purpose if wanted.

### Anything else?

If wanted, this could also increase scope to `modlog` where case entries could have their own pre-set formatted text within the embed/text sent to moderator channels for logging purposes. This would be separate from the DM command.

VasilisThePikachu:
Added a new option to the base mod cog to show a new field embed with customizable text on the ban embed. The intended reason in this example (and my usage) will be to link a ban appeal link or some other kind of information the banned user may want.

Previously to get this effect you needed to potentially set up a custom cog or another way to first send a message to the user before running the actual ban command.

Additional changes:
- TempBan now has an embed.

The code is... meh, I have not used python in years and this code seemed pretty old.

Fixes #5591

image

Yes I have, I have tested all cases I could think off.

### DIFF
--- a/docs/cog_guides/mod.rst
+++ b/docs/cog_guides/mod.rst
@@ -253,7 +253,23 @@ modset dm
 
 .. code-block:: none
 
-    [p]modset dm [enabled]
+    [p]modset dm
+
+**Description**
+
+Settings for messaging the user on moderation action.
+
+.. _mod-command-modset-dm-sendmessage:
+
+"""""""""
+modset dm sendmessage
+"""""""""
+
+**Syntax**
+
+.. code-block:: none
+
+    [p]modset dm sendmessage [enabled]
 
 **Description**
 
@@ -265,6 +281,71 @@ and reason as to why they were kicked/banned.
 **Arguments**
 
 * ``[enabled]``: Whether a message should be sent to a user when they are kicked/banned. |bool-input|
+
+.. _mod-command-modset-banshowextrafield:
+
+"""""""""
+modset dm banshowextrafield
+"""""""""
+
+**Syntax**
+
+.. code-block:: none
+
+    [p]modset dm banshowextrafield [enabled]
+
+**Description**
+
+
+Toggle whether to show an extra customizable field when banning. This is useful to add information such as a ban appeal link.
+
+**Arguments**
+
+* ``[enabled]``: If an extra customizable embed field should appear when banning. |bool-input|
+
+.. _mod-command-modset-banextrafieldtitle:
+
+"""""""""
+modset dm banextrafieldtitle
+"""""""""
+
+**Syntax**
+
+.. code-block:: none
+
+    [p]modset dm banextrafieldtitle [title]
+
+**Description**
+
+Set the title for the optional extra embed on ban
+
+Set to "clear" to remove.
+
+**Arguments**
+
+* ``[title]``: The title of the embed field. Can by any string of text.
+
+.. _mod-command-modset-banextrafieldcontents:
+
+"""""""""
+modset dm banextrafieldcontents
+"""""""""
+
+**Syntax**
+
+.. code-block:: none
+
+    [p]modset dm banextrafieldcontents [contents]
+
+**Description**
+
+Set the contents for the optional extra embed on ban
+
+Set to "clear" to remove.
+
+**Arguments**
+
+* ``[contents]``: The contents of the embed field. Can by any string of text.
 
 .. _mod-command-modset-requirereason:
 
--- a/docs/cog_guides/mod.rst
+++ b/docs/cog_guides/mod.rst
@@ -318,7 +318,7 @@ modset dm banextrafieldtitle
 
 **Description**
 
-Set the title for the optional extra embed on ban
+Set the title for the optional extra embed on ban.
 
 Cannot be over 252 characters long.
 
--- a/docs/cog_guides/mod.rst
+++ b/docs/cog_guides/mod.rst
@@ -257,13 +257,13 @@ modset dm
 
 **Description**
 
-Settings for messaging the user on moderation action.
+Settings for messaging the user when being kicked or banned.
 
 .. _mod-command-modset-dm-sendmessage:
 
-"""""""""
+"""""""""""""""""""""
 modset dm sendmessage
-"""""""""
+"""""""""""""""""""""
 
 **Syntax**
 
@@ -284,9 +284,9 @@ and reason as to why they were kicked/banned.
 
 .. _mod-command-modset-banshowextrafield:
 
-"""""""""
+"""""""""""""""""""""""""""
 modset dm banshowextrafield
-"""""""""
+"""""""""""""""""""""""""""
 
 **Syntax**
 
@@ -296,8 +296,9 @@ modset dm banshowextrafield
 
 **Description**
 
+Toggle whether to show an extra customizable field when banning.
 
-Toggle whether to show an extra customizable field when banning. This is useful to add information such as a ban appeal link.
+This can be used to add additional information for the banned user, such as a ban appeal link.
 
 **Arguments**
 
@@ -305,9 +306,9 @@ Toggle whether to show an extra customizable field when banning. This is useful 
 
 .. _mod-command-modset-banextrafieldtitle:
 
-"""""""""
+""""""""""""""""""""""""""""
 modset dm banextrafieldtitle
-"""""""""
+""""""""""""""""""""""""""""
 
 **Syntax**
 
@@ -319,17 +320,17 @@ modset dm banextrafieldtitle
 
 Set the title for the optional extra embed on ban
 
-Set to "clear" to remove.
+Cannot be over 252 characters long.
 
 **Arguments**
 
-* ``[title]``: The title of the embed field. Can by any string of text.
+* ``[title]``: The title of the embed field. Can by any string of text under 252 charcters long.
 
 .. _mod-command-modset-banextrafieldcontents:
 
-"""""""""
+"""""""""""""""""""""""""""""""
 modset dm banextrafieldcontents
-"""""""""
+"""""""""""""""""""""""""""""""
 
 **Syntax**
 
@@ -341,11 +342,11 @@ modset dm banextrafieldcontents
 
 Set the contents for the optional extra embed on ban
 
-Set to "clear" to remove.
+Cannot be over 1024 characters long.
 
 **Arguments**
 
-* ``[contents]``: The contents of the embed field. Can by any string of text.
+* ``[contents]``: The contents of the embed field. Can by any string of text under 1024 charcters long.
 
 .. _mod-command-modset-requirereason:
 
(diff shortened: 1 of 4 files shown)


Flame442:
This PR is currently failing the style check. This repo uses the Black autoformatter, and your changes need to be modified to comply with its specifications. Instructions on how to use it can be found in our [contributing.md file](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/CONTRIBUTING.md#43-style). Be sure to use the `-l 99` flag when running Black.

VasilisThePikachu:
Should I also update the docs for the mod page? I don't see it being a requirement in contrib.md but considering I changed how enabling dm on ban/kick works I feel like it's in order.

Flame442:
If any new commands are added or command docstrings are updated, it would be useful to also update the docs to have parity.

redbot/cogs/mod/settings.py
```diff
@@ -370,6 +388,62 @@ async def dm(self, ctx: commands.Context, enabled: bool = None):
...
+                _("The extra embed field is currently set to: {setting}").format(setting=setting)
+            )
+            return
+        await self.config.guild(guild).ban_show_extra.set(enabled)
+        if enabled:
+            await ctx.send(
+                _(
+                    "An extra field will be shown when banning. "
+                    f"Configure it with `{ctx.clean_prefix}modset dm banextrafieldtitle` "
+                    f"and `{ctx.clean_prefix}modset dm banextrafieldcontents`"
```
Flame442: Generally, you cannot use `f` strings in our repo. Our translation logic (in the `_` object, called on a string using `_("Example string")`) maps a template to a translated string. If the string is pre-interpolated, the translator cannot match to a valid template. For this reason, you will see `.format` used _after_ the translation throughout our repo:
```py
_("Example {variable}").format(variable=myvar)
```
This allows the string to be translated to another language with the variable indicator still present, while still allowing the variable to be formatted.

redbot/cogs/mod/settings.py
```diff
@@ -370,6 +388,62 @@ async def dm(self, ctx: commands.Context, enabled: bool = None):
                 _("Bot will no longer attempt to send a DM to user before kick and ban.")
             )
 
+    @dm.command(name="banshowextrafield")
+    async def dm_banshowextrafield(self, ctx: commands.Context, enabled: bool = None):
+        """
+        Toggle whether to show an extra customizable field when banning. This is useful to add information such as a ban appeal link.
```
Flame442:
```suggestion
        Toggle whether to show an extra customizable field when banning.

        This can be used to add additional information for the banned user, such as a ban appeal link.
```
Phrasing & formatting (should also be applied in the docs)

redbot/cogs/mod/settings.py
```diff
@@ -347,9 +359,15 @@ async def reinvite(self, ctx: commands.Context):
                 )
             )
 
-    @modset.command()
+    @modset.group()
     @commands.guild_only()
-    async def dm(self, ctx: commands.Context, enabled: bool = None):
+    async def dm(self, ctx: commands.Context):
+        """
+        Settings for messaging the user on moderation action.
```
Flame442: I believe this still only applies to kicks and bans, not all moderation actions

redbot/cogs/mod/settings.py
```diff
@@ -370,6 +388,62 @@ async def dm(self, ctx: commands.Context, enabled: bool = None):
...
+
+    @dm.command(name="banextrafieldtitle")
+    async def dm_banextrafieldtitle(self, ctx: commands.Context, *, title: str) -> None:
+        """
+        Set the title for the optional extra embed on ban
+
+        Set to "clear" to remove.
+        """
+        guild = ctx.guild
+        if title == "clear":
```
Flame442: (nitpick) Clearing just sets these two value to the defaults, which are mostly meaningless values intended to act as placeholders. I'm not sure if there is much value to allowing the setting to be reverted like that.

redbot/cogs/mod/kickban.py
```diff
@@ -658,16 +671,38 @@ async def tempban(
...
+                )
+            if extra_embed:
+                extra_embed_title = await self.config.guild(guild).ban_extra_embed_title()
+                extra_embed_contents = await self.config.guild(guild).ban_extra_embed_contents()
+
+                em.add_field(
+                    name=bold(extra_embed_title),
+                    value=extra_embed_contents,
+                    inline=False,
                 )
```
Flame442: There is a limit of `256` characters for `embed.name`s and `1024` characters for `embed.value`s. The `add_field` calls should not ever be passed a string that exceeds this limit.
<https://discord.com/developers/docs/resources/message#embed-object-embed-limits>

One gotcha is the fact that `bold` will escape existing markdown formatting unless the `escape_formatting` kwarg is set to `False`. This could cause the title to look strange, or cause the escapes to count as extra unexpected characters in whatever logic is used to limit the character count.

redbot/cogs/mod/kickban.py
```diff
@@ -658,16 +671,38 @@ async def tempban(
...
+                value=reason if reason is not None else _("No reason was given."),
+                inline=False,
             )
-            if guild_data["dm_on_kickban"] and reason:
-                msg += _("\n\n**Reason:** {reason}").format(reason=reason)
             if invite:
-                msg += _("\n\nHere is an invite for when your ban expires: {invite_link}").format(
-                    invite_link=invite
+                em.add_field(
+                    name=bold("Here is an invite for when your ban expires"),
```
Flame442:
```suggestion
                    name=bold(_("Here is an invite for when your ban expires")),
```

docs/cog_guides/mod.rst
```diff
@@ -253,7 +253,23 @@ modset dm
...
+
+**Description**
+
+Settings for messaging the user on moderation action.
+
+.. _mod-command-modset-dm-sendmessage:
+
+"""""""""
+modset dm sendmessage
+"""""""""
```
Flame442: You are currently failing the docs style check. The over and under bars on these headers need to be the same length as the text they surround
```suggestion
"""""""""""""""""""""""""""""
modset dm sendmessage
"""""""""""""""""""""""""""""
```

Flame442 (changes requested):

redbot/cogs/mod/kickban.py
Flame442: This also applies to the `Cleared embed title` and `Cleared embed contents` strings, if that option is not removed

redbot/cogs/mod/kickban.py
VasilisThePikachu: From what I can tell, `bold` just adds the 4 stars to bold a message so that leaves us with 252 characters. From some testing, `len` does not count escape characters and counts what they become (so \n would be 1)

redbot/cogs/mod/settings.py
```diff
@@ -370,6 +388,65 @@ async def dm(self, ctx: commands.Context, enabled: bool = None):
...
+            )
+            return
+        await self.config.guild(guild).ban_show_extra.set(enabled)
+        if enabled:
+            await ctx.send(
+                _(
+                    "An extra field will be shown when banning. Configure it with `{prefix}modset dm banextrafieldtitle` and `{prefix}modset dm banextrafieldcontents`".format(
+                        prefix=ctx.prefix
+                    )
+                )
```
Flame442: The format is currently applied before the translation.
```suggestion
                _(
                    "An extra field will be shown when banning. Configure it with `{prefix}modset dm banextrafieldtitle` and `{prefix}modset dm banextrafieldcontents`"
                ).format(prefix=ctx.prefix)
```

redbot/cogs/mod/settings.py
```diff
@@ -370,6 +388,65 @@ async def dm(self, ctx: commands.Context, enabled: bool = None):
...
+            )
+        else:
+            await ctx.send(_("An extra field will be no longer be shown when banning."))
+
+    @dm.command(name="banextrafieldtitle")
+    async def dm_banextrafieldtitle(self, ctx: commands.Context, *, title: str) -> None:
+        """
+        Set the title for the optional extra embed on ban
```
Flame442:
```suggestion
        Set the title for the optional extra embed on ban.
```
Nitpick

redbot/cogs/mod/kickban.py
```diff
@@ -658,16 +671,38 @@ async def tempban(
...
+                )
+            if extra_embed:
+                extra_embed_title = await self.config.guild(guild).ban_extra_embed_title()
+                extra_embed_contents = await self.config.guild(guild).ban_extra_embed_contents()
+
+                em.add_field(
+                    name=bold(extra_embed_title),
+                    value=extra_embed_contents,
+                    inline=False,
                 )
```
Flame442: > From what I can tell, `bold` just adds the 4 stars to bold a message so that leaves us with 252 characters. From some testing, `len` does not count escape characters and counts what they become (so \n would be 1)

In the other ban case, you passed `False` as the second parameter to `bold`. That is probably why your testing showed that escape characters were not counted, since they were never included at all. In this case (tempbans), formatting is escaped, which results in cases where the ban DM is not sent due to the suppressed `discord.HTTPException` triggering when the resulting title exceeds the character count. You need to also disable escaping logic here to avoid issues.

```suggestion
                em.add_field(
                    name=bold(extra_embed_title, escape_formatting=False),
                    value=extra_embed_contents,
                    inline=False,
                )
```

docs/cog_guides/mod.rst
```diff
@@ -266,6 +282,72 @@ and reason as to why they were kicked/banned.
...
+
+**Syntax**
+
+.. code-block:: none
+
+    [p]modset dm banextrafieldtitle [title]
+
+**Description**
+
+Set the title for the optional extra embed on ban
```
Flame442:
```suggestion
Set the title for the optional extra embed on ban.
```
Nitpick

redbot/cogs/mod/kickban.py
```diff
@@ -153,6 +155,17 @@ async def ban_user(
...
                         inline=False,
                     )
+                    if extra_embed:
+                        extra_embed_title = await self.config.guild(guild).ban_extra_embed_title()
+                        extra_embed_contents = await self.config.guild(
+                            guild
+                        ).ban_extra_embed_contents()
+
+                        em.add_field(
+                            name=bold(extra_embed_title, False),
```
Flame442:
```suggestion
                            name=bold(extra_embed_title, escape_formatting=False),
```
Nitpick

Flame442 (changes requested):

Flame442 (approved):
Thank you for contributing!

PR #6594, opened 2025-08-12
Fix fuzzy command search not ignoring blacklisted users

### LINKED ISSUE #6482: [Core] Fuzzy command help should ignore blocklisted individuals

Core/HelpFormatter, command_not_found function

When a Red is set to provide fuzzy help on a command, it should ignore blocklisted individuals. Currently it will answer anyway.

### Anything else?

_No response_

karlsbjorn:

Fixes fuzzy command search not ignoring blacklisted users
Closes #6482 

Yes


Jackenmen (dismissed):
This does technically solve the specific issue that it aims to, but the true reason this is happening is that the bot's global checks are not called - you can observe the same problem with e.g. the ignore list:
image
We *could* instead check the output of `bot.message_eligible_as_command`, but that may still be a band-aid fix? i.e. other cogs could be adding their own checks that wouldn't be called then (but then, on the other hand, maybe they shouldn't, since such checks might not expect the provided context to have no command? + we would deviate quite a bit from d.py's behaviour, if we called these checks)

Kowlin (approved):
Works like a charm.

PR #6609, opened 2025-08-27
Fix audio command docstrings not translating

### LINKED ISSUE #6608: Audio docstrings are not properly translated

3.5.20

### Description

Docstring translations occur by passing a `Translator` object to the `@cog_i18n()` decorator. The translator object used to translate docstrings is currently a single object which is attached to the cog class. Audio has its code split into a large number of smaller files, including several files which define commands located in `audio/core/commands/`. The `@cog_i18n()` call on the `Audio` class specifies the folder to search for docstring translations as `audio/core/`, which causes the bot to fail to locate the translations for docstrings correctly. 

https://github.com/Cog-Creators/Red-DiscordBot/blob/cc803579a0ef4c5ea88e9409efea81ab30888817/redbot/cogs/audio/core/__init__.py#L28-L39

https://imgur.com/a/fHOpIfU#qr2e1MJ

Flame442:
The `cog_i18n` translator object was referencing a directory which contains no commands. Moved where this is applied to the directory with commands.

Fixes #6608

Yes

### DIFF
--- a/redbot/cogs/audio/core/__init__.py
+++ b/redbot/cogs/audio/core/__init__.py
@@ -3,7 +3,6 @@ import datetime
 import json
 
 from collections import Counter, defaultdict
-from pathlib import Path
 from typing import Mapping, Dict
 
 import aiohttp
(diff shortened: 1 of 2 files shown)


Jackenmen (changes requested):
This means that Audio's cog docstring won't be picked up by redgettext. The simplest solution (even if a bit of a workaround) would be to move the docstring to the `Commands` class since the class docstrings are inherited when the subclass doesn't define its own anyway.

Jackenmen (approved):

PR #6611, opened 2025-09-01
[Audio] Update application.yml source options

aikaterna:
Changed the application.yml to allowed video metadata loading on all clients that support it.

Other changes added here:
Mirrored the settings from the youtube-source readme, that describes all current client options and their recommended settings.
* Enabled metadata searching on MWEB.
* Disabled playback on TV - playback only works with Oauth. All current settings are `False`, so it's going to be skipped. Should we remove it and the settings for now? Users that choose to leverage Oauth will need to add these values back into their copy and change a couple of the settings, so I'm inclined to leave it as-is so they have it easily available in their current copy from Red.
* Disabled playback on TVHTML5EMBEDDED - playback only works with Oauth.

Yesn't

The video metadata change was tested through a handful of users in the support channel, including myself on my own hardware. However, the last changes specifically mentioned here were not explicitly tested as my current hardware for today isn't experiencing the original issue.


Kowlin (approved):
Thank you as always Aik!

PR #6620, opened 2025-09-26
Move the reasoning for the audit log to the front

Kowlin:
Yes... I know it looks jank... This also has some other problems in regards to cutting off *who* did the action if it goes over 500 chars. But realistically speaking, this shouldn't happen that often, and if it does. Pray that the end user has a modlog setup!

Yes


Jackenmen:
What's the motivation for this? I'm unsure why this is better than the status quo.

PR #6628, opened 2025-11-04
fix(core): rpc reload has old cache

NNTin:

> When a cog is unregistered and re-registered, the RPC methods are being removed and added, but the JsonRpcMethod wrapper is storing a reference to the old method object from the original cog instance.
> - Investigate why `rpc_reload()` in `redbot/core/core_commands.py` doesn't properly refresh module code
> - The issue: `_cleanup_and_refresh_modules()` is called in `rpc_load()` but old method references may persist
> - Fix the reload sequence to ensure modules are refreshed before re-reigstering RPC handlers
> - Ensure `bot.remove_cog()` properly unregisters old RPC handlers before new ones are added

Yes, via tests
No, via manual reproduction of issue

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -12,6 +12,7 @@ import functools
 from collections import namedtuple, OrderedDict
 from datetime import datetime
 from importlib.machinery import ModuleSpec
+import importlib.util
 from pathlib import Path
 from typing import (
     Optional,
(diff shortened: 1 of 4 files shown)


NNTin:
```bash
            # Call RPC handler and verify it now executes new code
            result = await new_handler()
>           assert result == "version_2", "RPC handler should execute new code after reload"
E           AssertionError: RPC handler should execute new code after reload
E           assert 'version_1' == 'version_2'
E             - version_2
E             ?         ^
E             + version_1
E             ?         ^
```
from https://github.com/Cog-Creators/Red-DiscordBot/actions/runs/19065280048/job/54454289643?pr=6628
I can now reproduce the problem

NNTin:
The test with the version update was fixed in commit 9650a09ef944c15b7e0cd643056783c9ec6235e2 (?)
Issue was caching.

python -m pytest tests/core/test_rpc.py::test_rpc_handler_updates_on_reload -v -s --tb=short

Solving rpc reload via websocket, has the same root of error:
```
Before reload:
Original cog: 2280439307152
Original handler: 2280439127936
Original cog method: 2280439129536

After reload:
New cog: 2280439307152 ← SAME COG OBJECT!
New handler: 2280439127936 ← SAME HANDLER OBJECT!
New cog method: 2280360060864 ← DIFFERENT METHOD OBJECT!
```
Solving that was a mess. Month just started and my GitHub Copilot is already at 50% usage 😖

PR #6634, opened 2025-11-21
Add test: When updating a cog the requirements from info.json are updated as well

NNTin:
~~When a cog is installed the dependencies from info.json is automatically installed. When a cog is updated the dependency is now also updated.~~

~~Not yet. Test has been written but it is not tested on a Discord Bot. Still in draft.  Will have to review a bit if this is implemented right. Opening the PR now because I want to see the pipeline result~~

**Misunderstood requirement. This is already working. So just added a test. Can be merged or be abandoned.**

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -340,11 +340,8 @@ class Downloader(commands.Cog):
                 except ValueError:
                     # module wasn't modified - we just need to update its commit
                     module.commit = repo.commit
-                    if (
-                        module.type == InstallableType.COG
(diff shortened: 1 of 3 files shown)


NNTin:
Misunderstood requirement. This is already working. So just added a test. Can be merged or be abandoned.
Feedback is much appreciated. PR is now ready for review.

edit: putting on draft again. command "[p]cog update" is still not properly tested
edit2: **closing and abandoning PR. It does not perform a proper test. To test [p]cog update I need to mock too much at which point I am unsure if the cog update's functionality is actually verified or it is only green because the mocks make it green.**

PR #6641, opened 2025-12-10
`[p]mywarnings` improvements - Respect `show_mod` and allow sending self requested warnings to DMs

VasilisThePikachu:

The `[p]mywarnings` command will now respect `show_mod` and hide the moderator who made the warning from showing up.
image
This may require locale changes due to the way the message is now formatted. I can revert these changes and instead replace the moderator with some generic "anonymous" string or something. But IMO this looks pretier without the clutter if showing the mod is turning off.

It now also has the option to be sent in a dm instead of the current channel. Preventing clutter and members "flexing" their warnings.
image
(in dms)
image

Yes


Kowlin (approved):
LGTM

Image

PR #6646, opened 2026-01-29
fix: Handle permission errors when rotating log files

MeatyChunks:
## Summary

Fixes PermissionError crashes during Red startup when log files are locked by another process (common on Windows).

## Changes

- Added try/except around `path.unlink()` to handle PermissionError and FileNotFoundError
- Added try/except around `path.replace()` to handle PermissionError
- Logs warnings when files are in use instead of crashing
- Fixes crashes during `init_logging()` when `latest.log` is locked

## Testing

Tested on Windows with Python 3.12 and discord.py 2.7+

Fixes crashes with:
```
PermissionError: [WinError 32] The process cannot access the file because it is being used by another process
```

### DIFF
--- a/redbot/core/_events.py
+++ b/redbot/core/_events.py
@@ -434,7 +434,7 @@ def init_events(bot, cli_flags):
             msg = _("This command is on cooldown. Try again {relative_time}.").format(
                 relative_time=relative_time
             )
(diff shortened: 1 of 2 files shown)


MeatyChunks:
misclick, apologies.

PR #6651, opened 2026-02-12
Add warning to `[p]set api` if <>s are included in secret

### LINKED ISSUE #6253: Add warning to [p]set api if <>s are included in secret

Evanroby:
addressed review feedback, meaning I have:
- wrapped command examples with inline code marks for readability
- moved warning text inline instead of using an embed

related to #6265 
fixes #6253 

No

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3715,7 +3715,6 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
                         " was `HREDFGWE`, make sure to run `[p]set api {service} {token_name} HREDFGWE` and not "
                         "`[p]set api {service} {token_name} <HREDFGWE>`."
                     ).format(token_name=token_name, service=service)
-                    log.warning(angle_bracket_warning)
                     break
 
             await ctx.bot.set_shared_api_tokens(service, **tokens)
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3709,7 +3709,7 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
 
             for token_name, token in tokens.items():
                 if token.startswith("<") and token.endswith(">"):
-                    angle_bracket_warning = (
+                    angle_bracket_warning = _(
                         "You may have failed to properly format your `{token_name}`. If you were told to enter a token"
                         " with an example such as `[p]set api {service} {token_name} <your_{token_name}_here>`, and your {token_name}"
                         " was `HREDFGWE`, make sure to run `[p]set api {service} {token_name} HREDFGWE` and not "
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -3704,8 +3704,27 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         else:
             if ctx.bot_permissions.manage_messages:
                 await ctx.message.delete()
+
+            angle_bracket_warning = None
+
+            for api_service_name, token in tokens.items():
+                if token.startswith("<") and token.endswith(">"):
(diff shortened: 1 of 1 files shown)


redbot/core/core_commands.py
```diff
@@ -3704,8 +3704,27 @@ async def _set_api(
...
+
+            for token_name, token in tokens.items():
+                if token.startswith("<") and token.endswith(">"):
+                    angle_bracket_warning = (
+                        "You may have failed to properly format your `{token_name}`. If you were told to enter a key"
+                        " with an example such as `[p]set api {service} {token_name} <your_api_key_here>`, and your {token_name}"
+                        " was `HREDFGWE`, make sure to run `[p]set api {service} {token_name} HREDFGWE` and not "
+                        "`[p]set api {service} {token_name} <HREDFGWE>`."
+                    ).format(token_name=token_name, service=service)
+                    log.warning(angle_bracket_warning)
```
Jackenmen: I don't think it makes sense to log this when it's directly associated with the command that already outputs this warning.
```suggestion
```

redbot/core/core_commands.py
```diff
@@ -3704,8 +3704,27 @@ async def _set_api(
         else:
             if ctx.bot_permissions.manage_messages:
                 await ctx.message.delete()
+
+            angle_bracket_warning = None
+
+            for token_name, token in tokens.items():
+                if token.startswith("<") and token.endswith(">"):
+                    angle_bracket_warning = (
```
Jackenmen: i18n has to be applied to a literal string.
```suggestion
                    angle_bracket_warning = _(
```

redbot/core/core_commands.py
```diff
@@ -3704,8 +3704,27 @@ async def _set_api(
...
+                    ).format(token_name=token_name, service=service)
+                    log.warning(angle_bracket_warning)
+                    break
+
             await ctx.bot.set_shared_api_tokens(service, **tokens)
-            await ctx.send(_("`{service}` API tokens have been set.").format(service=service))
+
+            message = _("`{service}` API tokens have been set.").format(service=service)
+            if angle_bracket_warning:
+                message += "\n\n" + _("**Warning:** ") + _(angle_bracket_warning)
```
Jackenmen: No need for this with [the other comment](https://github.com/Cog-Creators/Red-DiscordBot/pull/6651/changes#r3278352310):
```suggestion
                message += "\n\n" + _("**Warning:** ") + angle_bracket_warning
```

Jackenmen (changes requested):

Evanroby:
Applied requested changes !

Jackenmen (approved):

PR #6671, opened 2026-03-04
Add Sphinx extension for extracting prompt contents

Jackenmen:

You can completely ignore this, as this extension does not do anything *unless* you explicitly specify to use the builder that it adds. I made this because I'm hoping to get the install instructions for automated install tests _directly_ from the docs.

But, in case you're interested, you can run the following commands to see how this works:
```
cd docs
make jsonprompt
make textprompt
```

Yes


Kowlin (approved):
Trusting you on this

PR #6680, opened 2026-03-04
Add --no-debug flag for resetting the verbosity level

Jackenmen:

I was a tad bit surprised that the `--debug` flag does not have a `--no-debug` counterpart. This PR implements one such that it resets the set verbosity level back to 0.

Yes

### DIFF
--- a/redbot/core/_cli.py
+++ b/redbot/core/_cli.py
@@ -244,6 +244,14 @@ def parse_cli_flags(args):
         dest="logging_level",
         help="Increase the verbosity of the logs, each usage of this flag increases the verbosity level by 1.",
     )
+    parser.add_argument(
+        "--no-verbose",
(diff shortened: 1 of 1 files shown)


Kowlin:
Don't we set the verbosity back to default already when we don't specify our flags? Or is this doing more here that is escaping my head?

Jackenmen:
The point is to support `--debug --no-debug` in a single invocation. This is common to support resetting what the switch did when using bash command aliases.
Say I had a `alias run-red=python -Om redbot my-instance --debug --dev`, I can then run `run-red --no-debug` to disable debug despite `--debug` already being passed to the command.

Kowlin:
Okay ye that's fair!

Kowlin (approved):

PR #6684, opened 2026-03-05
[Mutes] fix for timeout duration and add untimeout command

### LINKED ISSUE #6627: [Mutes] Unable to remove timeouts if a default mute duration is configured

Mutes

When a mod uses `[p]timeout @member` and if no time is passed then it doesn't uses the default mute time setting instead it removes the timeout incase user is already timedout.
Additionally, if we start using the default mute time setting, then there is no current way to remove the timeout.

### Anything else?

_No response_

Evanroby:
fixes #6627 

fixes an issue where `[p]timeout` with no duration and no default time set would silently remove the timeout instead of raising an error
also adds `[p]untimeout` to remove timeouts from members

No

### DIFF
--- a/docs/cog_guides/mutes.rst
+++ b/docs/cog_guides/mutes.rst
@@ -368,6 +368,34 @@ Examples:
 * ``<users...>``: A space separated list of usernames, ID's, or mentions.
 * ``[time_and_reason]``: The time and reason. If no time is provided, the mute will use the default set time or give an error if this hasn't been configured.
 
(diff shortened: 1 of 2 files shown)


TrustyJAID:
I haven't reviewed the code on this, but has discord fixed the issue where a user's `timeouted_out_until` is actually set to None after the time elapses?

Evanroby:
I'll just use `is_timed_out()` instead of checking if the timeout duration is None then, good catch

PR #6692, opened 2026-03-06
Use ubuntu-slim runners

Jackenmen:

I donno, if this will make sense, but I figured why not try?
https://github.blog/changelog/2026-01-22-1-vcpu-linux-runner-now-generally-available-in-github-actions/
IMO, almost every job our CI performs is lightweight, with the notable exception being testing and codeQL.

No

### DIFF
--- a/.github/workflows/auto_labeler_issues.yml
+++ b/.github/workflows/auto_labeler_issues.yml
@@ -8,7 +8,7 @@ permissions:
 
 jobs:
   apply_triage_label_to_issues:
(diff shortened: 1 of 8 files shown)


Jackenmen:
Admittedly, this could be a fluke, but this does not suggest it's in any way worth it, even on something simple like that:
image

Jackenmen:
And just as ~~quickly~~ slowly, we got it ~~open~~ to prove this PR should just be closed...

PR #6702, opened 2026-03-22
fix: allow setting embed colour to None for theme-adaptive color

### LINKED ISSUE #6618: [Set]:`None` colour not accepted in `set colour`

3.5.22

### Cog name

Core

### Command name

set colour

I expected that "None" colour for embeds would be allowed, tho it appears it’s not allowed by converter.

### What actually happened?

Image
Image
It said it isn’t allowed, tho we can use it when making embeds on our own.

1. Run `[p]set colour None` on your redbot instance.

### Anything else?

_No response_

Sim-hu:
Closes #6618

The `[p]set colour` command only accepted valid `discord.Colour` values, so passing
`none` would fail with a conversion error. This meant there was no way to set the embed
colour to null, which Discord renders as the user's theme colour.

Changes:
- Accept `none` as a special keyword in `[p]set colour` to clear the fixed colour
- Store `None` in config and propagate it through `get_embed_color` / `embed_colour`
- Show "Theme (auto)" in `[p]showsettings` when colour is unset
- Passing no argument still resets to the default (red), matching existing behavior

Following the approach suggested by Flame442 in the issue discussion.

### DIFF
--- a/redbot/core/bot.py
+++ b/redbot/core/bot.py
@@ -989,9 +989,7 @@ class Red(
         """
         await self._prefix_cache.set_prefixes(guild=guild, prefixes=prefixes)
 
-    async def get_embed_color(
-        self, location: discord.abc.Messageable
-    ) -> Optional[discord.Color]:
+    async def get_embed_color(self, location: discord.abc.Messageable) -> Optional[discord.Color]:
         """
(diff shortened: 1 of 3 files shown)


Jackenmen:
Superseded by #6707

PR #6704, opened 2026-03-24
Use Simple Repository API for fetching latest version

Jackenmen:

Things become a bit less... simple, but this makes use of [the API that's *actually* standardised](https://packaging.python.org/en/latest/specifications/simple-repository-api/). Side "benefit", perhaps, is the ability to point this at a PyPI mirror/private index with an env var. I suppose one could even (ab)use this to prevent getting version update notifications on startup, but :shushing_face: 

Also, we now prefer using `packaging.version.Version` for version numbers received from outside, *just in case* we wanted to make changes in our version scheme in the future.

Yes


palmtree5 (approved):
I don't see any issues with this off hand. Not sure really how to test this tbh or if I really need to test just this

PR #6707, opened 2026-03-29
[Core]: Allow setting embed color to none.

### LINKED ISSUE #6618: [Set]:`None` colour not accepted in `set colour`

Evanroby:

Fixes #6618
Rewrite of AI-looking PR #6702 

- `[p]set colour` converted to a group so missing args now errors instead of silently resetting to red
- `[p]set colour theme` sets embed colour to `None` (Discord's default theme colour)
- `[p]set colour reset` explicitly resets to Red's default (Flame Red, `0xE74C3C`, or `15158332`)
- `get_embed_color` and `embed_colour` return type updated to `Optional[discord.Colour]`
- `_pre_login` guarded against `None` color value to prevent crash on startup

Yes

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -4120,9 +4120,9 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         **Example:**
         - `[p]set colour reset`
         """
-        default_color = discord.Color(15158332)
+        await ctx.bot._config.color.clear()
+        default_color = await ctx.bot._config.color()
         ctx.bot._color = default_color
-        await ctx.bot._config.color.set(default_color.value)
         await ctx.send(_("The embed colour has been reset to the default."))
 
     @_set.command(
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -4089,10 +4089,6 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         - `[p]set color 0x#FDFEFE`
         - `[p]set color #7F8C8D`
 
-        **Subcommands:**
-        - `[p]set colour theme` - Use Discord's default embed colour (no colour set).
-        - `[p]set colour reset` - Reset to Red's default colour.
-
         **Arguments:**
         - `<colour>` - The colour to use for embeds.
         """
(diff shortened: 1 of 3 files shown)


redbot/core/core_commands.py
```diff
@@ -4090,16 +4090,41 @@ async def _set_colour(self, ctx: commands.Context, *, colour: discord.Colour = N
...
+    async def _set_colour_reset(self, ctx: commands.Context):
+        """
+        Resets the bot's embed colour to Red's default.
+
+        **Example:**
+        - `[p]set colour reset`
+        """
+        default_color = discord.Color(15158332)
+        ctx.bot._color = default_color
+        await ctx.bot._config.color.set(default_color.value)
```
```suggestion
        await ctx.bot._config.color.clear()
        default_color = await ctx.bot._config.color()
        ctx.bot._color = default_color
```
(untested suggestion)

`.clear` would be better, mostly to avoid inconsistencies if we ever decide to change the default color for some reason. That would require either fetching to know what the default was, or storing the default in some way that it can be imported. I would probably just go with the fetch, since this command does not particularly care about the mild performance hit.
https://github.com/Cog-Creators/Red-DiscordBot/blob/e2acec08623c6a31ed8f2c1134db269656e2d847/redbot/core/bot.py#L121

Flame442 (changes requested):

redbot/core/core_commands.py
Evanroby: I'll run testings with this, we'll see if it works. If it works correctly, I'll resolve

Evanroby: Works great for me.

redbot/core/core_commands.py
```diff
@@ -4090,16 +4090,41 @@ async def _set_colour(self, ctx: commands.Context, *, colour: discord.Colour = N
...
+    async def _set_colour_reset(self, ctx: commands.Context):
+        """
+        Resets the bot's embed colour to Red's default.
+
+        **Example:**
+        - `[p]set colour reset`
+        """
+        await ctx.bot._config.color.clear()
+        default_color = await ctx.bot._config.color()
+        ctx.bot._color = default_color
```
Flame442: Looking closer at the rest of the PR, would it be better for this to be `discord.Colour(default_color) if default_coloris not None else None`?

Evanroby: Great catch fixed it

PR #6708, opened 2026-03-29
[Core]: Fix for force enabled commands.

### LINKED ISSUE #6487: `[p]slash enablecog/disablecog` improperly handle force enabled commands

3.5.11

### Cog name

Core

### Command name

`[p]slash enablecog` and `[p]slash disablecog`

### Describe the issue

These commands still attempt to enable/disable commands which feature the `red_force_enable` key in `extras` set to `True` when their cog is used with one of these commands. These commands should be filtered out of consideration by checking for this key, similar to how `[p]slash disable` currently does.

1. `[p]repo add zeph https://github.com/Zephyrkul/FluffyCogs`
2. `[p]cog install zeph onetrueslash`
3. `[p]load onetrueslash`
4. `[p]slash enablecog onetrueslash` - Lists the command but should not, as it is force enabled
5 `[p]slash disablecog onetrueslash` - Lists the command but should not, as it is force enabled

### Anything else?

_No response_

Evanroby:

Fixes #6487
- `[p]slash enablecog` and `[p]slash disablecog` now skip commands with `red_force_enable` set in `extras`, for both slash commands and context menus
`[p]slash disable` already had this guard, `enablecog`/`disablecog` were just missed.

Yes

### DIFF
--- a/redbot/core/core_commands.py
+++ b/redbot/core/core_commands.py
@@ -2105,11 +2105,15 @@ class Core(commands.commands._RuleDropper, commands.Cog, CoreLogic):
         successful_cogs = set()
         # Fetch a list of command names to enable
         for name, com in self.bot.tree._disabled_global_commands.items():
+            if com.extras.get("red_force_enable", False):
+                continue
             for cog_name in cog_names:
                 if self._is_submodule(cog_name, com.module):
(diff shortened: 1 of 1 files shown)


Evanroby:
PR ready for review!

PR #3681, opened 2020-03-22
Add `redbot-setup restore` cli command for restoring instance from backup

Jackenmen:
### Type

- [ ] Bugfix
- [ ] Enhancement
- [x] New feature

### Description of the changes
Fixes #2947, depends on #6696

I added progress bars to indicate to the user that the command does something but they're surely not the most accurate progress bars - progress bar for (un)packing archives are based on file amount, not on file size and progress bar for downloading repos is based on repo amount. I'm open to improving it but I think it still serves its purpose even if it's not very linear.

Backup can now have a version of its own in case we need to change something about its format later.

#### Note:
Restoring cogs is still far from perfect as it requires running `[p]cog update` after the offline restoring process so that the previously installed cogs get reinstalled (it's totally possible to reinstall cogs offline but it would require rewriting more than I currently have time for).

Update: There's now work done on improving this as well:
https://github.com/Jackenmen/Red-DiscordBot/pull/47

PR #6709, opened 2026-03-29
Add `redbot-setup restore` cli command

### LINKED ISSUE #2947: [Backup/Downloader] Option to auto-reinstall cogs from backup
# Feature request

- [ ] Cog
- [x] Command
- [ ] API functionality

Currently `[p]backup` doesn't backup cogs from CogManager and RepoManager folder and backup repo urls in a `RepoManager/repos.json` file. It would be nice if the user would be able to automatically restore repos from this file and automatically install cogs that were previously installed.

Basically this is something that I want to see added to Core under `[p]restore` command or to Downloader under (I don't know what name of the command). If someone doesn't grab this, I eventually plan on writing this functionality but it may not happen any time soon. 
 
Related issue: #3421

Jackenmen:

Fixes #2947, depends on #6696, supersedes #3681

I added progress bars to indicate to the user that the command does something but they're surely not the most accurate progress bars - progress bar for (un)packing archives are based on file amount, not on file size and progress bar for downloading repos is based on repo amount. I'm open to improving it but I think it still serves its purpose even if it's not very linear.

Backup can now have a version of its own in case we need to change something about its format later.

Docs: https://red-discordbot--6709.org.readthedocs.build/en/6709/backup_red.html

Yes

### DIFF
--- a/redbot/setup.py
+++ b/redbot/setup.py
@@ -1,15 +1,19 @@
+from __future__ import annotations
+
 from redbot import _early_init
 
 # this needs to be called as early as possible
 _early_init()
 
 import asyncio
+import functools
 import json
 import logging
 import sys
 import re
 import tarfile
 from copy import deepcopy
+from dataclasses import dataclass
 from pathlib import Path
 from typing import Any, Dict, IO, Optional, Tuple, Union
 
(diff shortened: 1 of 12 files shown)


EternalllZM:
In regards to the docs on this, the current backup and restore page mentions (the possible need of) fixing permissions on Linux with the `sudo chown -R <user>:<user> ~/.local` command, that is removed in this PR.

`data_path` handles the restore location, would it be worth adding a check for file access? How would that be handled on a Windows to Linux or vise versa restore? Regardless, I think the command should remain in the backup/restore process docs just in case.

Jackenmen:
> In regards to the docs on this, the current backup and restore page mentions (the possible need of) fixing permissions on Linux with the `sudo chown -R <user>:<user> ~/.local` command, that is removed in this PR.

There should be no need for this unless the user has a seriously broken system. That's not on us to help with, and users will definitely just try running this command if something breaks for them due to unrelated reasons.

> `data_path` handles the restore location, would it be worth adding a check for file access?

Yes, I missed that earlier and have just noticed in testing that we are, in fact, not checking whether the path is accessible at any point. I just fixed that.

> How would that be handled on a Windows to Linux or vise versa restore

The updated code should make it so this fails on the `is_absolute()` call.

Jackenmen:
Okay, I added support for non-interactive use, and this is now (again) ready for review. Please note, however, that as described in the description, this PR depends on #6696 to work correctly, and the changeset from that PR is not included in this PR's branch.

Jackenmen:
I've made a slight change to metadata stored in the backup file - instead of a `backup.version` file storing just the version number, there's now a `backup_details.json` file with one `backup_version` key. While there was nothing stopping future versions of the backup format from storing additional metadata files, I figured it would be simpler if one could just use the same file for more purposes in the future.
Notably, I'm hoping this command could, in the future, have flags for making copies of installed cogs *and* installed repos, and the metadata would then have additional keys that tell you if such flags were used for the backup.
Here's an example of how such a change to the metadata would look in practice: https://github.com/Jackenmen/Red-DiscordBot/commit/3fb4816403d034ee7f08c8c59c603c303394ad65

Another small change is the addition of `dereference=True` parameter to `tarfile.open()` to ensure we backup the actual files rather than symlinks, if any are present in the data dir.

redbot/setup.py
```diff
@@ -372,6 +387,379 @@ async def remove_instance_interaction() -> None:
...
+        if not self.restore_downloader:
+            ignored_members |= {
+                "cogs/RepoManager/repos.json",
+                "cogs/RepoManager/settings.json",
+                "cogs/Downloader/settings.json",
+            }
+        return [member for member in self.all_tar_members if member.name not in ignored_members]
+
+    def print_instance_data(self) -> None:
+        print("\nWhen the instance was backuped, it was using these settings:")
```
palmtree5: Can we change `backuped` to `backed up` here?

palmtree5:
I tested this just now backing up my bot on Ubuntu 24.04 and restoring on Windows 11, it worked just fine and correctly told me it could not install 2 of the cogs, one because the repo wasn't setup on the bot at time of backup and one because the cog doesn't exist in the repo anymore. I did leave a comment on one minor thing in the code with wording on output of the settings as they were at time of backup though

redbot/setup.py
Jackenmen: Absolutely!

palmtree5 (approved):

PR #6710, opened 2026-03-29
Add `Environment` abstraction to internal Downloader API

### LINKED ISSUE #6718: max_bot_version from info.json is ignored, if min_bot_version is not specified

3.5.24

I tried to install a cog from a specific revision, which had `"max_bot_version": "3.4.99"` specified in `info.json`, but did not specify the `min_bot_version` key.

I expected the install to fail.

### What actually happened?

The install succeeded.

```
load downloader
repo add aika-repro https://github.com/aikaterna/aikaterna-cogs v3.4
cog install aika-repro reminder
```
You'll see that it successfully installs despite `max_bot_version` being set in `info.json`:
https://github.com/aikaterna/aikaterna-cogs/blob/e7bf1f22e656fd991ca3b13fc02aae23e34d7445/reminder/info.json#L9

### Anything else?

_No response_

Jackenmen:

This streamlines the version attributes to use `packaging.version.Version`, similarly to how #6704 did for `fetch_latest_red_version()` internal utility *and* it adds `Environment` that can be used to check against an environment that differs from the current one (which will be useful for the update command).

Fixes #6718 

Yes

### DIFF
--- a/redbot/cogs/downloader/downloader.py
+++ b/redbot/cogs/downloader/downloader.py
@@ -941,9 +941,7 @@ class Downloader(commands.Cog):
             ) + humanize_list(
                 [
                     inline(cog.name)
-                    + _(" (Minimum: {min_version})").format(
-                        min_version=".".join([str(n) for n in cog.min_python_version])
-                    )
+                    + _(" (Minimum: {min_version})").format(min_version=cog.min_python_version)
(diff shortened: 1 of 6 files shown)


palmtree5 (dismissed):
Don't see any issues here!

palmtree5 (approved):

PR #6724, opened 2026-04-01
Handle responding to interactions in help formatter for hybrids

### LINKED ISSUE #5916: Using `ctx.send_help()` in a hybrid command's slash command causes a timeout if the help menu is set to pagified

3.5.0.dev, commit bbb15924b969eb6fa170b3603556b5bcb9830a0f

Use `ctx.send_help()` to send a message as a response to a slash command while `[p]helpset usemenus` is set to `disable`.

Send the help message without any errors, like a `ctx.send()` or `[p]help test`.

### What actually happened?

The bot sent the help menu as its own message, but it also didn't respond to the interaction, causing it to time out.
Using `ctx.send_help()` with any other `[p]helpset usemenus` option works as intended.

eui9gtmx7F

1. `[p]helpset usemenus disable`
2. Add `ctx.send_help()` to a command
3. Trigger it

```py
from redbot.core.commands import commands


class Test(commands.Cog):
    def __init__(self, bot):
        self.bot = bot

    @commands.hybrid_command()
    async def test(self, ctx):
        await ctx.send_help()
```
The above code sample needs to be invoked as a slash command.

### Anything else?

Adding a `ctx.defer()` before `ctx.send_help()` does not fix this, it only extends the time it takes for the time out to happen.

Evanroby:

Fixes #5916 

This PR fixes an issue when using `ctx.send_help()` in a hybrid command's slash command, it causes a timeout if the help menu is set to not use menus. We are simply sending with `ctx.send()` instead of `destination.send()` on first page when it's an interaction and bot does not send help through dms.

Yes

### DIFF
--- a/redbot/core/commands/help.py
+++ b/redbot/core/commands/help.py
@@ -872,7 +872,7 @@ class RedHelpFormatter(HelpFormatterABC):
             and help_settings.use_menus is HelpMenuSetting.reactions
         ):
             use_DMs = help_settings.max_pages_in_guild == 0
-            destination = ctx.author if use_DMs else ctx.channel
+            destination = ctx.author if use_DMs and ctx.interaction is None else ctx
             # Specifically ensuring the menu's message is sent prior to returning
             m = await (destination.send(embed=pages[0]) if embed else destination.send(pages[0]))
             c = menus.DEFAULT_CONTROLS if len(pages) > 1 else {"\N{CROSS MARK}": menus.close_menu}
@@ -892,15 +892,13 @@ class RedHelpFormatter(HelpFormatterABC):
             delete_delay = help_settings.delete_delay
 
             messages: List[discord.Message] = []
-            for i, page in enumerate(pages):
+            page_destination = ctx if ctx.interaction and not use_DMs else destination
+            for page in pages:
                 try:
-                    use_ctx_send = ctx.interaction and not use_DMs and i == 0
                     if embed:
-                        msg = await (
-                            ctx.send(embed=page) if use_ctx_send else destination.send(embed=page)
-                        )
+                        msg = await page_destination.send(embed=page)
                     else:
-                        msg = await (ctx.send(page) if use_ctx_send else destination.send(page))
+                        msg = await page_destination.send(page)
                 except discord.Forbidden:
                     return await ctx.send(
                         _(
@@ -910,8 +908,9 @@ class RedHelpFormatter(HelpFormatterABC):
                     )
                 else:
                     messages.append(msg)
+                    page_destination = destination
             if ctx.interaction and use_DMs:
-                await ctx.interaction.response.send_message(
+                await ctx.send(
                     _("I have sent the help message to your DMs."), ephemeral=True
                 )
             elif use_DMs and help_settings.use_tick:
--- a/redbot/core/commands/help.py
+++ b/redbot/core/commands/help.py
@@ -890,14 +890,19 @@ class RedHelpFormatter(HelpFormatterABC):
             use_DMs = len(pages) > max_pages_in_guild
             destination = ctx.author if use_DMs else ctx.channel
             delete_delay = help_settings.delete_delay
+            is_interaction = ctx.interaction is not None
+            if is_interaction and use_DMs:
+                if not ctx.interaction.response.is_done():
+                    await ctx.defer(ephemeral=True)
 
             messages: List[discord.Message] = []
-            for page in pages:
+            for i, page in enumerate(pages):
                 try:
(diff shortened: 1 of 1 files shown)


redbot/core/commands/help.py
```diff
@@ -890,14 +890,21 @@ async def send_pages(
             use_DMs = len(pages) > max_pages_in_guild
             destination = ctx.author if use_DMs else ctx.channel
             delete_delay = help_settings.delete_delay
+            is_interaction = ctx.interaction is not None
+            if is_interaction and use_DMs:
+                if not ctx.interaction.response.is_done():
+                    await ctx.defer(ephemeral=True)
```
karlsbjorn: This will just have the bot be in an infinite `thinking` state since defer expects another interaction response, which is never provided in the case of a DM. It's better to just directly respond with an ephemeral message saying that the bot sent a DM.

Evanroby:
Both fixes implemented, let me know if there's more !

redbot/core/commands/help.py
Jackenmen: While this does only have one message, that page is not sent as a response to the interaction, so it's affected by the same issue too:
Image

The way to resolve this should be simple, though:
```suggestion
            destination = ctx.author if use_DMs and ctx.interaction is None else ctx
```

redbot/core/commands/help.py
```diff
@@ -907,7 +910,11 @@ async def send_pages(
                     )
                 else:
                     messages.append(msg)
```
Jackenmen: This seems simpler / avoids a conditional in every iteration and having to use `enumerate()`:
```suggestion
            page_destination = ctx if ctx.interaction and not use_DMs else destination
            for page in pages:
                try:
                    if embed:
                        msg = await page_destination.send(embed=page)
                    else:
                        msg = await page_destination.send(page)
                except discord.Forbidden:
                    return await ctx.send(
                        _(
                            "I couldn't send the help message to you in DM. "
                            "Either you blocked me or you disabled DMs in this server."
                        )
                    )
                else:
                    messages.append(msg)
                    page_destination = destination
```

---

By the way, I want to note that when there are fewer than 6 pages, we could _theoretically_ use `ctx.send()` because aside from the interaction response, it can send up to 5 follow-up responses. I'm not actually suggesting doing that here because it would need further discussion, but I just thought it was interesting.

redbot/core/commands/help.py
```diff
@@ -907,7 +910,11 @@ async def send_pages(
...
-            if use_DMs and help_settings.use_tick:
+            if ctx.interaction and use_DMs:
+                await ctx.interaction.response.send_message(
+                    _("I have sent the help message to your DMs."), ephemeral=True
+                )
```
Jackenmen: Considering that the command can use `await ctx.send_help()` after doing something else first, this should be a more universal solution.

Specficially, this will work better in case the interaction is expired (in which case, this will ensure the message is sent to the channel rather than as a response) *or* if the user has already responded to the interaction (in which case, this will ensure the message is sent through an interaction follow-up):
```suggestion
                await ctx.send(
                    _("I have sent the help message to your DMs."), ephemeral=True
                )
```

Jackenmen (changes requested):

Evanroby:
Updated files to match your suggestions !

redbot/core/commands/help.py
```diff
@@ -907,7 +908,12 @@ async def send_pages(
                     )
                 else:
                     messages.append(msg)
-            if use_DMs and help_settings.use_tick:
+                    page_destination = destination
+            if ctx.interaction and use_DMs:
+                await ctx.send(
+                    _("I have sent the help message to your DMs."), ephemeral=True
+                )
```
Jackenmen:
```suggestion
                await ctx.send(_("I have sent the help message to your DMs."), ephemeral=True)
```

redbot/core/commands/help.py
```diff
@@ -907,7 +908,10 @@ async def send_pages(
                     )
                 else:
                     messages.append(msg)
-            if use_DMs and help_settings.use_tick:
+                    page_destination = destination
+            if ctx.interaction and use_DMs:
+                await ctx.send(_("I have sent the help message to your DMs."), ephemeral=True)
+            elif use_DMs and help_settings.use_tick:
                 await ctx.tick()
```
Jackenmen: These `if`s have a common condition:
```suggestion
            if use_DMs:
                if ctx.interaction:
                    await ctx.send(_("I have sent the help message to your DMs."), ephemeral=True)
                elif help_settings.use_tick:
                    await ctx.tick()
```

Jackenmen (approved):

PR #6725, opened 2026-04-01
Fix handling of word boundaries in Filter cog

### LINKED ISSUE #3682: Filter Inconsistencies
This is on a fresh install of 3.3.2 but also existed on my 3.3.1 install, I imagine it's been like this a while since filter's been untouched for a bit.

With a filter list of `(h), test, ﷽, <h>` it will only remove "test" by default it seems. The only way I can get it to remove the unicode character or the brackets is by removing the boundary checks [here](https://github.com/Cog-Creators/Red-DiscordBot/blob/V3/develop/redbot/cogs/filter/filter.py#L356), which makes it a little too sensitive unfortunately. Are there any other workarounds for this?

Evanroby:

Fixes #3682 

refactors word list pattern generation to use a dedicated `_build_word_pattern` helper, for a more flexible and accurate word boundary handling

Yes

### DIFF
--- a/redbot/cogs/filter/filter.py
+++ b/redbot/cogs/filter/filter.py
@@ -475,7 +475,7 @@ class Filter(commands.Cog):
 
             if word_list:
                 pattern = re.compile(
-                    "|".join(_build_word_pattern(w) for w in word_list), flags=re.I
+                    "|".join(rf"(?<!\w){re.escape(w)}(?!\w)" for w in word_list), flags=re.I
                 )
             else:
                 pattern = None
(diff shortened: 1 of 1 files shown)


Jackenmen (changes requested):
Using negative lookahead/lookbehind with `\w` for word boundaries seems like an improvement over `\b` since, unlike `\b`, it *always* checks that the preceding character is not a word character instead of determining the behaviour based on the first/last character of the word. 

For example, the current implementation (as seen in stable releases) for a word such as `<h>` has:
- false positives: `"x<h>y"`
- false negatives: `" <h> "`

This PR handles both of these correctly.

With that said, this case is also handled fine by a simpler:
```python
rf"(?<!\w){re.escape(w)}(?!\w)"
```

I don't think we should really be deciding whether a word boundary should be checked based on whether the first/last character is a word character. Do you have some concrete cases where this would actually be an improvement rather than just make things possibly more confusing?

Evanroby:
Applied change!

Jackenmen (approved):

PR #6726, opened 2026-04-02
Prefer using repr() on the Dev cog results (excl. str instances)

Jackenmen:

This has been an annoyance for ages, and the simpler alternative of *always* using `repr()` on the object was rejected due to practicality of `return "some\nstring"`, sending back:
```
some
string
```
rather than `"some\nstring"`. That's why this PR special cases `str` instances.

No

### DIFF
--- a/redbot/core/dev_commands.py
+++ b/redbot/core/dev_commands.py
@@ -158,7 +158,7 @@ class DevOutput:
             output.append(self.formatted_exc)
         elif self.always_include_result or self.result is not None:
             try:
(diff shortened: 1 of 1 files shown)


TrustyJAID (approved):
I like this change, the repr gives more useful information when returning discord objects specifically. Now I don't need to return the guild.id if I want it I can just return the guild and pull the ID and name as I need it.

Kowlin (approved):
casually +1'ing this after I hated todays debugging session 🙃

PR #6727, opened 2026-04-04
[i18n]: fix sr-SP not being accepted

### LINKED ISSUE #5812: `[p]set locale` does not accept "Serbian (Cyrillic)" as valid

3.4.17

### Cog name

Core

### Command name

set locale

After typing `[p]set locale sr-SP` the bot should respond with `Locale has been set.` and switch to using the Serbian Cyrillic locale.

### What actually happened?

The bot responds with `Invalid language code. Use format: en-US` and the locale does not change.

1. Type `[p]set locale sr-SP`
2. Wait for bot to respond

### Anything else?

I'm aware that the "Serbian (Cyrillic)" translation for core Red is currently at 0%, but that shouldn't matter since "Serbian (Latin)" is also at 0% and yet the bot accepts that locale (with `[p]set locale sr-CS`) just fine.

There are also technically other locale codes for "Serbian (Cyrillic)" but since `sr-SP` is the only one crowdin uses (and the other locale codes don't work with crowdin), the bot should allow it.

Evanroby:

Fixes #5812 

As `sr-SP` is not a valid ISO 3166-1 code, Babel is throwing an UnknownLocaleError. The fix I have made makes an override to that, so `sr-SP` gets accepted now

Yes

### DIFF
--- a/redbot/core/_i18n.py
+++ b/redbot/core/_i18n.py
@@ -30,6 +30,10 @@ current_regional_format_default = None
 
 translators: List[Translator] = []
 
+_CROWDIN_LOCALE_OVERRIDES: dict[str, str] = {
+    "sr-SP": "sr-SP",
+}
+
 
 def _reload_locales() -> None:
     for translator in translators:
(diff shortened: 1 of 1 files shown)


Jackenmen:
Closing per https://github.com/Cog-Creators/Red-DiscordBot/issues/5812#issuecomment-4504763371

PR #6730, opened 2026-04-05
[Audio]: Add button support for `[p]now`

### LINKED ISSUE #6133: [p]now doesn't use buttons when buttons are enabled

3.5.1

### Cog name

Audio

### Command name

now

`[p]now` should use buttons like other menus in builtin cogs when `[p]set usebuttons` is toggled on

### What actually happened?

`[p]now` still uses reactions

1. `[p]set usebuttons`
2. `[p]play` a song
3. `[p]now`

### Anything else?

image

Evanroby:

Fixes #6133 

Adds a button menu in `utilities/menus` directory, that is used on the `[p]now` command when `[p]set usebuttons` is true instead of the old reaction based menu

Yes


redbot/cogs/audio/core/commands/controller.py
```diff
@@ -143,22 +143,29 @@ async def command_now(self, ctx: commands.Context):
...
-        if (
-            (dj_enabled or vote_enabled)
-            and not await self._can_instaskip(ctx, ctx.author)
-            and not await self.is_requester_alone(ctx)
-        ):
+        can_control = await self._can_instaskip(ctx, ctx.author) or await self.is_requester_alone(
+            ctx
+        )
+        has_queue = bool(player.queue) or autoplay
+        if await ctx.bot._config.use_buttons():
```
Jackenmen:
```suggestion
        if await ctx.bot.use_buttons():
```

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -143,22 +143,29 @@ async def command_now(self, ctx: commands.Context):
...
         )
         vote_enabled = await self.config.guild(ctx.guild).vote_enabled()
-        if (
-            (dj_enabled or vote_enabled)
-            and not await self._can_instaskip(ctx, ctx.author)
-            and not await self.is_requester_alone(ctx)
-        ):
+        can_control = await self._can_instaskip(ctx, ctx.author) or await self.is_requester_alone(
+            ctx
+        )
```
Jackenmen: You can just check for everything in here and then simplify the later if checks to `if not can_control:`
```suggestion
        can_control = (
            not (dj_enabled or vote_enabled)
            or await self._can_instaskip(ctx, ctx.author)
            or await self.is_requester_alone(ctx)
        )
```

redbot/cogs/audio/core/utilities/menus/now_playing.py
Jackenmen, line 1: "menu" in the name of the file seems a bit redundant when it's already the package name. Maybe rename to something like `now_playing.py`?

redbot/cogs/audio/core/commands/controller.py
```diff
@@ -82,7 +83,6 @@ async def command_disconnect(self, ctx: commands.Context):
     @commands.command(name="now")
     @commands.guild_only()
     @commands.bot_has_permissions(embed_links=True)
-    @commands.bot_can_react()
```
Jackenmen, line 85: This still needs to be checked for, just not if `usebuttons` is turned on.

Jackenmen (changes requested):
Aside from the review comments, I do feel it's a bit weird that we stop accepting actions after one action, but admittedly, this is consistent with the behaviour of how it works with reactions-based `[p]now`, so it makes more sense to address this in a separate PR.

Evanroby:
Applied suggested changes

PR #6731, opened 2026-04-06
[Mute]: Make `unmute` dms independant of `mute` dms.

### LINKED ISSUE #6262: Separate muteset senddm for mutes and unmutes

Evanroby:

Fixes #6262 

This PR introduces a new feature - mute dms and unmute dms fully independent toggles

## New command :
`[p]muteset unmutesenddm true/false` - controls unmute dms independently

`dm_on_unmute` defaults to `True` so existing behavior is preserved for guilds that had `dm=True`

No

### DIFF
--- a/docs/cog_guides/mutes.rst
+++ b/docs/cog_guides/mutes.rst
@@ -294,9 +294,9 @@ Decide whether the name of the moderator muting a user should be included in the
 
 .. _mutes-command-muteset-unmutesenddm:
 
-""""""""""""""""""""""
+""""""""""""""""""""
 muteset unmutesenddm
-""""""""""""""""""""""
+""""""""""""""""""""
 
 .. note:: |mod-lock|
 
--- a/docs/cog_guides/mutes.rst
+++ b/docs/cog_guides/mutes.rst
@@ -39,7 +39,7 @@ activemutes
 
 .. code-block:: none
 
-    [p]activemutes 
+    [p]activemutes
 
 **Description**
 
@@ -113,7 +113,7 @@ muteset
 
 .. code-block:: none
 
-    [p]muteset 
+    [p]muteset
 
 **Description**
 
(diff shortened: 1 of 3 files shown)


docs/cog_guides/mutes.rst
```diff
@@ -292,6 +292,31 @@ Decide whether the name of the moderator muting a user should be included in the
 
 * ``<true_or_false>``: Whether to enable or disable this setting, must provide ``true`` or ``false``.
 
+.. _mutes-command-muteset-unmutesenddm:
+
+""""""""""""""""""""""
+muteset unmutesenddm
+""""""""""""""""""""""
```
Jackenmen: We prefer to keep the number of header markers the same as the header's length:
```suggestion
""""""""""""""""""""
muteset unmutesenddm
""""""""""""""""""""
```

Jackenmen (changes requested):
I think a better default would be *not* to send DMs on unmute. To keep the behaviour unchanged for existing users, you can implement a migration that sets `dm_on_unmute` to `True` when `dm` is `True`.

Jackenmen (changes requested):
The migration part of my review remains unaddressed

Evanroby:
My bad I skipped that comment, will do on my free time.

Evanroby:
Applied requested changes

PR #6734, opened 2026-04-11
Add `redbot-update` command for updating Red

### LINKED ISSUE #3103: [Core] Add `redbot-update` command for updating Red
# Feature request

- [ ] Cog
- [x] Command (it's a cli command tho)
- [ ] API functionality

I think it would be nice to have an easier option to update than running `python3.7 -m pip install -U Red-DiscordBot[maybe_some_extra]`, it's long, not really something that a regular user would remember, the method of update differs a little on different OSes (executable name), it could just be easier to have a command like `redbot-update` (or if having something that simple in separate command is not good with you then perhaps `redbot --update` could work instead). I don't like `redbot-launcher` but that is one thing that is certainly easier for non-experienced user with it and we're gonna remove it at some point, I think it would be nice to have some good replacement for that functionality.

Jackenmen:

Depends on #6711, #6704, #6710, #6705, #6709
Fixes #3103

Here's an asciicast previewing the update process (note that some parts are skipped depending on the conditions + you can skip a lot of steps with command flags):
[asciicast](https://asciinema.org/a/iC9iu3XOLotS5okc)

### Notes

1. While I did test that it works on the legacy Windows (i.e. conhost rather than Windows Terminal), this looks much better in Windows Terminal and I might in the future change some of our docs to start pointing people towards using Windows Terminal.
2. On Windows, when using Python 3.8 or 3.9, you might find that the command may exit with `RuntimeError: Event loop is closed` **ignored exception**. This is something we work around in `redbot`, but I don't think it's worth complicating `redbot-update` to do the same. The workaround is not needed on 3.10+, so the problem will fully disappear with Red 3.6. This is cosmetic only; it does not affect the exit code.

Yes

### How to test this?

Due to the fact that there are no releases with redbot-update in them *yet*, it's not possible to test some of this without additional dev configuration.
Here are env vars available for overriding:
- `_RED_UPDATE_PRETEND_VERSION` - sets the version that `redbot-update` will assume to be the current version
- `_RED_UPDATE_PRETEND_SPECIFIER_TEMPLATE` - a dependency specifier that will be passed to `pip install` when "updating"
    - This string is `.format()`-ted with `{extras}` (e.g. `[postgres]`) and `{versionspec}` (e.g. `==3.5.24`) placeholders
    - By default, the specifier is `Red-DiscordBot {extras} {versionspec}`
    - Link to method: https://github.com/Cog-Creators/Red-DiscordBot/blob/22fca3c589d0cd057f74689d368e6d8a11c56969/redbot/_update/common.py#L50-L59
- `_RED_RTD_CANONICAL_URL` - base URL of the docs (`https://docs.discord.red/en/stable/` by default)
- `RED_SIMPLE_API_URL`, `_RED_SIMPLE_API_ENDPOINT_PATH` - allow to specify a different location that should be queried for available versions (by default, `https://pypi.org/simple/` and `Red-DiscordBot` respectively)

Basic env var configuration may look like this (note that I did not include any of the `export `/`$env:`/`set ` prefixes that are needed for these to actually be set in a shell, make sure to do that!):
```
_RED_UPDATE_PRETEND_VERSION='3.5.22'
_RED_UPDATE_PRETEND_SPECIFIER_TEMPLATE='Red-DiscordBot {extras} @ https://github.com/Jackenmen/Red-DiscordBot/archive/redbot_update.tar.gz'
_RED_RTD_CANONICAL_URL='https://jack1142-red-prs.readthedocs.io/en/make_changelog_easy_to_parse/'
```

If you want to test what happens when the current Python interpreter is incompatible or when a major update is released, you'll want to set the following to a custom Simple API endpoint:
```
RED_SIMPLE_API_URL='https://jackenmen.github.io/simple-api-red-testing/'
_RED_SIMPLE_API_ENDPOINT_PATH='Red-DiscordBot.json'
```
The above URL hosts a JSON file with an example `3.5.25` and `3.6.0` restricted to Python >= 3.10 (for demonstration purposes), which means you can test updating to a Red version requiring a different Python version from what you have (assuming you, e.g. use Python 3.8) *and* updating to a new major release.

You can also locally host a dir with a JSON file as described in the comment, if there are other scenarios you are interested in:
https://github.com/Jackenmen/Red-DiscordBot/blob/5849354a4429319270c9e674263dfe5bec82acef/redbot/core/utils/_internal_utils.py#L76-L82


palmtree5 (dismissed):
I tested and found this to be working (on Python 3.11 for the purposes of testing this)

palmtree5 (approved):

PR #6747, opened 2026-04-22
Fix files perms in send_interactive

palmtree5:

So I found a bug when doing `[p]eval` in a DM where it would respond, the immediately give an error message on the file perms check. DM channels are in fact `discord.channel.DMChannel`, not `discord.abc.User` it seems.

I have not tested this yet though


Jackenmen:
See #6697

PR #6756, opened 2026-05-16
Handle cog requirement reinstall in redbot-update

Jackenmen:

I forgot to do this when this command was introduced in #6734.

Yes


Jackenmen:
This has now been tested for the following cases:
- Python version changed during update (I also tested that Red doesn't remove the lib on startup, if it already got regenerated)
  image
- Python version did not change during update
  image

And everything works as expected.

Flame442 (approved):

PR #6765, opened 2026-05-19
Always allow installing a pre-release if explicitly requested

Jackenmen:

`redbot-update --version 3.6.0rc1` should allow installation of the specified pre-release regardless of what the current version is.

Yes

### DIFF
--- a/redbot/_update/updater.py
+++ b/redbot/_update/updater.py
@@ -302,7 +302,8 @@ class Updater:
             available_versions = await fetch_available_red_versions(
                 include_prereleases=(
                     self.options.red_version.is_prerelease
-                    or common.get_current_red_version().is_prerelease
+                    if self.options.red_version is not None
+                    else common.get_current_red_version().is_prerelease
                 )
             )
             latest_major = available_versions[0]
--- a/redbot/_update/updater.py
+++ b/redbot/_update/updater.py
@@ -300,7 +300,10 @@ class Updater:
         interpreter_info = self.options.new_python_interpreter or PythonInfo.current_system()
         with self.console.status("Checking latest version..."):
             available_versions = await fetch_available_red_versions(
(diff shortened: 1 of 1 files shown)


redbot/_update/updater.py
```diff
@@ -300,7 +300,10 @@ async def _prepare_metadata(self) -> None:
         interpreter_info = self.options.new_python_interpreter or PythonInfo.current_system()
         with self.console.status("Checking latest version..."):
             available_versions = await fetch_available_red_versions(
-                include_prereleases=common.get_current_red_version().is_prerelease
+                include_prereleases=(
+                    self.options.red_version.is_prerelease
```
Flame442, line 304: If a `red_version` option is not specified, is `self.options.red_version` a kind of object that has an `.is_prerelease` attribute? This line is before the `if self.options.red_version` check, so I don't think this is a safe line.

Flame442 (changes requested):

redbot/_update/updater.py
Jackenmen: It most definitely isn't, I'm very surprised that I never checked a simple `redbot-update` when testing it... Will fix

Jackenmen: The issue with `red_version` being `None` should be fixed now.
I should note that the behaviour of the PR is slightly different from before - we no longer fallback to `common.get_current_red_version().is_prerelease` when `self.options.red_version.is_prerelease` is `False`. I'm not sure if either is really more "correct" - we care for `--red-version 3.6.0rc1` to include pre-releases, but it doesn't really matter whether `--red-version 3.6.0` does, since we just iterate through versions for an exact match. This should mean that either of the implementations resulted in the same behaviour from the user's perspective.

Flame442 (approved):

PR #6768, opened 2026-05-21
Fix disconnect allowing DC from VC that user has no access to

### LINKED ISSUE #6365: [Audio] `disconnect` command should check if the user is actually in the VC

Jackenmen:

Fixes #6365, (supersedes) fixes #6385 
I added a new condition disallowing disconnect when the user:
- is not in a VC that the bot is in,
- cannot connect to that VC,
- and the VC has non-bot users in it (i.e. someone is *actually* listening to music in there)

Note that this PR might seem larger than it is - the first commit is a formatting-only change (removed `else` and dedented all lines), see how it looks with whitespaces hidden: https://github.com/Cog-Creators/Red-DiscordBot/pull/6768/changes/cf6055f92ed84eff62526a7e54a176c71f65043f

The actual changeset is the 2 commits that come afterwards, see diff:
https://github.com/Cog-Creators/Red-DiscordBot/pull/6768/changes/cf6055f92ed84eff62526a7e54a176c71f65043f..456b93e6559e59447c245072566b4208744ee955

Yes

### DIFF
--- a/redbot/cogs/audio/core/commands/controller.py
+++ b/redbot/cogs/audio/core/commands/controller.py
@@ -55,12 +55,6 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
                 title=_("Unable To Disconnect"),
                 description=_("You need the DJ role to disconnect."),
             )
-        if dj_enabled and not can_skip:
-            return await self.send_embed_msg(
-                ctx,
-                title=_("Unable to Disconnect"),
-                description=_("You need the DJ role to disconnect."),
-            )
 
         await self.send_embed_msg(ctx, title=_("Disconnecting..."))
         self.bot.dispatch("red_audio_audio_disconnect", ctx.guild)
--- a/redbot/cogs/audio/core/commands/controller.py
+++ b/redbot/cogs/audio/core/commands/controller.py
@@ -32,52 +32,50 @@ class PlayerControllerCommands(MixinMeta, metaclass=CompositeMetaClass):
         """Disconnect from the voice channel."""
         if not self._player_check(ctx):
             return await self.send_embed_msg(ctx, title=_("Nothing playing."))
-        else:
-            dj_enabled = self._dj_status_cache.setdefault(
-                ctx.guild.id, await self.config.guild(ctx.guild).dj_enabled()
-            )
-            vote_enabled = await self.config.guild(ctx.guild).vote_enabled()
-            player = lavalink.get_player(ctx.guild.id)
-            can_skip = await self._can_instaskip(ctx, ctx.author)
-            if (
-                (vote_enabled or (vote_enabled and dj_enabled))
-                and not can_skip
-                and not await self.is_requester_alone(ctx)
-            ):
-                return await self.send_embed_msg(
-                    ctx,
-                    title=_("Unable To Disconnect"),
-                    description=_("There are other people listening - vote to skip instead."),
-                )
-            if dj_enabled and not vote_enabled and not can_skip:
-                return await self.send_embed_msg(
-                    ctx,
-                    title=_("Unable To Disconnect"),
-                    description=_("You need the DJ role to disconnect."),
-                )
-            if dj_enabled and not can_skip:
-                return await self.send_embed_msg(
-                    ctx,
-                    title=_("Unable to Disconnect"),
-                    description=_("You need the DJ role to disconnect."),
-                )
 
-            await self.send_embed_msg(ctx, title=_("Disconnecting..."))
-            self.bot.dispatch("red_audio_audio_disconnect", ctx.guild)
-            self.update_player_lock(ctx, False)
-            eq = player.fetch("eq")
-            player.queue = []
-            player.store("playing_song", None)
-            player.store("autoplay_notified", False)
-            if eq:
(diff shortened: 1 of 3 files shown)


redbot/cogs/audio/core/commands/controller.py
```diff
@@ -33,17 +33,20 @@ async def command_disconnect(self, ctx: commands.Context):
...
         vote_enabled = await self.config.guild(ctx.guild).vote_enabled()
-        player = lavalink.get_player(ctx.guild.id)
         can_skip = await self._can_instaskip(ctx, ctx.author)
-        if (
-            (vote_enabled or (vote_enabled and dj_enabled))
-            and not can_skip
-            and not await self.is_requester_alone(ctx)
-        ):
+        is_alone = await self.is_requester_alone(ctx, channel)
+        if (vote_enabled or (vote_enabled and dj_enabled)) and not can_skip and not is_alone:
```
Flame442: Unless I am missing something, `(vote_enabled or (vote_enabled and dj_enabled))` is equivalent to `vote_enabled`. I'm not sure if this is supposed to be here for readability, if it is a behavior mistake, or if it was just an oopsie.

redbot/cogs/audio/core/commands/controller.py
Flame442: This check does not have an `is_alone` condition, so if there is a DJ role set but not voting, users that are alone in voice channels cannot skip. I assume that is unintended behavior?

Because every condition is going to need to check `is_alone` for the expected behavior, it would be \*nice\* to shortcut that condition earlier in the logic, but that would probably require either an ugly large-scoped `if` or a helper function just for this function to check if it should be able to skip, neither of which I love.

Flame442 (dismissed):
Nothing wrong with your changes (unless the way you refactored caused any of these issues I didn't really look closely), but some feedback on the existing implementation of the checks in this command while I am looking at them.

redbot/cogs/audio/core/commands/controller.py
Jackenmen: They are equivalent, yes. I did notice it, but since I'm not entirely certain that I know everything about how the DJ role and voting are intended to work, I didn't want to simplify the code just for the sake of it. I'm basically slightly worried that some information will be lost to the future reader if this weirdness isn't retained - i.e. if someone else looks at this weirdness in the future, they might think "hmm, I wonder why it is like this" and maybe then they'll finally add some documentation in the comments so the intention is clearer. Basically, I'm too lazy to dig into this myself for now, so I'm hoping some person in the future will be less lazy...

redbot/cogs/audio/core/commands/controller.py
Jackenmen: > This check does not have an `is_alone` condition, so if there is a DJ role set but not voting, users that are alone in voice channels cannot skip

_You'd think_ that is the case but instaskip actually checks for this too:
https://github.com/Cog-Creators/Red-DiscordBot/blob/443fa9f64fe2770510bbef324900c9bc5eb76334/redbot/cogs/audio/core/utilities/player.py#L103-L104
https://github.com/Cog-Creators/Red-DiscordBot/blob/443fa9f64fe2770510bbef324900c9bc5eb76334/redbot/cogs/audio/core/utilities/player.py#L707

If I had more interest in Audio, I'd probably want to make this less convoluted.

Flame442: https://github.com/Cog-Creators/Red-DiscordBot/blob/443fa9f64fe2770510bbef324900c9bc5eb76334/redbot/cogs/audio/core/utilities/player.py#L707
This line specifically only succeeds if the user is in a DIFFERENT voice channel. If a user is alone *with the bot*, it will not succeed, and they will not be able to `[p]disconnect`.

Jackenmen:
I've decided to rework the `if` statements further with some actual documentation in the code comments, which should actually address *both* of the review comments:
- future readers will no longer be confused because the expected behaviour is clearly stated, meaning I can remove `(vote_enabled or (vote_enabled and dj_enabled))`
- `is_alone` is checked, along with `can_skip` - if either of these is true, the user should be allowed to DC the bot; if not, we have `elif` conditions that check the specific early-return conditions that we previously had
    - as you might notice, this is a slight change in behaviour - when `can_instaskip` is `True`, we do not early return regardless of whether the user can connect to/is in the player's current VC; this is because `instaskip` is only `True` if one of the following is true:
        - the user has a DJ role
        - the user is a mod/guild owner/bot owner
        - the player's channel is equal to the user's channel (i.e. the `maybe_move_player()` function returned `True`, meaning the player moved to the user's channel)

PR #6782, opened 2026-05-26
Fix check_100_plus return value and missing i18n in cleanup cog

Evanroby:

this PR fixes 2 small bugs inside the cleanup cog, first one was reported by someone making testings for me, and tried to funnily do some impossibles things, this being that even when a number was above `2**63 - 1`, bot would delete messages still, exemple with `[p]cleanup bot 9999999999999999999`. the second bug is two small i18n translating issues, I could see that the parentheses were wrongly put and fixed it

Yes


Jackenmen (approved):
Looks good, thanks.

PR #6784, opened 2026-06-11
Privlieged intent changes

EternalllZM:

Update the documentation surrounding privileged intents based off Discord's latest announcement.

I am not sure of whether or not bots can now be in more than 100 servers without verification as long as the unique member count remains below 10,000. I have went ahead and removed it until clarification is added on that. I don't see it any longer.

This also updates the bot verification article which currently leads to a 404.

Yes

### DIFF
--- a/docs/intents.rst
+++ b/docs/intents.rst
@@ -12,11 +12,11 @@ our stance regarding "public bots", and the discord bot verification process.
 
 To clarify:
 
-- **Small bots** are bots under 100 servers. They currently do not need to undergo Discord's
-  bot verification process
-- **Public bots** (or big bots) are bots that have reached 100 servers. They need to be
+- **Small bots** are bots accessible to fewer than 10,000 unique users. They currently do not need to undergo Discord's
+  bot verification process.
+- **Public bots** (or big bots) are bots that have reached 10,000 total unique users. They need to be
   `verified <https://support-dev.discord.com/hc/en-us/articles/23926564536471-How-Do-I-Get-My-App-Verified>`_
-  by Discord to join more than 100 servers and gain privileged intents
+  by Discord to gain privileged intents.
 
 .. warning::
 
@@ -51,7 +51,7 @@ Bot verification process
 ------------------------
 
 When your bot ceases to be a small bot Discord will require you to verify your bot before allowing
-it to join more servers and gain privileged intents. If you've read the previous section,
+it to continue using privileged intents. If you've read the previous section,
 you will know that we do **not** support public bots. Logically, we also do not provide help for
 the verification process.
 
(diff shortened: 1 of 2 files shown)


docs/intents.rst
```diff
@@ -12,11 +12,11 @@ our stance regarding "public bots", and the discord bot verification process.
 
 To clarify:
 
-- **Small bots** are bots under 100 servers. They currently do not need to undergo Discord's
-  bot verification process
-- **Public bots** (or big bots) are bots that have reached 100 servers. They need to be
+- **Small bots** are bots accessible to fewer than 10,000 unique users. They currently do not need to undergo Discord's
```
EternalllZM: I was conflicted on whether to say "across all servers the bot is in" as I don't know whether that is implied. Red is meant to be installed to guilds, not as a user app.

yamikaitou:
> I am not sure of whether or not bots can now be in more than 100 servers without verification as long as the unique member count remains below 10,000

They are now used for different things. The 100 servers is used for App Verification and the 10k users is used for Privileged Intents. Under this new system, it does not appear (based on the docs that have been updated) you need a Verified App to request access to Privileged Intents.

EternalllZM:
> > I am not sure of whether or not bots can now be in more than 100 servers without verification as long as the unique member count remains below 10,000
> 
> 
> 
> They are now used for different things. The 100 servers is used for App Verification and the 10k users is used for Privileged Intents. Under this new system, it does not appear (based on the docs that have been updated) you need a Verified App to request access to Privileged Intents.

You are correct and that's an oversight on my part. Thanks for double checking. I will revise this PR to better differentiate.

Jackenmen:
Notably, privileged intents do not need ID verification but still need you to go through their abysmal application process.

EternalllZM:
Since bot verification and intent request are now separate, is needing to submit for intent access treated the same as now being a "public bot"?

Would it make sense for
> - **Small bots** are bots under 100 servers. They currently do not need to undergo Discord's
  bot verification process.

to be changed to say

> - **Small bots** are bots under 100 servers or 10,000 unique users. They currently do not need to undergo Discord's bot verification or intent review process.


A statement further down:

> Red was designed with one single goal in mind: a bot that you can host on your own hardware
and customize to your needs, making it really *your* bot. **The target audience of Red are server
owners with a few servers**

Except now when those few (or even just one) server(s) add up to 10,000+ unique users, which isn't all that difficult. Even the Red server has more than 10K users at this moment. Red itself used in the Red server is not something I would consider a "public bot".

Also, is there a need for more detail about the privileged intent application/process? The workflow isn't even there yet to apply, but it allegedly looks like this:
image

I think the "Intents" section further down covers it pretty well as a baseline already.

EternalllZM:
I'm going to close this pull request, not because I believe it isn't necessary to update and inform others about, but instead due to the fact I think the org should decide what are the best changes and apply them on their own timeline.

PR #6798, opened 2026-08-20
[Mod] Add per-channel overrides for deleterepeats

### LINKED ISSUE #4721: Add ability of using `[p]modset deleterepeats` per channel

Evanroby:

resolves #4721.

adds optional channel args to `deleterepeats` so you can override the repeat threshold per channel instead of just guild-wide. 0 clears an override back to the server default.

No

### DIFF
--- a/redbot/cogs/mod/events.py
+++ b/redbot/cogs/mod/events.py
@@ -19,21 +19,23 @@ class Events(MixinMeta):
     """
 
     async def check_duplicates(self, message):
-        guild = message.guild
-        author = message.author
-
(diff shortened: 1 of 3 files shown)


Evanroby:
I'm going to rework on it a bit more later, draft for now until I decide on some changes

PR #6803, opened 2026-08-31
docs: Document SetApiView/SetApiModal in Shared API Keys guide

### LINKED ISSUE #6114: Update Shared API Keys document with the new View functionality

Shared API Keys documentation

Currently, [this document](https://docs.discord.red/en/latest/framework_apikeys.html) describes the old, less fun way of setting API keys. I think it would be great if it could be updated with some examples of the newer functionality - the `SetApiModal`/`SetApiView` classes.

### Anything else?

_No response_

ktriponis:
Adds a "Prompting for API Keys" section to the Shared API Keys guide, covering `SetApiView` / `SetApiModal`.

Closes #6114

The guide only covers the `[p]set api` route and reading tokens back in code - nothing about `SetApiView`/`SetApiModal` in `redbot.core.utils.views`, which are a handy way to let a cog prompt the owner for keys via a secure button + modal instead of typing them in chat.

The new section covers:

- `SetApiView` with `default_service` / `default_keys` to prompt for a service's keys
- the modal's `key value`-per-line format, and that you keep the pre-filled key label (typing just the value gets rejected - this caught me out at first)
- `SetApiModal` for embedding in a custom `discord.ui.View`
- both being owner-only

Docs only, no code changes.

Yes

### DIFF
--- a/docs/framework_apikeys.rst
+++ b/docs/framework_apikeys.rst
@@ -54,7 +54,7 @@ Prompting for API Keys
 
 The primary way to set keys is ``[p]set api``. Run on its own, with no service or tokens, it opens a secure modal for the owner to fill in, so a cog does not need to add its own command for this.
 
-``SetApiView`` (from ``redbot.core.utils.views``) is most useful as a fallback when a required key is missing. When a command needs a key that has not been set, reply with the error and attach the button so the owner can set it in place. ``default_service`` and ``default_keys`` pre-fill the modal with the service and key names the cog expects.
+`SetApiView` is most useful as a fallback when a required key is missing. When a command needs a key that has not been set, a reply with the error message that includes this view will include a button that allows the owner to instantly open the secure modal. The parameters ``default_service`` and ``default_keys`` can be configured to pre-fill the modal with the service and key names the cog expects.
 
 .. code-block:: python
 
@@ -64,7 +64,7 @@ The primary way to set keys is ``[p]set api``. Run on its own, with no service o
         @commands.command()
         async def weather(self, ctx: commands.Context, *, city: str):
             tokens = await self.bot.get_shared_api_tokens("weather")
-            if not tokens.get("api_key"):
+            if tokens.get("api_key") is None:
                 view = SetApiView(default_service="weather", default_keys={"api_key": ""})
                 await ctx.send(
                     "The weather API key has not been set. "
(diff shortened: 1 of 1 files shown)


docs/framework_apikeys.rst
```diff
@@ -48,6 +48,31 @@ Basic Usage
             # Use the API key to access content as you normally would
 
 
+**********************
+Prompting for API Keys
+**********************
+
+Instead of asking the bot owner to type ``[p]set api ...`` by hand, a cog can prompt them with a secure button using ``SetApiView`` from ``redbot.core.utils.views``. It shows an owner-only button that opens a modal where the keys are entered, and saves them with ``set_shared_api_tokens`` when submitted.
```
Flame442: I think `[p]set api` should be the "primary" way that users interface with the API key system, with the View already spawning from that command if no inputs are given (albeit without the service specific hints). Instead, I think the View is better used as a fallback included with the error message when the owner attempts to use a command requiring an API where the key has not been set. Adding an extra command in your cog that does the same thing as the existing command just bloats the bot with no real benefit, but dynamically responding to commands that fail due to missing keys can save a step for the owner.

Changing the example to that kind of setup would be more impactful.

Flame442 (changes requested):

docs/framework_apikeys.rst
ktriponis: Good point - I hadn't clocked that `[p]set api` already opens the modal when run bare, so a separate command really is just duplication. Reworked the example around the missing-key fallback instead: the command checks for the key and, if it isn't set, replies with the error plus the button so the owner can set it in place. Better fit, thanks!

docs/framework_apikeys.rst
```diff
@@ -48,6 +48,37 @@ Basic Usage
             # Use the API key to access content as you normally would
 
 
+**********************
+Prompting for API Keys
+**********************
+
+The primary way to set keys is ``[p]set api``. Run on its own, with no service or tokens, it opens a secure modal for the owner to fill in, so a cog does not need to add its own command for this.
+
+``SetApiView`` (from ``redbot.core.utils.views``) is most useful as a fallback when a required key is missing. When a command needs a key that has not been set, reply with the error and attach the button so the owner can set it in place. ``default_service`` and ``default_keys`` pre-fill the modal with the service and key names the cog expects.
```
Flame442:
```suggestion
`SetApiView` is most useful as a fallback when a required key is missing. When a command needs a key that has not been set, a reply with the error message that includes this view will include a button that allows the owner to instantly open the secure modal. The parameters ``default_service`` and ``default_keys`` can be configured to pre-fill the modal with the service and key names the cog expects.
```

docs/framework_apikeys.rst
```diff
@@ -48,6 +48,37 @@ Basic Usage
...
+
+.. code-block:: python
+
+    from redbot.core.utils.views import SetApiView
+
+    class MyCog(commands.Cog):
+        @commands.command()
+        async def weather(self, ctx: commands.Context, *, city: str):
+            tokens = await self.bot.get_shared_api_tokens("weather")
+            if not tokens.get("api_key"):
```
Flame442:
```suggestion
            if tokens.get("api_key") is None:
```

docs/framework_apikeys.rst
```diff
@@ -48,6 +48,37 @@ Basic Usage
...
+                view = SetApiView(default_service="weather", default_keys={"api_key": ""})
+                await ctx.send(
+                    "The weather API key has not been set. "
+                    "Use the button below to set it (bot owner only).",
+                    view=view,
+                )
+                return
+            # use tokens["api_key"] as normal
+
+``default_service`` pre-fills and locks the service name. ``default_keys`` is a mapping of the key names the service expects (values may be empty) and pre-populates the modal. The owner enters one ``key value`` pair per line, keeping the key label; for the example above the field is pre-filled with ``api_key YOUR_API_KEY``, and replacing the line with just the value is rejected. On submit the tokens are saved with ``set_shared_api_tokens``.
```
Flame442:
```suggestion
``default_service`` pre-fills and locks the service name. ``default_keys`` is a mapping of the key names the service expects (values may be empty) and pre-populates the modal. The owner enters one ``key value`` pair per line, keeping the key label; for the example above the field is pre-filled with ``api_key YOUR_API_KEY``, and replacing the line with just the value is rejected. On submit the tokens are saved with `Red.set_shared_api_tokens`.
```

docs/framework_apikeys.rst
```diff
@@ -48,6 +48,37 @@ Basic Usage
...
+
+To embed the prompt in a custom ``discord.ui.View``, use ``SetApiModal`` directly and send it with ``interaction.response.send_modal(...)``. ``SetApiView`` and ``SetApiModal`` are both owner-only.
```
Flame442:
```suggestion
To embed the prompt in a custom `discord.ui.View`, use `SetApiModal` directly and send it with ``interaction.response.send_modal(...)``. `SetApiView` and `SetApiModal` are both owner-only.
```

Flame442 (changes requested):
Attempting to make some things render nicer

Flame442 (approved):
Thank you for contributing!

